PR #348, opened 2019-03-12
Update to use axe-core 3.2.1

waabid:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1411573

#### Description of changes

Update usage of axe-core to 3.2.1 (latest). Also includes changes to tests that were broken as a result (some axe APIs now rely on axe tree to be setup).


src/scanner/rule-to-links-mappings.ts
```diff
@@ -87,4 +87,8 @@ export const ruleToLinkConfiguration: DictionaryStringTo<HyperlinkDefinition[]>
     'aria-allowed-role': [BestPractice],
     'autocomplete-valid': [BestPractice],
     'css-orientation-lock': [BestPractice],
+    'aria-hidden-focus': [link.WCAG_4_1_2],
+    'form-field-multiple-labels': [BestPractice],
+    'label-content-name-mismatch': [BestPractice],
+    'landmark-complementary-is-top-level': [link.WCAG_1_3_1],
```
dbjorge: >'landmark-complementary-is-top-level': [link.WCAG_1_3_1], [](start = 4, length = 57)

Can you mark this as `[link.WCAG_1_3_1, BestPractice]` off the bat, per #390?

package.json
```diff
@@ -87,7 +87,7 @@
     "dependencies": {
         "@types/react-helmet": "^5.0.8",
         "applicationinsights-js": "^1.0.20",
-        "axe-core": "3.1.1",
+        "axe-core": "3.2.1",
```
dbjorge: I feel worried about the impact this might have on our persisted assessment data; have you already audited for whether there are any changes to the axe-generated bits of our persistent data that would need any special version migration considerations?

waabid: Discussed offline: for the moment, we will keep as is since this change does not affect users. Moving forward, we may need to warn users or provide an additional UX if/when this happens.

src/scanner/rule-to-links-mappings.ts
waabid: Done 😄

PR #356, opened 2019-03-13
[TSLint] Fix no-typedef rule - II

Shobhit1:
#### Pull request checklist

#### Description of changes

Continuation of changes in fixes for `no-typedef` lint rules.

#### Notes for reviewers

Part 2 of the series of PRs to get rid of `no-typedef` lint error


src/assessments/types/report-instance-field.ts
```diff
@@ -14,6 +14,7 @@ export type ReportInstanceFieldMap = { [KEY in string]?: ReportInstanceField };
 export type ReportInstanceFields = ReportInstanceField[];
 
 type HasPropertyBag<PB> = { propertyBag: PB };
+type PropertyBagKey<PB> = string | PB[keyof PB & string];
```
smoralesd: this looks not to strong of a type... if the value is a `string` then it can be any string right? or am I missing something?

src/background/browser-adapter.ts
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 import { ClientBrowserAdapter, ClientChromeAdapter } from '../common/client-browser-adapter';
 
+// tslint:disable: typedef
```
smoralesd: was this added on a lapsus?

src/assessments/types/report-instance-field.ts
Shobhit1: The way I am normally assigning types is to check what the function is normally returning as of now and then just add it like it.

src/background/browser-adapter.ts
Shobhit1: Yes. Will remove it.

src/assessments/types/report-instance-field.ts
Shobhit1: Does it look okay now?

src/tests/unit/tests/scanner/widget-function.test.ts
```diff
@@ -95,7 +95,7 @@ describe('verify widget function data', () => {
 
     const context = {
         _data: null,
-        data: function(d: any) {
+        data: function(d): void {
```
lisli1: I'm not sure why `d: any` is not necessary here but it is used in `src/tests/unit/tests/scanner/unique-landmark-check.test.ts`(above)?

Shobhit1: If not necessary here because implicitly it is taking it as `any`. The case in the `unique-landmark-check.test.ts` was me being overly specific and just putting it there.

PR #380, opened 2019-03-18
Improve default logger debuggability

dbjorge:
#### Pull request checklist

- [x] **N/A** Addresses an existing issue: Fixes #0000
- [x] **N/A** Added screenshots/GIFs for UI changes.

#### Description of changes

Previously, the devtools console's file/line printout/link next to any logger-generated error or log line would always point to default-logger.ts, which is not useful. With this change, that link will instead go to the line that actually logged the message/error, which is what you actually want in 99.9% of cases.


src/common/logging/default-logger.ts
```diff
@@ -4,7 +4,11 @@ import { Logger } from './logger';
 // Licensed under the MIT License.
 export const createDefaultLogger = (): Logger => {
     return {
-        log: (message?: any, ...otherArgs: any[]) => console.log(message, ...otherArgs),
-        error: (message?: any, ...otherArgs: any[]) => console.error(message, ...otherArgs),
+        // It's important that these use the console functions directly rather than wrapping them
```
smoralesd: base on the comments on the test itself, I don't think you need this comments... looks like over documenting why you're doing this.
Also, how about typings for `Logger`? does the types show up properly on VSCode?

dbjorge: Fair enough, removing redundant comment. The types still show up fine in VSCode at the consuming classes (they're all written in terms of the `Logger` type, which this PR doesn't impact)

PR #382, opened 2019-03-18
Move common test helper code out of individual test files

dbjorge:
#### Pull request checklist

- [x] **N/A** Addresses an existing issue: Fixes #0000
- [x] **N/A** Added relevant unit test for your changes. (`npm run test`)
- [x] **N/A** Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] **N/A** Added screenshots/GIFs for UI changes.

#### Description of changes

Wallaby complains with warnings when test files import other test files, since it causes it to double-run tests. This fixes the existing cases of these warnings by extracting the common test helpers into their own files.

### DIFF
--- a/src/tests/unit/common/visual-helper-toggle-config-builder.ts
+++ b/src/tests/unit/common/visual-helper-toggle-config-builder.ts
@@ -1,3 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
 import { VisualHelperToggleConfig } from '../../../assessments/types/test-step';
 import { ManualTestStatus } from '../../../common/types/manual-test-status';
 import {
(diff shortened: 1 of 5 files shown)


src/tests/unit/common/visual-helper-toggle-config-builder.ts
```diff
@@ -0,0 +1,99 @@
...
+                    html: 'html',
+                    propertyBag: {},
+                    target: ['element2'],
+                } as IGeneratedAssessmentInstance,
+            } as DictionaryStringTo<IGeneratedAssessmentInstance>,
+            isStepEnabled: true,
+            isStepScanned: false,
+        };
+    }
+    public withActionMessageCreator(actionMessageCreator: DetailsViewActionMessageCreator): VisualHelperToggleConfigBuilder {
```
smoralesd: We can use  `public with<P extends keyof T>(property: P, value: T[P])` directly from `BaseDataBuilder`; `with` has proper typings. In general, do you feel is better to have specific "with" functions or better to use the generic one?

This "with" function should be on the code since before we introduce the `BaseDataBuilder`, that may be the reason we have this but we don't really need them (from my point of view). Then again, I'm not sure how many references we may have to this functions. If we have too many, I'm fine with not changing this right now.

dbjorge: Yeah, I think using the generic one is an improvement, but I don't want to change it as part of this PR; I'd rather keep this one closer to a simple move

PR #383, opened 2019-03-19
[TSLint] Fix no typedef issues in some of our unit tests

Shobhit1:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000 - N/A
- [ ] Added relevant unit test for your changes. (`npm run test`) - N/A
- [ ] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage` - N/A
- [ ] Added screenshots/GIFs for UI changes. - N/A

#### Description of changes

Continuation of changes to add return typedef information on all our functions. In this PR, a lot of tests are covered.


src/tests/unit/tests/injected/analyzer-controller.test.ts
```diff
@@ -243,10 +243,10 @@ describe('AnalyzerControllerTests', () => {
 });
 
 class AnalyzerStub implements IAnalyzer<any> {
-    public analyze() {
+    public analyze(): Error {
         throw new Error('Method not implemented.');
     }
-    public teardown() {
+    public teardown(): Error {
```
smoralesd: shouldn't this be `void`?

Shobhit1: It doesn't matter for this case, where `throw new Error()` is the only command in the function and I am fine with this function returning `Error`. Until you feel very strongly about me returning `void` here, I guess we can keep this as `Error`

Shobhit1: changed it to "void"

src/tests/unit/tests/injected/visualization/drawer.test.ts
```diff
@@ -1153,7 +1153,7 @@ describe('Drawer', () => {
                 throw new Error('Not implemented');
             }
 
-            public getDialogRenderer() {
+            public getDialogRenderer(): any {
```
dbjorge: > any [](start = 39, length = 4)

Can you instead update `Formatter` to use `DialogRenderer?` and match that type here?

src/tests/unit/tests/injected/visualization/svg-drawer-v2.test.ts
```diff
@@ -996,7 +996,7 @@ describe('SVGDrawerV2Tests', () => {
             });
     }
 
-    function setupFilterFactoryDefault(dom) {
+    function setupFilterFactoryDefault(dom): void {
```
dbjorge: >dom [](start = 39, length = 3)

Should the parameter also have a type? I'm surprised that the line passes the lint check without that

dbjorge:
>     function setupSVGshapeFactoryDefault(dom): void {

Should this param have a type?

---
Refers to: src/tests/unit/tests/injected/visualization/svg-drawer-v2.test.ts:1010 in 9b3f69f. [](commit_id = 9b3f69fb3471fb2f60b1bbb740556bbc8eba9fef, deletion_comment = False)

src/tests/unit/tests/injected/visualization/svg-drawer-v2.test.ts
Shobhit1: There is no check as of yet to do that and personally I have not been touching those, So I would rather take that as a separate task with different specific PRs for those ones. 👍

dbjorge: Do you mean "we haven't enabled such a check"? I think the check exists (as the `parameter` argument to the `typedef` rule)

dbjorge: (regardless, fine with doing as a separate PR)

Shobhit1: Discussed offline, this should be discussed in our next engineering discussion meeting and hopefully come up with some solution for this 😄

src/tests/unit/tests/injected/visualization/drawer.test.ts
Shobhit1: Done.

PR #384, opened 2019-03-19
Clear prompt to go to settings when user starts typing a GitHub URL

peterdur:
#### Pull request checklist

- [ ] Addresses an existing issue: (opportunistic cleanup of UX imperfection)
- [X] Added relevant unit test for your changes. (`npm run test`)
- [X] Ran precheckin (`npm run precheckin`)
- [X] Added screenshots/GIFs for UI changes. (not applicable)

#### Description of changes

Previous behavior:
- Start with GitHub repo link blank in settings.
- Select an issue in details view and click File Issue button to show help text saying go to settings.
- Go to settings and specify a repo link
- Observe: help text is hidden. 
- Delete the link
- Observe: The help text is shown again, so it’s announced again. 

Change:
- When the repo link is changed to be non-empty, reset the state of the help text so that another button click is required before it is shown again, even if the repo URL is empty.


src/common/components/file-issue-details-button.tsx
```diff
@@ -36,6 +37,12 @@ export class FileIssueDetailsButton extends React.Component<FileIssueDetailsButt
         this.state = { showingFileIssueDialog: false, showingHelpText: false };
     }
 
+    private static getDerivedStateFromProps(props: FileIssueDetailsButtonProps): FileIssueDetailsButtonState {
+        if (isString(props.issueTrackerPath) && props.issueTrackerPath.length > 0) {
```
Shobhit1: Should we trim the `issueTrackerPath` variable before checking for its length to check if some user didn't just put any spaces in?

src/tests/unit/tests/common/components/file-issue-details-button.test.tsx
```diff
@@ -90,4 +90,22 @@ describe('FileIssueDetailsButtonTest', () => {
         bugActionMessageCreatorMock.verifyAll();
         expect(wrapper.getElement()).toMatchSnapshot();
     });
+
+    test('getDerivedStateFromProps clears state when issueTrackerPath set', () => {
+        const props = { issueTrackerPath: ' ' };
```
Shobhit1: If we check for trimmed value, this test will fail!

src/common/components/file-issue-details-button.tsx
peterdur: If the user types anything into settings (even a space), they no longer need to be told to go to settings.

src/tests/unit/tests/common/components/file-issue-details-button.test.tsx
peterdur: If we were to check for trimmed value, we'd want to keep this test case (and change the expectation), and add another test case for a value that's not empty after trimming.

src/common/components/file-issue-details-button.tsx
Shobhit1: Discussed offline, Scope of this bug and PR is just to check if there is anything in the url repo to make the settings text disappear! 👍

src/common/components/file-issue-details-button.tsx
```diff
@@ -36,6 +37,12 @@ export class FileIssueDetailsButton extends React.Component<FileIssueDetailsButt
         this.state = { showingFileIssueDialog: false, showingHelpText: false };
     }
 
+    private static getDerivedStateFromProps(props: FileIssueDetailsButtonProps): FileIssueDetailsButtonState {
```
dbjorge: This is the first usage of `getDerivedStateFromProps` in our codebase and it confused me a lot as a reader (it's really not apparent if you don't already know about this feature that a private static method is going to be involved in the render/update pipeline without doing any particular registration steps). It also confuses VS code (which thinks this is dead code).

Would you consider doing this with `onComponentDidUpdate` instead? I understand that doing so would trade adding an extra re-render in this path, but I think I'd much rather take that hit if it helps future maintainers to understand the code more easily.

haonliu: I'm wondering if there is a reason that you are avoiding using ```ComponentDidUpdate``` ?

peterdur: `getDerivedStateFromProps` looked like a closer match for what's happening here, and helped avoid the re-render, but I can take a look at using `componentDidUpdate` and see if that works out more familiarly.

peterdur: OK, pushed changes to do this with `componentDidUpdate`. Not much different in the main code (calls `setState` rather than returning new state, and adds an extra condition to the test to prevent looping). 

The test code is a little bit more complex in that we need to be mounted to be able to call `setState`, whereas `getDerivedStateFromProps` could be tested in isolation.

PR #387, opened 2019-03-19
[Lint] Fix no typedef errors

Shobhit1:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000 - N/A
- [ ] Added relevant unit test for your changes. (`npm run test`)- N/A
- [ ] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`- N/A
- [ ] Added screenshots/GIFs for UI changes.- N/A

#### Description of changes 

Lint Errros after this PR: **342 ▶️  273**

### DIFF
--- a/src/common/components/selector-input-list.tsx
+++ b/src/common/components/selector-input-list.tsx
@@ -146,7 +146,7 @@ export class SelectorInputList extends React.Component<SelectorInputListProps, S
     }
 
     @autobind
-    private getDeleteSelectorHandler(item: string[]): (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void {
+    private getDeleteSelectorHandler(item: string[]): (event: React.MouseEvent<HTMLButtonElement>) => void {
         return (event: React.MouseEvent<HTMLButtonElement>) => {
             this.deleteSelector(event, item);
         };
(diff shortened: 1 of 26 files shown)


src/common/components/selector-input-list.tsx
```diff
@@ -146,14 +146,14 @@ export class SelectorInputList extends React.Component<SelectorInputListProps, S
     }
 
     @autobind
-    private getDeleteSelectorHandler(item: string[]) {
+    private getDeleteSelectorHandler(item: string[]): (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void {
```
smoralesd: do we need the `MouseEvent` for the type?

src/injected/analyzer-controller.ts
```diff
@@ -95,7 +95,7 @@ export class AnalyzerController {
         });
     }
 
-    private getAnalyzerByIdentifier(key: string) {
+    private getAnalyzerByIdentifier(key: string): IAnalyzer<string> {
```
smoralesd: This is a little tricky by I think `<string>` is not the correct type for the `IAnalyzer` return type here.

Shobhit1: We are returning `IAnalyzer` using `key: string`. That was my thinking for this

src/common/components/selector-input-list.tsx
Shobhit1: Updated.

src/injected/analyzer-controller.ts
Shobhit1: Updated.

PR #393, opened 2019-03-20
Stubs clean up

smoralesd:
#### Pull request checklist

#### Description of changes

Remove unneeded/unused Stubs

#### Notes for reviewers

**N/A**


src/tests/unit/tests/injected/client-view-controller.test.ts
```diff
@@ -37,11 +35,10 @@ describe('ClientViewControllerTest', () => {
         let privateListenerVisStore;
         let privateListenerVisScanStore;
         let privateListenerAssessmentStore;
-        let privateListenerTabStore;
 
         const visualizationStore = Mock.ofType(VisualizationStore, MockBehavior.Strict);
         const visualizationScanResultStoreMock = Mock.ofType(VisualizationScanResultStore, MockBehavior.Strict);
-        const assessmentStoreMock = Mock.ofType<IBaseStore<IAssessmentStoreData>>(StoreStub, MockBehavior.Strict);
+        const assessmentStoreMock = Mock.ofType<IBaseStore<IAssessmentStoreData>>(null, MockBehavior.Strict);
```
waabid: Do we need null here? Can we mention the AssessmentStore?

smoralesd: I prefer not to have direct dependency with the concrete class, we should be good using the interface only (since now we have access to Proxies from javascript), but in this case, I didn't want to change/remove `MockBehavior.Strict` and thus, I used null. I guess we should clean this things up soon.

PR #396, opened 2019-03-20
[Lint] More typedef lint error resolution

Shobhit1:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000 **- N/A**
- [ ] Added relevant unit test for your changes. (`npm run test`) **- N/A**
- [ ] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage` **- N/A**
- [ ] Added screenshots/GIFs for UI changes. **- N/A**

#### Description of changes

Continuation of the tslint fix for `no-typedef` rule.
This basically covers all of the `TS` files that are straight forward to fix. Will start working on fixing all the tests after this

### DIFF
--- a/src/common/indexedDB/idb.ts
+++ b/src/common/indexedDB/idb.ts
@@ -29,6 +29,14 @@ export class Store {
     }
 }
 
+type IDBResultTarget = EventTarget & {
+    result: IDBValidKey[];
+};
+
+type IDBEvent = {
+    target: IDBResultTarget;
+} & Event;
+
 let store: Store;
 
 function getDefaultStore(): Store {
@@ -64,7 +72,7 @@ export function keys(defaultStore = getDefaultStore()): Promise<IDBValidKey[]> {
 
     return defaultStore
         ._withIDBStore('readonly', (s: any) => {
-            s.getAllKeys().onsuccess = function(event) {
+            s.getAllKeys().onsuccess = function(event: IDBEvent): void {
                 keyArray = event.target.result;
             };
         })
--- a/src/common/indexedDB/idb.ts
+++ b/src/common/indexedDB/idb.ts
@@ -29,9 +29,9 @@ export class Store {
     }
 }
 
-type IDBResultTarget = EventTarget & {
+type IDBResultTarget = {
     result: IDBValidKey[];
-};
+} & EventTarget;
 
 type IDBEvent = {
     target: IDBResultTarget;
--- a/src/injected/shadow-initializer.ts
+++ b/src/injected/shadow-initializer.ts
@@ -33,7 +33,7 @@ export class ShadowInitializer {
         container.appendChild(styleElement);
     }
 
-    private createShadowHost() {
+    private createShadowHost(): HTMLDivElement {
         const rootContainer = this.htmlElementUtils.querySelector(`#${rootContainerId}`);
 
         const shadowHostElement = this.createDivWithId('insights-shadow-host');
@@ -43,11 +43,11 @@ export class ShadowInitializer {
         return shadowHostElement;
     }
 
-    private removeExistingShadowHost() {
+    private removeExistingShadowHost(): void {
         this.htmlElementUtils.deleteAllElements('#insights-shadow-host');
     }
 
-    private createShadowContainer() {
+    private createShadowContainer(): HTMLElement {
         this.removeExistingShadowHost();
 
         const shadowHostElement = this.createShadowHost();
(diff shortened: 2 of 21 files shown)


src/common/indexedDB/idb.ts
```diff
@@ -29,6 +29,14 @@ export class Store {
...
 }
 
+type IDBResultTarget = EventTarget & {
+    result: IDBValidKey[];
+};
+
+type IDBEvent = {
+    target: IDBResultTarget;
+} & Event;
+
```
peterdur: Is there a reason `IDBResultTarget` takes the form `otherType & { … }` but `IDBEvent` takes the form `{ … } & otherType`? If we'd like them to be similar, do we have a preference for either order?

Shobhit1: There is no preference. I am not sure but I think they mean the same. So I have no preference and I have seen both used in our code.

smoralesd: We should have a preference! And of course consistency in the file is the minimum bar!

Shobhit1: From preference I meant was a repo-wide preference and that I dont have a personal pref for this. As far as consistency in the file is concerned, I will change this instance.

Shobhit1: Done.

src/injected/visualization/failure-instance-formatter.ts
```diff
@@ -1,10 +1,17 @@
...
+import { DialogRenderer } from '../dialog-renderer';
 import { AxeResultsWithFrameLevel, IAssessmentVisualizationInstance } from '../frameCommunicators/html-element-axe-results-helper';
 import { Formatter } from './formatter';
 
+type failureBoxConfigType = {
+    background: string;
+    fontColor: string;
+    text: string;
+    boxWidth: string;
+};
```
dbjorge: formatter.ts already defines a BoxConfig type identical to this; why not reuse that?

src/injected/shadow-initializer.ts
```diff
@@ -33,7 +33,7 @@ export class ShadowInitializer {
         container.appendChild(styleElement);
     }
 
-    private createShadowHost() {
+    private createShadowHost(): HTMLDivElement {
```
dbjorge: >HTMLDivElement [](start = 32, length = 14)

I think it would be a slight improvement for this function to encapsulate the implementation detail of which element type happens to be used for the shadow host container (ie, that this method would do better to return HTMLElement rather than HTMLDivElement)

Shobhit1: Tbh I don't want to change the implementation of this function in this PR. In my opinion, thats not the meaning or scope of this PR, which is specific to only putting a typedef where it is not. I dont mind looking into changing some other time.

src/injected/visualization/failure-instance-formatter.ts
Shobhit1: Done. Missed that.

src/injected/shadow-initializer.ts
dbjorge: I think you're misunderstanding; this wouldn't change the implementation, just the type annotation you've added as part of this PR.

---
In reply to: [267558414](https://github.com/Microsoft/accessibility-insights-web/pull/396#discussion_r267558414) [](ancestors = 267558414)

Shobhit1: ah, i think `createDivWithId` returns a `div` element and it made sense for me to actually be very specific here that it returns a `div` rather than make it `HTMLElement` which is like returning a parent class than actually be specific. I don't really know or may be I am not able to understand this, how is `HTMLElement` better in this case. Sorry.

dbjorge: The responsibility of this function is to create the HTML element that will represent our shadow host. Its consumers don't need to know about exactly what tags/attributes/etc are involved in "being a shadow host element" - that's this function's job to encapsulate. Returning HTMLDivElement leaks an implementation detail unnecessarily

Shobhit1: But if the responsibility of the function was to create an HTML element, the implementation should have something resembling that, it is specifically creating a `div`. I guess changing the name of the function will make more sense in that scenario. For eg: `createShadowContainer` does exactly that and hence it is returning a `HTMLElement` and not any specific `div` or `anchor` or any other element.

Shobhit1: Changed

PR #398, opened 2019-03-20
Added distinguished, date-time-based filenames to reports

markreay:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1443097

#### Description of changes

No UI changes strictly. But the file names have been changed for both reports as illustrated in the following screenshot.

image

The example file names are 

* AssessmentReport_20190320_AccessibleUniversity.html
* AutomatedChecksReport_20190320_AccessibleUniversity.html

The file names include the date and an abbreviated version of the site title.

#### Notes for reviewers

I leveraged and enhanced existing report name generation code.


src/DetailsView/reports/components/report-file-name-generator.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+type DateProvider = () => Date;
```
smoralesd: we already have something similar to this (a class though, not a type)

src/tests/unit/tests/DetailsView/components/export-dialog.test.tsx
```diff
@@ -4,8 +4,13 @@ import { IMock, Mock, MockBehavior, Times } from 'typemoq';
 
 import { DetailsViewActionMessageCreator } from '../../../../../DetailsView/actions/details-view-action-message-creator';
 import { ExportDialog, ExportDialogProps } from '../../../../../DetailsView/components/export-dialog';
+import { ReportFileNameGenerator } from '../../../../../DetailsView/reports/components/report-file-name-generator';
 
 describe('ExportDialog', () => {
+    const reportFileName = 'THE REPORT FILE NAME';
+    const reportFileNameGeneratorMock = Mock.ofType(ReportFileNameGenerator);
```
smoralesd: you probably want to use `Mock.ofType<ReportFileNameGenerator>();` instead... works fine with interfaces and types, don't need a concrete class

src/DetailsView/reports/components/report-file-name-generator.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type ReportFileNameGeneratorDeps = {
+    dateProvider: DateProvider;
+};
+
+export class ReportFileNameGenerator {
```
dbjorge: >ReportFileNameGenerator [](start = 13, length = 23)

Discussed offline with mark that we already have a ReportNameGenerator with very similar functionality (but which seems to be dead code right now, that shows up as issues table state that never gets read). Mark's going to look into consolidating

src/tests/unit/tests/DetailsView/reports/components/report-file-name-generator.test.ts
```diff
@@ -0,0 +1,21 @@
+import { ReportFileNameGenerator } from '../../../../../../DetailsView/reports/components/report-file-name-generator';
```
smoralesd: copyright header should be on top

src/DetailsView/components/issues-table.tsx
```diff
@@ -125,6 +126,7 @@ export class IssuesTable extends React.Component<IssuesTableProps, IssuesTableSt
             <ExportDialog
                 deps={this.props.deps}
                 isOpen={this.state.isExportDialogOpen}
+                fileNameBase="AutomatedChecksReport"
```
peterdur: It looks like when I converged both reports on Export Dialog as part of the OF6 update, I [didn't wire up](https://github.com/Microsoft/accessibility-insights-web/commit/a1bc41e5b6247a30b16d634af8e04ec3cf6a3fc3#diff-0d3bccd06335566e1695cedac734218cL158
) the generated file name for the automated checks report. That's actually still available here at `this.state.exportName`. It contains the date (not the time) from the report and the target page name. There might be something reusable here.

src/DetailsView/reports/components/report-file-name-generator.ts
```diff
@@ -0,0 +1,19 @@
...
+    constructor(private readonly deps: ReportFileNameGeneratorDeps) {}
+
+    public getFileName(baseName: string, extension: string): string {
+        const { dateProvider } = this.deps;
+        const isoString = dateProvider().toISOString();
+        const dateTime = `${isoString.substr(0, 10)} ${isoString.substr(11, 8)}`;
```
peterdur: I noticed that this generates a GMT date/time for the file name, whereas the date/time shown in the report is local time. Should the two match?

peterdur: Another reason the timestamps might diverge for automated checks between the report and the file name: the report shows the time of the scan, and the file name is the time of the export. (Assessment reports currently display the time of report generation, so they'd only be different if a displayed time boundary is crossed.)

src/DetailsView/components/issues-table.tsx
markreay: I believe I have wired it up as you expected. Let me know if it isn't right.

---
In reply to: [267592312](https://github.com/Microsoft/accessibility-insights-web/pull/398#discussion_r267592312) [](ancestors = 267592312)

src/tests/unit/tests/DetailsView/components/export-dialog.test.tsx
markreay: I removed that mock so the comment isn't relevant anymore. I noted your feedback though, thanks.

---
In reply to: [267545106](https://github.com/Microsoft/accessibility-insights-web/pull/398#discussion_r267545106) [](ancestors = 267545106)

src/DetailsView/reports/components/report-file-name-generator.ts
markreay: I consolidated.

src/tests/unit/tests/DetailsView/reports/components/report-file-name-generator.test.ts
markreay: File removed.

src/DetailsView/reports/components/report-file-name-generator.ts
markreay: Back to local time.

src/DetailsView/components/export-dialog.tsx
```diff
@@ -8,10 +8,12 @@ import * as React from 'react';
 
 import { ExportResultType } from '../../common/telemetry-events';
 import { DetailsViewActionMessageCreator } from '../actions/details-view-action-message-creator';
+import { ReportNameGenerator } from '../reports/report-name-generator';
```
peterdur: This import is no longer needed.

src/DetailsView/reports/report-generator.ts
```diff
@@ -18,24 +18,22 @@ export class ReportGenerator {
...
 
     public generateAssessmentHtml(
-        deps: ReportGeneratorDeps,
         assessmentStoreData: IAssessmentStoreData,
         assessmentsProvider: IAssessmentsProvider,
         featureFlagStoreData: FeatureFlagStoreData,
         tabStoreData: ITabStoreData,
         description: string,
     ): string {
         return this.assessmentReportHtmlGenerator.generateHtml(
```
peterdur: In a more consistent world, we'd pass the `scanDate` to `assessmentReportHtmlGenerator.generateHtml` to ensure that the same date is used in both places (and from there to `assessmentReportModelBuilderFactory.create`) - but given that we're now only using the date (in local time zone) in the file name, there's not much of a chance of a user-visible mismatch. Given that this is an existing inconsistency between the assessment and the automated checks report code, not a blocker.

src/DetailsView/components/export-dialog.tsx
markreay: Removed.

src/DetailsView/reports/report-generator.ts
markreay: Agreed. Discussed offline with @peterdur and we will save this for the future.

PR #404, opened 2019-03-21
[tslint] switch-default

smoralesd:
#### Pull request checklist

- [x] Addresses an existing issue: Related to #1468692

#### Description of changes

Move switch-default rule to build enforced config

#### Notes for reviewers

**N/A**

### DIFF
--- a/src/tests/unit/common/visualization-scan-result-store-data-builder.ts
+++ b/src/tests/unit/common/visualization-scan-result-store-data-builder.ts
@@ -24,7 +24,6 @@ export class VisualizationScanResultStoreDataBuilder extends BaseDataBuilder<IVi
     }
 
     public withSelectorMap(type: VisualizationType, selectorMap: any): VisualizationScanResultStoreDataBuilder {
-        // tslint:disable-next-line: switch-default
         switch (type) {
             case VisualizationType.Headings:
                 this.data.headings.fullAxeResultsMap = selectorMap;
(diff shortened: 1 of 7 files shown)


src/tests/unit/common/visualization-scan-result-store-data-builder.ts
```diff
@@ -53,6 +55,7 @@ export class VisualizationScanResultStoreDataBuilder extends BaseDataBuilder<IVi
     }
 
     public withSelectedIdToRuleResultMap(type: VisualizationType, selectedIdToRuleResultMap: any): VisualizationScanResultStoreDataBuilder {
+        // tslint:disable-next-line: switch-default
```
Shobhit1: are we okay with adding `tslint:disable` ?
I was also looking at fixing the `switch-default` issue but could not find a default value for these switch cases, so didn't try this! 😛

src/tests/unit/common/visualization-scan-result-store-data-builder.ts
```diff
@@ -24,6 +24,7 @@ export class VisualizationScanResultStoreDataBuilder extends BaseDataBuilder<IVi
     }
 
     public withSelectorMap(type: VisualizationType, selectorMap: any): VisualizationScanResultStoreDataBuilder {
+        // tslint:disable-next-line: switch-default
```
dbjorge: >// tslint:disable-next-line: switch-default [](start = 8, length = 43)

I think it might be better to just make tests fail in the unsupported cases, rather than suppressing the rule, eg:

```
            default:
                throw new Error(`Test logic error: VisualizationType ${type} does not support fullAxeResultsMap`);
```

smoralesd: We are passing `-1 as VisualizationType` in so many places that this is not an option for me at this time! 

I'm considering other options as @Shobhit1 also raise the flag on this.

dbjorge: I think I'm maybe misunderstanding; I just tried out using this strategy and there were no instances where a test failed due to passing -1. However, there *were* a few instances of failures that look like they're exposing legitimate test bugs (that is, a test's logic is incorrectly assuming it should be able to use this builder on a visualization type that isn't supported), eg, chrome-command-handler.test.ts at line 130 has an each case that assumes VisualizationStoreDataBuilder.withDisable support TabStops even though it doesn't.

Even if there was such a case, I think it would still be a legitimate test bug we should look into fixing; if a test is invoking this builder and we're silently throwing away the input, it probably means the test isn't testing what it thinks it is.

Can you show a specific example where using a `throw`ing default case in the instances this PR is touching causes a failure for a test using the -1 thing in a reasonable way?

---
In reply to: [267848311](https://github.com/Microsoft/accessibility-insights-web/pull/404#discussion_r267848311) [](ancestors = 267848311)

src/tests/unit/common/visualization-scan-result-store-data-builder.ts
markreay: The switch-default conversation is dancing around the fact that we really should be using a different design pattern here. This, every-method-has-an-enum-that-we-switch-on anti-pattern really should be addressed soon.

src/tests/unit/common/visualization-scan-result-store-data-builder.ts
smoralesd: @dbjorge  I'm with you in the fact that having an explicit default -> throw statement will help us catch issues. So that's what I did.

src/tests/unit/common/visualization-scan-result-store-data-builder.ts
smoralesd: This as now been removed

src/tests/unit/tests/background/stores/visualization-scan-result-store.test.ts
```diff
@@ -289,8 +289,8 @@ describe('VisualizationScanResultStoreTest', () => {
 
         const expectedState = new VisualizationScanResultStoreDataBuilder()
             .withSelectorMap(type, selectorMap)
-            .withFullIdToRuleResultMap(type, expectedFullIdToResultMap)
-            .withSelectedIdToRuleResultMap(type, expectedFullIdToResultMap)
+            .withFullIdToRuleResultMapForIssues(expectedFullIdToResultMap)
```
dbjorge: It feels a little off for half of the builder invocation to use `type` and the other half to use hardcoded Issues based things. I think it would be nicer to inline the `type` variable (like the `onUpdateIssuesSelectedTargets` test does on line 395) to make it more apparent that they're hardcoded in sync.

smoralesd: The point is we are not passing any other `VisualizationType` except for `Issues` hence the name change.

PR #405, opened 2019-03-21
remove feature flag high contrast mode

lisli1:
#### Pull request checklist

#### Description of changes

Removed highContrastMode from FeatureFlags and all references to FeatureFlags.highContrastMode

#### Notes for reviewers

(put down any questions or special points for reviewers)

### DIFF
--- a/src/tests/unit/tests/DetailsView/components/settings-panel.test.tsx
+++ b/src/tests/unit/tests/DetailsView/components/settings-panel.test.tsx
@@ -4,6 +4,7 @@ import { IChoiceGroupOption } from 'office-ui-fabric-react';
 import * as React from 'react';
 import { IMock, Mock, Times } from 'typemoq';
 
+import { shallow } from 'enzyme';
 import { FeatureFlags } from '../../../../../common/feature-flags';
 import { UserConfigMessageCreator } from '../../../../../common/message-creators/user-config-message-creator';
 import { BugServicePropertiesMap, UserConfigurationStoreData } from '../../../../../common/types/store-data/user-configuration-store';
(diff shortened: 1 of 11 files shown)


src/tests/unit/tests/DetailsView/components/settings-panel.test.tsx
```diff
@@ -107,11 +108,11 @@ describe('SettingsPanelTest', () => {
                 userConfigMessageCreator: userConfigMessageCreatorMock.object,
             },
             userConfigStoreState: userConfigStoreData,
-            featureFlagData: { [FeatureFlags.highContrastMode]: false },
+            featureFlagData: { 'test-flag': false },
```
dbjorge: >featureFlagData: { 'test-flag': false }, [](start = 12, length = 40)

This shouldn't have any effect on the tests that are passing `'test-flag'`, right? can you instead just omit it?

lisli1: The problem is that SettingsPanelProps expects a featureFlagData element. I have tried removing it completely but it gives me an error, so I decided to replace the feature flag string with 'test-flag'

smoralesd: @dbjorge the reason I asked @lisli1 to do this is a bit more profound:
1. As part of this PR, `SettingsPanel` component is no longer using `FeatureFlags.highContrastMode` directly.
2. `SettingsPanelProps` still have a `featureFlagData` property, which we are passing through to the `FlaggedComponent` to render the bug filling setting.
3. As far as I know (please correct me if I'm wrong), from the snapshot test point of view, all empty objects are made equal.

Point 3. is the main reason for having this "fake" feature flag key under `featureFlagData`. In order to properly assert the `featureFlagData` is indeed being passed through to the `FlaggedComponent` we should not use an empty object. Instead, we should a test object, with test data, which will make it easier for the future dev to check that this condition is actually being met. Does this make sense?

src/tests/unit/tests/DetailsView/components/settings-panel.test.tsx
```diff
@@ -107,11 +108,11 @@ describe('SettingsPanelTest', () => {
...
         };
 
-        const testSubject = new TestableSettingsPanel(testProps);
-        expect(testSubject.render()).toMatchSnapshot();
+        const wrapped = shallow(<SettingsPanel {...testProps} />);
+        expect(wrapped.getElement()).toMatchSnapshot();
```
Shobhit1: Love this. `shallow` is a much better way to test with snapshot!

PR #428, opened 2019-03-22
[TSLint] Final no-typedef fixes

Shobhit1:
#### Pull request checklist

#### Description of changes
Removes remaning `no-typedef` lint errors

#### Notes for reviewers

There are 2 peculiar cases in this PR:
  - base-data-builder.ts : that class is abstract and has a function `with`, where if I specify `BaseDataBuilder<T>` as return type; a lot of other tests start to fail. So,
     - First way of fixing is to keep the class abstract and specify `this` as return type
     - Second is to make it a normal class and then add `BaseDataBuilder<T>` as its return type. This option builds and compiles the code at the moment but then with newer version of TS(3.3.3), gives an error.

I chose the first one as the simpler fix. Open to change if someone knows what is the significance of `abstract` in that class and is it safe to remove it.

 - ~with-store-subscription.tsx: the function `withStoreSubscription` returns anonymous class which I dont know how to put in `typedef` info. So disabling that line~

### DIFF
--- a/src/common/components/with-store-subscription.tsx
+++ b/src/common/components/with-store-subscription.tsx
@@ -15,6 +15,7 @@ export type WithStoreSubscriptionDeps<T> = {
     storeActionMessageCreator: IStoreActionMessageCreator;
 };
 
+// tslint:disable-next-line: typedef
 export function withStoreSubscription<P extends WithStoreSubscriptionProps<S>, S>(WrappedComponent: React.ComponentType<P>) {
     return class extends React.Component<P, S> {
         constructor(props: P) {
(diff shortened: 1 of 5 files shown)


src/common/components/with-store-subscription.tsx
```diff
@@ -15,6 +15,7 @@ export type WithStoreSubscriptionDeps<T> = {
     storeActionMessageCreator: IStoreActionMessageCreator;
 };
 
+// tslint:disable-next-line: typedef
```
dbjorge: I think `React.ComponentClass<P, S>` is the return type you're looking for here

Shobhit1: Thank you. Worked like a charm.

PR #432, opened 2019-03-22
Ubuntu build

smoralesd:
#### Pull request checklist

- [ ] Ran precheckin (`npm run precheckin`) **N/A**

#### Description of changes

Add new job section to build.yaml in order to build our extension on ubuntu agent.

#### Notes for reviewers

General idea is to make sure our extension build on linux. Running this job should increase our confidence that our extension actually run on linux/chrome.
First build run this job in 1m 43s. Also notice it runs in parallel with the other existing jobs.

##### Open question
I'm basically running:
- `npm ci`
- `npm run build:all`

No tests, not tslint, no format check and no header checks since this steps are currently being run on the other jobs. **Is this ok, or do you think we should run addition steps on this job?**

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -117,3 +117,26 @@ jobs:
                 pathtoPublish: '$(System.DefaultWorkingDirectory)/drop'
                 artifactName: 'drop'
             displayName: publish drop
+
+    - job: 'ubuntu_build'
+
+      pool:
+          vmImage: 'ubuntu-16.04'
+
+      steps:
+          - task: NodeTool@0
+            inputs:
+                versionSpec: '10.15.0'
+            displayName: use node 10.15.0
+
+          - task: Npm@1
+            inputs:
+                command: custom
+                customCommand: ci
+            displayName: npm ci
+
+          - task: Npm@1
+            inputs:
+                command: custom
+                customCommand: run build:all
+            displayName: npm run build:all


peterdur:
I think it would be good to run the e2e tests on Ubuntu, if only to check that the extension loads successfully, given that it was broken for a while.

build.yaml
```diff
@@ -117,3 +117,26 @@ jobs:
                 pathtoPublish: '$(System.DefaultWorkingDirectory)/drop'
                 artifactName: 'drop'
             displayName: publish drop
+
+    - job: 'ubuntu_build'
+
+      pool:
```
devaradhanm: did u check if there is a way to import yaml file to avoid duplicate code?

smoralesd: Nop. It's a good point, a little out of the discussion scope but I'll have a look

smoralesd: Done... I think we can take a second look a this later as right now my objective is to make sure we add a new job to run things under ubuntu!

smoralesd:
> I think it would be good to run the e2e tests on Ubuntu, if only to check that the extension loads successfully, given that it was broken for a while.

Agree. Based on @devaradhanm comment, I used template feature to extract the build and run e2e test to a separated file and reuse the template on both, macOS & ubuntu runs. Only the macOS run publish test results and the drop for now.

PR #450, opened 2019-03-26
Remove Tslint disable comment for BaseStore interface

Shobhit1:
#### Pull request checklist

#### Description of changes

Remove Tslint disable comment for BaseStore interface


src/background/stores/base-store.ts
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { Store } from '../../common/flux/store';
-import { IBaseStore } from '../../common/istore';
+import { BaseStore } from '../../common/istore';
 import { StoreNames } from '../../common/stores/store-names';
 
-export abstract class BaseStore<TState> extends Store implements IBaseStore<TState> {
+export abstract class BaseStoreImpl<TState> extends Store implements BaseStore<TState> {
```
smoralesd: I'm wondering if we should update the file name too... it should only affect the files that are already part of this PR, right?

Shobhit1: I think so. I will check and update this PR.

src/common/istore.d.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-// tslint:disable-next-line:interface-name
-export interface IBaseStore<TState> {
+
+export interface BaseStore<TState> {
```
smoralesd: should we change the file name here? and move from `.d.ts` to `.ts` is a good idea too

Shobhit1: I want to, but  again only if It doesn't increase the number of changes too much!

Shobhit1: Done. Changed it to `base-store.ts`

src/background/stores/base-store.ts
Shobhit1: changed it to `base-store-impl.ts`

Shobhit1:
> I have some comments about file names

Done. Thanks for the feedback. 😄

PR #452, opened 2019-03-26
Build pipeline improvements

smoralesd:
#### Pull request checklist

#### Description of changes

Use templates to reduce duplication on our build pipeline definition.

#### Notes for reviewers

The end goal should be adding a ubuntu job. I have found issues running unit and e2e test so we need more time to fix those issues. This is preparation work and anything related to running our build under ubuntu is out of scope.

### DIFF
--- a/pipeline/e2e-test.yaml
+++ b/pipeline/e2e-test.yaml
@@ -1,15 +1,15 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 steps:
-  - task: Npm@1
-    inputs:
-        command: custom
-        customCommand: run build:all
-    displayName: npm run build:all
+    - task: Npm@1
+      inputs:
+          command: custom
+          customCommand: run build:all
+      displayName: npm run build:all
 
-  # e2e tests depend on npm run drop
-  - task: Npm@1
-    inputs:
-        command: custom
-        customCommand: run test:e2e -- --ci
-    displayName: run e2e tests
+    # e2e tests depend on npm run drop
+    - task: Npm@1
+      inputs:
+          command: custom
+          customCommand: run test:e2e -- --ci
+      displayName: run e2e tests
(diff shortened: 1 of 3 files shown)


pipeline/e2e-test.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+steps:
+    - task: Npm@1
+      inputs:
+          command: custom
+          customCommand: run build:all
+      displayName: npm run build:all
+
+    # e2e tests depend on npm run drop
```
peterdur: Do the e2e tests really depend on `npm run drop`? Because we don't seem to be doing that.

smoralesd: Comment is misleading. The actual dependency is on the `drop` folder, not on `npm run drop`. So, we need to run some of our steps to generate that `drop` folder and I guess the simplest, most straightforward way to do this is by running `npm run build:all`.

I updated the comment to reflect this.

pipeline/e2e-test.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+steps:
+    # e2e tests depend drop folder. run build:all will create said folder
```
dbjorge: I think including the 2 "publish" tasks for the e2e results/screenshots is at least as reasonable as including build:all here :)

smoralesd: I though about that... in the long run though, if we are able to run unit/e2e test on ubuntu, would we want to publish test results from that run too??? So I didn't move it, it should be a problem we solve in the future.

dbjorge: I think we definitely would want to publish test results from both places; we'd need to in order to debug failures effectively.

smoralesd: I prefer we make those changes when we get there, basically YAGNI

PR #454, opened 2019-03-27
Rename TestStep interface

haonliu:
#### Pull request checklist

#### Description of changes

#### Notes for reviewers

(put down any questions or special points for reviewers)


haonliu:
> Hey @haonliu can you break this PR in smaller PR's? 91 files is a little hard to review.

I know it is hard to review... However, the renaming of 'TestStep' interface itself is being used in every requirement configuration file, that's why this PR is so big. and I cannot break that down any more.

src/assessments/types/requirement.ts
```diff
@@ -55,6 +55,6 @@ export interface VisualHelperToggleConfig {
     assessmentNavState: AssessmentNavState;
     instancesMap: DictionaryStringTo<IGeneratedAssessmentInstance>;
     actionMessageCreator: DetailsViewActionMessageCreator;
-    isStepEnabled: boolean;
-    isStepScanned: boolean;
+    isRequirementEnabled: boolean;
```
dbjorge: This is a breaking change to our persisted store data format. We can think about how to manage this, but it's a MUCH more frightening change than updating type and file names. Can you please revert the property rename from this PR to be considered separately?

src/assessments/assessment-builder.tsx
```diff
@@ -239,12 +239,12 @@ export class AssessmentBuilder {
         } as Assessment;
     }
 
-    private static getStepConfig(steps: TestStep[], testStep: string): TestStep {
+    private static getStepConfig(steps: Requirement[], testStep: string): Requirement {
```
peterdur: It's _so_ tempting to suggest renaming methods like `getStepConfig` and parameters like `steps`... but this PR is plenty big already, so it makes sense to just do the single mechanical rename now to keep the immediate change as simple as possible (91 files!) and defer additional renames to future PRs.

src/assessments/audio-video-only/test-steps/audio-only-equivalent.tsx
```diff
@@ -5,7 +5,7 @@ import { link } from '../../../content/link';
 import * as content from '../../../content/test/audio-video-only/audio-only-equivalent';
 import { ManualTestRecordYourResults } from '../../common/manual-test-record-your-results';
 import * as Markup from '../../markup';
-import { TestStep } from '../../types/test-step';
+import { Requirement } from '../../types/requirement';
 import { AudioVideoOnlyTestStep } from './test-steps';
```
peterdur: Similarly, a future separate PR should rename the test-steps folders to requirements, and the contained test-steps.ts files to requirements.ts.

src/assessments/types/requirement.ts
haonliu: Good point. But VisualHelperToggleConfig seems not being used in stores or in background. It is only used in views. Let me know if I'm misunderstanding.

src/assessments/assessment-builder.tsx
haonliu: I agree. there's a lot more place to change, we cannot fit them in a single PR.

src/assessments/types/requirement.ts
dbjorge: I'm sorry, you're right; I saw `isStepScanned` in the persisted data and got concerned, but it comes from the property by that name in `manual-test-status.ts`, not from here. Still, I'd rather keep PRs this big to single mechanical refactorings at a time

haonliu: Cool, thanks.

PR #463, opened 2019-03-27
Re-factor assessment store to inject the initialAssessmentStoreDataGenerator dependency

haonliu:
#### Pull request checklist

#### Description of changes

Since a lot of tests are dependent on assessment store data builder, we have to make changes to many tests to ensure that they don't break. A lot of them are dependent on **some state** being given by the store data builder so we've gone with an internal mock approach, although not fully ideal, to return the state as expected.


src/tests/unit/common/assessment-store-data-builder.ts
```diff
@@ -5,12 +5,26 @@ import { AssessmentDataConverter } from '../../../background/assessment-data-con
 import { AssessmentStore } from '../../../background/stores/assessment-store';
 import { IAssessmentData, IAssessmentStoreData } from '../../../common/types/store-data/iassessment-result-data';
 import { VisualizationType } from '../../../common/types/visualization-type';
+import { InitialAssessmentStoreDataGenerator } from './../../../background/intial-assessment-store-data-generator';
 import { BaseDataBuilder } from './base-data-builder';
 
 export class AssessmentsStoreDataBuilder extends BaseDataBuilder<IAssessmentStoreData> {
-    constructor(provider: AssessmentsProvider, dataConverter: AssessmentDataConverter) {
+    constructor(
+        provider: AssessmentsProvider,
```
smoralesd: I'm wondering if this should be private?

src/tests/unit/tests/background/stores/assessment-store.test.ts
```diff
@@ -93,6 +95,11 @@ describe('AssessmentStoreTest', () => {
             } as IndexedDBAPI,
             MockBehavior.Strict,
         );
+        initialAssessmentStoreDataGeneratorMock = Mock.ofType(InitialAssessmentStoreDataGenerator);
+    });
+
+    afterEach(() => {
+        initialAssessmentStoreDataGeneratorMock.verifyAll();
```
smoralesd: I'm not sure I agree we should be using 'afterEach' to run the "assert" step of our tests! I know it seems easier to put this step in here but my take is this functions `afterEach`/`afterAll` should be used for clean up @dbjorge @peterdur @devaradhanm any thoughts on this?

src/tests/unit/common/assessment-store-data-builder.ts
```diff
@@ -5,12 +5,26 @@ import { AssessmentDataConverter } from '../../../background/assessment-data-con
...
-        this.data = new AssessmentStore(null, null, dataConverter, null, provider, null, null).getDefaultState();
+        this.data = new AssessmentStore(
+            null,
+            null,
+            dataConverter,
+            null,
+            provider,
+            null,
+            null,
+            initialAssessmentStoreDataGenerator || new InitialAssessmentStoreDataGenerator(provider.all()),
```
waabid: Why are we doing this?

markreay: Can we mock that?

src/background/stores/assessment-store.ts
```diff
@@ -278,14 +277,12 @@ export class AssessmentStore extends BaseStoreImpl<IAssessmentStoreData> {
...
-        if (assessmentData.generatedAssessmentInstancesMap == null) {
-            return;
-        }
-
         payload.payloadBatch.forEach(updateInstanceVisibilityPayload => {
+            const config = this.assessmentsProvider.forType(updateInstanceVisibilityPayload.test).getVisualizationConfiguration();
+            const assessmentData = config.getAssessmentData(this.state);
+            if (assessmentData.generatedAssessmentInstancesMap == null) {
+                return;
+            }
```
waabid: This change is not absolutely necessary in this PR but I figured to make it since I was working within the test for this and found it odd that it was not behaving per whatever was sent in the payload. Since this **is hidden behind a non-public feature flag**, I would feel comfortable moving forward with this.

src/tests/unit/tests/background/stores/assessment-store.test.ts
dbjorge: I agree; I would much rather make individual tests be explicit about exactly what they're arranging+acting+asserting, as a rule

Shobhit1: I don't know but I think the verification for the mocks is one thing that I am okay with putting in the `afterEach`.  Just saves me a sure shot repetition for something that I want to do in each test anyway!

Shobhit1: Putting this item as a discussion point for next engineering discussion! Will link this PR as well.

src/tests/unit/common/assessment-store-data-builder.ts
waabid: I believe we do not keep it, so we don't need to do that 😄

src/tests/unit/common/test-assessment-provider.tsx
```diff
@@ -27,6 +28,9 @@ const assessmentWithColumns: Assessment = {
     title: 'assessment 1',
     gettingStarted: null,
     guidance: content.assessment1.guidance,
+    initialDataCreator: () => {
```
smoralesd: does it make sense to extract this function?

src/tests/unit/tests/background/initial-assessment-store-data-generator.test.ts
```diff
@@ -1,32 +1,60 @@
...
-    it.each([[undefined], [null], [{}]])('outputs default assessment data if persistedData.assessments is %p', persistedAssessments => {
+    it.each([[undefined], [null]])('outputs default assessment data if persistedData.assessments is %p', persistedAssessments => {
+        const generatedState = generator.generateInitialState({
+            assessments: persistedAssessments,
+        } as IAssessmentStoreData);
+
+        expect(generatedState.assessments).toEqual(defaultState.assessments);
+    });
+
+    it('outputs default assessment data if persistedData.assessments is non-nullable', () => {
```
dbjorge: >outputs default assessment data if persistedData.assessments is non-nullable [](start = 8, length = 76)

Can you improve this test's name? I don't think it's expressing a real requirement correctly (the name as written suggests that it should always be generating default data and never be using persisted data, even if present)

waabid: Whoops, correct. It should be returning whatever data is given by initial data creator.

src/tests/unit/tests/background/__snapshots__/initial-assessment-store-data-generator.test.ts.snap
```diff
@@ -7,67 +7,8 @@ Object {
...
-          "stepFinalResult": 1,
-        },
-        "assessment-2-step-2": Object {
-          "isStepScanned": false,
-          "stepFinalResult": 1,
-        },
-      },
-    },
+    "assessment-1": null,
+    "assessment-2": null,
```
dbjorge: This isn't a very good test right now; it'd be easy for the class under test to have a bug that doesn't trigger a test failure. Can this use a more reasonable stub object?

waabid: Sure

PR #478, opened 2019-03-29
Move content from ./src/popup/scripts/ to ./src/popup/

smoralesd:
#### Description of changes

Our `popup` folder is the only one where we have a `scripts` folder inside. This is usual at the root level (similar to our `src` folder) but not necessary on the lower level folder. 
Moving the content inside `scripts` folder one level up with this PR.

#### Pull request checklist


src/DetailsView/components/left-nav/nav-link-handler.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { autobind } from '@uifabric/utilities';
-
+import * as React from 'react';
```
Shobhit1: Do we need `React` for a `ts` file?

src/tests/unit/tests/popup/scripts/components/header.test.tsx
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
```
peterdur: Should we also remove the scripts folder from the test tree?
src/tests/unit/tests/popup/**scripts**/components/header.test.tsx

smoralesd: Good catch... totally forgot about the tests :(

src/DetailsView/components/left-nav/nav-link-handler.ts
smoralesd: done... we should have a closer look at this though

src/tests/unit/tests/popup/scripts/components/header.test.tsx
smoralesd: @peterdur Updated

Issue #474, opened 2019-03-28
TabStore onTabUpdate gives an exception


Shobhit1:
## Describe the bug

**To Reproduce**

1. Go to `Chrome`. Open a new tab.
2. Open `background.html` devtools console.
3. Type `google.com` (website url doesn't matter here). Type any url.
3. Check `console` to get the below mentioned error

## Expected behavior

There should not be a console error before I even start to use my accessibility Insights extension.

## GIF

![tab-id-undefined](https://user-images.githubusercontent.com/4496335/55196635-30970700-516d-11e9-9995-66d47e5d94f1.gif)

## Extension (please complete the following information)

- OS: Windows 10
- Version:1.0.4
- Powered by axe-core 3.2.1
- Chrome Version: Version 74.0.3729.40 (Official Build) beta (64-bit)
- Target Page: N/A

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

Yes

ferBonnin:
@haonliu is this resolved?

haonliu:
> @haonliu is this resolved?

Yes. Should the person who opened this issue close it or I should?

ferBonnin:
great!
the person who open it. @Shobhit1 could you please verify?

Shobhit1:
Fixed. I dont see the console error anymore. Closing the issue.

PR #484, opened 2019-03-30
BugFix - TabStore onTabUpdate gives an exception

haonliu:
#### Description of changes

 I think this is an issue from chrome, where onUpdated Events should provide with three variables: 
```
- tabId: number,  // the id of the tab get updated
- changeInfo: chrome.tabs.TabChangeInfo, 
- tab: chrome.tabs.Tab. // the tab instance that get updated
```
However, when I debug in chrome, the later two can be null in some cases: 
image

Thus I would suggest handle null case in our getTab function. And we do not have to log error since chrome will do it:
image

P.S. No unit test are changed since browser adapter has no UTs.

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #[474](https://github.com/Microsoft/accessibility-insights-web/issues/474)


src/background/browser-adapter.ts
```diff
@@ -94,7 +94,14 @@ export class ChromeAdapter extends ClientChromeAdapter implements BrowserAdapter
     }
 
     public getTab(tabId: number, callback: (tab: chrome.tabs.Tab) => void): void {
-        chrome.tabs.get(tabId, callback);
+        chrome.tabs.get(tabId, tab => {
+            if (tab === null) {
+                console.log(`Cannot find tab with id ${tabId}`);
```
Shobhit1: Should we be using the `Logger` class?

https://github.com/Microsoft/accessibility-insights-web/blob/master/src/common/logging/logger.ts

haonliu: chrome will log the error

src/background/browser-adapter.ts
```diff
@@ -1,7 +1,7 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
waabid: This change seems unintended.

haonliu: thanks.

src/background/browser-adapter.ts
```diff
@@ -94,7 +94,11 @@ export class ChromeAdapter extends ClientChromeAdapter implements BrowserAdapter
     }
 
     public getTab(tabId: number, callback: (tab: chrome.tabs.Tab) => void): void {
```
dbjorge: I don't think this is an appropriate error contract; if I'm trying to call `getTab` from somewhere else in the code, and I call it with an invalid tab, I want the callsite to be responsible for deciding whether it's acceptable to allow that and move on or propogate an error with more context.

I think the ideal way to handle that would be to make this an async function and emit an exception that the caller can choose to handle or not, but if we don't want to do that for just this one particular function, I'd do it with the javascript-typical resolve/reject callback pattern.

haonliu: Great point, I'm also thinking doing that.

src/background/browser-adapter.ts
Shobhit1: I get that `chrome` will log the errors but since we have wrapped all our logs using `Logger` class; does it make sense to be consistent here as well and use that same class?

src/background/browser-adapter.ts
```diff
@@ -24,7 +24,7 @@ export interface BrowserAdapter extends ClientBrowserAdapter {
     createInactiveTab(url: string, callback: (tab: chrome.tabs.Tab) => void): void;
     closeTab(tabId: number): void;
     switchToTab(tabId: number): void;
-    getTab(tabId: number, callback: (tab: chrome.tabs.Tab) => void): void;
+    getTab(tabId: number, onResolve: (tab: chrome.tabs.Tab) => void, onReject?: () => void): void;
```
dbjorge: There are 4 other callsites for `BrowserAdapter.getTab` (across `details-view-initializer.ts`, `assessment-store.ts`, `tab-controller.ts`, and `target-tab-finder.ts`) that should probably be updated to account for this signature change.

src/popup/target-tab-finder.ts
```diff
@@ -40,9 +40,15 @@ export class TargetTabFinder {
...
-                this.browserAdapter.getTab(tabIdInUrl, (tab: Tab) => {
-                    resolve(tab);
-                });
+                this.browserAdapter.getTab(
+                    tabIdInUrl,
+                    (tab: Tab) => {
+                        resolve(tab);
+                    },
+                    () => {
+                        throw new Error(`Tab with Id ${tabIdInUrl} not found`);
```
dbjorge: This should instead be `reject`ing with the message in question.

haonliu: updated!

PR #496, opened 2019-04-03
tslint rule no-string-based-set-*

smoralesd:
#### Description of changes

SDT build is flagging our mock setup for `setTimeout` because we're using `It.is(isFunction)` which has a return type of any making tslint to flag this uses. 

On this PR I'm adding a wrapper to check the `setTimeout` parameter is a function and make this wrapper to return type of `Function`.

#### Pull request checklist

- [x] Addresses an existing issue: Fixes **several**

### DIFF
--- a/src/tests/unit/common/is-function.ts
+++ b/src/tests/unit/common/is-function.ts
@@ -1,5 +1,5 @@
-import { It } from 'typemoq';
 import { isFunction as checkIsFunction } from 'lodash';
+import { It } from 'typemoq';
 
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
--- a/src/tests/unit/common/is-function.ts
+++ b/src/tests/unit/common/is-function.ts
@@ -1,6 +1,0 @@
-import { isFunction as checkIsFunction } from 'lodash';
-import { It } from 'typemoq';
-
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-export const isFunction = It.is<Function>(checkIsFunction);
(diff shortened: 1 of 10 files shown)


src/tests/unit/common/is-function.ts
```diff
@@ -0,0 +1,6 @@
+import { isFunction as checkIsFunction } from 'lodash';
+import { It } from 'typemoq';
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dbjorge: Copyright header should go above imports

src/tests/unit/tests/background/injector-controller.test.ts
```diff
@@ -177,7 +177,7 @@ class InjectorControllerValidator {
 
     public setupTimeoutHandler(times: number): InjectorControllerValidator {
         this.mockWindowUtils
-            .setup(x => x.setTimeout(It.is(isFunction), It.isAnyNumber()))
+            .setup(x => x.setTimeout(isFunction, It.isAnyNumber()))
```
dbjorge: When I look at a test using this new helper, it kind of looks like the test has a bug because the normal usage of these mocks is usually of the form "It.isSomething". Maybe it would be more clear to use a name like "itIsFunction", even though this would be an exception to our normal conventions about starting functions with verbs?

src/tests/unit/common/is-function.ts
smoralesd: Ups

src/tests/unit/tests/background/injector-controller.test.ts
smoralesd: Agreed

smoralesd: @dbjorge Updated

PR #500, opened 2019-04-04
introduce the bug filing option interface

haonliu:
#### Description of changes

introduce the bug filing option interface.
This is in a very initial state, we will be changing this interface as needed.

Consider for future work: 
refactor behaviors like renderSettingsForm, buildStoreData, isSettingsValid, fileBug as Extensions.

#### Pull request checklist


src/bug-filing/types/bug-filing-option.ts
```diff
@@ -0,0 +1,12 @@
...
+import { CreateIssueDetailsTextData } from './../../common/types/create-issue-details-text-data';
+
+export interface BugFilingOption<Settings = {}> {
+    key: string;
+    displayName: string;
+    renderSettingsForm: React.SFC;
+    buildStoreData: (params: any[]) => Settings;
+    isSettingsValid: (data: Settings) => boolean;
+    fileBug: (data: Settings, bugData: CreateIssueDetailsTextData) => void;
+}
```
karanbirsingh: I'm curious how the various setting types (e.g. in GitHub there is a single string but Azure Devops has two strings) will be persisted - who is in charge of saving these settings, and will `renderSettingsForm` have enough information to render the form pre-populated? If I'm understanding it correctly, it seems like `fileBug` and `renderSettingsForm` are similar in the sense that they both need the settings information - LMK if I'm missing something

PR #509, opened 2019-04-09
Add GitHub bug filing service configuration

haonliu:
#### Description of changes

Add GitHub bug filing service configuration

#### Pull request checklist


src/bug-filing/github/github-bug-filing-service.tsx
```diff
@@ -0,0 +1,51 @@
...
+    };
+}
+
+function isSettingsValid(data: GithubBugFilingSettings): boolean {
+    return !!data && !_.isEmpty(data.repository.trim());
+}
+
+const renderSettingsForm = NamedSFC<SettingsFormProps<GithubBugFilingSettings>>('BugFilingSettings', props => {
+    const onGitHubRepositoryChange = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
+        props.deps.userConfigMessageCreator.setBugServiceProperty('gitHub', 'repository', newValue);
```
smoralesd: This is the `key` of this service. Should we have a const in this file and use it on both places?

smoralesd: I'm also worried about `repository` being pass as an string here, as this is the name of the only property of `GithubBugFilingSettings`. Any way we can tide those two value/name together?

src/bug-filing/github/github-bug-filing-service.tsx
```diff
@@ -0,0 +1,51 @@
...
+import * as _ from 'lodash';
+import { TextField } from 'office-ui-fabric-react/lib/TextField';
+import * as React from 'react';
+
+import { NamedSFC } from '../../common/react/named-sfc';
+import { CreateIssueDetailsTextData } from '../../common/types/create-issue-details-text-data';
+import { BugFilingService } from '../types/bug-filing-service';
+import { SettingsFormProps } from '../types/settings-form-props';
+
+export type GithubBugFilingSettings = {
```
smoralesd: Should this be 'Git**H**ubBugFilingSettings'?

src/bug-filing/github/github-bug-filing-service.tsx
haonliu: Yes we should. This snippet was copied from bug-filing-settings. I'll make the changes to address this issue.

haonliu: After a closer look, I think we should not pass service key in the payload. instead the store should take the selected bug service key to update its data. But for now, I'll leave it as it is.

smoralesd: That's an assumption we're always updating settings for the currently selected bug filing service. Having the option to update settings for any service (selected or not) is not too expensive (we already have it) and it makes the API more flexible.

src/bug-filing/github/github-bug-filing-service.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+function buildStoreData(repository: string): GitHubBugFilingSettings {
+    return {
+        repository,
+    };
+}
+
+function isSettingsValid(data: GitHubBugFilingSettings): boolean {
+    return !!data && !_.isEmpty(data.repository.trim());
```
Shobhit1: - this is going to fail if data is `{}`
- we either need to put a check for `data.repository` or
- we check `isEmpty(data)`

Basically, this is missing some cases. Please add `unit tests` for those cases.

For example: 
- data is `{}`
- data is `null` || `undefined`
- data is not empty but doesn't contain `repository` property

src/tests/unit/tests/bug-filing/github/github-bug-filing-service.test.tsx
```diff
@@ -37,15 +37,25 @@ describe('GithubBugFilingServiceTest', () => {
     });
 
     test('isSettingsValid', () => {
-        const emptySettings: GitHubBugFilingSettings = null;
-        const invalidSettings: GitHubBugFilingSettings = {
+        const nullSettings: GitHubBugFilingSettings = null;
```
smoralesd: does it make sense to use `it.each` to separate all this scenarios now?

src/tests/unit/tests/bug-filing/github/github-bug-filing-service.test.tsx
```diff
@@ -37,15 +37,25 @@ describe('GithubBugFilingServiceTest', () => {
...
+        const invalidSettings2: GitHubBugFilingSettings = {
             repository: '  ',
         };
+        const invalidSettings3: GitHubBugFilingSettings = {
+            repository: 3,
+        } as any;
         const validSettings: GitHubBugFilingSettings = {
             repository: 'repository',
         };
+        expect(GitHubBugFilingService.isSettingsValid(nullSettings)).toBe(false);
```
Shobhit1: In addition to @smoralesd comment, it might make sense to separate the `true` and `false` expectation test cases into different test cases. Just a thought. Not blocking, hence approving!

src/tests/unit/tests/bug-filing/github/github-bug-filing-service.test.tsx
haonliu: updated

PR #517, opened 2019-04-09
[Refactor] Introduce ActionMessageDispatcher

smoralesd:
#### Description of changes

Introduce `ActionMessageDispatcher`. The general idea is to not use inheritance on our action message creator. This could be part of a longer refactor on the action message creators... if I have time to complete it.

#### Pull request checklist


waabid:
My first thought for this PR was "oh, is this to get rid of the inheritance in our action message creators?" and lo and behold the description was perfect; love it when it works out that way! 😄

src/common/message-creators/action-message-dispatcher.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Message } from '../message';
+import { Messages } from '../messages';
+import { TelemetryData } from '../telemetry-events';
+
+export class ActionMessageDispatcher {
+    constructor(private postMessageDelegate: (message: Message) => void, private tabId: number) {}
+
+    public dispatchMessage(message: Message): void {
+        if (this.tabId != null) {
+            message.tabId = this.tabId;
```
dbjorge: Based on the naming, it was surprising to me that this function would mutate its input. Consider `{ ...message, this.tabId }` instead

src/common/message-creators/action-message-dispatcher.ts
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT License.
+import { PayloadWithEventName } from '../../background/actions/action-payloads';
+import { Message } from '../message';
+import { Messages } from '../messages';
+import { TelemetryData } from '../telemetry-events';
+
+export class ActionMessageDispatcher {
+    constructor(private postMessageDelegate: (message: Message) => void, private tabId: number) {}
+
+    public dispatchMessage(message: Message): void {
```
dbjorge: The reuse of the same `Message` type feels misleading, here; it isn't apparent to me as a reader from the names/types here that the caller of `dispatchMessage` should not fill in the `tabId` property of the input object.

src/tests/unit/tests/common/message-creators/action-message-dispatcher.test.ts
```diff
@@ -0,0 +1,87 @@
...
+import { It, Mock, Times } from 'typemoq';
+import { Message } from '../../../../../common/message';
+import { ActionMessageDispatcher } from '../../../../../common/message-creators/action-message-dispatcher';
+import { Messages } from '../../../../../common/messages';
+import { BaseTelemetryData, TelemetryEventSource } from '../../../../../common/telemetry-events';
+
+describe('ActionMessageDispatcher', () => {
+    const postMessageMock = Mock.ofInstance((message: Message) => {});
+
+    const tabIds = [-1, null];
```
dbjorge: It feels off to test *only* edge cases and no realistic ids

src/tests/unit/tests/common/message-creators/action-message-dispatcher.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+            const testObject = new ActionMessageDispatcher(postMessageMock.object, tabId);
+
+            testObject.dispatchMessage(message);
+
+            const expectedMessage = {
+                ...message,
+            };
+
+            if (tabId != null) {
```
dbjorge: repeating the implementation in the test like this isn't great; it's disproportionately likely for a bug in this sort of logic to be in the choice of which comparator to use, for example. I think I'd prefer to see this sort of test *not* use `it.each`, and to have the individual tests write out explicitly the exact output they're expecting.

For example, consider how close the techinique of dynamically building up the message comes here to making the test accidentally depend on the way the implementation accidentally mutates the input message.

src/common/message-creators/action-message-dispatcher.ts
smoralesd: Done. I added a related type: `InterpreterMessage` that "extend" `Message` and add an optional `tabId` property. Now `Message` doesn't have the `tabId?` property. In a following PR I'll rename `Message` to `ActionMessage` to match the naming of our `XActionMessageCreator`'s

src/common/message-creators/action-message-dispatcher.ts
smoralesd: I added a related type: `InterpreterMessage` that "extend" `Message` and add an optional `tabId` property. Now `Message` doesn't have the `tabId?` property. In a following PR I'll rename `Message` to `ActionMessage` to match the naming of our `XActionMessageCreator`'s

src/tests/unit/tests/common/message-creators/action-message-dispatcher.test.ts
smoralesd: Separate this cases to remove the `if` block.

src/tests/unit/tests/common/message-creators/action-message-dispatcher.test.ts
smoralesd: `if` block has been removed now

src/common/message-creators/action-message-dispatcher.ts
devaradhanm: how about having two message types? GlobalMessage & TabContextMessage?

---
In reply to: [274691083](https://github.com/Microsoft/accessibility-insights-web/pull/517#discussion_r274691083) [](ancestors = 274691083)

PR #524, opened 2019-04-10
add reference to the bug that prevents us from using idb-keyval package

Shobhit1:
#### Description of changes

Adds reference and a comment to why we are not able to use `idb-keyval` package and have to rely on the code rather than the package.

#### Pull request checklist

### DIFF
--- a/src/common/indexedDB/idb.ts
+++ b/src/common/indexedDB/idb.ts
@@ -1,5 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+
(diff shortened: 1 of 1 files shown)


src/common/indexedDB/idb.ts
```diff
@@ -1,5 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+
+// reference: https://github.com/jakearchibald/idb-keyval/blob/master/idb-keyval.ts
+// idb-keyval has an issue which prevents us using the `idb-keyval` npm package
```
smoralesd: should mention we cannot use it in combination with uglify right?

PR #525, opened 2019-04-10
build issue url for github

haonliu:
#### Description of changes

#### Pull request checklist


src/bug-filing/common/bug-filing-url-string-utils.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { EnvironmentInfo } from '../../common/environment-info-provider';
+import { CreateIssueDetailsTextData } from './../../common/types/create-issue-details-text-data';
+
+export class BugFilingUrlStringUtils {
```
smoralesd: shouldn't be 'IssueFilingUrlStringUtils'? I know we're already using 'bug' but on this PR you added 'IssueFilingDialogDeps' so better to be consistent, at least at the PR level and later we should have a PR to rename all the bug's to issue's

src/bug-filing/common/bug-filing-url-string-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+    public static footer(environmentInfo: EnvironmentInfo): string {
+        return (
+            'This accessibility issue was found using Accessibility Insights for Web ' +
```
smoralesd: instead of 'Accessibility Insights for Web' we should use `title` from `src\content\strings\application.ts`

src/bug-filing/github/create-gitHub-bug-filing-url.ts
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
smoralesd: not sure the casing in this filename follows our pattern. the 'H' in git**H**ub caught my attention here

src/tests/unit/tests/bug-filing/github/github-bug-filing-service.test.tsx
```diff
@@ -75,8 +75,6 @@ describe('GithubBugFilingServiceTest', () => {
     });
 
     describe('create bug filing url', () => {
-        it.each(invalidTestSettings)('with %o', settings => {
-            expect(GitHubBugFilingService.createBugFilingUrl(settings, null)).toBeNull();
-        });
+        expect(GitHubBugFilingService.createBugFilingUrl).not.toBeNull();
```
smoralesd: This doesn't looks correct... I guess the test should assert the createBugFilingUrl on GitHubFilingService is the same that we expect it to be... just not being null doesn't seems enough.

src/bug-filing/common/bug-filing-url-string-utils.ts
haonliu: yeah, I have create a work item for this. that would be the last work item before we complete this feature.

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
```diff
@@ -25,13 +26,22 @@ describe('IssueFilingDialog', () => {
     let deps: IssueFilingDialogDeps;
     let bugFilingServiceStub: BugFilingService;
     let props: IssueFilingDialogProps;
+    let envInfoProviderMock: IMock<EnvironmentInfoProvider>;
+    const envInfo = {
```
waabid: As per discussion: this should be in before Each.

src/tests/unit/tests/bug-filing/github/create-github-bug-filing-url.test.ts
```diff
@@ -0,0 +1,85 @@
...
+// Licensed under the MIT License.
+import { GlobalMock, GlobalScope, It } from 'typemoq';
+
+import { createGitHubBugFilingUrl } from '../../../../../bug-filing/github/create-github-bug-filing-url';
+import { IssueFilingUrlStringUtils } from './../../../../../bug-filing/common/issue-filing-url-string-utils';
+import { GitHubBugFilingSettings } from './../../../../../bug-filing/github/github-bug-filing-service';
+import { EnvironmentInfo } from './../../../../../common/environment-info-provider';
+
+describe('createGitHubBugFilingUrlTest', () => {
+    const environmentInfo: EnvironmentInfo = {
```
waabid: As per eng discussion: this should be in before Each.

src/tests/unit/tests/bug-filing/common/issue-filing-url-string-utils.test.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { IssueFilingUrlStringUtils } from './../../../../../bug-filing/common/issue-filing-url-string-utils';
+import { EnvironmentInfo } from './../../../../../common/environment-info-provider';
+
+describe('BugFilingUrlStringUtilsTest', () => {
+    const environmentInfo: EnvironmentInfo = {
```
waabid: As per eng discussion: this should be in before Each.

haonliu:
regarding the utils class: 
did some search, functions in utility classes (and they are not supposed to be changed.) generally should be static IMO. 
It make sense for a function like 
````
public static collapseConsecutiveSpaces(input: string): string {
        return input.replace(/\s+/g, ' ');
}
````
should be static, it is just like Math.max(). 
We should not avoid using them only because they are hard to mock. Maybe we need a better way to mock static functions.

File: src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx [hunk 8]
```diff
@@ -25,13 +26,23 @@ describe('IssueFilingDialog', () => {
     let deps: IssueFilingDialogDeps;
     let bugFilingServiceStub: BugFilingService;
     let props: IssueFilingDialogProps;
+    let envInfoProviderMock: IMock<EnvironmentInfoProvider>;
+    let envInfo;
```
waabid: missing typing

smoralesd:
The mock part is just one point on the list of reason why not to use static functions of classes.
Another important point to consider is the fact that static functions/classes are not pass as parameters to the constructors or functions, making it non-obvious from the public API of the user class that they indeed have a dependency on another class or function. Doing this (using "globals") goes against open-close & dependency inversion principles

PR #527, opened 2019-04-11
BugFilingSettingContainer gets visible services + issue filing dialog uses bug filing setting container

waabid:
#### Description of changes

Title says it all. Two main changes here: bug filing setting container uses bug service provider to get all visible services (new change) and then using the container in the issue filing dialog.

#### Pull request checklist

### DIFF
--- a/src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
+++ b/src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
@@ -40,7 +40,7 @@ describe('IssueFilingDialog', () => {
         };
         deps = {
             bugFilingServiceProvider: null,
-        };
+        } as IssueFilingDialogDeps;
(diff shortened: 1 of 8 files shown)


src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
```diff
@@ -40,7 +40,7 @@ describe('IssueFilingDialog', () => {
         };
         deps = {
             bugFilingServiceProvider: null,
```
haonliu: Might be better to pass something other than null here.

waabid: The reason I passed as null is because we don't care about what the actual deps are, just that they are the same object passed into children. Hence, if deps was {} or null, I would agree that is not okay but I'm fine with this.

PR #537, opened 2019-04-12
[Refactor] String utils to non-static class

smoralesd:
#### Description of changes

In order to use string utils as a "global", this PR is a refactor of the original class in a way that we can "normally" mock it's dependencies.

#### Pull request checklist


src/bug-filing/null-issue-filing-service/null-issue-filing-service.tsx
```diff
@@ -16,5 +16,5 @@ export const NullIssueFilingService: BugFilingService = {
     renderSettingsForm,
     buildStoreData: () => null,
     isSettingsValid: () => false,
-    createBugFilingUrl: () => null,
+    issueFilingUrlProvider: () => null,
```
devaradhanm: method names should be verb

devaradhanm: depends of what this method does, may be call it getIssueFilingUrlProvider or createIssueFilingUrlProvider

src/bug-filing/github/create-github-bug-filing-url.ts
```diff
@@ -2,57 +2,57 @@
...
-    const encodedIssue = `/new?title=${encodeURIComponent(title)}&body=${encodeURIComponent(body)}`;
-    return `${settingsData.repository}${encodedIssue}`;
+    return `${prefix}${data.ruleResult.help} (${selectorLastPart})`;
 }
+
+function buildGithubText(stringUtils: StringUtils, environmentInfo: EnvironmentInfo, data: CreateIssueDetailsTextData): string {
+    const result = data.ruleResult;
+
+    const text = [
+        `**Issue**: \`${result.help}\` ([\`${result.ruleId}\`](${result.helpUrl}))`,
```
devaradhanm: if this is going to be a key value pair, can you create a method that generates the line given a kv pair?

src/bug-filing/null-issue-filing-service/null-issue-filing-service.tsx
smoralesd: it doesn't return a provider... it provides the url, hence the wording

src/bug-filing/github/create-github-bug-filing-url.ts
smoralesd: I'm not touching this logic as part of this PR... I'm just moving the code out of and static function in a class. Does it make sense to refactor this as part of this PR?

src/bug-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -4,36 +4,35 @@ import { EnvironmentInfo } from '../../common/environment-info-provider';
...
-    public static footer(environmentInfo: EnvironmentInfo): string {
+export type StringUtils = {
+    footer: (info: EnvironmentInfo) => string;
+    collapseConsecutiveSpaces: (input: string) => string;
+    markdownEscapeBlock: (input: string) => string;
+    getSelectorLastPart: (selector: string) => string;
+    standardizeTags: (data: CreateIssueDetailsTextData) => string[];
+};
+
+export const IssueFilingUrlStringUtils: StringUtils = {
```
Shobhit1: Do we really need a object with a type that just contains 5 functions? I was thinking that these 5 functions as separate functions in `issue-filling-url-string-utils.ts` file should be good enough. Thoughts?

smoralesd: I see your point. The functions are stand alone from the execution point of view (i.e. they don't share any kind of state, that's why they were static functions). But, from the usage point of view, we're using this functions pretty close together. If you look at `src\bug-filing\github\create-github-bug-filing-url.ts` you'll see I'm creating a function by passing an instance of `StringUtils`, which looks more convenient than passing 5 separated (but usage-related functions).

src/bug-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -4,36 +4,35 @@ import { EnvironmentInfo } from '../../common/environment-info-provider';
 import { title } from '../../content/strings/application';
 import { CreateIssueDetailsTextData } from './../../common/types/create-issue-details-text-data';
 
-export class IssueFilingUrlStringUtils {
-    public static footer(environmentInfo: EnvironmentInfo): string {
+export type StringUtils = {
```
waabid: StringUtils seems too generic a type for this concern. I think something like IssueUrlCreationUtils or something _else_ that indicates the usage of this class is for this particular context would be nice.

smoralesd: Update!

src/tests/unit/tests/bug-filing/github/__snapshots__/create-github-bug-filing-url.test.ts.snap
```diff
@@ -1,5 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`createGitHubBugFilingUrlTest createGitHubBugFilingUrl: no tag 1`] = `"test repo/new?title=RR-help%20(last%20part)&body=**Issue**%3A%20%60RR-help%60%20(%5B%60RR-rule-id%60%5D(RR-help-url))%0A%0A**Target%20application**%3A%20%5BpageTitle%3Cx%3E%5D(pageUrl)%0A%0A**Element%20path**%3A%20RR-selector%3Cx%3E%0A%0A**Snippet**%3A%0A%0A%20%20%20%20collapesd%0A%0A**How%20to%20fix**%3A%0A%0Aescaped%0A%0A**Environment**%3A%0Atest%20spec%0A%0A%3D%3D%3D%3D%0A%0Atest%20footer"`;
+exports[`createGitHubBugFilingUrlTest createGitHubBugFilingUrl: no tag 1`] = `"test repo/new?title=RR-help%20(last%20part)&body=**Issue**%3A%20%60RR-help%60%20(%5B%60RR-rule-id%60%5D(RR-help-url))%0A%0A**Target%20application**%3A%20%5BpageTitle%3Cx%3E%5D(pageUrl)%0A%0A**Element%20path**%3A%20RR-selector%3Cx%3E%0A%0A**Snippet**%3A%0A%0A%20%20%20%20collapsed%0A%0A**How%20to%20fix**%3A%0A%0Aescaped%0A%0A**Environment**%3A%0Atest%20spec%0A%0A%3D%3D%3D%3D%0A%0Atest%20footer"`;
```
haonliu: I'm confused: what's changed here?

smoralesd: `collapesd`: mispelled `collapsed`

src/bug-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -4,36 +4,35 @@ import { EnvironmentInfo } from '../../common/environment-info-provider';
...
-    public static footer(environmentInfo: EnvironmentInfo): string {
+export type IssueUrlCreationUtils = {
+    footer: (info: EnvironmentInfo) => string;
+    collapseConsecutiveSpaces: (input: string) => string;
+    markdownEscapeBlock: (input: string) => string;
+    getSelectorLastPart: (selector: string) => string;
+    standardizeTags: (data: CreateIssueDetailsTextData) => string[];
+};
+
+export const IssueFilingUrlStringUtils: IssueUrlCreationUtils = {
```
markreay: These are more about URL creation than strings. Perhaps `IssueCreationUrlUtils` would be a better name.

smoralesd: That's the name of the type... words scrambled but same words...

src/bug-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -4,36 +4,35 @@ import { EnvironmentInfo } from '../../common/environment-info-provider';
...
-
-    public static collapseConsecutiveSpaces(input: string): string {
-        return input.replace(/\s+/g, ' ');
-    }
-
-    public static markdownEscapeBlock(input: string): string {
-        return input
+    },
+    collapseConsecutiveSpaces: (input: string): string => input.replace(/\s+/g, ' '),
+    markdownEscapeBlock: (input: string): string =>
```
dbjorge: "escape" strongly implies that you would be sanitizing any characters in `input` that markdown would treat specially by replacing them with an escape sequence. This method would be better named something like `encloseInMarkdownCodeBlock` or `formatAsMarkdownCodeBlock`.

I know you didn't introduce this name in your PR, but misusing the term "escape" is so likely to result in a security vulnerability that I want to point it out anyway while I'm noticing it. We need to rename this and audit all of the use sites for any that are expecting it to actually result in the sanitization of untrusted input.

src/bug-filing/types/bug-filing-service.ts
```diff
@@ -6,6 +6,12 @@ import { ReactSFCWithDisplayName } from './../../common/react/named-sfc';
 import { CreateIssueDetailsTextData } from './../../common/types/create-issue-details-text-data';
 import { SettingsFormProps } from './settings-form-props';
 
+export type IssueFilingUrlProvider<TSettings> = (
```
dbjorge: Do we have a team standard about whether we prefer the `T` prefix for type parameters? It's surprising to me that we'd want to avoid `I` but keep `T`. I note that `BugFilingService<Settings>` on line 15 doesn't use `T`.

I couldn't care less whether we use `T` or not, personally, but it'd be nice to be consistent.

smoralesd: That's a good point. I don't think we have talk about it. I'll got with consistency at the PR level and I guess we should check the current state is right now for this names.

src/bug-filing/common/issue-filing-url-string-utils.ts
smoralesd: Rename!
We're applying this function to data from we get from axe (`failureSummary`). Not sure if we need to actually escape this or not. We're using this info to create a url to create the issue/bug against github/azure dev ops so basically we're transferring this risk to a 3rd party. If you still want this to be done, I'm fine doing it on a separated PR (as this one is getting old and I would really like to getting completed soon)

PR #543, opened 2019-04-12
Clean up unused variables/imports - Part 3

smoralesd:
#### Description of changes

Clean up unused variables/imports. General idea is finish the clean up and then enable `noUnusedParameters` & `noUnusedLocals`  options on tsconfig.json

#### Pull request checklist


src/DetailsView/components/details-view-right-panel.tsx
```diff
@@ -1,7 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import * as React from 'react';
```
Shobhit1: We dont need `react` for a `tsx` file? Just curious. 😄

smoralesd: You only need to import `react` if you're using `jsx` syntax (like `return <Component .../>` which is not the case on this file. I didn't want to change the file extension as this will affect more files with a change that is not strictly "remove unused vars and imports"

smoralesd: @Shobhit1 on second though... I'll try the change an see how it looks

smoralesd: Updated!

Shobhit1: Thanks for taking this up. I think its just easier to comprehend a file without `JSX` as a `ts` file than a `tsx`.

PR #545, opened 2019-04-13
[spike] adding dependency injection support for devtools scripts

devaradhanm:
#### Description of changes

#### Pull request checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4168,8 +4168,7 @@
                 "ansi-regex": {
                     "version": "2.1.1",
                     "bundled": true,
-                    "dev": true,
-                    "optional": true
+                    "dev": true
                 },
                 "aproba": {
                     "version": "1.2.0",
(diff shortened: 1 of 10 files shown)


devaradhanm:
@Shobhit1 , i agree. i wanted to see how it looks with this dep injection package & share it with the team. that's y i have marked this as spike & donot merge.

smoralesd:
I kind of agree with both of you... I'm partial to use aa dependency injection container if this let us simplify our initializeres and factories, specially if we can add proper unit tests to the container. Then again, we have live a long time without an dependency container and it seems we have achieve a certain discipline about where to use `new` to create new instances. I'm still open to use a container though.

devaradhanm:
@smoralesd & @Shobhit1 , i am thinking of talking about this in our next engineering meeting.

dbjorge:
I appreciate that you've done the spike to try this out; I like trying out new ideas to see if we would like them on principle, and I think this would be a good engineering discussion item.

That said, looking at these changes, my opinion on this particular DI implementation is the same as my usual opinions on DI frameworks in general:

* If you assume infamiliarity with the DI patter/framework in use (ie, my expectation for most external contributors), I think it makes it *dramatically* harder to understand how the code works
* If you assume familiarity with the DI pattern/framework in use (ie, my expectation for ourselves after a few weeks, if we were to adopt it), I think it still makes it somewhat harder to follow the flow of code

Secondarily, I note that this PR adds more code than it removes, but I think making it easy to understand is so important that I would be unwilling to sacrifice it even if it made the code substantially more concise. I would be strongly opposed to integrating a DI framework like this on that basis.

PR #542, opened 2019-04-12
backgroud support for issue filing

haonliu:
#### Description of changes

backgroud support for issue filing

#### Pull request checklist

- [x] Added relevant unit test for your changes. (`npm run test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] Ran precheckin (`npm run precheckin`)

PR #546, opened 2019-04-15
replace issue filing button in details pane (under feature flag)

haonliu:
#### Description of changes

As discussed offline with Deva/Mark/Fer/Chelsea on Friday, opening issue filing settings dialog from target page might get cut from the feature. Whether the UX of that part changes, open dialog from issues details pane will not change. This PR wires up the components we built earlier to enable file bug from issues details pane. The dialog state is local, since we may not need to store it in background. If we still want to support open issue filing dialog from target page, I have another PR (https://github.com/Microsoft/accessibility-insights-web/pull/542) that implements the background support.

replace issue filing button in details pane (under feature flag)
image
issue

#### Pull request checklist


src/DetailsView/components/Issues-details-pane.tsx
```diff
@@ -22,7 +24,7 @@ export type IssuesDetailsPaneDeps = ToastDeps &
     FileIssueDetailsButtonDeps & {
         issueDetailsTextGenerator: IssueDetailsTextGenerator;
         detailsViewActionMessageCreator: DetailsViewActionMessageCreator;
-    };
+    } & IssueFilingButtonDeps;
```
smoralesd: can you move so we have a group of "dendency" deps and then the "direct" deps?

src/DetailsView/components/Issues-details-pane.tsx
```diff
@@ -68,14 +71,29 @@ export class IssuesDetailsPane extends React.Component<IssuesDetailsPaneProps, I
     }
 
     private getFileIssueDetailsButton(issueData: CreateIssueDetailsTextData): JSX.Element {
-        return (
+        const oldButton: JSX.Element = (
```
smoralesd: not thrill with the names but I don't have any better suggestion. Maybe `oldExperienceButton` and `newExperienceButton`?

src/bug-filing/github/create-github-bug-filing-url.ts
```diff
@@ -53,6 +54,7 @@ export function createGitHubBugFilingUrl(
 
     const title = buildTitle(bugData);
     const body = buildGithubText(bugData);
+    const repository = IssueFilingUrlStringUtils.appendSuffixToUrl(settingsData.repository, 'issues');
```
smoralesd: Should way for the `IssueFilingUrlStringUtils` refactor before completing this PR. We should not use objects as globals, this should be pass as a dependency.

src/bug-filing/github/create-github-bug-filing-url.ts
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { endsWith } from 'lodash';
```
smoralesd: this is not being used.

src/DetailsView/components/action-and-cancel-buttons-component.tsx
```diff
@@ -26,6 +27,7 @@ export class ActionAndCancelButtonsComponent extends React.Component<ActionAndCa
                         onClick={this.props.primaryButtonOnClick}
                         disabled={this.props.primaryButtonDisabled}
                         href={this.props.primaryButtonHref}
+                        target={isEmpty(this.props.primaryButtonHref) ? '_self' : '_blank'}
```
smoralesd: I'm not sure we do we need this `target` here. Can you clarify? (I'm not saying we don't need it, I just don't get the why)

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
```diff
@@ -66,23 +66,18 @@ describe('IssueFilingDialog', () => {
...
-            .verifiable();
     });
 
-    it('render open', () => {
-        const testSubject = shallow(<IssueFilingDialog {...props} />);
+    it.each([true, false])('isSettingsValid: %s', isSettingsValid => {
+        isSettingsValidMock
+            .setup(isValid => isValid(selectedServiceData))
+            .returns(() => isSettingsValid)
+            .verifiable();
```
smoralesd: I don't think you need to verify this call as you are returning a boolean that definitely change what's in the snapshot

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
```diff
@@ -66,23 +66,18 @@ describe('IssueFilingDialog', () => {
...
 
-        isSettingsValidMock.verifyAll();
-        createBugFilingUrlMock.verifyAll();
-        expect(testSubject.getElement()).toMatchSnapshot();
-    });
+        createBugFilingUrlMock
+            .setup(createBugFilingUrl => createBugFilingUrl(selectedServiceData, selectedBugDataStub, envInfo))
```
smoralesd: should this return a value? I think it needs to and then you probably don't need the verifiable as the result is being use as the href for `ActionAndCancelButtonsComponent`. If this props shows up on the snapshot you can drop the verify; if it doesn't shows up then you probably want to keep it.

src/DetailsView/components/action-and-cancel-buttons-component.tsx
haonliu: If we have a href on the primary button, it means we are using the button as a link, which should open in a new tab. Having target=_blank takes care of that.

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
```diff
@@ -91,12 +86,15 @@ describe('IssueFilingDialog', () => {
...
+        isSettingsValidMock
+            .setup(isSettingsValid => isSettingsValid(selectedServiceData))
+            .returns(() => true)
+            .verifiable();
         createBugFilingUrlMock
             .setup(createBugFilingUrl => createBugFilingUrl(selectedServiceData, selectedBugDataStub, envInfo))
             .verifiable();
         telemetryCallbackMock.setup(telemetryCallback => telemetryCallback(eventStub)).verifiable();
-        onCloseMock.setup(onClose => onClose(null)).verifiable();
+        onCloseMock.setup(onClose => onClose(It.isAny())).verifiable(Times.exactly(2));
```
smoralesd: why use `It.isAny()`? I think we are explicitly passing `null` on line 101, right?

src/tests/unit/tests/DetailsView/components/issues-details-pane.test.tsx
```diff
@@ -3,6 +3,9 @@
 import { shallow } from 'enzyme';
 import * as React from 'react';
 
+import { FlaggedComponent } from '../../../../../common/components/flagged-component';
+import { FeatureFlags } from '../../../../../common/feature-flags';
```
smoralesd: I don't think `FeatureFlags` is actually being used in this file.

src/tests/unit/tests/DetailsView/components/issues-table.test.tsx
```diff
@@ -164,6 +166,9 @@ describe('IssuesTableTest', () => {
                 issueDetailsTextGenerator: null,
                 windowUtils: null,
                 bugActionMessageCreator: null,
+                environmentInfoProvider: null,
```
smoralesd: I guess we need this because of type changes. None of this props show up on the snapshots though. It would be interesting to check if we need to make sure they show up and if we don't need that maybe find a way to not add this (as in, avoid adding new props -because type changes- that don't show up on the snapshots, making coder live easier in the future)

src/tests/unit/tests/bug-filing/bug-filing-service-provider.test.ts
```diff
@@ -36,4 +36,22 @@ describe('BugFilingServiceProviderTest', () => {
 
         expect(new BugFilingServiceProvider(givenTestOptions).allVisible()).toEqual(expectedTestOptions);
     });
+
+    test('forKey', () => {
```
smoralesd: are we missing a test for the "key not found" case?

src/tests/unit/tests/DetailsView/details-view-container.test.tsx
```diff
@@ -271,6 +271,7 @@ describe('DetailsViewContainer', () => {
                 issueTrackerPath={issueTrackerPath}
                 rightPanelConfiguration={rightPanelConfiguration}
                 switcherNavConfiguration={switcherNavConfiguration}
+                userConfigurationStoreData={storeMocks.userConfigurationStoreData}
```
smoralesd: I'd seen this change on a lot of files... any thoughts on how to avoid this multiple-files change when we need to add store data to the props of a leaf component?

src/tests/unit/tests/bug-filing/github/github-bug-filing-service.test.tsx
```diff
@@ -74,6 +74,13 @@ describe('GithubBugFilingServiceTest', () => {
         expect(wrapper.getElement()).toMatchSnapshot();
     });
 
+    it('renderSettingsForm: no valid settings object', () => {
```
smoralesd: base on the name... and looking at the tested file... should we use `isSettingsValid` instead of `isEmpty`?
file: src\bug-filing\github\github-bug-filing-service.tsx
line: 9

src/tests/unit/tests/DetailsView/components/issues-table.test.tsx
haonliu: I think this is a good point. I think we should not have to touch this file when we change the deps on a child component. I think it is good to cast deps to 'IssuesTableDeps' and leave the deps that is not directly used in this component.

src/tests/unit/tests/DetailsView/details-view-container.test.tsx
haonliu: Yeah, this is painful for me, but currently I cannot think of a solution other than use Context.

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+                };
+            })
+            .verifiable();
+        bugFilingServiceProviderMock
+            .setup(bp => bp.forKey(testKey))
+            .returns(() => testBugService)
+            .verifiable();
+    });
+
+    test.each([true, false])('render: isSettingsValid: %s', isSettingsValid => {
```
smoralesd: can you rename this to something like `render with isSettingsValid`? the name makes me thing that there is an `isSettingsValid` function that render something, which is misleading.

src/tests/unit/tests/bug-filing/github/github-bug-filing-service.test.tsx
haonliu: line 9? I think you are talking about line 43 right? 
If so, no, we should not. Reason is, if a user types say 5 space in the text field, although this is an invalid setting, we should let the text field display 5 spaces, instead of clear the text field.

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+        environmentInfoProviderMock
+            .setup(envp => envp.getEnvironmentInfo())
+            .returns(() => {
+                return {
+                    extensionVersion: '1',
+                    axeCoreVersion: '2',
+                    browserSpec: 'spec',
+                };
+            })
+            .verifiable();
```
smoralesd: Usually, you don't need `verifiable` when the mock has been setup to return a value. This is specially through if that return value heavily influence the end result your asserting for. 
Are this values (versions and spec) being assert by another mock expectations? If that's true, you probably don't need the verifiable!

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+        bugFilingServiceProviderMock
+            .setup(bp => bp.forKey(testKey))
+            .returns(() => testBugService)
+            .verifiable();
+    });
+
+    test.each([true, false])('render: isSettingsValid: %s', isSettingsValid => {
+        testBugService.isSettingsValid = () => isSettingsValid;
```
smoralesd: Why are we not using a mock for `testBugService`?

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+        const wrapper = shallow(<IssueFilingButton {...props} />);
+        expect(wrapper.debug()).toMatchSnapshot();
+
+        environmentInfoProviderMock.verifyAll();
+        bugFilingServiceProviderMock.verifyAll();
+    });
+
+    test('onclick: valid settings, file bug', () => {
+        bugActionMessageCreatorMock
+            .setup(messageCreator => messageCreator.trackFileIssueClick(It.isAny(), testKey as any))
```
smoralesd: can you make the assertion of the first parameter more robust? `It.isAny()` doesn't let you validate much and we already have a `EventStubFactory` ready to use on our tests.

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+                pageUrl: 'http://pageUrl',
+                ruleResult: null,
+            },
+            userConfigurationStoreData: userConfigurationStoreData,
+        };
+        const wrapper = shallow(<IssueFilingButton {...props} />);
+
+        wrapper.find(DefaultButton).simulate('click');
+
+        environmentInfoProviderMock.verifyAll();
```
smoralesd: Why are we verifying `environmentInfoProviderMock` and `bugFilingServiceProviderMock`? the function being tested here (`trackFileIssueClick`) doesn't directly use any of this deps. Are we over verifying things in this test? Things that we are also verifying on a different test - one with a better context about the usage of the 2 mentioned mocks?

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
haonliu: I personally prefer not using mock for this kind of object.

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
haonliu: we are using simulate('click') to test on click, which we cannot pass an event stub.

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
haonliu: right, it is not necessary.

src/bug-filing/github/create-github-bug-filing-url.ts
haonliu: Sure.

src/DetailsView/components/Issues-details-pane.tsx
```diff
@@ -19,7 +21,8 @@ import { DetailsViewActionMessageCreator } from '../actions/details-view-action-
 import { GuidanceLinks } from './guidance-links';
 
 export type IssuesDetailsPaneDeps = ToastDeps &
-    FileIssueDetailsButtonDeps & {
+    FileIssueDetailsButtonDeps &
```
Shobhit1: I think there is this style that we follow where we mention the object before and then the `&` dependencies. I don't have a preference but It would be better if we are consistent in our style on that. 
For example, this `IssuesDetailsPaneDeps` would become something like this:

```
export type IssuesDetailsPaneDeps = {
        issueDetailsTextGenerator: IssueDetailsTextGenerator;
        detailsViewActionMessageCreator: DetailsViewActionMessageCreator;
} & ToastDeps  & FileIssueDetailsButtonDeps  & IssueFilingButtonDeps 
```

src/DetailsView/components/action-and-cancel-buttons-component.tsx
Shobhit1: In normal scenarios: `target = _blank` makes the link open the `href` in a new tab. Is that the same behavior here? I am a little confused.

src/DetailsView/details-view-initializer.ts
```diff
@@ -275,6 +277,12 @@ if (isNaN(tabId) === false) {
                     urlParser,
                     dateProvider: DateProvider.getDate,
                     settingsProvider: SettingsProviderImpl,
+                    environmentInfoProvider: new EnvironmentInfoProvider(
```
Shobhit1: nit: Is it better to create this object in a new line and use that variable/object here?

src/tests/unit/tests/bug-filing/common/issue-filing-url-string-utils.test.ts
```diff
@@ -26,6 +26,12 @@ describe('BugFilingUrlStringUtilsTest', () => {
         expect(IssueFilingUrlStringUtils.markdownEscapeBlock('hello\nworld')).toEqual('    hello\n    world');
     });
 
+    test('appendSuffixToUrl', () => {
+        expect(IssueFilingUrlStringUtils.appendSuffixToUrl('repo', 'hello')).toEqual('repo/hello/');
+        expect(IssueFilingUrlStringUtils.appendSuffixToUrl('repo/hello', 'hello')).toEqual('repo/hello');
+        expect(IssueFilingUrlStringUtils.appendSuffixToUrl('repo/hello/', 'hello')).toEqual('repo/hello/');
```
Shobhit1: Is there a scenario where you `append` something different from the same word to the end of url? Maybe add an assertion for that case as well, even though its trivial!

src/tests/unit/tests/bug-filing/github/github-bug-filing-service.test.tsx
```diff
@@ -74,6 +74,13 @@ describe('GithubBugFilingServiceTest', () => {
...
+        const Component = GitHubBugFilingService.settingsForm;
+        props.settings = null;
+        const wrapper = shallow(<Component {...props} />);
+        expect(wrapper.find(TextField).props().value).toEqual('');
```
Shobhit1:
```suggestion
        const textValue = wrapper.find(TextField).props().value
        expect(textValue).toEqual('');
```

Shobhit1: It might also be better to assert that you found that element from `wrapper`.
For example:

```
    const textField = wrapper.find(TextField);
    expect(textField).toBeDefined();
```

src/DetailsView/components/action-and-cancel-buttons-component.tsx
haonliu: Yes, it is the same behavior here.

smoralesd: So, what's the scenario where we use `'_self'`???

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
smoralesd: Resolved without change. Can you clarify?

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
smoralesd: So.... does the returned value make a difference on the snapshot?

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
```diff
@@ -66,42 +66,59 @@ describe('IssueFilingDialog', () => {
...
+            .returns(() => true)
+            .verifiable(Times.once());
         createBugFilingUrlMock
             .setup(createBugFilingUrl => createBugFilingUrl(selectedServiceData, selectedBugDataStub, envInfo))
-            .verifiable();
-        telemetryCallbackMock.setup(telemetryCallback => telemetryCallback(eventStub)).verifiable();
-        onCloseMock.setup(onClose => onClose(null)).verifiable();
+            .verifiable(Times.once());
+        telemetryCallbackMock.setup(telemetryCallback => telemetryCallback(eventStub)).verifiable(Times.once());
+        onCloseMock.setup(onClose => onClose(It.isAny())).verifiable(Times.once());
```
smoralesd: why use It.isAny()? I think we are explicitly passing null on line 101, right?

src/tests/unit/tests/bug-filing/bug-filing-service-provider.test.ts
```diff
@@ -36,4 +36,40 @@ describe('BugFilingServiceProviderTest', () => {
...
+                isHidden: true,
+            } as BugFilingService,
+        ];
+
+        const expectedTestOption = givenTestOptions[0];
+
+        expect(new BugFilingServiceProvider(givenTestOptions).forKey('github')).toEqual(expectedTestOption);
+    });
+
+    test('forKey', () => {
```
smoralesd: same name on 2 different tests

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
haonliu: Sorry, yes the url appears in the snapshot.

src/tests/unit/tests/DetailsView/components/issue-filing-dialog.test.tsx
haonliu: yes, it makes 'isHidden' different in snapshot.

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { IssueFilingButton, IssueFilingButtonDeps, IssueFilingButtonProps } from '../../../../../common/components/issue-filing-button';
+import { EnvironmentInfoProvider } from '../../../../../common/environment-info-provider';
+import { BugActionMessageCreator } from '../../../../../common/message-creators/bug-action-message-creator';
+import { NamedSFC } from '../../../../../common/react/named-sfc';
+import { UserConfigurationStoreData } from '../../../../../common/types/store-data/user-configuration-store';
+import { EventStubFactory } from '../../../common/event-stub-factory';
+
+describe('IssueFilingButtonTest', () => {
+    const testKey: string = 'test';
+    const eventStubFactory = new EventStubFactory();
```
smoralesd: `eventStubFactory` doesn't seems to be in used on this file

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
smoralesd: So are we still 'verifying` this? why?

haonliu: I replied to this comment, but now it disappears!
So I think what you said make sense. In this particular case, we are not verifying the return value on other mocks, so I kept verifiable here.

devaradhanm:
>     public static footer(environmentInfo: EnvironmentInfo): string {

getFooterContent

---
Refers to: src/bug-filing/common/issue-filing-url-string-utils.ts:10 in c1b6aa5. [](commit_id = c1b6aa5c9b09388e681c475955bf654dab87645a, deletion_comment = False)

devaradhanm:
>         }

wat about iframe?

---
Refers to: src/bug-filing/common/issue-filing-url-string-utils.ts:34 in c1b6aa5. [](commit_id = c1b6aa5c9b09388e681c475955bf654dab87645a, deletion_comment = False)

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
smoralesd: We can. Look for examples on our own code.

src/tests/unit/tests/bug-filing/github/github-bug-filing-service.test.tsx
smoralesd: You're correct!

src/bug-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -36,4 +38,12 @@ export class IssueFilingUrlStringUtils {
     public static standardizeTags(data: CreateIssueDetailsTextData): string[] {
         return data.ruleResult.guidanceLinks.map(tag => tag.text.toUpperCase());
     }
+
+    public static appendSuffixToUrl(url: string, suffix: string): string {
+        if (endsWith(url, suffix) || endsWith(url, `${suffix}/`)) {
+            return url;
+        }
```
devaradhanm: can you see this?

src/tests/unit/tests/common/components/issue-filing-button.test.tsx
haonliu: Done.

src/DetailsView/components/action-and-cancel-buttons-component.tsx
haonliu: '_self' is the default option, it's the same as not have a target attr.

PR #550, opened 2019-04-15
Introduce Azure Boards Settings Form

smoralesd:
#### Description of changes

New component for Azure Boards settings.
Not yet wired up.

#### Pull request checklist

- [x] Addresses an existing issue: Related WI #1505514


src/bug-filing/azure-boards/azure-boards-bug-filing-service.tsx
```diff
@@ -1,52 +1,52 @@
...
+import { IDropdownOption } from 'office-ui-fabric-react/lib/Dropdown';
 import { EnvironmentInfo } from '../../common/environment-info-provider';
-import { NamedSFC } from '../../common/react/named-sfc';
 import { CreateIssueDetailsTextData } from '../../common/types/create-issue-details-text-data';
 import { BugServicePropertiesMap } from '../../common/types/store-data/user-configuration-store';
 import { BugFilingService } from '../types/bug-filing-service';
-import { SettingsFormProps } from '../types/settings-form-props';
+import { AzureBoardsSettingsForm } from './azure-boards-settings-form';
+
+const AzureDevOpsBugFilingServiceKey = 'azureBoards';
```
peterdur: Missed rename here
```suggestion
const AzureBoardsBugFilingServiceKey = 'azureBoards';
```

src/bug-filing/azure-boards/azure-boards-bug-filing-service.tsx
```diff
@@ -1,52 +1,52 @@
...
     return {
         projectURL,
-        issueDetailsLocationField,
+        issueDetailsField,
     };
 }
 
-function getSettingsFromStoreData(bugServicePropertiesMap: BugServicePropertiesMap): AzureDevOpsBugFilingSettings {
-    return bugServicePropertiesMap[AzureDevOpsBugFilingServiceKey] as AzureDevOpsBugFilingSettings;
+function getSettingsFromStoreData(bugServicePropertiesMap: BugServicePropertiesMap): AzureBoardsBugFilingSettings {
```
peterdur: Comparing to getSettingsFromStoreData in github-bug-filing-service.tsx, two thoughts:
1. The parameter type in that function is not correct (it's `GitHubBugFilingSettings`, should be `BugServicePropertiesMap`)
2. Except for return type, it's the same function - does every service need to implement this?

src/bug-filing/azure-boards/azure-boards-bug-filing-service.tsx
```diff
@@ -1,52 +1,52 @@
...
+}
+
+export type AzureBoardsBugFilingSettings = {
     projectURL: string;
-    issueDetailsLocationField: string;
+    issueDetailsField: AzureBoardsIssueDetailField;
 };
 
-function buildStoreData(projectURL: string, issueDetailsLocationField: string): AzureDevOpsBugFilingSettings {
+function buildStoreData(projectURL: string, issueDetailsField: AzureBoardsIssueDetailField): AzureBoardsBugFilingSettings {
```
peterdur: `buildStoreData` seems to only be used in tests, for now at least... curious who is going to call it generically while knowing the right order for parameters?

src/bug-filing/azure-boards/azure-boards-bug-filing-service.tsx
haonliu: For 2: yes, every service need to implement the same function. We talked about having a builder to put the same function on every service object, but feel like it is a bit over engineering for now.

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { NamedSFC } from '../../common/react/named-sfc';
+import { SettingsFormProps } from '../types/settings-form-props';
+import {
+    AzureBoardsBugFilingService,
+    AzureBoardsBugFilingSettings,
+    AzureBoardsIssueDetailField,
+    AzureBoardsIssueDetailLocationDropdownOption,
+} from './azure-boards-bug-filing-service';
+
+export const AzureBoardsSettingsForm = NamedSFC<SettingsFormProps<AzureBoardsBugFilingSettings>>('AzureDevOpsSettingsForm', props => {
```
peterdur: Missed rename
```suggestion
export const AzureBoardsSettingsForm = NamedSFC<SettingsFormProps<AzureBoardsBugFilingSettings>>('AzureBoardsSettingsForm', props => {
```

src/bug-filing/azure-boards/azure-boards-bug-filing-service.tsx
smoralesd: We've been building this piece by piece. Wiring up stories should use this in the near future. If we get to a point when we don't need this, we'll clean it up.

src/bug-filing/azure-boards/azure-boards-bug-filing-service.tsx
smoralesd: Updated

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
```diff
@@ -0,0 +1,58 @@
...
+            propertyName,
+            newValue.key as AzureBoardsIssueDetailField,
+        );
+    };
+
+    return (
+        <>
+            <TextField
+                className="issue-setting"
+                label="Enter the desired Azure DevOps project link:"
```
peterdur: Should be Boards
```suggestion
                label="Enter the desired Azure Boards project link:"
```

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
```diff
@@ -0,0 +1,58 @@
...
+                value={props.settings ? props.settings.projectURL : null}
+                placeholder="https://dev.azure.com/org/project"
+                onChange={onProjectURLChange}
+            />
+            <Dropdown
+                options={options}
+                label="File issue details in:"
+                onChange={onIssueDetailLocationChange}
+                selectedKey={props.settings ? props.settings.issueDetailsField : null}
```
peterdur: It looks like we're forcing the user to pick a field. Does it make sense to have a default?

src/bug-filing/azure-boards/azure-boards-bug-filing-service.tsx
smoralesd: As a middle ground solution: extracted the "duplicated" code to a creational function. As Haonan mentioned, we discussed several options to facilitate the creation of new issue filing services but decide against it at this point as there is a lot of refactoring going on which make it a tad hard to visualize the end result. And as with other pieces of this feature, we'll introduce some better way to handle this if we needed as part of our clean up work at the end of the feature.

src/tests/unit/tests/bug-filing/azure-devops/__snapshots__/azure-boards-bug-filing-service.test.tsx.snap
```diff
@@ -0,0 +1,30 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
peterdur: Need to rename folder from `\src\tests\unit\tests\bug-filing\azure-devops\` to `azure-boards`

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
smoralesd: Updated

src/tests/unit/tests/bug-filing/azure-devops/__snapshots__/azure-boards-bug-filing-service.test.tsx.snap
smoralesd: updated

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
```diff
@@ -0,0 +1,58 @@
...
+            key: 'reproSteps',
+            text: 'Repro steps (default)',
+        },
+        {
+            key: 'description',
+            text: 'Description',
+        },
+    ];
+
+    const onProjectURLChange = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
```
peterdur: Per bug 1512520 we'll likely want this behavior to be different between the settings panel and the "Specify issue filing location" dialog. It would be great if we could hide that difference from bug filing services and forms.

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
smoralesd: Make sense. Updated.

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
smoralesd: Not taking care of that bug on this PR... sounds like something we want to take care of as part of wiring up.

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
```diff
@@ -0,0 +1,58 @@
...
+        const propertyName: keyof AzureBoardsBugFilingSettings = 'projectURL';
+        props.deps.userConfigMessageCreator.setBugServiceProperty(AzureBoardsBugFilingService.key, propertyName, newValue);
+    };
+
+    const onIssueDetailLocationChange = (event: React.FormEvent<HTMLElement>, newValue?: IDropdownOption) => {
```
peterdur: `newValue` is declared as optional here but we assume it's valid below. Should a check be added?

peterdur: Can we declare the second parameter as `AzureBoardsIssueDetailLocationDropdownOption` instead of `IDropdownOption`?

src/tests/unit/tests/bug-filing/azure-boards/azure-boards-bug-filing-service.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    beforeEach(() => {
+        projectStub = 'some project';
+        issueDetailsLocationStub = 'some location' as AzureBoardsIssueDetailField;
+        userConfigMessageCreatorMock = Mock.ofType(UserConfigMessageCreator);
+        props = {
+            deps: {
+                userConfigMessageCreator: userConfigMessageCreatorMock.object,
+            },
+            settings: {
+                projectURL: 'some project',
```
peterdur: Should we use the stub above to prevent divergence here? (Similarly for `issuesDetailsField` on next line?)
```suggestion
                projectURL: projectStub,
```

peterdur:
Does it make sense to add `AzureBoardsBugFilingService` to `BugFilingServiceProviderImpl` at this point? This enables the form to be displayed and interacted with (even if there's no `issueFilingUrlProvider` yet, but that's OK since we're behind a FF, right?)

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
```diff
@@ -0,0 +1,59 @@
...
+            AzureBoardsBugFilingService.key,
+            propertyName,
+            newValue.key as AzureBoardsIssueDetailField,
+        );
+    };
+
+    return (
+        <>
+            <TextField
+                className="issue-setting"
```
peterdur: This formatting attached to this class is currently under `.settings-panel` so that will need to be tweaked. Figma has more space above both the TextField and the Dropdown.

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
```diff
@@ -0,0 +1,59 @@
...
+                label="Enter the desired Azure Boards project link:"
+                value={props.settings ? props.settings.projectURL : null}
```
peterdur: `null` isn't valid here - results in a console error (below).
```
Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
    in input (created by TextFieldBase)
    in div (created by TextFieldBase)
    in div (created by TextFieldBase)
    in div (created by TextFieldBase)
    in TextFieldBase (created by Context.Consumer)
    in StyledTextFieldBase (created by AzureBoardsSettingsForm)
    in AzureBoardsSettingsForm (created by BugFilingSettingsContainer)
    in BugFilingSettingsContainer (created by IssueFilingDialog)
```

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
smoralesd: That sound as fit and finish

smoralesd:
> Does it make sense to add `AzureBoardsBugFilingService` to `BugFilingServiceProviderImpl` at this point? This enables the form to be displayed and interacted with (even if there's no `issueFilingUrlProvider` yet, but that's OK since we're behind a FF, right?)

This is behind a feature flag. Adding it to the provider make sense so we can actually see what we're missing UX-wise

src/tests/unit/tests/bug-filing/azure-boards/azure-boards-bug-filing-service.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+        };
+        const givenData: BugServicePropertiesMap = {
+            'some other service': {},
+            [AzureBoardsBugFilingService.key]: expectedStoreData,
+        };
+        expect(AzureBoardsBugFilingService.getSettingsFromStoreData(givenData)).toEqual(expectedStoreData);
+    });
+
+    describe('check for invalid settings', () => {
+        it.each(invalidTestSettings)('with %o', settings => {
```
peterdur: I like keeping the list of test cases close to the `each` so it's easy to read through the test for each case without scrolling.

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
smoralesd: Remove optionality. This should not happen as we're the ones setting the options and the options are not configurable by the user or outside the code.

smoralesd: I guess you're concern is related to the cast `newValue.key as AzureBoardsIssueDetailField` a couple of lines below. That I understand but the signature of this function is tide to the `Dropdown` component `onChange` definition, which does require the `newValue` to be of type `IDropdownOption`.

src/tests/unit/tests/bug-filing/azure-boards/azure-boards-bug-filing-service.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+            issueDetailsField: issueDetailsLocationStub,
+        };
+        const givenData: BugServicePropertiesMap = {
+            'some other service': {},
+            [AzureBoardsBugFilingService.key]: expectedStoreData,
+        };
+        expect(AzureBoardsBugFilingService.getSettingsFromStoreData(givenData)).toEqual(expectedStoreData);
+    });
+
+    describe('check for invalid settings', () => {
```
peterdur: Dan's comment yesterday makes me think - would it be a little tidier to have a `describe('isSettingsValid')` around the `it.each` for invalid settings and an `it` for valid settings?

src/tests/unit/tests/bug-filing/azure-boards/azure-boards-bug-filing-service.test.tsx
smoralesd: Not sure if I follow but I update to use one `describe` to group all the `isSettingsValid` test cases (valid and invalid settings)

src/tests/unit/tests/bug-filing/azure-boards/azure-boards-bug-filing-service.test.tsx
smoralesd: Updated

src/tests/unit/tests/bug-filing/azure-boards/azure-boards-bug-filing-service.test.tsx
smoralesd: Done

smoralesd: Consider I didn't touch this file much. I only updated on test. Github shows the file is new because I change the containing folder name. So all this comments are basically not part of the original PR.

src/bug-filing/azure-boards/azure-boards-settings-form.tsx
smoralesd: Where are you getting this?

peterdur: I tried adding `AzureBoardsBugFilingService` to `BugFilingServiceProviderImpl` so I could see the UX. Then I cleaned settings, did a FastPass with failures, clicked File Issue in the failure details pane, selected Azure Boards, and saw this error on the console.

smoralesd: Sure... it's fixed now.

PR #553, opened 2019-04-16
[Refactor] BugActionMessageCreator using ActionMessageDispatcher

smoralesd:
#### Description of changes

 Simple refactor to update BugActionMessageCreator to use ActionMessageDispatcher (instead of inherit from BaseActionMessageCreator)

#### Pull request checklist


src/common/message-creators/bug-action-message-creator.ts
```diff
@@ -1,37 +1,32 @@
...
 import { FILE_ISSUE_CLICK, FileIssueClickService, TelemetryEventSource } from '../telemetry-events';
-import { BaseActionMessageCreator } from './base-action-message-creator';
+import { ActionMessageDispatcher } from './action-message-dispatcher';
 
-export class BugActionMessageCreator extends BaseActionMessageCreator {
+export class BugActionMessageCreator {
     constructor(
-        postMessage: (message: Message) => void,
-        tabId: number,
+        private readonly dispatcher: ActionMessageDispatcher,
```
peterdur: Not blocking: we could make the other constructor parameters here `readonly` too, yes?

src/tests/unit/tests/common/message-creators/bug-action-message-creator.test.ts
```diff
@@ -1,77 +1,70 @@
...
+            dispatcher.dispatchMessage({
+                messageType: Messages.SettingsPanel.OpenPanel,
+                payload: {
+                    telemetry,
+                },
+            }),
+        );
 
-        testSubject.openSettingsPanel(event);
+        testSubject.openSettingsPanel(eventStub);
```
peterdur: If I comment out this line, the test still passes. Need a `.verifiable(Times.once())` on the mock setup.

peterdur: (The `telemetryFactoryMock` is also not verified, but I can see that as an implementation detail of the test and not part of the contract being examined.)

src/tests/unit/tests/common/message-creators/bug-action-message-creator.test.ts
```diff
@@ -1,77 +1,70 @@
...
-            payload,
-        };
+    it('dispatch message for trackFileIssueClick', () => {
+        const testService: FileIssueClickService = 'test file issue service' as FileIssueClickService;
+        const telemetry = { ...telemetryStub, service: testService };
+
+        telemetryFactoryMock.setup(factory => factory.forFileIssueClick(eventStub, source, testService)).returns(() => telemetry);
 
-        postMessageMock.setup(pm => pm(It.isValue(expectedMessage))).verifiable();
+        dispatcherMock.setup(dispatcher => dispatcher.sendTelemetry(FILE_ISSUE_CLICK, telemetry));
```
peterdur: (Same note here as on line 53 above)

src/common/message-creators/bug-action-message-creator.ts
smoralesd: Make sense. Updated.

src/tests/unit/tests/common/message-creators/bug-action-message-creator.test.ts
smoralesd: Good catch on the missing `verifiable`
As per the `telemetryFactoryMock`, in general, use of `verifiable` is not needed when the return value of the mock alter the result of the tested function. In this case, verifying the dispatcher is proof enough the telemetry factory was used indeed.

src/tests/unit/tests/common/message-creators/bug-action-message-creator.test.ts
smoralesd: Updated

Issue #547, opened 2019-04-15
Inspect HTML button should not be disabled


haonliu:
## Describe the bug

The Inspect HTML button in the details dialog on target page should not be disabled, for accessibility concerns.

**To Reproduce**
Open
Run automated checks.

1. Go to any web page with a11y issue.
2. Run automated checks.
3. Click the '!' icon in the target page.
4. See Inspect HTML button is disabled.

## Expected behavior

Inspect HTML button should not be disabled. User should be able to interact with it.

## Screenshots

![image](https://user-images.githubusercontent.com/15974344/56166473-91ef1080-5f8a-11e9-86f8-8ed62690ea74.png)

## Extension (please complete the following information)

- OS: Windows
- Version: staging 2.1.2
- Chrome Version [e.g 71.0]
- Target Page: any

## Are you willing to submit a PR?

yes

## Did you search for similar existing issues?
yes

Did you search for similar issues in our github issues or on [stackoverflow](https://stackoverflow.com/questions/tagged/accessibility-insights)
yes

Shobhit1:
@haonliu Isn't the `Inspect HTML` button supposed to be disabled till the time user manually opens the dev tools (`F12`) ? Can you please provide more context on this issue?

ferBonnin:
@Shobhit1 behaviour should be similar to the file bug button, the message and alert that you need to open dev tools appears after you interact with the button

Shobhit1:
> @Shobhit1 behaviour should be similar to the file bug button, the message and alert that you need to open dev tools appears after you interact with the button

Thanks @ferBonnin . I will take a look if I can take this one up today. 😺

Shobhit1:
Taking this one.

Shobhit1:
Created PR #859 for it.

Shobhit1:
@ferBonnin / @haonliu Can you please verify this and close the bug if its fixed?

ferBonnin:
Looks good to me :)

PR #561, opened 2019-04-17
wire up target page issue filing button ux

haonliu:
#### Description of changes

wire up target page issue filing button ux
issue2

#### Pull request checklist

- [ ] Work item 1512484


src/injected/components/details-dialog.tsx
```diff
@@ -162,12 +167,28 @@ export class DetailsDialog extends React.Component<DetailsDialogProps, DetailsDi
...
 
     private renderFileIssueButton(issueData: CreateIssueDetailsTextData): JSX.Element {
-        return (
+        const oldExperienceButton: JSX.Element = (
             <FileIssueDetailsButton
                 deps={this.props.deps}
                 issueDetailsData={issueData}
                 issueTrackerPath={this.state.issueTrackerPath}
-                restoreFocus={false}
+                restoreFocus={true}
```
dbjorge: >true [](start = 30, length = 4)

Is it intentional to change this as part of this PR?

haonliu: No, will fix.

src/injected/components/details-dialog.tsx
```diff
@@ -162,14 +167,30 @@ export class DetailsDialog extends React.Component<DetailsDialogProps, DetailsDi
...
             />
         );
+        const newExperienceButton: JSX.Element = this.state.userConfigurationStoreData ? (
+            <IssueFilingButton
+                deps={this.props.deps}
+                issueDetailsData={issueData}
+                userConfigurationStoreData={this.state.userConfigurationStoreData}
+                needsSettingsContentRenderer={IssueFilingNeedsSettingsHelpText}
+            />
+        ) : null;
```
dbjorge: >) : null; [](start = 8, length = 9)

This feels inconsistent with how the other buttons work; for the inspect, back, and next buttons, the dialog handler exposes a "isButtonTypeDisabled" that owns deciding whether to disable a button and the buttons themselves are passed a disabled prop so they can decide how to render a "disabled" vs "enabled" state, but here, the dialog's renderer takes on both of those 2 responsibilities.

Do the other buttons treat "disabled" as meaning "don't render at all" or "render as disabled"? I think it would be good to be consistent in both code layout and UX.

haonliu: Yeah, I also think that we should be consistent. We have created an issue (https://github.com/Microsoft/accessibility-insights-web/issues/547) to change the other buttons' UX to remove their disabled state.

haonliu: Updated!

src/common/components/issue-filing-button.tsx
```diff
@@ -38,7 +42,12 @@ export class IssueFilingButton extends React.Component<IssueFilingButtonProps, I
     }
 
     public render(): JSX.Element {
+        if (isEmpty(this.props.userConfigurationStoreData)) {
```
dbjorge: I think this latest change is an improvement (in that I think it makes more sense for this component to own knowledge of how to handle the data == null state), but I still think I would prefer to see the component render in a disabled or less-functional state, rather than not rendering anything at all; I think rendering null her is both inconsistent with the other buttons and prone to making the UI jittery during initial load.

haonliu: Oh, I misunderstood your request. I'm hesitate to make UX changes here without PM/Desginer's input, I'd rather make changes to make sure that the user config store state are loaded (the same way how we make sure other store data that the client view is depend on are loaded).

smoralesd:
I see no checks on the checklist... I think we do have a WI associated with this, right?

src/injected/client-view-controller.ts
```diff
@@ -27,10 +28,12 @@ export class ClientViewController {
     private visualizationStore: BaseStore<VisualizationStoreData>;
     private assessmentStore: BaseStore<AssessmentStoreData>;
     private tabStore: BaseStore<TabStoreData>;
+    private userConfigurationStore: BaseStore<UserConfigurationStoreData>;
     private scanResultStore: BaseStore<VisualizationScanResultData>;
     private currentScanResultState: VisualizationScanResultData;
     private currentAssessmentState: AssessmentStoreData;
     private currentTabState: TabStoreData;
+    private currentUserConfigStoreState: UserConfigurationStoreData;
```
smoralesd: How is this being used? I didn't find meaningful usages of this, just assign the state from the store and check for nullity

src/tests/unit/tests/injected/client-view-controller.test.ts
```diff
@@ -64,6 +68,10 @@ describe('ClientViewControllerTest', () => {
             .setup(sm => sm.addChangedListener(It.isAny()))
             .callback(listener => {})
             .verifiable();
+        userConfigStoreMock
+            .setup(sm => sm.addChangedListener(It.isAny()))
+            .callback(listener => {})
```
smoralesd: what's the point of adding this `callback` call in here?

src/tests/unit/tests/injected/client-view-controller.test.ts
```diff
@@ -64,6 +68,10 @@ describe('ClientViewControllerTest', () => {
...
+            .verifiable();
```
smoralesd: missing `Times.x`

src/common/components/issue-filing-button.tsx
```diff
@@ -39,6 +42,7 @@ export class IssueFilingButton extends React.Component<IssueFilingButtonProps, I
 
     public render(): JSX.Element {
         const { issueDetailsData, userConfigurationStoreData, deps } = this.props;
+
```
waabid: Extra line is unnecessary here.

src/tests/unit/tests/injected/components/details-dialog.test.tsx
```diff
@@ -4,26 +4,39 @@ import { shallow } from 'enzyme';
...
+type DetailsDialogTestCase = {
+    isDevToolsOpen: boolean;
+    shadowDialog: boolean;
+};
 
-    test('render: isDevToolsOpen = false, shadowDialog = true', () => {
-        testRender(true, false);
-    });
+describe('DetailsDialogTest', () => {
+    const testCases = [
```
smoralesd: should this use the `DetailsDialogTestCase ` type?

src/tests/unit/tests/injected/components/__snapshots__/details-dialog.test.tsx.snap
```diff
@@ -432,7 +642,115 @@ exports[`DetailsDialogTest render: isDevToolsOpen = false, shadowDialog = true 1
 </div>
 `;
 
-exports[`DetailsDialogTest render: isDevToolsOpen = false, shadowDialog = true 2`] = `
+exports[`DetailsDialogTest render: { isDevToolsOpen: false, shadowDialog: true }: new isssue filing bug UI 1`] = `
```
waabid: issue*

src/tests/unit/tests/injected/components/details-dialog.test.tsx
```diff
@@ -115,5 +131,14 @@ describe('DetailsDialogTest', () => {
...
         }
+
+        const userConfigStoreDataStub = {
+            bugService: 'service',
+            bugServicePropertiesMap: {},
+        };
+
+        wrapper.setState({ userConfigurationStoreData: userConfigStoreDataStub });
+
+        expect(wrapper.find(FlaggedComponent).getElement()).toMatchSnapshot('new isssue filing bug UI');
```
waabid: issue*

src/injected/client-view-controller.ts
haonliu: resolve as discussed offline

PR #562, opened 2019-04-17
[Refactor] Make typings on withStoreSubscription more strict

haonliu:
#### Description of changes

Make typings on withStoreSubscription more strict, so that we don't need to pass storeState = null from parent level.

#### Pull request checklist


src/tests/unit/tests/common/components/with-store-subscription.test.tsx
```diff
@@ -41,6 +42,7 @@ describe('withStoreSubscription', () => {
                 storeActionMessageCreator: storeActionCreatorMock.object,
                 storesHub: storesHubStub,
             },
+            storeState: null,
```
Shobhit1: I thought you made this change so that `storeState` cannot be `null`! I am a little confused with this.

haonliu: I made the change so that the storeState is not needed in the props after wrapped. the actual component to be wrapped (like details-view-container) will always need a storeState props.

PR #567, opened 2019-04-18
background support for save issue filing settings 

haonliu:
#### Description of changes

Background support for save issue filing settings. This will be used in Wahaj's change to fix the cancel button's UX.
No UX changes.

#### Pull request checklist

- [x] WI - 1502923

### DIFF
--- a/src/background/actions/action-payloads.ts
+++ b/src/background/actions/action-payloads.ts
@@ -129,7 +129,7 @@ export interface SetBugServicePayload extends BaseActionPayload {
     bugServiceName: string;
 }
 
-export interface SaveBugFilingSettingsPayload extends SetBugServicePayload {
+export interface SaveIssueFilingSettingsPayload extends SetBugServicePayload {
     bugFilingSettings: BugServiceProperties;
 }
 
--- a/src/background/actions/action-payloads.ts
+++ b/src/background/actions/action-payloads.ts
@@ -4,10 +4,10 @@ import * as TelemetryEvents from '../../common/telemetry-events';
 import { BaseTelemetryData, TelemetryData, ToggleTelemetryData } from '../../common/telemetry-events';
 import { DetailsViewPivotType } from '../../common/types/details-view-pivot-type';
 import { ManualTestStatus } from '../../common/types/manual-test-status';
+import { BugServiceProperties } from '../../common/types/store-data/user-configuration-store';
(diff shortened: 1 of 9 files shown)


src/background/stores/global/user-configuration-store.ts
```diff
@@ -86,6 +88,14 @@ export class UserConfigurationStore extends BaseStoreImpl<UserConfigurationStore
         this.saveAndEmitChanged();
     }
 
+    @autobind
+    private onSaveIssueSettings(payload: SaveBugFilingSettingsPayload): void {
+        const bugService = payload.bugServiceName;
+        this.state.bugService = payload.bugServiceName;
```
waabid: You can reuse your variable 'bugService' here.

src/background/actions/action-payloads.ts
```diff
@@ -128,6 +129,10 @@ export interface SetBugServicePayload extends BaseActionPayload {
     bugServiceName: string;
 }
 
+export interface SaveBugFilingSettingsPayload extends SetBugServicePayload {
```
peterdur: Should this be Save_Issue_FilingSettingsPayload given that the message is `SaveIssueFilingSettings`

haonliu: We have a work item (1506355) to rename all at once. In this PR there are several places that I don't really know which one should I use at present.

src/background/actions/action-payloads.ts
```diff
@@ -128,6 +129,10 @@ export interface SetBugServicePayload extends BaseActionPayload {
     bugServiceName: string;
 }
 
+export interface SaveIssueFilingSettingsPayload extends SetBugServicePayload {
```
peterdur: Having read more of the code now, and particularly thinking about the `extends SetBugServicePayload` here, I'm thinking that my earlier suggestion about renaming went the wrong way. It looks like this message should have the same semantics as `SetBugService` but also set the bundle of properties. If that's correct, I would suggest the following:

- Message to be named `SetBugServiceAndProperties`
- Payload to be named `SetBugServiceAndPropertiesPayload`
- Property within payload to be named `bugServiceProperties` to match type `BugServiceProperties`
- Methods related to `SetBugServiceAndProperties` should immediately follow their `SetBugService` analogs.

Thoughts?

haonliu: Make sense!

PR #568, opened 2019-04-18
Azure boards new bug url provider

smoralesd:
#### Description of changes

Add azure boards file issue url provider.

#### Pull request checklist

- [x] Addresses an existing issue: Part of WI #1505514


src/bug-filing/common/get-issue-details-html.ts
```diff
@@ -0,0 +1,64 @@
...
+): string => {
+    const body =
+        buildBodySection('Issue Details', buildIssueDetails(data.ruleResult.help, data.ruleResult.helpUrl, data.ruleResult.ruleId)) +
+        buildBodySection('Application', buildApplicationHtml(data.pageTitle, data.pageUrl)) +
+        buildBodySection('Element Path', escape(data.ruleResult.selector)) +
+        buildBodySection('Snippet', buildSnippetHtml(data.ruleResult.html)) +
+        buildBodySection('How to fix', buildHowToFixHtml(data.ruleResult.failureSummary)) +
+        buildBodySection(
+            'To reproduce',
+            'Use <a href="http://aka.ms/AccessibilityInsights">Accessibility Insights for Web</a> to investigate the issue details',
```
haonliu: should we get the string values from application.ts?

src/bug-filing/common/http-query-builder.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        return this.truncate(fullUrl);
+    }
+
+    private truncate(url: string): string {
+        if (url.length <= HTTPQueryBuilder.maxUrlLength) {
+            return url;
+        }
+
+        const truncateIndex = url.substr(0, HTTPQueryBuilder.maxUrlLength).length;
```
haonliu: Won't this always return HTTPQueryBuilder.maxUrlLength? I'm confused.

src/tests/unit/tests/bug-filing/azure-boards/create-azure-boards-issue-filing-url.test.ts
```diff
@@ -0,0 +1,138 @@
...
+            queryBuilderMock
+                .setup(builder => builder.withParam('[Microsoft.VSTS.TCM.ReproSteps]', testIssueDetails))
+                .returns(() => queryBuilderMock.object);
+
+            const result = testSubject(settingsData, sampleIssueDetailsData, environmentInfo);
+
+            expect(result).toMatchSnapshot();
+        });
+
+        it('uses repro steps field, with tabs', () => {
```
haonliu: with tags?

src/bug-filing/common/http-query-builder.ts
peterdur: [Original code](https://dev.azure.com/mseng/1ES/_git/accessibility-insights-web/commit/7d3c4e5e17ebf93535c22691732ebe73fbfe9efa?path=%2Fsrc%2Fbackground%2Fbug-url-builder.ts&gridItemType=2&_a=contents
) had a little extra logic here to truncate before a %3C (<) so we didn't cut things off in the middle of a parameter name, HTML element, or escape sequence. We probably still want to do something similar here.

src/bug-filing/common/get-issue-details-html.ts
smoralesd: updated.

src/bug-filing/common/http-query-builder.ts
smoralesd: It didn't occur to me at first reading the code. I remove it while writing the tests for it but now I see why we need it. So I added it back

src/tests/unit/tests/bug-filing/azure-boards/create-azure-boards-issue-filing-url.test.ts
smoralesd: updated

src/bug-filing/azure-boards/create-azure-boards-issue-filing-url.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const buildHowToFixHtml = (failureSummary: string): string => {
+    return _.escape(failureSummary)
+        .replace(/\n  /g, '<br>- ')
+        .replace(/\n /g, '<br> ')
+        .replace(/\n/g, '<br>');
+};
+
+const buildBody = (
+    createBugData: CreateIssueDetailsTextData,
```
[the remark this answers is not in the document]
smoralesd: Just did. Some thoughts about this:
- our html output doesn't have much interactive elements (expect for the links)
- not much html is used expect for `<div>`'s and `<br>`'s
- if we find issues, we may be not be able to resolve them since Azure Boards is the one rendering our content.
I'll check with an AT before resolving this comment.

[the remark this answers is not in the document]
smoralesd: I think we have some issues with our links but I would like to show the repro steps content to Fernanda, finalize the details and then fix any issue we may have. Sounds good @haonliu ?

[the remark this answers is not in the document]
haonliu: Sounds good!

src/bug-filing/azure-boards/create-azure-boards-issue-filing-url.ts
```diff
@@ -0,0 +1,49 @@
...
+    return tags.join('; ');
+};
+
+export const createAzureBoardsIssueFilingUrlProvider = (
+    stringUtils: IssueUrlCreationUtils,
+    issueDetailsGetter: IssueDetailsGetter,
+    queryBuilderProvider: () => HTTPQueryBuilder,
+) => {
+    return (settingsData: AzureBoardsBugFilingSettings, bugData: CreateIssueDetailsTextData, environmentInfo: EnvironmentInfo) => {
+        const standardTags = stringUtils.standardizeTags(bugData);
```
peterdur: There's an odd division of responsibility between `standardizeTags` and `buildTags` (which was present in the old code too, I see). 

- `IssueFilingUrlStringUtils.standardizeTags` gets the WCAG standard tags out of a `CreateIssueDetailsTextData` and capitalizes them
- `buildTags` collects the rule tag out of a `CreateIssueDetailsTextData`, the WCAG standard tags passed in, and some hardcoded tags. 

Might it be cleaner to split as: 
- `standardizeTags` to be something like `getTagsFromIssueDetails` and have it return all the tags that depend on data from `CreateIssueDetailsTextData`
- `buildTags` takes that list (and doesn't need to take a `CreateIssueDetailsTextData`), adds the hardcoded tags, and return the joined string?

src/bug-filing/common/issue-details-getter.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { EnvironmentInfo } from '../../common/environment-info-provider';
+import { CreateIssueDetailsTextData } from '../../common/types/create-issue-details-text-data';
+import { IssueUrlCreationUtils } from './issue-filing-url-string-utils';
+export type IssueDetailsGetter = (
```
peterdur: What do you think of calling this `IssueDetailsFormatter` instead? Getter sounds more like retrieving, and one of these things is given the raw details and formats them.

src/bug-filing/azure-boards/create-azure-boards-issue-filing-url.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        const titleField = stringUtils.getTitle(bugData);
+        const tags = buildTags(bugData, standardTags);
+        const body = issueDetailsGetter(stringUtils, environmentInfo, bugData);
+
+        let bodyField: string = '[Microsoft.VSTS.TCM.ReproSteps]';
+
```
peterdur: For me, this blank line makes it harder to follow that lines 29-33 are about setting `bodyField`.

src/bug-filing/common/issue-details-getter.ts
smoralesd: But is not a formatter. I though about provider or `buildIssueDetails`. In the end, I decided to just move it and not rename it.

src/bug-filing/azure-boards/create-azure-boards-issue-filing-url.ts
smoralesd: Not sure what's the suggestion here. I could remove the extra line but in my case, I usually prefer to have an empty line before "block-ed" pieces of code, live `if`'s and `for`'s. I guess is something more of a personal choice.

src/bug-filing/common/get-issue-details-html.ts
```diff
@@ -0,0 +1,65 @@
...
+};
+
+const buildHowToFixHtml = (failureSummary: string): string => {
+    return escape(failureSummary)
+        .replace(/\n  /g, '<br>- ')
+        .replace(/\n /g, '<br> ')
+        .replace(/\n/g, '<br>');
+};
+
+export const getIssueDetailsHtml = (
```
peterdur: I would be tempted to explicitly declare 'getIssueDetailsHtml' as a `IssueDetailsGetter` since it's intended to be used as one.

(Parenthetical thought: after years of naming classes as nouns and methods as verbs, it feels weird to pick either part of speech as the type name for a function.)

src/bug-filing/azure-boards/create-azure-boards-issue-filing-url.ts
smoralesd: `standardizeTags` is also being used on the git hub issue title, which don't include the rule id.

I think this will be reworked as part of our fit and finish. There are some thing I need to discuss with PM's before finalize the issue details for azure boards as they are not the same as for git hub right now but I don't have specs said issue details.

src/bug-filing/common/get-issue-details-html.ts
```diff
@@ -0,0 +1,65 @@
...
+import { EnvironmentInfo } from '../../common/environment-info-provider';
+import { CreateIssueDetailsTextData } from '../../common/types/create-issue-details-text-data';
+import { title } from '../../content/strings/application';
+import { IssueUrlCreationUtils } from './issue-filing-url-string-utils';
+
+const buildBodySection = (headingHtml: string, contentHtml: string): string => {
+    return `<div><b>${headingHtml}</b></div><div>${contentHtml}</div><br>`;
+};
+
+const buildIssueDetails = (help: string, helpUrl: string, ruleId: string): string => {
```
peterdur: I'd be tempted to call this `buildIssueDetailsHtml` as this makes it consistent with the other methods that return HTML that's ready to be used (untrusted data escaped).

(Also exists in the old code)

src/bug-filing/common/get-issue-details-html.ts
smoralesd: The general idea was to let the user of this function to enforce the type. I added though

src/bug-filing/common/get-issue-details-html.ts
```diff
@@ -0,0 +1,65 @@
...
+    const helpEscapedForHtml = escape(help);
+    const ruleIdEscapedForHtml = escape(ruleId);
+
+    return `${helpEscapedForHtml}<br><a href="${helpUrl}">${ruleIdEscapedForHtml}</a>`;
```
peterdur: Hmm, looks like we're not encoding `helpUrl`. (Problem also exists in the old code.) Probably should have an `encodeURI` here.

src/bug-filing/common/http-query-builder.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { size, values } from 'lodash';
+import { DictionaryStringTo } from '../../types/common-types';
+
+export type QueryParam = {
```
peterdur: It feels redundant to have name duplicated in the dictionary key and value:
```
{
    n1: { name: n1, value: v1 },
    n2: { name: n2, value: v2 },
}
```
Seems like it would be simpler to have
```
{
    n1: v1,
    n2: v2,
}
```
Key to doing this is using lodash's `toPairs` rather than `values` at line 39.

src/bug-filing/common/get-issue-details-html.ts
smoralesd: updated.

src/bug-filing/common/get-issue-details-html.ts
smoralesd: updated

src/bug-filing/azure-boards/create-azure-boards-issue-filing-url.ts
peterdur: Oh, wait... `standardizeTags` is also used to generate part of the title. (It is a happy coincidence that we use the same delimeter between a list of tags to put on an issue, and the list of WCAG standards we put in the title.)

src/bug-filing/common/http-query-builder.ts
smoralesd: Using such a dictionary and `toPairs` render me to this:

```
const queryParameters = toPairs(this.parameters)
    .map(pair => '${encodeURIComponent(pair[0])}=${encodeURIComponent(pair[1])}')
    .join('&');
```

where accessing the name and the value doesn't look too elegant.

The other option is to use `Object.keys(this.parameters)` and use the resulting array with a `forEach` or `for` loop. `forEach` will land on the same place as `toPairs` while `for` you have one more line of code to access the current element in the array using the current key.

I'll prefer to keep this little redundancy (not to bad really) so I can have the more verbose version of the `map` usage.

PR #573, opened 2019-04-20
[bug fix] Handle new property added in persisted data

haonliu:
#### Description of changes

 Handle new property added in persisted data

#### Pull request checklist

### DIFF
--- a/src/background/stores/global/user-configuration-store.ts
+++ b/src/background/stores/global/user-configuration-store.ts
@@ -35,8 +35,14 @@ export class UserConfigurationStore extends BaseStoreImpl<UserConfigurationStore
         super(StoreNames.UserConfigurationStore);
     }
 
+    private generateDefaultState(persisted: UserConfigurationStoreData): UserConfigurationStoreData {
+        const persistedState = cloneDeep(persisted);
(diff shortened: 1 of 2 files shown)


src/background/stores/global/user-configuration-store.ts
```diff
@@ -28,15 +28,21 @@ export class UserConfigurationStore extends BaseStoreImpl<UserConfigurationStore
...
         private readonly userConfigActions: UserConfigurationActions,
         private readonly indexDbApi: IndexedDBAPI,
     ) {
         super(StoreNames.UserConfigurationStore);
     }
 
+    private generateDefaultState(persisted: UserConfigurationStoreData): UserConfigurationStoreData {
+        const persistedState = cloneDeep(persisted);
+        const defaultState = cloneDeep(UserConfigurationStore.defaultState);
+        return Object.assign(defaultState, persistedState);
```
waabid: We usually just use the spread operator; why not here?

haonliu: If there are common properties, for me it's more clear that assign() will take the value from the later one; Also assign has intellisense in the IDE that tell me which is source, will is target, while spread operator does not have that.

PR #581, opened 2019-04-22
[Refactor] Issue details builder

smoralesd:
#### Description of changes

Refactor how we create the issue details content using Abstract Factory Pattern.
As a consequence of this refactor, the content for both, github and azure boards, is the same except for the format, as github use markdown while azure boards use html.

**Github**
**before**
01 - github - old
**after**
02 - github - new

**Azure boards**
**before**
03 - azure boards - old
**after**
04 - azure boards - new

#### Pull request checklist


src/bug-filing/common/markup/html-factory.ts
```diff
@@ -0,0 +1,51 @@
...
+        return `<b>${text}</b>`;
+    };
+
+    const snippet = (text: string): string => {
+        const maxSnippetLength = 256;
+        let constrainedSnippet = text;
+
+        if (text.length > maxSnippetLength) {
+            constrainedSnippet = text.substr(0, maxSnippetLength) + '...';
+        }
```
dbjorge: The truncating part of this is duplicated between the two factory implementations; perhaps that belongs in a more generic string utils instead?

src/bug-filing/azure-boards/create-azure-boards-issue-filing-url.ts
```diff
@@ -16,14 +17,14 @@ const buildTags = (createBugData: CreateIssueDetailsTextData, standardTags: stri
 
 export const createAzureBoardsIssueFilingUrlProvider = (
     stringUtils: IssueUrlCreationUtils,
-    issueDetailsGetter: IssueDetailsGetter,
+    issueDetailsGetter: IssueDetailsBuilder,
```
haonliu: should property name match the type?

src/bug-filing/common/markup/markup-factory.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export type MarkupFactory = {
```
dbjorge: >Factory [](start = 18, length = 7)

I understand that you're using an abstract factory pattern, but I think "factory" is maybe not the most intuitive name for this; I think maybe `Formatter` expresses the responsibility more specifically

src/bug-filing/common/markup/html-factory.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        return `<code>${escape(constrainedSnippet)}</code>`;
```
dbjorge: >escape [](start = 24, length = 6)

It feels off that snippet and howToFix in this factory use escape, but none of the other functions in either factory perform escaping. Should we be consistent?

src/bug-filing/common/create-issue-details-builder.ts
```diff
@@ -0,0 +1,40 @@
...
+import { title } from '../../content/strings/application';
+import { IssueDetailsBuilder } from './issue-details-builder';
+import { MarkupFactory } from './markup/markup-factory';
+
+export const createIssueDetailsBuilder = (markup: MarkupFactory): IssueDetailsBuilder => {
+    const getter = (environmentInfo: EnvironmentInfo, data: CreateIssueDetailsTextData): string => {
+        const result = data.ruleResult;
+
+        const text = [
+            `${markup.bold('Issue')}: ${markup.snippet(result.help)} (${markup.link(result.helpUrl, result.ruleId)})`,
```
dbjorge: >`${markup.bold('Issue')}: ${markup.snippet(result.help)} (${markup.link(result.helpUrl, result.ruleId)})`, [](start = 12, length = 106)

Personally, I think I liked the old Azure DevOps style better (where there were line breaks between the section headers and the single-line contents); description fields are usually on the thin side in a bug viewer, so I prefer keeping the line lengths short. Maybe we should run it by Chelsea?

dbjorge: >bold [](start = 22, length = 4)

Is "bold" actually the ideal thing for us to be using here? I think we had gotten guidance from design when we made our PR/Bug templates that `####` was the preferred section heading distinguisher in markdown; would it be better to do a `markup.sectionHeader` here and have the 2 factories use `h4` and `####` instead?

src/bug-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -1,60 +1,50 @@
...
-    formatAsMarkdownCodeBlock: (input: string) => string;
     getSelectorLastPart: (selector: string) => string;
     standardizeTags: (data: CreateIssueDetailsTextData) => string[];
     appendSuffixToUrl: (url: string, suffix: string) => string;
 };
 
-export const IssueFilingUrlStringUtils: IssueUrlCreationUtils = {
-    getTitle: (data: CreateIssueDetailsTextData): string => {
-        const standardTags = IssueFilingUrlStringUtils.standardizeTags(data);
+export const createUtils = (): IssueUrlCreationUtils => {
```
dbjorge: >export [](start = 0, length = 6)

is exporting this necessary?

src/bug-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -1,60 +1,50 @@
...
+
+    return {
+        getTitle,
+        getSelectorLastPart,
+        appendSuffixToUrl,
+        standardizeTags,
+    };
 };
+
+export const IssueFilingUrlStringUtils = createUtils();
```
dbjorge: What does this indirection add vs just exporting the object returned by createUtils directly?

src/bug-filing/common/create-issue-details-builder.ts
smoralesd: I run it by Fernanda and she was ok with how it looks. Remember we don't have full control of how the content looks as it is ultimately rendered by Azure boards/Github

src/bug-filing/common/markup/html-factory.ts
smoralesd: Updated. Ended up using lodash truncate.

src/bug-filing/azure-boards/create-azure-boards-issue-filing-url.ts
smoralesd: Updated.

src/bug-filing/common/markup/markup-factory.ts
smoralesd: Make sense. Updated.

src/bug-filing/common/issue-filing-url-string-utils.ts
smoralesd: It makes it easier to reuse functions inside the object being created. Look at `getTitle` for example, it use `standardizeTags`. For me it reads better like this instead of `IssueFilingUrlStringUtils.standardizeTag`

src/bug-filing/common/issue-filing-url-string-utils.ts
smoralesd: Same as the comment for the indirection. It makes it easier to reuse functions inside the object being created. Look at `getTitle` for example, it use `standardizeTags`. For me it reads better like this instead of `IssueFilingUrlStringUtils.standardizeTag`

src/bug-filing/common/issue-filing-url-string-utils.ts
dbjorge: I agree it reads a little better, but I don't think it's enough better to be worth having to type the names multiple times and losing F12/codelens.

src/bug-filing/common/issue-filing-url-string-utils.ts
dbjorge: That answers why you'd use a separate function for it, but not why the function itself has to be exported (as opposed to it being a non-exported function and only the result of calling it is exported).

src/bug-filing/common/create-issue-details-builder.ts
smoralesd: This is an interesting point. I'm not sure what the context and scope of that guidance is, and I'm not sure it affects content we generate but we don't directly render.
That being said, I'll explore using your suggestions, see how it behave/looks and decide base on what I gather.

src/bug-filing/common/issue-filing-url-string-utils.ts
smoralesd: No misunderstanding on your side. You're totally right, we don't need to export the `createUtils` function, which means you're probably right about not actually need the function in the first place. That just slip my mind. I wrote it like that because of other files having the same coding style where I need to expose the create function in order to pass mocked dependencies; that's clearly not this case so I'll update it following your comments.

src/bug-filing/common/issue-filing-url-string-utils.ts
smoralesd: Not sure if I follow you're F12/codelens comment. That's around IDE not about the code itself so it feels more subjective than not. I'm still updating to not expose the function just the result and looking into actually removing the function itself (not sure if I actually do that last part though)

smoralesd: Updated. Does this work better?

src/bug-filing/common/issue-filing-url-string-utils.ts
smoralesd: Updated. Does this work better?

src/bug-filing/common/create-issue-details-builder.ts
smoralesd: @ferBonnin: Dan has a good point here about making the content accessible by using headers instead of just bold text. I made the change and update the screenshots. Can you have a look? Let me know what you think. Thanks

src/bug-filing/common/markup/html-factory.ts
smoralesd: how to fix and snippet are the only places we expect to have html code we may need to scape, so not too sure about the consistency point here.

src/tests/unit/tests/bug-filing/common/issue-details-builder.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    const environmentInfo: EnvironmentInfo = {
+        extensionVersion: '1.1.1',
+        axeCoreVersion: '2.2.2',
+        browserSpec: 'test spec',
+    };
+
+    const markupMock = Mock.ofType<MarkupFormatter>(null, MockBehavior.Strict);
+
+    beforeEach(() => {
+        markupMock.reset();
```
waabid: As per our engineering discussion (?), should we not be creating the Mock here instead of having to reset?

src/tests/unit/tests/bug-filing/common/markup/html-formatter.test.ts
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { It, Mock, MockBehavior } from 'typemoq';
+import { createFormatter } from '../../../../../../bug-filing/common/markup/html-formatter';
+
+describe('HTMLFormatter', () => {
+    const truncateMock = Mock.ofInstance((text: string) => text, MockBehavior.Strict);
```
waabid: I think we should have this creation in a before each so we can be sure of no side-effects.

src/tests/unit/tests/bug-filing/common/markup/markdown-formatter.test.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { It, Mock, MockBehavior } from 'typemoq';
+import { createFormatter } from '../../../../../../bug-filing/common/markup/markdown-formatter';
+
+describe('MarkdownFormatter', () => {
+    const truncateMock = Mock.ofInstance((text: string) => text, MockBehavior.Strict);
```
waabid: Same thing about using beforeEach.

src/tests/unit/tests/bug-filing/common/issue-details-builder.test.ts
smoralesd: our engineering discussion was about not using `after` to run verify on our mocks... I don't think we land on anything about the use of `reset` or where to create the mocks

src/tests/unit/tests/bug-filing/common/markup/markdown-formatter.test.ts
smoralesd: updated

src/tests/unit/tests/bug-filing/common/markup/html-formatter.test.ts
smoralesd: updated

src/tests/unit/tests/bug-filing/common/issue-details-builder.test.ts
smoralesd: updated

PR #582, opened 2019-04-22
Block running extension on Edge Insider addon pages

peterdur:
#### Description of changes

We block running scans on the Chrome Web Store since the browser does not allow extensions to execute script on these pages. The same limitation applies to the Edge Insider Addons store pages (when visited from Edge Insider), so we need to block those as well. To do this, I modified `UrlValidator.isSupportedUrl` to use a new private helper `hasSupportedPrefix` which checks a list of unsupportedPrefixes and returns `true` if none of them match the passed URL.

#### Pull request checklist

- [X] Addresses an existing issue: Fixes 1516239
- [X] Added relevant unit test for your changes. (`npm run test`)
- [X] Ran precheckin (`npm run precheckin`)
- [X] Added screenshots/GIFs for UI changes.
image


src/common/url-validator.ts
```diff
@@ -6,14 +6,19 @@ export class UrlValidator {
     public async isSupportedUrl(url: string, chromeAdapter: BrowserAdapter): Promise<boolean> {
         const lowerCasedUrl: string = url.toLowerCase();
         if (lowerCasedUrl.match('http://*/*') || lowerCasedUrl.match('https://*/*')) {
```
dbjorge: While you're here, this regex looks like it's trying to test the wrong thing but happens to be working by accident (I bet it would fail a test case like `http:/oops` and would only accidentally pass a test case like `http://domain-with-no-trailing-slash`) :)

dbjorge: probably accidentally passes `oops_http://example.com` too

src/common/url-validator.ts
```diff
@@ -6,14 +6,19 @@ export class UrlValidator {
...
+            return this.hasSupportedPrefix(lowerCasedUrl);
         } else if (UrlValidator.isFileUrl(lowerCasedUrl)) {
             return await this.checkAccessToFileUrl(chromeAdapter);
         } else {
             return false;
         }
     }
 
+    private hasSupportedPrefix(lowerCasedUrl: string): boolean {
+        const unsupportedPrefixes = ['https://chrome.google.com/', 'https://microsoftedge.microsoft.com/'];
```
smoralesd: Currently on master/production we are able to scan https://microsoftedge.microsoft.com using our chrome extension. This change means we won't. The opposite is also true: we won't be able to scan urls that contains 'chrome' using the edge extension. Is that the correct intention? This sounds like one of several behaviors we should consider (and implement) as differential behavior between chrome and edge extension.

peterdur: It is true that Chrome can currently scan https://microsoftedge.microsoft.com and would not be able to after this change. Edge Insider, on the other hand, is prevented from scanning both Chrome Web Store pages and Edge Insider Addon pages. I think things will be simpler if we can avoid differential behavior between the two browsers if at all possible.

src/common/url-validator.ts
peterdur: You're right - added these test cases and updated to fix.

PR #597, opened 2019-04-23
Revert visual helper text change

Shobhit1:
#### Description of changes

Revert Visual helper string change till we figure out specific cases where the term "Highlight" doesn't make sense. For e.g: Color Visualizations.

#### Pull request checklist

#### Screenshots:

image

image

### DIFF
--- a/src/tests/unit/tests/assessments/automated-checks/automated-checks-visualization-toggle.test.tsx
+++ b/src/tests/unit/tests/assessments/automated-checks/automated-checks-visualization-toggle.test.tsx
@@ -7,6 +7,7 @@ import { IMock, Mock } from 'typemoq';
 import { AutomatedChecksVisualizationToggle } from '../../../../../assessments/automated-checks/automated-checks-visualization-enabled-toggle';
 import { VisualizationToggle, VisualizationToggleProps } from '../../../../../common/components/visualization-toggle';
 import { DetailsViewActionMessageCreator } from '../../../../../DetailsView/actions/details-view-action-message-creator';
+import { visualHelperText } from '../../../../../DetailsView/components/base-visual-helper-toggle';
 import { VisualHelperToggleConfigBuilder } from '../../../common/visual-helper-toggle-config-builder';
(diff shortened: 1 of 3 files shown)


src/tests/unit/tests/assessments/automated-checks/automated-checks-visualization-toggle.test.tsx
```diff
@@ -7,7 +7,6 @@ import { IMock, Mock } from 'typemoq';
 import { AutomatedChecksVisualizationToggle } from '../../../../../assessments/automated-checks/automated-checks-visualization-enabled-toggle';
 import { VisualizationToggle, VisualizationToggleProps } from '../../../../../common/components/visualization-toggle';
 import { DetailsViewActionMessageCreator } from '../../../../../DetailsView/actions/details-view-action-message-creator';
-import { visualHelperText } from '../../../../../DetailsView/components/base-visual-helper-toggle';
```
waabid: The changes in this file don't seem to be necessary.

Shobhit1: I realized that I was importing the `visualHelper` variable to make an assertion while the assertion should be for the content of the variable. It made sense to me to test that rather than relying on the same variable because now if I change the variable name, this test will break; which was not the case earlier.

waabid: Since this UI content, I don't think we should be testing whether the UI content _is_ the UI content. I'm _fine_ with this for now but I'd like to cover this in an engineering discussion.

If anything maybe this should be done via snapshot testing? Thoughts?

Shobhit1: Snapshot testing makes sense here 100%. I will refactor these tests to make sure that happens. 

For the time being I will revert this test change and cover all of it to make this a snapshot test.

PR #609, opened 2019-04-25
[Refactor] Issue filing handling

smoralesd:
#### Description of changes

Refactor how we are handling 'file issue' button activation.

Currently, the button has a href property on it and also an onClick property (where we send telemetry).

Instead, the button should only send an action message to the background, which should send telemetry and handle the logic to file the issues against the proper service.

Also: remove unneeded custom string type; add missing test to ensure key uniqueness across issue filing services.

#### Pull request checklist

- [x] Addresses an existing issue: WI 1518658

### DIFF
--- a/src/background/background-init.ts
+++ b/src/background/background-init.ts
@@ -3,14 +3,18 @@
 import { AppInsights } from 'applicationinsights-js';
 
 import { Assessments } from '../assessments/assessments';
+import { BugFilingServiceProviderImpl } from '../bug-filing/bug-filing-service-provider-impl';
+import { AxeInfo } from '../common/axe-info';
 import { VisualizationConfigurationFactory } from '../common/configs/visualization-configuration-factory';
+import { EnvironmentInfoProvider } from '../common/environment-info-provider';
 import { IndexedDBAPI, IndexedDBUtil } from '../common/indexedDB/indexedDB';
+import { InsightsFeatureFlags } from '../common/insights-feature-flag';
 import { createDefaultLogger } from '../common/logging/default-logger';
+import { NavigatorUtils } from '../common/navigator-utils';
 import { NotificationCreator } from '../common/notification-creator';
 import { TelemetryDataFactory } from '../common/telemetry-data-factory';
 import { UrlValidator } from '../common/url-validator';
 import { WindowUtils } from '../common/window-utils';
-import { Window } from '../Scripts/Window';
 import { ChromeAdapter } from './browser-adapter';
 import { ChromeCommandHandler } from './chrome-command-handler';
 import { DetailsViewController } from './details-view-controller';
@@ -31,7 +35,7 @@ import { TelemetryLogger } from './telemetry/telemetry-logger';
 import { TelemetryStateListener } from './telemetry/telemetry-state-listener';
 import { UserStoredDataCleaner } from './user-stored-data-cleaner';
 
-declare var window: Window;
+declare const window: InsightsFeatureFlags & Window;
 const browserAdapter = new ChromeAdapter();
 const urlValidator = new UrlValidator();
 const backgroundInitCleaner = new UserStoredDataCleaner(browserAdapter);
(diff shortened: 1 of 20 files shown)


src/bug-filing/common/issue-filing-controller-impl.ts
```diff
@@ -0,0 +1,32 @@
...
+import { BrowserAdapter } from '../../background/browser-adapter';
+import { BaseStore } from '../../common/base-store';
+import { EnvironmentInfo } from '../../common/environment-info-provider';
+import { FileIssueClickService } from '../../common/telemetry-events';
+import { CreateIssueDetailsTextData } from '../../common/types/create-issue-details-text-data';
+import { BugServiceProperties, UserConfigurationStoreData } from '../../common/types/store-data/user-configuration-store';
+import { BugFilingServiceProvider } from '../bug-filing-service-provider';
+
+export type IssueFilingController = {
+    fileIssue: (serviceKey: FileIssueClickService, issueData: CreateIssueDetailsTextData) => void;
```
peterdur: `FileIssueClickService` was defined in telemetry-events.ts as the type for the `service` in `FileIssueClickTelemetryData`, but now it's being used more broadly. I think Telemetry code's parameter type should depend on the File Issue code's service type, not the other way around. Probably also rename it to `IssueFilingServiceKey` as `FileIssueClickService` is confusing (to me, at least) in this context. Oh, and we should add `azureBoards` as a possible value.

smoralesd: I was going to clean up that type on a different PR, but since it surface here...
The type is not giving us any real advantage. Having a test to ensure uniqueness of the key across the services is enough. Base on this, I remove the type and use `string` instead.

src/bug-filing/common/issue-filing-controller-impl.ts
```diff
@@ -0,0 +1,31 @@
...
+import { EnvironmentInfo } from '../../common/environment-info-provider';
+import { CreateIssueDetailsTextData } from '../../common/types/create-issue-details-text-data';
+import { BugServiceProperties, UserConfigurationStoreData } from '../../common/types/store-data/user-configuration-store';
+import { BugFilingServiceProvider } from '../bug-filing-service-provider';
+
+export type IssueFilingController = {
+    fileIssue: (serviceKey: string, issueData: CreateIssueDetailsTextData) => void;
+};
+
+export class IssueFilingControllerImpl implements IssueFilingController {
```
peterdur: Here `IssueFilingControllerImpl` is a class that implements abstract type `IssueFilingController`.  (`StoreActionMessageCreatorImpl` is another example of this naming pattern). We also have `BugFilingServiceProviderImpl` which is an instance of concrete type `BugFilingServiceProvider`. These seem like different concepts that should have different names, rather than both being `Impl`. (I think the usage is good here, not as good for `BugFilingServiceProvider`, but would like to hear what others think...)

haonliu:
Sorry for the conflicts :-)

src/background/background-init.ts
```diff
@@ -3,14 +3,18 @@
...
+import { EnvironmentInfoProvider } from '../common/environment-info-provider';
 import { IndexedDBAPI, IndexedDBUtil } from '../common/indexedDB/indexedDB';
+import { InsightsFeatureFlags } from '../common/insights-feature-flag';
 import { createDefaultLogger } from '../common/logging/default-logger';
+import { NavigatorUtils } from '../common/navigator-utils';
 import { NotificationCreator } from '../common/notification-creator';
 import { TelemetryDataFactory } from '../common/telemetry-data-factory';
 import { UrlValidator } from '../common/url-validator';
 import { WindowUtils } from '../common/window-utils';
-import { Window } from '../Scripts/Window';
```
peterdur: I think this is the last reference to this file, so it can be deleted.

src/background/background-init.ts
```diff
@@ -31,7 +35,7 @@ import { TelemetryLogger } from './telemetry/telemetry-logger';
 import { TelemetryStateListener } from './telemetry/telemetry-state-listener';
 import { UserStoredDataCleaner } from './user-stored-data-cleaner';
 
-declare var window: Window;
+declare const window: InsightsFeatureFlags & Window;
```
peterdur: Tiny optional comment: I think this would be a little clearer as `Windows & InsightsFeatureFlags` (it's a window with the feature flag property added to it...)

src/background/global-context-factory.ts
```diff
@@ -26,6 +30,8 @@ export class GlobalContextFactory {
         telemetryDataFactory: TelemetryDataFactory,
         indexedDBInstance: IndexedDBAPI,
         persistedData: PersistedData,
+        issueFilingServiceProvider?: BugFilingServiceProvider,
```
peterdur: I think these parameters shouldn't be optional (the one call in production code passes them, they're skipped in test code - maybe the test should be more complete?).

PR #617, opened 2019-04-26
Remove overriding window type

smoralesd:
#### Description of changes

Instead of "overwriting" `Window` type having a proper type with the property we need. In case we need the original `Window` we should be using intersection types, e.g
```typescript
declare var window: Window & InsightsFeatureFlag;
```

#### Pull request checklist


src/background/background-init.ts
```diff
@@ -31,7 +31,7 @@ import { TelemetryLogger } from './telemetry/telemetry-logger';
 import { TelemetryStateListener } from './telemetry/telemetry-state-listener';
 import { UserStoredDataCleaner } from './user-stored-data-cleaner';
 
-declare var window: Window;
+declare var window: InsightsFeatureFlags;
```
peterdur: Debated about whether this should be
`declare var window: Window & InsightsFeatureFlags;`
so it doesn't confuse someone who later needs to get to other properties of `window`. However, PR description (and this comment) should cover the case of someone looking at this and thinking "why is this the wrong type?" and doing a blame to find out.

smoralesd: @peterdur 
For the record:

- We are only directly using the global `window` object, i.e. we are not passing it as a parameter to any class/instance.
- We don't use any window properties expect for one `insightsFeatureFlags`, which we are using to expose our feature flags controller
- Using `InsightsFeatureFlags` for the `window` object is better suit to abide with **Interface segregation principle**

The last point is the main reason I decided to got with this type and not an intersection type between `Window` and `InsightsFeatureFlags`

PR #603, opened 2019-04-24
Using HTTPQueryBuilder for github issue filing url

smoralesd:
#### Description of changes

For github issue filing we are manually building the final url.
On this PR, I updated that logic to use `HTTPQueryBuilder` (which take care of url truncation too).
Updated the max url length to 2000 (the de facto value).
Updated label and placeholder to mention issues is a necessary part of the github repo url for issue filing.

Annotation 2019-04-24 164032

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [x] Added relevant unit test for your changes. (`npm run test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] Ran precheckin (`npm run precheckin`)
- [x] Added screenshots/GIFs for UI changes.

PR #621, opened 2019-04-29
Add url rectifier to add /issues if the base url does not have it

smoralesd:
#### Description of changes

Re-adding some code to ensure the github url ends with `/issues`.
The main reason to do this is the difference in behaviors between the first and second versions of issue file: on the first version, we only ask the user to provider the repo url (without /issues) and we were appending the proper suffix, while the on second version we ask the user for the repo issues url (including /issues).
By updating the extension, and old version of the url will not work as of now. This small PR takes care of it while also supporting non-standard github urls (like in the case of github enterprise)

Original commit [here](https://github.com/Microsoft/accessibility-insights-web/pull/603/commits/f18746d82dba69727de966ee76d4db2567d2f779)

#### Pull request checklist

### DIFF
--- a/src/issue-filing/github/github-url-rectifier.ts
+++ b/src/issue-filing/github/github-url-rectifier.ts
@@ -8,7 +8,7 @@ export const rectify: UrlRectifier = (url: string): string => {
         return `${url}/issues`;
     }
 
-    const ownerAndRepoAndSlash = new RegExp('^https?://github.com/[^/]+/[^/]+/?$');
+    const ownerAndRepoAndSlash = new RegExp('^https?://github.com/[^/]+/[^/]+/$');
     if (ownerAndRepoAndSlash.test(url)) {
         return `${url}issues`;
     }
--- a/src/issue-filing/github/github-url-rectifier.ts
+++ b/src/issue-filing/github/github-url-rectifier.ts
@@ -1,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export type UrlRectifier = (url: string) => string;
+
+export const rectify: UrlRectifier = (url: string): string => {
+    const ownerAndRepo = new RegExp('^https?://github.com/[^/]+/[^/]+$');
+    if (ownerAndRepo.test(url)) {
+        return `${url}/issues`;
+    }
+
+    const ownerAndRepoAndSlash = new RegExp('^https?://github.com/[^/]+/[^/]+/?$');
+    if (ownerAndRepoAndSlash.test(url)) {
+        return `${url}issues`;
+    }
+
+    return url;
+};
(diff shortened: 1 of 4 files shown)


src/issue-filing/github/github-url-rectifier.ts
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License.
+export type UrlRectifier = (url: string) => string;
+
+export const rectify: UrlRectifier = (url: string): string => {
+    const ownerAndRepo = new RegExp('^https?://github.com/[^/]+/[^/]+$');
+    if (ownerAndRepo.test(url)) {
+        return `${url}/issues`;
+    }
+
+    const ownerAndRepoAndSlash = new RegExp('^https?://github.com/[^/]+/[^/]+/?$');
```
peterdur: The last ? here is not correct because this RegExp is looking for a URL that ends with a slash
```suggestion
    const ownerAndRepoAndSlash = new RegExp('^https?://github.com/[^/]+/[^/]+/$');
```
(Doesn't cause a bug because the case without a slash is handled by lines 6-9 above)

src/issue-filing/github/github-url-rectifier.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export type UrlRectifier = (url: string) => string;
+
+export const rectify: UrlRectifier = (url: string): string => {
+    const ownerAndRepo = new RegExp('^https?://github.com/[^/]+/[^/]+$');
```
peterdur: Including `github.com` in the regular expression means we won't add a missing /issues to a GitHub Enterprise URL (but it also means we won't break a GitHub Enterprise URL that's formatted differently than we expect, which is a nice effect of how the tests are structured).

smoralesd: That was part of the rationale. I don't have much info about how github enterprise works so I'm assuming the url can be fully customized, in case, the rectifier will not touch it, it will be up to the user to write down the correct url, which is the same right now on master anyway so this is definitely an improvement.

src/issue-filing/github/github-url-rectifier.ts
smoralesd: Good catch

src/issue-filing/github/github-url-rectifier.ts
haonliu: Does this support ````https://www.github.com/owner/project```` ?

smoralesd: Yes, there is a test case with `'http://github.com/me/my-repo'` (I'm guessing you're question points to the fact you have no ending `/`). You can see the test cases for more scenarios.

haonliu: I'm actually thinking the case where user put 'www' before 'github.com', should we support that?

smoralesd:
@waabid : Check with @ferBonnin for the github enterprise scenario not being address on this PR and we're good to go.

src/issue-filing/github/github-url-rectifier.ts
smoralesd: In that case, the user is responsible to provide the correct url. Also remember we have a placeholder with and example url which doesn't have `www` in it. That being said, if the user is on github enterprise account where the url start with `www` this rectifier will not change the url but it won't "throw" any error neither.

PR #635, opened 2019-04-30
run npm upgrade

Shobhit1:
#### Description of changes

Run npm-upgrade after release. This upgrades some of our dependencies and some of the tests were breaking. Fixed those as well.

#### Pull request checklist


src/tests/unit/tests/DetailsView/components/target-change-dialog.test.tsx
```diff
@@ -26,7 +25,7 @@ describe('TargetChangeDialog test set for prev tab null', () => {
         };
 
         urlParserMock
-            .setup(urlParserObject => urlParserObject.areURLHostNamesEqual(prevTab, newTab.url))
+            .setup(urlParserObject => urlParserObject.areURLHostNamesEqual(It.isAny(), newTab.url))
```
haonliu: why do you need to make this change?

Shobhit1: Because the `urlParser` accepts a `string` and since this code checks that the mocked is never called (see `Times.never()`). This change actually setups the test correctly, since now it represents that whatever the value the mock is never called.

haonliu: Thanks, it make total sense when verifying a mock is never called.

haonliu:
BTW (not related to this PR) should we look into update office-fabric-react soon?

PR #637, opened 2019-04-30
Show graceful incompatible message for when popup is opened in older Microsoft Edge browsers (below v75)

waabid:
#### Description of changes

Instead of going through the normal popup initializing flow, we instead return/render an incompatible browser component into the popup container that will point users to accessibilityinsights.io for information.

Also includes minor changes related to manifest allowing the extension to load appropriately with correct icons and minor changes to CSS targeting the Microsoft Edge (v74 and below) browser specifically.

In Microsoft Edge (v74 and below):

image

In Chrome:

image

In Microsoft Edge (v75+):

image

#### Pull request checklist


src/manifest.json
```diff
@@ -13,7 +18,8 @@
     },
     "content_security_policy": "script-src 'self' 'unsafe-eval' https://az416426.vo.msecnd.net; object-src 'self'",
     "background": {
-        "page": "background/background.html"
+        "page": "background/background.html",
+        "persistent": false
```
smoralesd: This seems to be a sensible change. Do we need to fully test our extension? I've been looking at the auto update feature and this came up... wondering how "dangerous" is to make this change, in the sense of the unintended consequences it may have...

src/popup/incompatible-browser-renderer.tsx
```diff
@@ -0,0 +1,36 @@
...
+        const container = this.dom.querySelector('#popup-container');
+
+        this.renderer(
+            <>
+                <div className="ms-Fabric unsupported-browser-info-panel">
+                    <Header title={title} />
+                    <div className="ms-Grid main-section">
+                        <div className="launch-panel-general-container">
+                            Thanks for your interest in Accessibility Insights!
+                            <div className="some-class">
```
smoralesd: is the class name intended to be `"some-class"`?

src/tests/unit/tests/popup/incompatible-browser-renderer.test.tsx
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import ReactDOM from 'react-dom';
+import { It, Mock } from 'typemoq';
+
+import { IncompatibleBrowserRenderer } from '../../../../popup/incompatible-browser-renderer';
+
+describe('IncompatibleBrowserRenderer', () => {
+    it('render', () => {
```
smoralesd: typo: render**s**

src/tests/unit/tests/popup/incompatible-browser-renderer.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+        const renderMock = Mock.ofType<typeof ReactDOM.render>();
+        const containerStub = Mock.ofType<HTMLElement>();
+        const documentMock = Mock.ofType<NodeSelector & Node>();
+
+        documentMock.setup(mock => mock.querySelector('#popup-container')).returns(() => containerStub.object);
+
+        renderMock
+            .setup(mock => mock(It.isAny(), containerStub.object))
```
smoralesd: should the first parameter be `containerStub`?

src/manifest.json
peterdur: The EdgeHTML browser [requires](https://docs.microsoft.com/en-us/microsoft-edge/extensions/troubleshooting) a value be set here. The conservative approach would be to specify the default value to preserve the existing state. Google is so insistent on wanting you to set `"persistent": false` that they don't really [document](https://developer.chrome.com/extensions/manifest) what the default is. [This random blog post](http://coryg89.github.io/technical/2013/08/13/how-to-create-your-own-chrome-extensions/) from 2013 suggests that the default is `"persistent": true`. We do want to get to `"persistent": false` eventually, but I agree with @smoralesd that we'd need to test thoroughly before making this change.

smoralesd: I'll do some testing as part of my spike... I'm fine with having this change as part of this PR I just want us all to be aware that maybe unintended consequences :(

src/tests/unit/tests/popup/incompatible-browser-renderer.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License.
+import ReactDOM from 'react-dom';
+import { It, Mock } from 'typemoq';
+
+import { IncompatibleBrowserRenderer } from '../../../../popup/incompatible-browser-renderer';
+
+describe('IncompatibleBrowserRenderer', () => {
+    it('render', () => {
+        const renderMock = Mock.ofType<typeof ReactDOM.render>();
+        const containerStub = Mock.ofType<HTMLElement>();
```
peterdur: Since this is a mock, should it be called `containerMock`?

src/tests/unit/tests/popup/incompatible-browser-renderer.test.tsx
peterdur: I think this is correct - this is mocking [ReactDOM.render](https://reactjs.org/docs/react-dom.html#render) where the first parameter is the element and the second parameter is the container.

src/manifest.json
waabid: @peterdur I was under the assumption it was false since they seem very insistent on telling us to set it to false 😆 . I've changed it to true now.

src/popup/incompatible-browser-renderer.tsx
waabid: Good catch! Yes, I will set a more meaningful name.

src/tests/unit/tests/popup/incompatible-browser-renderer.test.tsx
waabid: Nope! The first is the react component. The 2nd is the container to which it will be added.

src/tests/unit/tests/popup/incompatible-browser-renderer.test.tsx
waabid: I'm not using any mock capabilities with it so that's why I called it that; I can change it to mock.

src/tests/unit/tests/popup/incompatible-browser-renderer.test.tsx
smoralesd: Right! I'm reading backwards. My bad.

PR #646, opened 2019-05-01
refactor: Introduce Storage Adapter

smoralesd:
#### Description of changes

Currently, we use `BrowserAdapter` to encapsulate our access to the chrome API. Having the adapter is great as it place all `chrome.` calls in a single place. The down side of doing this is: our `BrowserAdapter` type expose 30 different functions. This means we have "user" classes that depend on `BrowesrAdapter` but hardly use 5 of this 30 functions. Additionally, by looking at who is using the adapter, how the adapter is being used and which underlying chrome API each function uses, is possible to see `BrowserAdapter` is not cohesive. In other words, `BrowserAdapter` breaks the [Interface Segregation Principle](https://hackernoon.com/interface-segregation-principle-bdf3f94f1d11).

This is a first step into breaking `BrowserAdapter` into smaller, more cohesive types. I started with the local storage API because:
- all of this functions use `chrome.storage.local`
- all the "user" classes don't use anything else from the `BrowserAdapter`, just need access to the local storage.
- just 3 functions needed to be extracted from `BrowserAdapter`

#### Pull request checklist

### DIFF
--- a/src/background/install-data-generator.ts
+++ b/src/background/install-data-generator.ts
@@ -5,18 +5,14 @@ import { InstallationData } from './installation-data';
 import { LocalStorageDataKeys } from './local-storage-data-keys';
 
 export class InstallDataGenerator {
-    private generateGuid: () => string;
-    private dateGetter: () => Date;
     private installationData: InstallationData;
 
     constructor(
-        initialInstallationData: InstallationData,
-        generateGuid: () => string,
-        dateGetter: () => Date,
+        readonly initialInstallationData: InstallationData,
+        private readonly generateGuid: () => string,
+        private readonly dateGetter: () => Date,
         private readonly storageAdapter: StorageAdapter,
     ) {
-        this.generateGuid = generateGuid;
-        this.dateGetter = dateGetter;
         this.installationData = initialInstallationData;
     }
 
--- a/src/background/install-data-generator.ts
+++ b/src/background/install-data-generator.ts
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { StorageAPI } from './browser-adapters/storage-adapter';
+import { StorageAdapter } from './browser-adapters/storage-adapter';
 import { InstallationData } from './installation-data';
 import { LocalStorageDataKeys } from './local-storage-data-keys';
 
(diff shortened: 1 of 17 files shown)


src/background/user-stored-data-cleaner.ts
```diff
@@ -1,20 +1,16 @@
...
-import { BrowserAdapter } from './browser-adapter';
+import { StorageAPI } from './browser-adapters/storage-adapter';
 
 export class UserStoredDataCleaner {
-    private _browserAdapter: BrowserAdapter;
-
-    constructor(adapter: BrowserAdapter) {
-        this._browserAdapter = adapter;
-    }
+    constructor(private readonly storageAdapter?: StorageAPI) {}
```
lisli1: I'm curious why we want `storageAdapter` to be optional here, since `cleanUserData` will use it.

src/background/install-data-generator.ts
```diff
@@ -1,25 +1,23 @@
...
     private dateGetter: () => Date;
     private installationData: InstallationData;
-    private browserAdapter: BrowserAdapter;
 
     constructor(
         initialInstallationData: InstallationData,
         generateGuid: () => string,
         dateGetter: () => Date,
-        browserAdapter: BrowserAdapter,
+        private readonly storageAdapter: StorageAPI,
```
lisli1: For consistency, should we also change the other parameters in the constructor to look the same as the one for `storageAdapter`?

src/background/user-stored-data-cleaner.ts
smoralesd: Good catch!

src/background/install-data-generator.ts
smoralesd: all except `initialInstallationData`, which is use to initialize an internal, private property which is not `readonly`.

src/tests/unit/tests/background/global-context-factory.test.ts
```diff
@@ -20,6 +21,7 @@ import { CreateTestAssessmentProvider } from '../../common/test-assessment-provi
 
 describe('GlobalContextFactoryTest', () => {
     let _mockChromeAdapter: IMock<BrowserAdapter>;
+    let _mockStorageAdapter: IMock<StorageAdapter>;
```
Shobhit1: nit: do you want to get rid of the leading `_` ?

smoralesd: not as part of this PR... keeping the in-file style consistent though

PR #647, opened 2019-05-02
chore: Introduce Yarn as our package manager

Shobhit1:
#### Description of changes

Introduces Yarn as our package manager. 

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000 - n/a
- [ ] Added relevant unit test for your changes. (`npm run test`) - n/a
- [ ] Ran precheckin (`npm run precheckin`) - n/a
- [ ] (UI changes only) Added screenshots/GIFs to description above - n/a
- [ ] (UI changes only) Verified usability with NVDA/JAWS - n/a


README.md
```diff
@@ -28,7 +28,6 @@ You can install the extension from one of the following links
 Please ensure that you have at least the **minimum** recommended versions
 
 -   Node - 10.15.0 (Check by running `node --version`) - This is the version being enforced on our builds
--   NPM - 6.4.1 (Check by running `npm --version`)
```
dbjorge: Users need a global install of yarn, no? Shouldn't we note that here?

README.md
```diff
@@ -48,28 +47,35 @@ Please ensure that you have at least the **minimum** recommended versions
 #### 2. Install packages
 
 -   Install the packages
+
+    ```bash
+    yarn install
```
dbjorge: Just pick one to document, don't make an unfamiliar user have to think/research the difference :)

README.md
Shobhit1: Yes. Let me make a note on that. User do need `yarn` for this to work.

build.yaml
```diff
@@ -9,28 +9,16 @@ jobs:
       steps:
           - template: pipeline/prerequisites.yaml
 
-          - task: Npm@1
-            inputs:
-                command: custom
-                customCommand: run tslint
-            displayName: npm run tslint
+          - script: yarn tslint
+            displayName: check lint errors
```
dbjorge: Is the displayName necessary when using script? If it defaults to displaying the script being run I think that's equally readable and less duplicative

README.md
Shobhit1: Cool. I will pick `yarn install` because its more specific. If anyone else feels otherwise, please let me know!

.prettierignore
```diff
@@ -7,6 +7,7 @@ test-results/
 copyright-header.txt
 LICENSE
 package-lock.json
```
dbjorge: Remove this?

build.yaml
Shobhit1: it just displays `yarn tslint` which didn't look nice. I mean, I thought display name looked better; but thats my opinion. 
Removing `displayName` will make the yaml file a lot shorter though.

dbjorge: I don't have a strong preference, whichever way you like is fine for this PR imo

dbjorge:
Consider adding a [engines entry for yarn in our package.json](https://yarnpkg.com/en/docs/package-json#toc-engines) to enforce a consistent version

dbjorge:
We use `npm-run-all` in a few places in package.json; this probably needs to be updated

dbjorge:
There are a few other places in .md files (and one error message in `browser-factory.ts`) that also have references to the term "npm" that should be updated

Shobhit1:
> We use `npm-run-all` in a few places in package.json; this probably needs to be updated

This is unavoidable. Since `yarn-run-all` package itself runs `npm-run-all`.
https://www.npmjs.com/package/yarn-run-all

dbjorge:
> > We use `npm-run-all` in a few places in package.json; this probably needs to be updated
> 
> This is unavoidable. Since `yarn-run-all` package itself runs `npm-run-all`.
> [npmjs.com/package/yarn-run-all](https://www.npmjs.com/package/yarn-run-all)

Ah, I see, https://github.com/mysticatea/npm-run-all#yarn-compatibility explains that `npm-run-all` understands to use yarn instead of npm if it is itself invoked via yarn instead of npm. So this is fine. [concurrently](https://www.npmjs.com/package/concurrently) would also be an okay option, but I don't see any compelling reason to switch away from `npm-run-all`

README.md
```diff
@@ -28,7 +28,9 @@ You can install the extension from one of the following links
 Please ensure that you have at least the **minimum** recommended versions
 
 -   Node - 10.15.0 (Check by running `node --version`) - This is the version being enforced on our builds
--   NPM - 6.4.1 (Check by running `npm --version`)
+-   Yarn - Version >= v1.12.1 (Check by running `yarn --version`)
+
+> In case you dont have yarn, please install from: [Yarn](https://www.npmjs.com/package/yarn)
```
peterdur: https://www.npmjs.com/package/yarn suggests installing Yarn with `npm i yarn` but also links to a [Yarn install page](https://yarnpkg.com/en/docs/install) which provides a variety of options. Do we want to suggest `npm i -g yarn` or point directly to the Yarn install page instead?

README.md
```diff
@@ -28,7 +28,9 @@ You can install the extension from one of the following links
 Please ensure that you have at least the **minimum** recommended versions
 
 -   Node - 10.15.0 (Check by running `node --version`) - This is the version being enforced on our builds
--   NPM - 6.4.1 (Check by running `npm --version`)
+-   Yarn - Version >= v1.12.1 (Check by running `yarn --version`)
```
dbjorge: Why 1.12.1? Current is 1.15.2

build.yaml
dbjorge: I tried without displayname in axe-sarif-converter and it actually just shows as "CmdLine", so we should definitely keep the displayNames

README.md
Shobhit1: Fixed.

README.md
Shobhit1: I changed the link to point to `yarn's` own website instead of `npm` listing to avoid this confusion.

.prettierignore
Shobhit1: Done. 👍

smoralesd:
> There are a few other places in .md files (and one error message in browser-factory.ts) that also have references to the term "npm" that should be updated

Found some files mentioning `npm` that I don't see on the PR's file changes tab:
- docs\setup-tslint-on-vs-code.md
- docs\workflow.md

Shobhit1:
> > There are a few other places in .md files (and one error message in browser-factory.ts) that also have references to the term "npm" that should be updated
> 
> Found some files mentioning `npm` that I don't see on the PR's file changes tab:
> 
> * docs\setup-tslint-on-vs-code.md
> * docs\workflow.md

updated.

PR #649, opened 2019-05-03
chore: Update typescript & q versions

smoralesd:
#### Description of changes

Update typescript & q version.
Update types:
- Using `MessageEventSource` instead of `Window`
- Using `HTMLCollectionOf` instead of `NodeListOf`
- Casting `'*'` as `string & Transferable[]` (on `postMessage`)

All the type changes came from typescript type updates.

#### Pull request checklist


Shobhit1:
I think we should wait till #647 is merged as we got rid of `package-lock` 😄

smoralesd:
> I think we should wait till #647 is merged as we got rid of `package-lock` 😄

Doesn't seem necessary to explicitly wait. Whoever goes second needs to update his PR.

src/injected/tab-stops-listener.ts
```diff
@@ -131,7 +131,7 @@ export class TabStopsListener {
     @autobind
     private onStopListenToTabStops(): void {
         this.removeListeners();
-        const iframes: NodeListOf<HTMLIFrameElement> = this.getAllFrames();
```
lisli1: I noticed that in other places, `NodeListof<HTMLIFrameElements>` is replaced with `HTMLCollectionOf<HTMLIFrameElements>`, so I am wondering why that is not the case here?

src/tests/unit/tests/injected/frameCommunicators/window-message-handler.test.ts
```diff
@@ -9,6 +9,7 @@ import { FrameMessageResponseCallback, WindowMessageHandler } from '../../../../
 import { WindowMessageMarshaller } from '../../../../../injected/frameCommunicators/window-message-marshaller';
 
 describe('WindowMessageHandlerTests', () => {
+    const starOrigin = '*' as (string & Transferable[]);
```
waabid: What is this Transferable[] about? Should we establish a type for this; seems to be used in a few places.

smoralesd: Not sure what it is, it came up from an error while updating typescript so I'm just casting to the type to fix all errors.

src/injected/tab-stops-listener.ts
smoralesd: This seems outdated, I don't see the explicit type for the const

src/tests/unit/tests/injected/frameCommunicators/window-message-handler.test.ts
smoralesd: `Transferable` documentation [here](https://developer.mozilla.org/en-US/docs/Web/API/Transferable)

smoralesd:
@microsoft/accessibility-insights-web-code-owners 
@dbjorge was looking at this issue and found out we're getting "out of memory" issues. This are related to typescript 3.3.x and 3.4.x so I'm closing this PR for now as it requires some more investigation.

PR #650, opened 2019-05-03
refactor: introduce details-view-action-creator

smoralesd:
#### Description of changes

Refactor `ActionCreator` by extracting functions related to `DetailsViewActions` (no other actions involved) to a new `DetailsViewActionCreator`.

#### Pull request checklist

   - does not apply
   - does not apply
   - does not apply


src/tests/unit/tests/background/actions/details-view-action-creator.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        detailsViewControllerMock.verify(controller => controller.showDetailsView(tabId), Times.once());
+        telemetryEventHandlerMock.verify(handler => handler.publishTelemetry(SETTINGS_PANEL_OPEN, payload), Times.once());
+    });
+
+    test('on SettingsPanel.ClosePanel', () => {
+        const closeSeetingsPanelMock = Mock.ofType<Action<null>>();
+
+        detailsViewActionsMock.setup(actions => actions['closeSettingsPanel']).returns(() => closeSeetingsPanelMock.object);
+
+        const payload: BaseActionPayload = {
```
waabid: Seems to be the same payload as above; maybe you can move this to beforeEach.

src/tests/unit/tests/background/actions/details-view-action-creator.test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        setSelectedPanelMock.verify(action => action.invoke(payload), Times.once());
+    });
+
+    test('on Visualization.DetailsView.GetState', () => {
+        const getCurrentStateMock = Mock.ofType<Action<null>>();
+
+        detailsViewActionsMock.setup(actions => actions['getCurrentState']).returns(() => getCurrentStateMock.object);
+
+        registerCallbackMock
```
waabid: Might not be a bad idea to move this to a private method; seems duplicated in this test file.

smoralesd: the setting of the `registerCallbackMock`?? we have 3 different ways of calling the listener (no params, 1 and 2 params). Tests are short enough that I don't think this is a big deal. I'm planning on coming back to the action creator tests (once I finish breaking up the current action creator and global action creator) and apply something similar to what we have for the original action creator, i.e. an action creator tester class. But that seems to be out of scope for this PR.

src/tests/unit/tests/background/actions/details-view-action-creator.test.ts
smoralesd: Updated.

PR #656, opened 2019-05-07
refactor: extract message prefix to const

smoralesd:
#### Description of changes

Extract common prefix to decrease duplication.

#### Pull request checklist

   - does not apply
   - does not apply
   - does not apply
   - does not apply
   - does not apply

### DIFF
--- a/src/common/messages.ts
+++ b/src/common/messages.ts
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
 export interface StateMessages {
     GetCurrentVisualizationToggleState: string;
     GetCurrentVisualizationResultState: string;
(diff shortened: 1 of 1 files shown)


src/common/messages.ts
```diff
@@ -49,150 +48,148 @@ export interface VisualizationMessages {
...
 
+const messagePrefix = 'insights';
+
 export class Messages {
     public static readonly Visualizations: VisualizationMessages = {
         Common: {
-            Toggle: 'insights/visualization/toggle',
-            ScanCompleted: 'insights/visualization/scanCompleted',
-            ScrollRequested: 'insights/visualization/scrollRequested',
+            Toggle: `${messagePrefix}/visualization/toggle`,
```
haonliu: would it be better if we use a function to add prefix to all messages in order to reduce dupilication?

smoralesd: Any specific suggestions?

PR #664, opened 2019-05-09
refactor: convert layered details dialog component to named sfc

smoralesd:
#### Description of changes

Refactor `LayeredDetailsDialogComponent`:

- Update test to use enzyme's `shallow` renderer function
- Update the component to be a `NamedSFC` instead of a `class` extending `React.Component`
- Update the component deps to be a type instead of an interface 

#### Pull request checklist

   - does not apply
   - does not apply
   - does not apply

### DIFF
--- a/src/tests/unit/tests/injected/__snapshots__/layered-details-dialog-component.test.tsx.snap
+++ b/src/tests/unit/tests/injected/__snapshots__/layered-details-dialog-component.test.tsx.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`render component when shadow dom is disabled 1`] = `
+exports[` render component when shadow dom is disabled 1`] = `
 <LayerHost
   dir="ltr"
   id="insights-dialog-layer-host"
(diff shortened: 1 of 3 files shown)


src/tests/unit/tests/injected/__snapshots__/layered-details-dialog-component.test.tsx.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`LayeredDetailsDialogComponent render component when shadow dom is disabled 1`] = `
+exports[` render component when shadow dom is disabled 1`] = `
```
peterdur: Test name is missing from snapshot - does
```
describe(LayeredDetailsDialogComponent, () => {
```
in ...test.tsx below need to have quotes around `LayeredDetailsDialogComponent`?

smoralesd: Good catch. I notice the change but didn't realize the reason: it seems that for classes (actual `class` ...) jest gets the name from the class but in this PR, I changed the component to be a const = NamedSFC and thus jest can't get the name from the const!

PR #667, opened 2019-05-10
Spike: Using CSS Modules in React

waabid:
#### Description of changes

This change introduces the use of CSS modules in react.

Please look at this example commit for how to add styles in the future:

https://github.com/microsoft/accessibility-insights-web/commit/1adba2c8ada50b4a50e5e7f2a90bdb760198d7f9

Here's the basic idea:

We should be able to link our code/ts files with our styles so that when styles are modified (specifically when we do not link classes correctly), we break at compile time rather than finding out about it at run-time. Additionally, the use of CSS modules will/should encourage us to have less dead CSS and smaller CSS files (that will compiled to larger CSS files by webpack).

This PR is doing this by using webpack loaders. The first is sass-loader which will transpile our scss files to css files, which will then be consumed by css loader which resolves our css imports, followed by a typescript loader to generate typings for said CSS. Finally, we use the extract loader suggested by webpack to create the bundled-per-entry css files.

Note: while the 3/4 loaders are commonly used, the typescript loader is not as heavily used as other packages. However, it is getting recent updates/releases and allows us to consume css-loader's newest version which others do not (specifically the most consumed typescript loader has not been updated in 2 years). It is also a fairly simple loader, so I feel fairly comfortable using it. Finally, currently this would require us to use header copy script as the file generated does not contain it; I have opened a feature request on the typescript loader for this so in the future we do not have to do this.

Here is a screen-cap of the generated CSS in use:

image

#### Pull request checklist


dbjorge:
The example commit includes checking in generated code, which I'd prefer to see us avoid as a rule. Did you do it that way for example purposes or would you expect we could avoid doing that in actual practice?

src/tests/jest.common.config.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
         'office-ui-fabric-react/lib/(.*)$': 'office-ui-fabric-react/lib-commonjs/$1',
         '@uifabric/utilities': '@uifabric/utilities/lib-commonjs',
         '@uifabric/styling': '@uifabric/styling/lib-commonjs',
+        '\\.(scss)$': 'identity-obj-proxy',
```
dbjorge: This is unclear enough that it'd be nice to include a comment explaining it (maybe a link to an article explaining the technique?)

package.json
```diff
@@ -63,19 +65,23 @@
...
         "node-sass": "^4.12.0",
         "npm-run-all": "^4.1.5",
         "preprocess": "^3.1.0",
         "prettier": "^1.17.0",
         "puppeteer": "^1.15.0",
+        "sass-loader": "^7.1.0",
         "script-loader": "0.7.2",
         "serve-static": "^1.13.2",
         "source-map-loader": "^0.2.4",
+        "style-loader": "^0.23.1",
```
dbjorge: I'm not seeing where this dep is used; is it leftover from other experimenting?

waabid: Correct!

src/tests/jest.common.config.js
waabid: Good catch! I'll add some info.

waabid:
Closing this PR and opening a new one with an updated description/notes.

PR #668, opened 2019-05-10
refactor: introduce notification, runtime & injection adapter

smoralesd:
#### Description of changes

More refactoring on the browser/chrome adapter.
- introduce notification adapter (to create notification on the task bar)
- introduce runtime adapter (a couple of functions that explicitly use the word 'runtime' in their name)
- introduce injection adapter (to inject javascript and css into the target page)

**next in the list**:
We should keep breaking down `BrowserAdapter`. I can see this adapter as candidates for future refactorings:
- windows & tab adapter (to manage, create and query windows and tabs)
- event listener adapter (to add/remove listeners to specific events)
- communication adapter (to send messages between background, tabs and frames)

#### Pull request checklist

   - does not apply
- [x] ~~Added~~Updated relevant unit test for your changes. (`npm run test`)
   - does not apply
   - does not apply

### DIFF
--- a/src/background/browser-adapters/injector-adapter.ts
+++ b/src/background/browser-adapters/injector-adapter.ts
@@ -1,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export type InjectorAdapter = {
+    injectJs(tabId, file: string, callback: Function): void;
+    injectCss(tabId, file: string, callback: Function): void;
+};
--- a/src/background/browser-adapters/runtime-adapter.ts
+++ b/src/background/browser-adapters/runtime-adapter.ts
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 export type RuntimeAdapter = {
-    getRunTimeId(): string;
+    getRuntimeId(): string;
     getRuntimeLastError(): chrome.runtime.LastError;
 };
--- a/src/background/browser-adapters/runtime-adapter.ts
+++ b/src/background/browser-adapters/runtime-adapter.ts
@@ -1,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export type RuntimeAdapter = {
+    getRunTimeId(): string;
+    getRuntimeLastError(): chrome.runtime.LastError;
+};
--- a/src/background/background-init.ts
+++ b/src/background/background-init.ts
@@ -36,9 +36,9 @@ import { TelemetryStateListener } from './telemetry/telemetry-state-listener';
 import { UserStoredDataCleaner } from './user-stored-data-cleaner';
 
 declare var window: Window & InsightsFeatureFlags;
-const browserAdapter = new ChromeAdapter();
-const urlValidator = new UrlValidator(browserAdapter);
-const backgroundInitCleaner = new UserStoredDataCleaner(browserAdapter);
+const chromeAdapter = new ChromeAdapter();
+const urlValidator = new UrlValidator(chromeAdapter);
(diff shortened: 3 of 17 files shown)


src/background/browser-adapters/runtime-adapter.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export type RuntimeAdapter = {
+    getRunTimeId(): string;
```
peterdur: Capitalization of Runtime on line 4 (RunTime) is inconsistent with lines 3 and 5.
```suggestion
    getRuntimeId(): string;
```

src/background/browser-adapters/runtime-adapter.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export type RuntimeAdapter = {
```
peterdur: Currently `getRuntimeId` seems to only be used from `DetailsViewController` and `getRuntimeLastError` seems to only be called from `TabController` - this seems to be contrary to interface segregation.

smoralesd: Done

src/background/browser-adapters/runtime-adapter.ts
smoralesd: Done

src/background/browser-adapters/runtime-adapter.ts
smoralesd: Should I break `RuntimeAdapter` into two types, one function each? My original approach for this type was on the semantic of what we are using from chrome API more than who's using it. I do see your point though, I'm just not sure I like the idea to break this into one-function-type

peterdur: I suppose that if we were to follow ISP strictly, we'd essentially have a `DetailsViewControllerBrowserAdapter` interface that contains everything that `DetailsViewController` consumes (that isn't otherwise in a common interface that DVC fully consumes), and a `TabControllerBrowserAdapter` that contains everything that `TabController` consumes (that isn't otherwise in a common interface that TC fully consumes)... not sure if we need/want to get to that point in a single refactoring.

smoralesd: I agree I would prefer to do this on a separated PR as the current `RuntimeAdapter` is already small than the original `ChromeAdapter`, i.e. the classes that you mention "depend" on less things that they don't use, or from another point of view, this is already better than what we had before

src/background/browser-adapters/injector-adapter.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export type InjectorAdapter = {
+    injectJs(tabId, file: string, callback: Function): void;
```
dbjorge: I'm worried that this may be a leaky abstraction; it essentially cannot represent a unit of functionality that can be used in isolation because the opaque tabId must come from a different, coupled implementation of a different interface. I think perhaps it would be more appropriate for these to either be methods on a "tab" interface or to be part of the same interface that the tabId comes from.

src/background/browser-adapters/runtime-adapter.ts
```diff
@@ -0,0 +1,6 @@
...
+    getRuntimeId(): string;
+    getRuntimeLastError(): chrome.runtime.LastError;
```
dbjorge: Similar to above comment for `getRuntimeLastError`; for our purposes, a call to `getRuntimeLastError()` is only meaningful in the context of a coupled implementation of `getSelectedTabInWindow`, so I don't think it makes sense to decouple their interfaces.

PR #671, opened 2019-05-10
refactor: move file issue behavior to the service

smoralesd:
#### Description of changes

As part of solving issue 588 (not linking so it doesn't get close automatically), I'm moving the responsibility of filing a bug from `IssueFilingControllerImpl` to each service. This way, we give each service the flexibility of filing issue as they need (as opposed to: each service file issues by generating a url which we -basically- use for an HTTP GET request). This way, we may be able to use HTTP POST request or not even HTTP request but system calls (like open notepad for example)

#### Pull request checklist

   - does not apply
   - does not apply
   - does not apply

### DIFF
--- a/src/issue-filing/common/create-file-issue-handler.ts
+++ b/src/issue-filing/common/create-file-issue-handler.ts
@@ -7,8 +7,8 @@ import { IssueFilingServicePropertiesMap } from '../../common/types/store-data/u
 import { IssueFilingUrlProvider } from '../types/issue-filing-service';
 
 export const createFileIssueHandler = <Settings>(
-    urlProvider: IssueFilingUrlProvider<Settings>,
-    settingsGetter: (data: IssueFilingServicePropertiesMap) => Settings,
+    getUrl: IssueFilingUrlProvider<Settings>,
+    getSettings: (data: IssueFilingServicePropertiesMap) => Settings,
 ) => {
     return (
         browserAdapter: BrowserAdapter,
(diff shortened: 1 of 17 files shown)


src/issue-filing/common/create-file-issue-handler.ts
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License.
+import { BrowserAdapter } from '../../background/browser-adapters/browser-adapter';
+import { EnvironmentInfo } from '../../common/environment-info-provider';
+import { CreateIssueDetailsTextData } from '../../common/types/create-issue-details-text-data';
+import { IssueFilingServicePropertiesMap } from '../../common/types/store-data/user-configuration-store';
+import { IssueFilingUrlProvider } from '../types/issue-filing-service';
+
+export const createFileIssueHandler = <Settings>(
+    urlProvider: IssueFilingUrlProvider<Settings>,
+    settingsGetter: (data: IssueFilingServicePropertiesMap) => Settings,
```
devaradhanm: should this be getSettings? method names should be verb. this is not a class

smoralesd: Make sense

src/issue-filing/common/create-file-issue-handler.ts
```diff
@@ -7,18 +7,18 @@ import { IssueFilingServicePropertiesMap } from '../../common/types/store-data/u
...
         browserAdapter: BrowserAdapter,
         servicePropertiesMap: IssueFilingServicePropertiesMap,
         issueData: CreateIssueDetailsTextData,
         environmentInfo: EnvironmentInfo,
     ): void => {
-        const serviceConfig = settingsGetter(servicePropertiesMap);
+        const serviceConfig = getSettings(servicePropertiesMap);
 
-        const url = urlProvider(serviceConfig, issueData, environmentInfo);
+        const url = getUrl(serviceConfig, issueData, environmentInfo);
```
peterdur: Oh, that's much nicer!

PR #675, opened 2019-05-14
refactor:  moving types to their own files

smoralesd:
#### Description of changes

On some files we have a class and a type (or several types). In some cases this means you can see circular imports which is not nice. On this PR I'm moving some types to their own files to avoid this "circular dependencies" (they are not really circular dependencies as those are catch by `tsc`)

#### Pull request checklist

   - does not apply
   - does not apply
   - does not apply
   - does not apply
   - does not apply


src/DetailsView/components/overview-content/overview-help-links.tsx
```diff
@@ -2,13 +2,14 @@
...
+import { ExternalLink, ExternalLinkDeps } from '../../../common/components/external-link';
 import { NamedSFC } from '../../../common/react/named-sfc';
 import { HyperlinkDefinition } from '../../../views/content/content-page';
-import { HelpLinkDeps } from './overview-help-section';
+
+export type HelpLinksDeps = ExternalLinkDeps;
 
 export interface HelpLinksProps {
-    deps: HelpLinkDeps;
+    deps: HelpLinksDeps;
```
Shobhit1: Can we directly use `ExternalLinkDeps` here? I dont think it will matter much.

src/DetailsView/components/overview-content/overview-help-section.tsx
```diff
@@ -2,12 +2,11 @@
...
 import * as React from 'react';
 
-import { ExternalLinkDeps } from '../../../common/components/external-link';
 import { NamedSFC } from '../../../common/react/named-sfc';
 import { HyperlinkDefinition } from '../../../views/content/content-page';
-import { HelpLinks } from './overview-help-links';
+import { HelpLinks, HelpLinksDeps } from './overview-help-links';
 
-export type HelpLinkDeps = ExternalLinkDeps;
+export type HelpLinkDeps = HelpLinksDeps;
```
Shobhit1: same here.

src/DetailsView/components/overview-content/overview-help-links.tsx
smoralesd: The patter we're trying to follow is each component declare its own, direct dependencies and we use the '&' operator to bring the children's dependencies, In case the component the component doesn't have any direct dependency, we still declare the type. I guess this make for less changes if we need to add direct dependencies in the future and also it's nice to have one pattern only.

src/DetailsView/components/overview-content/overview-help-section.tsx
smoralesd: same answer.

peterdur: Seems like this should be `OverviewHelpSectionDeps` (and props at line 11 should be `OverviewHelpSectionProps`)?

src/DetailsView/components/overview-content/overview-help-links.tsx
```diff
@@ -2,13 +2,14 @@
...
     linkInformation: HyperlinkDefinition[];
 }
```
peterdur: Should we change the name of the file to be help-links to match the primary thing contained in the file (there's no overview here)?

src/DetailsView/components/overview-content/overview-help-links.tsx
Shobhit1: So, I guess my question here will be then: do we need to export this type that is only used here?

src/DetailsView/components/overview-content/overview-help-section.tsx
peterdur: If this is the deps for `OverviewHelpSection` below, should this be called `OverviewHelpSectionDeps` (and props at line 11 should be `OverviewHelpSectionProps`)?

smoralesd: That make sense, except I was not refactoring in a way that I was changing this component directly (and thus, I didn't really touch the left assign here). I understand that we want to go ahead and clean up more things but I would also like for reviewers to see the increased value of applying a PR like this where, yes, you can find related places where you can do more clean up, those places are not in such a good shape -so we want to clean up those places at some point- but are not really being changed on the PR. It seems to this kind of make for bigger than needed PR's and also,  take away part of the speed of using small PR's.

All that being said, I updated this file (and a second one) according to this comments.

src/DetailsView/components/overview-content/overview-help-links.tsx
smoralesd: Same thoughts here as with the comment on `overview-help-section` deps and props types.

src/tests/unit/tests/DetailsView/components/overview-content/help-links.test.tsx
```diff
@@ -4,14 +4,17 @@ import { shallow } from 'enzyme';
...
+import { HelpLinks } from '../../../../../../DetailsView/components/overview-content/help-links';
+import {
+    OverviewHelpSectionDeps,
+    OverviewHelpSectionProps,
+} from '../../../../../../DetailsView/components/overview-content/overview-help-section';
 
-describe('OverviewHelpLinks', () => {
-    const deps = Mock.ofType<HelpLinkDeps>().object;
+describe('HelpLinks', () => {
+    const deps = Mock.ofType<OverviewHelpSectionDeps>().object;
```
dbjorge: Did you mean for this test of HelpLinks to have a dependency on OverviewHelpSectionProps/Deps?

src/tests/unit/tests/DetailsView/components/overview-content/overview-content-container.test.tsx
```diff
@@ -30,7 +30,7 @@ describe('OverviewContainer', () => {
         actionInitiators: {
             openExternalLink,
         },
-    } as HelpLinkDeps;
+    } as OverviewHelpSectionDeps;
```
dbjorge: The name of the variable echoes the name of the type here; should it be updated alongside the type update?

src/tests/unit/tests/DetailsView/components/overview-content/help-links.test.tsx
smoralesd: Nop, my mistake. Good catch!!!

src/tests/unit/tests/DetailsView/components/overview-content/overview-content-container.test.tsx
smoralesd: Make sense.

PR #677, opened 2019-05-15
fix: update subtitle text

smoralesd:
#### Description of changes

Update subtitle for automated checks (on FastPass) to match new requirement. Update container size to control text wrapping

**text content**
01 - updated subtitle

#### Pull request checklist

   - does not apply
   - does not apply
   - does not apply
   - does not apply

### DIFF
--- a/src/DetailsView/Styles/detailsview.scss
+++ b/src/DetailsView/Styles/detailsview.scss
@@ -892,7 +892,6 @@ div.insights-file-issue-details-dialog-container {
                     }
                     .issues-table-subtitle {
                         color: $secondary-text;
-                        max-width: 561px;
                         font-size: 14px;
                         margin-top: 4px;
                         margin-bottom: 16px;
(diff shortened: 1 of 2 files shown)


smoralesd:
make sure @ferBonnin has a chance to review before you merge

smoralesd:
@msft-github-bot make sure @ferBonnin has a chance to review before you merge

smoralesd:
@msft-github-bot do not merge unless @ferBonnin approves.

ferBonnin:
is the line break set always? would be better to have assessment in the same line

smoralesd:
> is the line break set always? would be better to have assessment in the same line

@ferBonnin There is no line break. The text is wrapping because we have set the max width of the text container (it's like this from master, not sure that was intended or not). I updated the max width so the text itself will use the available space and wrap accordingly (please see the animated gif)

PR #686, opened 2019-05-20
feat: build electron app

smoralesd:
#### Description of changes

This is mostly a PoC for building an electron app on our current repo. We may still want to move this code to master.

- Add super basic electron app that load AU before on a new window
- Using our current webpack/grunt/yarn scripts to set up building and running the electron app

#### Pull request checklist

   - **n/a**
   - **n/a**
   - **n/a**
   - **n/a**
   - **n/a**


Shobhit1:
This looks good to me.

Gruntfile.js
```diff
@@ -208,40 +209,78 @@ module.exports = function(grunt) {
...
-                        },
-                    ],
-                },
-            },
             clean: {
                 [targetName]: dropPath,
             },
         });
+
+        if (targetName === 'electron') {
```
dbjorge: The two branches of this if look pretty redundant; perhaps it would be better to replace the "debug" config option with a "bundleFolder" or something and use the original version with 		`config.options.bundleFolder` replacing the ternary that was originally on line 215?

(I think this would also simplify the drop task registration further down slightly)

targets.config.js
```diff
@@ -102,4 +102,16 @@ module.exports = {
...
         },
     },
+    electron: {
+        config: {
+            options: {
+                ...internalOptions,
+                ...icons.dev,
+                debug: true,
+                extensionFullName: 'Accessibility Insights for Web - Dev',
+                telemetryBuildName: 'Dev',
```
dbjorge: Should we use a unique name here?

webpack.config.js
```diff
@@ -28,42 +28,70 @@ const commonEntryFiles = {
...
+            setImmediate: false,
+        },
+        performance: {
+            // We allow higher-than-normal sizes because our users only have to do local fetches of our bundles
+            maxEntrypointSize: 10 * 1024 * 1024,
+            maxAssetSize: 10 * 1024 * 1024,
+        },
+    };
+
+    if (isElectron) {
```
dbjorge: Why do this in createCommonConfig rather than overriding it in electronConfig?

targets.config.js
smoralesd: updated!

webpack.config.js
smoralesd: Updated!

Gruntfile.js
smoralesd: The result of applying this idea is so much better. Lest code changes, less bifurcations. Thanks.

webpack.config.js
```diff
@@ -28,42 +28,69 @@ const commonEntryFiles = {
...
+const electronConfig = {
+    ...createCommonConfig(electronEntryFiles, true),
+    name: 'electron',
+    mode: 'development',
+    output: {
+        path: path.join(__dirname, 'extension/electronBundle'),
+        filename: '[name].bundle.js',
     },
-    plugins: commonPlugins,
     node: {
```
dbjorge: is omitting the `setImmediate: false` from commonConfig intentional?

webpack.config.js
```diff
@@ -28,42 +28,69 @@ const commonEntryFiles = {
...
+            maxEntrypointSize: 10 * 1024 * 1024,
+            maxAssetSize: 10 * 1024 * 1024,
+        },
+    };
+
+    return baseConfig;
+};
+
+const electronConfig = {
+    ...createCommonConfig(electronEntryFiles, true),
```
dbjorge: The true is unnecessary now, right?

webpack.config.js
```diff
@@ -28,42 +28,69 @@ const commonEntryFiles = {
...
-        maxEntrypointSize: 10 * 1024 * 1024,
-        maxAssetSize: 10 * 1024 * 1024,
+    optimization: {
+        splitChunks: false,
     },
+    target: 'electron-main',
 };
 
 const devConfig = {
-    ...commonConfig,
```
dbjorge: Would it be more consistent with how the individual configs customize other config-specific properties to do

```ts
const individualConfig = {
    ...commonConfig,
    entry: individualEntryFiles,
    someConfigSpecificProperty: 'individual-setting',
    // ...
}
```

instead of 

```ts
const individualConfig = {
    ...createCommonConfig(individualEntryFiles),
    someConfigSpecificProperty: 'individual-setting',
    // ...
}
```

PR #688, opened 2019-05-21
Spike: Using CSS Modules in React (with generated and ignored typings)

waabid:
#### Description of changes

This change introduces the use of CSS modules in react.

Please look at this example commit for how to add styles in the future:

https://github.com/microsoft/accessibility-insights-web/pull/688/commits/4afae527c21ecb473ff9352820235c5b2e2f9c01

Here's the basic idea:

We should be able to link our code/ts files with our styles so that when styles are modified (specifically when we do not link classes correctly), we break at compile time rather than finding out about it at run-time. Additionally, the use of CSS modules will/should encourage us to have less dead CSS and smaller CSS files (that will compiled to larger CSS files by webpack).

This PR is doing this in two parts:

1. Use typed-scss-modules to generate typings for the scss files to link our styles with typed code (i.e. typescript helps prevent code breaks based on CSS)
2. In webpack: use sass-loader, css-loader, and MiniCssExtractPlugin to take scss files referenced in our code (i.e. tsx files), which now have typings due to step 1, and create bundled css files per entry point.

Note: the typed-scss-modules package is fairly new and does not have a lot of users (~1500 downloads per week). However, I feel comfortable taking it since it's not too complicated and if we needed to, we could swap out for a sass parser with typed-css-modules (which is a bit more used).

Here is a screen-cap of the generated CSS in use:

image

#### Pull request checklist


src/DetailsView/Styles/detailsview.scss
```diff
@@ -215,8 +215,8 @@ div.insights-file-issue-details-dialog-container {
 }
 
 .failure-instance-panel {
-    .action-and-cancel-buttons-component {
-        padding-top: 12px;
+    .observed-failure-textfield {
```
waabid: This is modified since the failure instance panel component does not hold responsibility for action and cancel buttons styling; as a result this style is modified to something that this component DOES control.

src/tests/end-to-end/tests/details-view/preview-features.test.ts
```diff
@@ -76,7 +76,7 @@ describe('Preview Features Panel', () => {
 
         const detailsViewPage = await waitForDetailsViewWithPreviewFeaturesPanel(browser, popupPage, targetTabId);
 
-        await detailsViewPage.waitForSelector(detailsViewSelectors.noPreviewFeaturesMessage);
+        await detailsViewPage.waitForSelector(detailsViewSelectors.previewFeaturesPanel);
```
waabid: This is an important change that I'm open to discussing.

Since there are some E2E tests that are dependent on css classnames, it might be a good idea to think about the pattern in which we do this. Should we start adding IDs for end to end tests, since the generated css selectors will not be something that jest can reproduce* (since it is not dependent on compiled code); the IDs would not be used for CSS but only for E2E tests, which might not be ideal for some.

Alternatively, we can also not use wait for selector and rely on another API.

And finally, there is also using this: https://www.npmjs.com/package/jest-transform-css.

Personally, I think the first solution is the easiest and would allow us to move forward without investing too much time.

* this is due to the suggested guidance on for using jest with CSS modules: https://jestjs.io/docs/en/webpack#mocking-css-modules.

Shobhit1: I think there is also something called `automation-id` which is similar to adding `id` but that is specific to our automated e2e tests. We can also check those and how that can fit in our e2e. Just my 2 cents here.

Gruntfile.js
```diff
@@ -169,7 +170,7 @@ module.exports = function(grunt) {
             },
             scss: {
                 files: ['src/**/*.scss'],
-                tasks: ['sass', 'copy:styles', 'embed-styles:code', 'drop:dev'],
+                tasks: ['exec:generate-scss-typings', 'sass', 'copy:styles', 'embed-styles:code', 'drop:dev'],
```
dbjorge: Isn't this redundant with `yarn watch:scss`?

src/tests/unit/tests/DetailsView/components/__snapshots__/action-and-cancel-buttons-component.test.tsx.snap
```diff
@@ -2,19 +2,19 @@
 
 exports[`ActionAndCancelButtonsComponent render with primary button href == null 1`] = `
 <div
-  className="action-and-cancel-buttons-component"
+  className="actionAndCancelButtonsComponent"
```
dbjorge: Is it intentional that it's camelCasing instead of kebab-casing? (Do we care?)

src/tests/end-to-end/tests/details-view/preview-features.test.ts
dbjorge: I like option 1 (just using id). automation-id is fine if there are id conflicts but I don't really see a reason to prefer it over id for the cases where id would work

src/tests/unit/tests/DetailsView/components/__snapshots__/action-and-cancel-buttons-component.test.tsx.snap
dbjorge: Oh, sorry, I understand now; this isn't the "real" output, it's the output from the identity-obj-proxy magic

build.yaml
```diff
@@ -18,6 +18,9 @@ jobs:
           - script: yarn copyrightheaders
             displayName: check copyrightheaders
 
+          - script: yarn generate-scss-typings
```
dbjorge: Why is this needed as a separate build.yaml step? I'd expect unit tests to be using the identity-object-proxy thing and for everything else to be having this happen via a `yarn build` task, no?

src/tests/end-to-end/tests/details-view/preview-features.test.ts
peterdur: There's also the possibility of continuing to specify the CSS class in the test code as a string (i.e. not using CSS modules to get the class name), yes? Or does CSS modules munge the class name?

waabid: @peterdur unfortunately there is no way to do this for E2E tests. This is because E2E tests depend on the produced bundle code when comparing screenshots VERSUS Jest which will be relying on `identity-object-proxy` to get the class-names, so in that scenario there would be soft-coupling.

Admittedly, we already have this but ideally we could improve this by relying on IDs that are accessible and guaranteed to be the same.

build.yaml
waabid: Correct! Will remove.

src/tests/unit/tests/DetailsView/components/__snapshots__/action-and-cancel-buttons-component.test.tsx.snap
waabid: Precisely! I think I may cover this in an engineering discussion. 😄

Gruntfile.js
waabid: Correct! Removed!

Gruntfile.js
```diff
@@ -335,6 +336,22 @@ module.exports = function(grunt) {
     grunt.registerTask('build-assets', ['sass', 'copy:code', 'copy:styles', 'embed-styles:code', 'copy:images']);
 
     // Main entry points for npm scripts:
+<<<<<<< HEAD
```
haonliu: unresolved merge conflict?

build.yaml
waabid: I've re-added this as per what I discovered yesterday.

For PR clarity:
We assumed that using the identity-obj-proxy would ensure that typings would not be required. This is not true.

The identity-obj-proxy is for Jest to be able to consume packages that it would normally not be able to consume (in this case scss). However, this has no links to ts-jest, which we use to transform our TS code. As a result, when the typings do not exist for ts-jest to validate, it will complain about not being able to find the module.

The suggested solution to this that I've seen online is to use the isolatedModules flag on ts-jest but that is not ideal as it disables type-checking on test files for jest runs; not necessarily the worst thing for us, since our webpack/build process would type-check those files anyways, but still not ideal.

PR #689, opened 2019-05-22
fix: switcher state is updated correctly (on component did update)

### LINKED ISSUE #659: Details view switcher is not updated when view changed from popup view
## Describe the bug

The details view switcher does not update when you click on fast pass or assessments in the popup view.

**To Reproduce**

1. Open popup view.
2. Click FastPass.
3. Open popup view again.
4. Click Assessment.

Note how the switcher is not updated and still says fast pass.

## Expected behavior

The switcher should state assessment when switched.

## Screenshots

switchernotupdated

- OS: Windows
- Version 2.3.0

Yes

waabid:
#### Description of changes

A change was made such that the state is being represented in the control instead of the props for the switcher. This was due to the nature of React Flux and how it works with accessibility concerns. However, a part was missed: when props are updated, the state should be updated (we have this in previous uses of the pattern). This is _that_ change.

switchernotupdatedbugfix

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #659

### DIFF
--- a/src/DetailsView/components/switcher.tsx
+++ b/src/DetailsView/components/switcher.tsx
@@ -29,7 +29,7 @@ export class Switcher extends React.Component<SwitcherProps, SwitcherState> {
     }
 
     public componentDidUpdate(prevProps: Readonly<SwitcherProps>): void {
-        if (isEqual(prevProps, this.props) === false) {
+        if (prevProps.pivotKey !== this.props.pivotKey) {
             this.setState(() => ({ selectedKey: this.props.pivotKey }));
         }
     }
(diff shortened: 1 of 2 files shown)


src/DetailsView/components/switcher.tsx
```diff
@@ -27,6 +28,12 @@ export class Switcher extends React.Component<SwitcherProps, SwitcherState> {
         this.state = { selectedKey: props.pivotKey };
     }
 
+    public componentDidUpdate(prevProps: Readonly<SwitcherProps>): void {
+        if (isEqual(prevProps, this.props) === false) {
```
Shobhit1: nit: do we want to check for all the props and perform a deep comparison here or can it be improved by comparing some specific prop or property?

PR #695, opened 2019-05-22
feat: build electron application from our existing configurations

Shobhit1:
#### Description of changes

Add basic Electron build stuff.
- loads AU home page in an electron window
- uses our webpack/grunt configs to build electron app as well.

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000 - n/a
- [ ] Added relevant unit test for your changes. (`yarn test`) - n/a


webpack.config.js
```diff
@@ -28,42 +28,69 @@ const commonEntryFiles = {
...
-                        loader: 'ts-loader',
-                        options: {
-                            transpileOnly: true,
-                            experimentalWatchApi: true,
+const electronEntryFiles = {
+    main: [path.resolve(__dirname, 'src/electron/main/main.ts')],
+    injected: [path.resolve(__dirname, 'src/injected/stylesheet-init.ts'), path.resolve(__dirname, 'src/injected/client-init.ts')],
+};
+
+const createCommonConfig = entry => {
```
peterdur: Echoing Dan's comment from #686:
`createCommonConfig` is mostly static, except for copying the `entry` parameter to the output as the value for `entry`. Is it more idiomatic to restore `commonConfig` to being a `const`, and change consumers from
```
const electronConfig = {
    ...createCommonConfig(electronEntryFiles),
    name: 'electron',
```
to
```
const electronConfig = {
    ...commonConfig,
    entry: electronEntryFiles,
    name: 'electron',
```

webpack.config.js
```diff
@@ -28,42 +28,69 @@ const commonEntryFiles = {
...
+    name: 'electron',
+    mode: 'development',
+    output: {
+        path: path.join(__dirname, 'extension/electronBundle'),
+        filename: '[name].bundle.js',
     },
-    plugins: commonPlugins,
     node: {
-        setImmediate: false,
+        __dirname: false,
```
peterdur: Echoing Dan's comment from #686 :
This value for `node` as an object containing `__dirname` and `__filename` properties overrides the `node` value containing a `setImmediate` property returned by `createCommonConfig`. Is that expected?

Shobhit1: Let me check on this and update you.

Shobhit1: So, it doesn't impact the process in electron but looks like according to https://webpack.js.org/configuration/node/#nodesetimmediate putting it to `false` might be better for us in terms of webpack performance; since we dont need any global node polyfills

Shobhit1: So changed it to include that property

webpack.config.js
Shobhit1: Updated.

webpack.config.js
```diff
@@ -28,6 +28,11 @@ const commonEntryFiles = {
     background: [path.resolve(__dirname, 'src/background/background-init.ts')],
 };
 
+const electronEntryFiles = {
+    main: [path.resolve(__dirname, 'src/electron/main/main.ts')],
+    injected: [path.resolve(__dirname, 'src/injected/stylesheet-init.ts'), path.resolve(__dirname, 'src/injected/client-init.ts')],
+};
+
 const commonConfig = {
     entry: commonEntryFiles,
```
peterdur: Optional suggestion: move `entry: commonEntryFiles` into `devConfig` and `prodConfig` because it's no longer totally common (overridden in `electronConfig`).

webpack.config.js
```diff
@@ -62,6 +67,26 @@ const commonConfig = {
...
+    ...commonConfig,
+    entry: electronEntryFiles,
+    name: 'electron',
+    mode: 'development',
+    output: {
+        path: path.join(__dirname, 'extension/electronBundle'),
+        filename: '[name].bundle.js',
+    },
+    node: {
+        setImmediate: false,
```
peterdur: A possible alternate:
```suggestion
        ...commonConfig.node,
```
(instead of manually copy/pasting the node settings we want to share in source and risking unintentional divergence, copy all the common node settings at run time and extend?)

PR #702, opened 2019-05-23
refactor: header icon refactor

smoralesd:
#### Description of changes

We're currently using a `` element with a `src` attribute and passing the path to the proper icon for the details view header icon (also on the insights guidance content).

We already have a white version of our heart as a react component (wrapping a svg version of the image).

On this PR:
 - Add blue version of our heart
- Using both, blue and white, react components for the icons on the header

#### Pull request checklist

   - there is no issue filed
   - does not apply as it looks the same
- [x] (UI changes only) Verified usability with NVDA/JAWS **=>** _icon is not being read (before, the img tag had an empty alt attribute, svg's don't have alt property but are not announced neither)_

### DIFF
--- a/src/icons/brand/blue/brand-blue.tsx
+++ b/src/icons/brand/blue/brand-blue.tsx
@@ -1,0 +1,91 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as React from 'react';
+
+import { NamedSFC } from '../../../common/react/named-sfc';
+
+export const BrandBlue = NamedSFC('BrandBlue', () => (
+    <svg className="header-icon" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
+        <mask id="mask0" mask-type="alpha" maskUnits="userSpaceOnUse" x="0" y="3" width="48" height="43">
+            <path
+                d="M43.8309 27.1383C49.3148 21.6544 49.3148 12.7632 43.8309 7.27931C38.347 1.7954 29.4558 1.7954 23.9719 7.27931C18.488 1.79541 9.59684 1.7954 4.11293 7.27931C-1.37098 12.7632 -1.37098 21.6544 4.11293 27.1383L21.986 45.0114C23.0828 46.1082 24.861 46.1082 25.9578 45.0114L43.8309 27.1383Z"
+                fill="white"
+            />
+        </mask>
+        <g mask="url(#mask0)">
+            <path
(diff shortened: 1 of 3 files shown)


src/tests/unit/tests/common/components/__snapshots__/header-icon.test.tsx.snap
```diff
@@ -1,17 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`HeaderIconComponent is high contrast mode enabled: false 1`] = `
-<img
-  alt=""
-  className="header-icon"
-  src="../../icons/brand/white/brand-white-48px.png"
-/>
-`;
+exports[`HeaderIconComponent is high contrast mode enabled: false 1`] = `<BrandWhite />`;
```
Shobhit1: Is there a way to make this more detailed? As in, if I change something inside the svg for `BrandWhite` and dont change the name of this component, this snapshot will still pass. So, this is not much  helpful imo.

dbjorge: IMO, that's a feature, not a bug; if we care to defend against regressions in the svg for BrandWhite, that should be in a brand-white.test.tsx, not in a test of its consumer.

smoralesd: Not really, using debug yields a very similar result. Also, look at the component and the original test... it was showing a `` tag (with some attributes). The change yield an equivalent result, except the icon components don't have any attribute.

src/icons/brand/blue/brand-blue.tsx
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as React from 'react';
+
+import { NamedSFC } from '../../../common/react/named-sfc';
+
+export const BrandBlue = NamedSFC('BrandBlue', () => (
+    <svg className="header-icon" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
dbjorge: My understanding is that the correct/standard-compliant equivalent of `alt=""` would be `role="presentation"` on the svg element (or if you wanted it to be non-presentational, to use `role="img"` and add a `<title>Accessibility Insights logo icon</title>` as the first child of the `<svg>` element).

smoralesd: We don't want to announce the icon... that's why the original `` tag had `alt=""`

smoralesd: I wasn't able to remove auto-merge label before the bot completed this so, PR #703 have a fix for this comment

PR #707, opened 2019-05-28
refactor: introduce report generator provider

smoralesd:
#### Description of changes

We are currently using the `ReportGenerator` directly on `IssuesTable` and `DetailsViewCommandBar`. We need to handle 2 different `ReportGenerator`'s based on the new feature flag we're introducing, ie, feature flag off => old report, feature flag on => new, under construction report.
Enter, `ReportGeneratorProvider`. Having a provider will let us separate the user classes (`IssuesTable` and `DetailsViewCommandBar`) from the actual `ReportGenerator` instance they use. This way, none of the user components need to now about the feature flag.
We should also introduce a new type/interface for the `ReportGenerator`.
We should also introduce the new `ReportGenerator`, it's ok if this generate dummy html report (that's the whole idea, to have a base, dummy report, under a feature flag, so we can parallelize the development of the internal components as much as we can).

#### Pull request checklist

- [x] Addresses an existing issue: wi # 1544549
   - not affected
   - not affected


src/DetailsView/details-view-initializer.ts
```diff
@@ -195,7 +195,13 @@ if (isNaN(tabId) === false) {
                 new NavigatorUtils(window.navigator).getBrowserSpec(),
                 assessmentDefaultMessageGenerator,
             );
-            const reportGenerator = new ReportGenerator(reportNameGenerator, reportHtmlGenerator, assessmentReportHtmlGenerator);
+
+            const reportGeneratorProvider = createReportGeneratorProvider(
+                reportNameGenerator,
+                reportHtmlGenerator,
```
haonliu: Do you want to rename this to reflect this is the report html generator for automated checks?

smoralesd: It includes both actually... I think this is a User Story we can thing about it at the end of the feature (have some sort of separation from the two reports we have now)

PR #713, opened 2019-05-29
fix: inspecting same element twice

### LINKED ISSUE #627: Inspect Element doesnt work if the same element is inspected again
## Describe the bug

**To Reproduce**

1. Go to any page with a11y issues
2. Run automated checks
3. click on one of the issues in the target page to open issue details dialog
4. Open devtools
5. Click on Inspect element
6. Select some other element in the devtools.
7. Inspect the same issue again.

## Screenshots
issue

**Expected**
The issue element should be highlighted in dev tools

**Actual**
The current selection in dev tools is not changed

smoralesd:
#### Description of changes

Make sure the state is always different (so we trigger the "emit change" on the store proxy side) by adding an auto increment count property to the dev tools store state.

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #627 
   - no UI/UX changes
   - no UI/UX changes

### DIFF
--- a/src/background/stores/dev-tools-store.ts
+++ b/src/background/stores/dev-tools-store.ts
@@ -47,6 +47,7 @@ export class DevToolStore extends BaseStoreImpl<DevToolState> {
     private onInspectElement(target: string[]): void {
         this.state.inspectElement = target;
         this.state.frameUrl = null;
+        // we're only using this to make sure the store proxy emit the change when the user inspect the same element twice
         this.state.inspectElementRequestId++;
         this.emitChanged();
     }
--- a/src/background/stores/dev-tools-store.ts
+++ b/src/background/stores/dev-tools-store.ts
@@ -19,7 +19,7 @@ export class DevToolStore extends BaseStoreImpl<DevToolState> {
     public getDefaultState(): DevToolState {
         const defaultValues: DevToolState = {
             isOpen: false,
-            count: 0,
+            inspectElementRequestId: 0,
         };
 
         return defaultValues;
@@ -47,7 +47,7 @@ export class DevToolStore extends BaseStoreImpl<DevToolState> {
     private onInspectElement(target: string[]): void {
         this.state.inspectElement = target;
         this.state.frameUrl = null;
-        this.state.count++;
+        this.state.inspectElementRequestId++;
         this.emitChanged();
     }
 
--- a/src/background/stores/dev-tools-store.ts
+++ b/src/background/stores/dev-tools-store.ts
@@ -47,7 +47,7 @@ export class DevToolStore extends BaseStoreImpl<DevToolState> {
     private onInspectElement(target: string[]): void {
         this.state.inspectElement = target;
         this.state.frameUrl = null;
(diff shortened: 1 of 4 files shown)


src/background/stores/dev-tools-store.ts
```diff
@@ -19,6 +19,7 @@ export class DevToolStore extends BaseStoreImpl<DevToolState> {
     public getDefaultState(): DevToolState {
         const defaultValues: DevToolState = {
             isOpen: false,
+            count: 0,
```
smoralesd: does it make sense to call this `eTag` and add a comment to clarify why do we need it?

peterdur: I think [eTag](https://en.wikipedia.org/wiki/HTTP_ETag) is not quite the right concept (it's about not doing work when the same page is requested), but agree that it would be good to have something more explicit than `count`. Maybe something like `inspectRequestId` or `distinctInspectId` or something like that?

src/tests/unit/tests/background/stores/dev-tools-store.test.ts
```diff
@@ -57,6 +57,7 @@ describe('DevToolsStoreTest', () => {
         const exepctedState = getDefaultState();
         exepctedState.inspectElement = payload;
         exepctedState.frameUrl = null;
+        exepctedState.count = 1;
```
peterdur: Typo: should be `expected`

src/common/types/store-data/idev-tool-state.d.ts
```diff
@@ -4,4 +4,5 @@ export interface DevToolState {
     isOpen: boolean;
     inspectElement?: string[];
     frameUrl?: string;
+    count?: number;
```
peterdur: Consider not making `count` optional in the interface, since it's already always there in non-test code (requires including it in stub in frame-url-search-initiator.test.ts)

smoralesd: updated!

src/tests/unit/tests/background/stores/dev-tools-store.test.ts
smoralesd: fixed

src/background/stores/dev-tools-store.ts
smoralesd: Updated

src/background/stores/dev-tools-store.ts
```diff
@@ -19,6 +19,7 @@ export class DevToolStore extends BaseStoreImpl<DevToolState> {
     public getDefaultState(): DevToolState {
         const defaultValues: DevToolState = {
             isOpen: false,
+            inspectElementRequestId: 0,
```
devaradhanm: can we just send a message back from devtools once the inspect is successful? if inspect is successfule, background can clear the inspectedElement value rite?

smoralesd: That's more complex to do... I now this seems hacky, but it's way simpler

PR #716, opened 2019-05-30
feat: add color box for color contrast fix instructions

### LINKED ISSUE #579: Add color box to the hex codes
## Is your feature request related to a problem? Please describe.

It should be easier to verify you've got the right colors and know how to act on it. 

Make it easier to see/understand color contrast problems. Sometimes what the color contrast tool is testing isn't what appears on the page.  It would be good to clearly show a box with the background/foreground color in the message. 

Might be also useful to remind folks that there may be additional testing required if there are gradients or background images. 

## Describe the desired outcome

There is a good description but most of haven't internalized the hex codes to know what foreground color: #00836d & background color: #fbf5ee look like. 

You could also possibly suggest colors like would meet the requirements like http://contrast-finder.tanaguru.com/

## Additional context

Screen Shot 2019-04-22 at 2 57 58 PM

smoralesd:
#### Description of changes

Adding a fix instruction "decorator" so we can add a little color box when foreground/background colors are mentioned on the fix instructions.

There are some assumptions for the implemented behavior, check `src\tests\unit\tests\injected\fix-instruction-processor.test.tsx` for more details.

~~**Open question**: should I add a little border around the boxes?~~ Solved on the discussion below.

**On target page**
01 - on target page issue details dialog

**On details view**
02 - on details view - failure details panel

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #579 

### DIFF
--- a/src/injected/fix-instruction-processor.tsx
+++ b/src/injected/fix-instruction-processor.tsx
@@ -66,7 +66,12 @@ export class FixInstructionProcessor {
 
         result.push(<span key={`instruction-split-${keyIndex++}`}>{coda}</span>);
 
-        return <>{result}</>;
+        return (
+            <>
+                <span aria-hidden="true">{result}</span>
+                <span className="screen-reader-only">{fixInstruction}</span>
+            </>
+        );
     }
 
     private createColorBox(colorHexValue: string, keyIndex: number): JSX.Element {
--- a/src/injected/fix-instruction-processor.tsx
+++ b/src/injected/fix-instruction-processor.tsx
@@ -1,0 +1,73 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as React from 'react';
+
+type ColorMatch = {
+    splitIndex: number;
+    colorHexValue: string;
+};
+
+export class FixInstructionProcessor {
+    private readonly colorGroupName = 'color';
+    private readonly colorValueMatcher = `(?<${this.colorGroupName}>#[0-9a-f]{6})`;
+    private readonly foregroundColorText = 'foreground color: ';
+    private readonly foregroundRegExp = new RegExp(`${this.foregroundColorText}${this.colorValueMatcher}`, 'i');
+    private readonly backgroundColorText = 'background color: ';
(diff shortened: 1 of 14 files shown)


smoralesd:
@ferBonnin @cheade Can you have a look at the color box style on the fix instructions for color contrast issues? I'm wondering if we want to add a small, thin border around them or some other similar style change.

ferBonnin:
@smoralesd @cheade  maybe a black border could work?

cheade:
Oh, I love this idea! 

One small ask, are we able to put the color box before the hex code? That's how it gets displayed in code :).

For the border, are we able to change the border dynamically based on the color inside? Black if the color needs it, no border if it doesn't? (It might not matter that there is not sufficient contrast between the color box and the border, as long as the contrast is between the color box and the pop-up?)

smoralesd:
@cheade I updated the position of the color box (before the hex value). As per the border, is there a proper way (a value, threshold or formula) we may use to decide when the border is needed? The most basic solution for the border is to always add it, that's easy to do; I can have a look at adding it as needed base on how hard is to decide when is needed though.

cheade:
I was thinking along the lines of If the color drops below 3:1 on #FFFFFF(pop-up background), then a border would be added. However, doing a thin black outline always would be fine if the color contrast isn't an issue for the darker colors. I would defer that to @ferBonnin.

ferBonnin:
black outline should be fine. The color box is added context from the Hex value we already provide :)

smoralesd:
Black border added (it's always on)

src/injected/fix-instruction-processor.tsx
```diff
@@ -0,0 +1,82 @@
...
+        if (!colorRegex.test(fixInstruction)) {
+            return null;
+        }
+
+        const match = colorRegex.exec(fixInstruction);
+
+        const groups = match['groups'];
+
+        return {
+            splitIndex: match.index + this.foregroundColorText.length + 7,
```
peterdur: 7 is a magic number here - is it the same as the 7 below? Is this the length of "#xxxxxx"?

smoralesd: updated

PR #721, opened 2019-05-30
Fix date provider

haonliu:
#### Description of changes

Currently the getDate() function is broken in master if the timestamp is not provided. 

#### Pull request checklist

- [x] Addresses an existing issue: Fixes an issue that is not in any environment yet.


src/common/date-provider.ts
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 export class DateProvider {
     public static getDate(timestamp?: string): Date {
-        return new Date(timestamp);
+        return timestamp == null ? new Date() : new Date(timestamp);
```
smoralesd: I guess is time to add some sort of test around this? is not longer a wrapper around `new Date`

peterdur: This code feels like `getDate` has two different responsibilities:
1. If a `timestamp` string is provided, it converts the string to a `Date`
2. If a `timestamp` string is not provided, it returns a `Date` with the current time.

From what I can see, callers are either wanting to do 1 or 2. Seems like it would be simpler and clearer to have two methods:
1. `getDateFromString` takes a string and converts it to a `Date`
2. `getDate` takes no arguments and returns the current time as a `Date`.

haonliu:
> We should unit test this...

Updated!

PR #728, opened 2019-06-03
feature: automated check report - new outcome badge

haonliu:
#### Description of changes

Introducing new outcome badge styling according to the mock for the new report.

Fail count badge in new automated checks report:
fail-in-new-automated-report

in assessment overview: 
count-badges-in-overview
with-high-contrast-mode

#### Pull request checklist

- [x] Addresses an existing issue: WI#1545861

### DIFF
--- a/src/common/styles/colors.scss
+++ b/src/common/styles/colors.scss
@@ -55,7 +55,7 @@
     --neutral-alpha-60: rgba(0, 0, 0, 0.6);
     --neutral-alpha-70: rgba(0, 0, 0, 0.7);
     --neutral-alpha-80: rgba(0, 0, 0, 0.8);
-    --neutral-alpha-90: rgba(0, 0, 0, 0.8);
+    --neutral-alpha-90: rgba(0, 0, 0, 0.9);
     // Outcome colors
     --positive-outcome: #107c10;
     --negative-outcome: #e81123;
--- a/src/common/styles/colors.scss
+++ b/src/common/styles/colors.scss
@@ -55,6 +55,7 @@
     --neutral-alpha-60: rgba(0, 0, 0, 0.6);
     --neutral-alpha-70: rgba(0, 0, 0, 0.7);
     --neutral-alpha-80: rgba(0, 0, 0, 0.8);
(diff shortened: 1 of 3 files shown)


src/common/styles/colors.scss
```diff
@@ -55,6 +55,7 @@
     --neutral-alpha-60: rgba(0, 0, 0, 0.6);
     --neutral-alpha-70: rgba(0, 0, 0, 0.7);
     --neutral-alpha-80: rgba(0, 0, 0, 0.8);
+    --neutral-alpha-90: rgba(0, 0, 0, 0.8);
```
smoralesd: value is same as previous line (0.8 alpha), is that intended?

smoralesd:
@Shobhit1 

> As discussed, lets keep the changes from the assessment report separate from this feature and only change automated check report.

The icons are used on both places (assessment overview and the new automated checks report). If updating/creating icons for the new report also update the assessment overview, that's fine. In the end, we want this changes on both places. As long as there is no extra effort into putting this changes into the assessment report, that's fine.

src/common/styles/colors.scss
haonliu: it was a mistake. good catch.

Shobhit1:
> @Shobhit1
> 
> > As discussed, lets keep the changes from the assessment report separate from this feature and only change automated check report.
> 
> The icons are used on both places (assessment overview and the new automated checks report). If updating/creating icons for the new report also update the assessment overview, that's fine. In the end, we want this changes on both places. As long as there is no extra effort into putting this changes into the assessment report, that's fine.

My comment came from @ferBonnin's comment on the feature, where she said that we needed to separate the scopes out. But thanks for clarification.

haonliu:
@smoralesd @Shobhit1  I can see your point in both ways, but I think the changes is very minor, so we can just go with the easier way.

smoralesd:
@Shobhit1 @haonliu I actually talked to Fernanda about this. Her point was not to forcefully add more work (from the assessment overview/report) to this feature. That's why the "no added effort" policy.

PR #729, opened 2019-06-03
add summary section to automated checks report

smoralesd:
#### Description of changes

Add summary section to the new automated checks report.

This PR does not include the changes on the icons, that will be part of a different PR.

08 - summary section

#### Pull request checklist


src/DetailsView/reports/components/report-sections/outcome-summary-bar.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { kebabCase } from 'lodash';
+import { CheckIcon } from '../../../../common/icons/check-icon';
+import { CircleIcon } from '../../../../common/icons/circle-icon';
+import { CrossIcon } from '../../../../common/icons/cross-icon';
+import { NamedSFC } from '../../../../common/react/named-sfc';
+import { SectionProps } from './report-section-factory';
+
+const iconMap = {
+    pass: <CheckIcon />,
+    'not applicable': <CircleIcon />,
```
haonliu: Is there any reason that only ``not applicable`` is inside quotes?

src/DetailsView/reports/components/report-sections/outcome-summary-bar.tsx
```diff
@@ -0,0 +1,53 @@
...
+    fail: <CrossIcon />,
+};
+
+const outcomeText = {
+    pass: 'Passed',
+    'not applicable': 'Not applicable',
+    fail: 'Failed',
+};
+
+type OutcomeType = 'pass' | 'fail' | 'not applicable';
```
haonliu: Can we use the type  ``OutcomeType`` from ``outcome-type.ts`` ?

haonliu: After a second look, the 'outcome' here is a different concept than the one in assessment. In that case, should we use another name to avoid confusion?

smoralesd: yeap... I though about refactoring the original component, make it generic (literally <TOutcomeType>)... but it has some hard dependencies so I decide to take a different approach for now. This is something that goes to the clean up & refactoring list though.

src/DetailsView/reports/components/report-sections/outcome-summary-bar.tsx
smoralesd: yeap... the space

smoralesd:
> Left some comments. One concern that I have is that the word 'outcome' confuses me with the other concept in assessment.

I'd say the concept is analogous but the actual values are different: there is not 'incomplete' on automated checks; there is no 'not applicable' under assessment; the concept is applicable thought: it's the outcome of using automated checks or working through assessments. I think this is something we want to refactor to avoid duplication but I would like to do that once we remove the code from the old report. The less code we need to update, the better.

src/DetailsView/reports/components/report-sections/outcome-summary-bar.tsx
haonliu: Sounds good. let me make a note in that story so that we don't forget.

src/DetailsView/reports/components/report-sections/outcome-summary-bar.tsx
```diff
@@ -0,0 +1,53 @@
...
+    return (
+        <div className="outcome-summary-bar">
+            {allOutcomeTypes.map(outcomeType => {
+                const text = outcomeText[outcomeType];
+                const outcomeIcon = iconMap[outcomeType];
+                const count = countSummary[outcomeType];
+
+                return (
+                    <span key={outcomeType} className={kebabCase(outcomeType)} style={{ flexGrow: count }}>
+                        {outcomeIcon} {count} {text}
```
haonliu: So if say 0 instances failed, the span for fails won't show in the summary bar? Is that expected?

haonliu: Never mind. Seems its minimum size is the size of the inner text.

src/DetailsView/reports/components/report-sections/outcome-summary-bar.tsx
haonliu: Maybe we should use ``resultGroups`` in ``axe.d.ts`` instead.

PR #739, opened 2019-06-05
Fix heading levels

smoralesd:
#### Description of changes

Using heading levels 1 to 3 as follows:
- Automated checks result (the title just after the header) is H1
- All section titles (summary, scan details, failed instances, passed checks, not applicable checks) are H2
- Each rule group under failed instances is an H3 (so, **only** rules with failed instances is announced as a header, but **no** rule under passed and not applicable checks section is announce)
- Update some stylines (mostly margins)

#### Pull request checklist

- [x] Addresses an existing issue: part of WI # 1544731

### DIFF
--- a/src/DetailsView/reports/components/instance-list-group-header.tsx
+++ b/src/DetailsView/reports/components/instance-list-group-header.tsx
@@ -12,7 +12,7 @@ import { InstanceOutcomeType } from './report-sections/outcome-summary-bar';
 export interface InstanceListGroupHeaderProps {
     ruleResult: RuleResult;
     outcomeType: InstanceOutcomeType;
-    'aria-level'?: number;
+    ariaLevel?: number;
 }
 
 export const InstanceListGroupHeader = NamedSFC<InstanceListGroupHeaderProps>('InstaceListGroupHeader', props => {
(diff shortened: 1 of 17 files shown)


src/DetailsView/reports/components/instance-list-group-header.tsx
```diff
@@ -12,6 +12,7 @@ import { InstanceOutcomeType } from './report-sections/outcome-summary-bar';
 export interface InstanceListGroupHeaderProps {
     ruleResult: RuleResult;
     outcomeType: InstanceOutcomeType;
+    'aria-level'?: number;
```
haonliu: why not ``ariaLevel``?

smoralesd: good point

src/DetailsView/reports/components/instance-list-group-header.tsx
```diff
@@ -46,8 +47,10 @@ export const InstanceListGroupHeader = NamedSFC<InstanceListGroupHeaderProps>('I
         return <span className="rule-details-description">{props.ruleResult.description}</span>;
     };
 
+    const ariaLevel = props['aria-level'] || undefined;
+
     return (
-        <div role="heading">
+        <div role="heading" aria-level={ariaLevel}>
```
haonliu: These two attr should come together. If the div have role=heading but no aria level it will fail a11y checks. But this can be taken care of in my PR.

src/DetailsView/reports/components/instance-list-group-header.tsx
smoralesd: updated.

src/DetailsView/reports/components/instance-list-group-header.tsx
smoralesd: Awesome... I didn't want to touch the already existing `role="heading"` part as I'm not fully aware where else we are using this component.

PR #740, opened 2019-06-05
Fix a11y issues in report

haonliu:
#### Description of changes
Before:
image

We have two a11y issues in the new report currently: 
1. group header doesn't have a valid heading level;
2. rule link does not have a valid aria description.

This PR fixes both of them.

After: 
image

#### Pull request checklist

### DIFF
--- a/src/DetailsView/reports/components/instance-list-group-header.tsx
+++ b/src/DetailsView/reports/components/instance-list-group-header.tsx
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { kebabCase } from 'lodash';
 import * as React from 'react';
 
 import { GuidanceLinks } from '../../../common/components/guidance-links';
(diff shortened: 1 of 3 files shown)


src/DetailsView/reports/components/instance-list-group-header.tsx
```diff
@@ -43,11 +48,15 @@ export const InstanceListGroupHeader = NamedSFC<InstanceListGroupHeaderProps>('I
...
+        return (
+            <span className="rule-details-description" id={ariaDescribedBy}>
+                {props.ruleResult.description}
+            </span>
+        );
     };
 
     return (
-        <div role="heading">
+        <div role="heading" aria-level={4}>
```
smoralesd: I'm guessing this line will have conflicts.
On my PR I'm passing the aria level value. I guess you can have a default value = 4 and override it if the `ariaLabel` property is not null.

haonliu: updated

PR #745, opened 2019-06-06
Introduce tags for WCAG 2.1

waabid:
#### Description of changes

This change introduces tags for 'New for WCAG 2.1'. They are attached to the success criteria. We have also surfaced this in the UI, however, **in this PR** no rules will have this visible (since we want to make sure master is always releasable). As a result, I have included screen-caps.

image

image

image

image

#### Pull request checklist


smoralesd:
Changes looks good.
I do have a question though: I'm wondering how are we going to handle the case when WCAG 2.1 is not new anymore and we'll need to remove the pill from the UI. I'm not sure we want to handle that on this PR, but maybe, give in it a little though and see how this scenario may affect the code in the future can give us a little insights if the current changes will make that update easier or not

waabid:
> Changes looks good.
> I do have a question though: I'm wondering how are we going to handle the case when WCAG 2.1 is not new anymore and we'll need to remove the pill from the UI. I'm not sure we want to handle that on this PR, but maybe, give in it a little though and see how this scenario may affect the code in the future can give us a little insights if the current changes will make that update easier or not

I agree. That's why I advocated for having an ID in the tag object; the display text can change (maybe back to "WCAG 2.1" or we could introduce other properties later as well.

If you have any other suggestions, I'm all ears too.

src/views/content/content-page.tsx
```diff
@@ -4,13 +4,14 @@ import { flatten, toPairs } from 'lodash';
 import * as React from 'react';
 
 import { NamedSFC } from '../../common/react/named-sfc';
+import { GuidanceTag } from '../../content/guidance-tags';
 import { createMarkup, Markup, MarkupDeps } from './markup';
 
-export type HyperlinkDefinition = { href: string; text: string };
+export type HyperlinkDefinition = { href: string; text: string; tags?: GuidanceTag[] };
```
smoralesd: It seems to me that altering the `HyperlinkDefinition` it's a little confusing. In general, I'd say, a hyperlink is just the text and the href (maybe just the href) but nothing more. It feels like we should have a new type where we have the `tags` (probably not optional).

Looking at the changes I found `getGuidanceTagsFromGuidanceLinks`, which suggest `GuidanceLink` for the new type... something like this maybe?
```typescript
type GuidanceLink = HyperlinkDefinition & { tags: GuidanceTag[] };
```

waabid: Updated!

PR #747, opened 2019-06-06
fix: accessible instance table

smoralesd:
#### Description of changes

Right now, we're using one table per failed instance (1 table of 3 rows -path, snippet, how to fix, and 2 columns -header and value-).

From Rob feedback on the first draft of the report: "it would be nice [better accessibility] to have all the failed instances [grouped by rule] in one table".

This is an attempt to implement Rob's feedback on the failed instance list.

I'm going to run some more test using JAWS and NVDA to see if this make sense and/or if we are missing anything on the new table (aria properties come to mind). I also send an example to Rob to get his feedback.

20 - accessible failed instance list

#### Pull request checklist

### DIFF
--- a/src/DetailsView/reports/automated-checks-report.scss
+++ b/src/DetailsView/reports/automated-checks-report.scss
@@ -286,3 +286,73 @@ $outcome-not-applicable-summary-color: $neutral-outcome;
         @include transform(rotate(-45deg));
     }
 }
+
+.a11y-report-instance-table {
+    table,
+    thead,
+    tbody,
+    th,
+    td,
+    tr {
+        display: block;
+    }
+
+    table {
+        font-size: 14px;
+        line-height: 20px;
+    }
+
+    thead tr {
+        position: absolute;
+        top: -9999px;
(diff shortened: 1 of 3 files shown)


src/DetailsView/reports/automated-checks-report.scss
```diff
@@ -250,3 +250,73 @@ $outcome-not-applicable-summary-color: $neutral-outcome;
...
+        width: 66px;
+        padding-right: 10px;
+        white-space: nowrap;
+
+        font-family: $semiBoldFontFamily;
+        color: $primary-text;
+    }
+
+    td:nth-of-type(1):before {
+        content: 'Path';
```
Shobhit1: any idea, why we need this?

smoralesd: If you look at `thead tr` style... we're displaying the headers off screen (but not using display:none so AT can read it anyway). The before selector on head `td` let us show the header (visually) and repeat it on each "card" without breaking the table semantics

PR #749, opened 2019-06-07
accessible instance details

smoralesd:
#### Description of changes

25 - accessible expand-collapse component

Improving accessibility on the new report:
- **Using a list for the failed instance details**: this help AT user to know about the amount of failed instances, where they are inside the list and quick navigation by AT shortcuts
- **Add custom component for collapsible/expandable section**: having a this component let us fully control styles and accessibility and it's also compatible with Edge (old Edge that is). As a note, previously we were using `<details>`, `<summary>` HTML 5 tags but this are not compatible with Edge and JAWS doesn't properly announce current state (expand/collapse of the section)
   - As part of this I had to add javascript code to handle expand/collapse and aria attributes, see src/DetailsView/reports/components/report-sections/collapsible-script-provider.tsx for more details

We run the changes with Rob and Adarsh (several times, basically after every minor improvement/tweak) so most of the decision on the code changes come from their feedback as heady AT users.

Main changes here are under:
- src/DetailsView/reports/components/report-sections/collapsible-script-provider.tsx: for the collapse/expand logic
- src/DetailsView/reports/components/report-sections/failed-instances-section.tsx: where we use the collapse/expand logic
- src/DetailsView/reports/components/report-sections/summary-details.tsx: new custom collapse/expand component
- src/DetailsView/reports/automated-checks-report.scss: proper styling of the new component

There rest of the changes are: unit test, pipe the dependencies from the composition root

#### Pull request checklist

- [x] Addresses an existing issue: part of WI # 1544731


src/DetailsView/reports/components/report-sections/summary-details.tsx
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as React from 'react';
+
+import { NamedSFC } from '../../../../common/react/named-sfc';
+
+export type SummaryDetailsProps = {
+    summaryProps?: Object;
```
Shobhit1: I can see why we would want to keep this loose but do you think until we are ready to expand on this component, we can keep strict type checking on this? Thoughts?

Issue #730, opened 2019-06-03
Download report gives an error on pages


virav:
While doing the assessment of a specific page, we get download error on "export result". I had discussed this issue with @smoralesd with screen share and as per discussion, it is an HTML file size issue. Please help me to overcome this issue for our portal to be accessibility compliant.

Thank you!

smoralesd:
hi @virav We have a fix for this issue now on [insider](https://chrome.google.com/webstore/detail/accessibility-insights-fo/nnmjfbmebeckhpejobgjjjnchlljiagp). It should go to [production](https://chrome.google.com/webstore/detail/accessibility-insights-fo/pbjjkligggfmakdaogkfomddhfmpjeni) presently (probably next week).

Please let us know if you have any  question about this bug fix.

virav:
It's working as expected. Thank you, Team!

PR #752, opened 2019-06-07
Increasing Report Download Capacities in Chrome

annikakouhia:
#### Description of changes

Previously, summary downloads were limited on Chrome to 2MB. We switched to using Blobs (https://chromium.googlesource.com/chromium/src/+/master/storage/browser/blob/README.md) to store the data and increase the upper bound on the size of our exported results downloads.

This works for both Fastpass (including both the original export and new automated export feature) and Assessment.

We included new mocks to handle this functionality in tests, and added a new method to our Window Utils file to extract a URL from the Blob.

#### Pull request checklist

- [X] Addresses an existing issue: Addresses Issue #730 


src/DetailsView/components/export-dialog.tsx
```diff
@@ -38,7 +40,8 @@ export const ExportDialog = NamedSFC<ExportDialogProps>('ExportDialog', props =>
         props.onDescriptionChange(value);
     };
 
-    const encodedHtml = encodeURIComponent(props.html);
+    const blob = new Blob([props.html], { type: 'text/html' });
```
smoralesd: I'm wondering if this deserve to be part of the `deps` of this class, so instead of using `new` we called the function from the `deps`. I'm a little concerned about how will this work on electron for example. I'm also concern on how this affect test (not sure about performance on creating blobs). Having this as a dependency allows to mock the dependency, and thus allow us to not really create a blob on the test, just a mock of a blob (or a stub for that matter)

karanbirsingh: If we take that approach, I wonder if it'd be better to include a `reportURLGetter(html: string)` prop that gets passed to `ExportDialog` and includes the blob creation & `windowUtils.createObjectURL`. Then the `ExportDialog` test can include a mock URL. How do you feel about this @smoralesd?

smoralesd: I think that would work even better! I think the name of the property should be verb-base, something in the lines of `getReportUrl` (just turning around your original wording). It also make sense to abstract away the fact that we are creating a blob (as it could be anything we could open through the browser, right)

src/DetailsView/components/export-dialog.tsx
```diff
@@ -4,6 +4,7 @@ import { PrimaryButton } from 'office-ui-fabric-react/lib/Button';
 import { Dialog, DialogFooter, DialogType } from 'office-ui-fabric-react/lib/Dialog';
 import { TextField } from 'office-ui-fabric-react/lib/TextField';
 import * as React from 'react';
+import { WindowUtils } from 'src/common/window-utils';
```
Shobhit1: This path needs to be relative. I am not sure why this is not failing the PR build.

annikakouhia:
Moving over to PR #766

PR #755, opened 2019-06-07
update footer for the new report 

haonliu:
#### Description of changes

update footer for the new report:
image

#### Pull request checklist

### DIFF
--- a/src/DetailsView/reports/components/report-sections/report-footer.tsx
+++ b/src/DetailsView/reports/components/report-sections/report-footer.tsx
@@ -1,0 +1,31 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as React from 'react';
+
+import { NewTabLink } from '../../../../common/components/new-tab-link';
+import { NamedSFC } from '../../../../common/react/named-sfc';
+import { toolName } from '../../../../content/strings/application';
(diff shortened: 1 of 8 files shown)


src/DetailsView/reports/components/report-sections/report-footer.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { NamedSFC } from '../../../../common/react/named-sfc';
+import { toolName } from '../../../../content/strings/application';
+import { SectionProps } from './report-section-factory';
+
+export type ReportFooterProps = Pick<SectionProps, 'environmentInfo'>;
+
+export const ReportFooter = NamedSFC<ReportFooterProps>('ReportFooter', ({ environmentInfo }) => {
+    const { extensionVersion, browserSpec, axeCoreVersion } = environmentInfo;
+    return (
+        <div className="report-footer-container">
```
smoralesd: This should a a change done to `src\DetailsView\reports\components\report-sections\footer-section.tsx`. The `FooterSection` component is new (just for the new report) and thus is not being used anywhere else, so it makes sense to update it instead of replacng it

haonliu: That one is also being used in src\DetailsView\reports\components\report-body.tsx, which is the v1 report.

PR #765, opened 2019-06-11
Fix icon not centered

haonliu:
#### Description of changes

Fix icons that are not centered.
image

image

#### Pull request checklist


src/common/icons/check-icon.tsx
```diff
@@ -6,6 +6,29 @@ import { NamedSFC } from '../react/named-sfc';
 
 export const CheckIcon = NamedSFC('CheckIcon', () => (
     <span className="check-container">
-        <span className="check-icon" />
+        <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+            <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
peterdur: Do we need nested `svg` elements here?

src/common/icons/check-icon.tsx
```diff
@@ -6,6 +6,29 @@ import { NamedSFC } from '../react/named-sfc';
...
+                <circle cx="8" cy="8" r="8" fill="#228722" />
```
peterdur: Does the fill color here need to change for high contrast mode?

src/DetailsView/reports/components/report-head-v2.tsx
```diff
@@ -8,7 +8,7 @@ import * as reportStyles from '../automated-checks-report.styles';
 export const ReportHeadV2 = NamedSFC('ReportHead', () => {
     return (
         <head>
-            <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
+            <meta httpEquiv="Content-Type" content="text/html;charset=utf-8" />
```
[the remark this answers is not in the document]
haonliu: That's true. These are some errors thrown on the console.

src/common/icons/check-icon.tsx
haonliu: removed

src/common/icons/check-icon.tsx
haonliu: Good point!

src/common/icons/cross-icon.tsx
```diff
@@ -6,6 +6,25 @@ import { NamedSFC } from '../react/named-sfc';
 
 export const CrossIcon = NamedSFC('CrossIcon', () => (
     <span className="check-container">
-        <span className="cross-icon" />
```
peterdur: Since we're using SVG instead of CSS to create the cross icon, can we remove the CSS rules for `cross-icon`?

src/common/styles/colors.scss
```diff
@@ -57,7 +57,7 @@
     --neutral-alpha-80: rgba(0, 0, 0, 0.8);
     --neutral-alpha-90: rgba(0, 0, 0, 0.9);
     // Outcome colors
-    --positive-outcome: #107c10;
+    --positive-outcome: #228722;
```
smoralesd: ??? is this intended? did you check with Chelsea about this?

src/views/content/content.scss
```diff
@@ -188,22 +188,6 @@
         }
     }
 
-    .do-header,
```
smoralesd: why are we removing this? how does this affect the icons?

src/common/styles/colors.scss
haonliu: This is the green color on the latest Figma for the new report.

src/views/content/content.scss
haonliu: Previously the check icon is just a check so we need the green background color for the circle. now the circle is part of the icon, so this is not needed.

PR #766, opened 2019-06-11
Increase Report Download Size for Chrome--Blobs

annikakouhia:
#### Description of changes

Previously, summary downloads were limited on Chrome to 2MB. We switched to using Blobs (https://chromium.googlesource.com/chromium/src/+/master/storage/browser/blob/README.md) to store the data and increase the upper bound on the size of our exported results downloads.
This works for both FastPass (including both the original export and new automated export feature) and Assessment.
We included new mocks to handle this functionality in tests, and added a new method to our Window Utils file to extract a URL from the Blob. We also created a BlobProvider file which abstracts the creation and provision of the Blob itself.
Note: Originally dealt with by PR #752, but branches got messy so switched to this one for clarity.

#### Pull request checklist

- [X] Addresses an existing issue: Addresses #730


src/common/blob-provider.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export class BlobProvider {
```
Shobhit1: Just curious if this can just be an exported function rather than a class with a static function.

src/common/window-utils.ts
```diff
@@ -22,6 +22,10 @@ export class WindowUtils {
         return window.setInterval(handler, timeout);
     }
 
+    public createObjectURL(ob: Blob | File | MediaSource): string {
```
Shobhit1: I feel we can rename `ob` to something that conveys more meaning. `sourceObject` maybe? thoughts?

src/tests/unit/common/blob-provider.test.tsx
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { provideBlob } from '../../../common/blob-provider';
+
+describe('BlobProviderTest', () => {
+    test('provideBlob', () => {
+        const mimeType = 'text/html';
+        const oMyBlob = provideBlob(['<a></a>'], mimeType);
```
Shobhit1: nit: This variable as well can have a more meaningful name.

src/DetailsView/components/export-dialog.tsx
```diff
@@ -38,7 +41,8 @@ export const ExportDialog = NamedSFC<ExportDialogProps>('ExportDialog', props =>
         props.onDescriptionChange(value);
     };
 
-    const encodedHtml = encodeURIComponent(props.html);
+    const blob = props.deps.provideBlob([props.html], 'text/html');
```
smoralesd: As per our conversation:
- `ExportDialog` should not know we're creating a Blob because that's an implementation detail
- We could move this 2 lines (creating the blob, creating a url for the blob) to a report provider (or file provider, the name can be generic enough that conveys we can create a url for any kind of content).

By doing this, we allow the implementation to change (maybe it's not a blob anymore, it's something else) without affecting the `ExportDialog` component.

I'm not too worried about this right now, so up to you if you want to make this change as part of this PR.

annikakouhia: We're going to submit for now to get it into Canary, but will put in additional changes in a second PR! Thank you!

src/tests/unit/common/blob-provider.test.tsx
annikakouhia: Ditto to my comment to Sebastian below--submitting for now to get into Canary, but will add in these suggestions in another PR! Thanks!

PR #770, opened 2019-06-11
fix: fix screen reader announcement of the details section content

Shobhit1:
#### Description of changes

Adds code that makes the screen reader properly announce the text in the scan details section.

The way Its done is to add a `screen-reader-only` span with properly formatted text, the `css` for that `span` will hide it from user's viewport but the screen-reader will read it since I have put the `ids` of those 
`span` in the `td` of the actual content's `aria-labelledby`.

Also since, now we are using `description` at two places; change the logic a little bit to make the replacement work better.

#### Pull request checklist

- [x] Addresses an existing issue: Fixes WI #1552587

#### Screenshot:
No-change in what we show here.

image

#### NVDA read out:
```
Target Page: https://www.chicos.com/store/
Scan date: 2019-06-11 9:51 PM UTC
Comment: Test
```

### DIFF
--- a/src/DetailsView/reports/components/report-head-v2.tsx
+++ b/src/DetailsView/reports/components/report-head-v2.tsx
@@ -8,7 +8,7 @@ import * as reportStyles from '../automated-checks-report.styles';
 export const ReportHeadV2 = NamedSFC('ReportHead', () => {
     return (
         <head>
-            <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
+            <meta httpEquiv="Content-Type" content="text/html;charset=utf-8" />
             <title>{title} automated checks result</title>
             <style dangerouslySetInnerHTML={{ __html: reportStyles.styleSheet }} />
         </head>
--- a/src/DetailsView/reports/components/report-sections/details-section.tsx
+++ b/src/DetailsView/reports/components/report-sections/details-section.tsx
@@ -36,7 +36,6 @@ export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', pr
                         <td className="icon" aria-hidden="true">
                             <UrlIcon />
                         </td>
-
                         <td className="text" aria-labelledby="target-page-text" aria-hidden="true">
                             <NewTabLink href={pageUrl} title="Navigate to target page">
                                 {pageUrl}
@@ -47,7 +46,6 @@ export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', pr
                         <td className="icon" aria-hidden="true">
                             <DateIcon />
                         </td>
-
                         <td className="text" aria-labelledby="scan-date-text" aria-hidden="true">
                             {scanDateUTC}
                         </td>
@@ -56,7 +54,6 @@ export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', pr
                         <td className="icon" aria-hidden="true">
                             <CommentIcon />
                         </td>
-
                         <td className="text description-text" aria-labelledby="comment-text" aria-hidden="true">
                             {description}
                         </td>
--- a/src/DetailsView/reports/components/report-sections/details-section.tsx
+++ b/src/DetailsView/reports/components/report-sections/details-section.tsx
@@ -21,20 +21,14 @@ export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', pr
     return (
         <div className="scan-details-section">
             <h2>Scan details</h2>
-            <span className="screen-reader-only" id="target-page-text">
-                {screenReaderTexts.targetPageLink}
-            </span>
-            <span className="screen-reader-only" id="scan-date-text">
-                {screenReaderTexts.scanDate}
-            </span>
-            <span className="screen-reader-only" id="comment-text">
-                {screenReaderTexts.comment}
-            </span>
             <table>
                 <tbody>
                     <tr>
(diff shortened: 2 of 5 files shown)


src/DetailsView/reports/components/report-head-v2.tsx
```diff
@@ -8,7 +8,7 @@ import * as reportStyles from '../automated-checks-report.styles';
 export const ReportHeadV2 = NamedSFC('ReportHead', () => {
     return (
         <head>
-            <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
+            <meta httpEquiv="Content-Type" content="text/html;charset=utf-8" />
```
Shobhit1: This was causing a console error.

smoralesd: wasn't this change made on one of Haonan's PR? seems duplicated... I guess it's fine as long as we don't see conflicts.

src/DetailsView/reports/components/report-sections/details-section.tsx
```diff
@@ -21,7 +27,10 @@ export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', pr
                         <td className="icon" aria-hidden="true">
                             <UrlIcon />
                         </td>
-                        <td className="text">
+                        <span className="screen-reader-only" id="target-page-text">
+                            {screenReaderTexts.targetPageLink}
+                        </span>
+                        <td className="text" aria-labelledby="target-page-text">
```
smoralesd: both NVDA and JAWS are reading this fields twice, one for the span `screen-reader-only` and then again for the td (minus the 'target page', 'scan date', 'comment' text)
I guess you may need to use aria-hidden on this td's and extensively test with both screen readers.

src/DetailsView/reports/components/report-sections/details-section.tsx
```diff
@@ -12,6 +12,12 @@ export type DetailsSectionProps = Pick<SectionProps, 'pageUrl' | 'description' |
...
 export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', props => {
     const { pageUrl, description, scanDate, toUtcString } = props;
+    const scanDateUTC: string = toUtcString(scanDate);
+    const screenReaderTexts = {
+        targetPageLink: `Target Page: ${pageUrl}`,
+        scanDate: `Scan date: ${scanDateUTC}`,
+        comment: `Comment: ${description}`,
+    };
     return (
         <div className="scan-details-section">
```
smoralesd: Not sure why but by using JAWS and navigating to the table, the AT is reading the whole table content at once. Later, using arrow keys (down arrow key) you can navigate to the different cells to get JAWS to read the content per cell. Worth looking at it since our 2 internal AT users use JAWS not NVDA.

src/DetailsView/reports/components/report-head-v2.tsx
Shobhit1: I added this previously to fix the special character issue in Edge, but I guess I messed up the naming of parameter.

src/DetailsView/reports/components/report-sections/details-section.tsx
```diff
@@ -12,6 +12,12 @@ export type DetailsSectionProps = Pick<SectionProps, 'pageUrl' | 'description' |
 
 export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', props => {
     const { pageUrl, description, scanDate, toUtcString } = props;
+    const scanDateUTC: string = toUtcString(scanDate);
+    const screenReaderTexts = {
+        targetPageLink: `Target Page: ${pageUrl}`,
```
smoralesd: I think you should drop the ':' here. NVDA doesn't read them so it's fine but JAWS **DO** read them which sounds weird: "target page colon http://...."

src/DetailsView/reports/components/report-sections/details-section.tsx
Shobhit1: I saw it with NVDA and it was not reading twice. Maybe need to check twice

src/DetailsView/reports/components/report-sections/details-section.tsx
Shobhit1: Thats what @ferBonnin told me to add

src/DetailsView/reports/components/report-sections/details-section.tsx
Shobhit1: I dont see NVDA repeating twice but JAWS certainly do read the whole table at once.

src/DetailsView/reports/components/report-head-v2.tsx
smoralesd: Sure... what I'm saying is this change is already on master... so why making it again? is one more file reviewers need to go through... also... this change have nothing to do with the scan details section accessibility... so... hard to figure out

Shobhit1: It was causing a console error even on master. Doesn't fail or anything, just to fix that red error! :)

src/DetailsView/reports/components/report-sections/details-section.tsx
Shobhit1: As discussed, I am trying it by removing the spans from `tr` and then checking it with JAWS

Shobhit1: Fixed

PR #774, opened 2019-06-12
fix: new automated checks report: how-to-fix conversion to list

haonliu:
#### Description of changes

build how to fix list in report: 
image

#### Pull request checklist


smoralesd:
I found some cases where the styles are not right. In the screenshot you can see bypass how to fix being cut off ("Page does not have a landmark regio" - missing "n").

Also - and we saw this before - on some cases the separator lines are cut of... looks like this are related. Can you have a look at this as part of this PR?
28 - how to fix styles

src/DetailsView/reports/components/report-sections/instance-details.tsx
```diff
@@ -22,12 +23,55 @@ export const InstanceDetails = NamedSFC<InstanceDetailsProps>('InstanceDetail',
             </tr>
         );
     };
+
+    const renderInstructions = (howToFixString: string) => {
```
smoralesd: 2 questions about the how to fix instructions code:
1. Can we move this to a separated component, it's implementation details the `InstanceDetails` component doesn't need to know
2. Any reason why we are implementing this from scratch instead of reusing `FixInstructionPanel` (src\injected\components\fix-instruction-panel.tsx)

src/DetailsView/reports/components/instance-details.scss
```diff
@@ -8,7 +8,9 @@
     background: $neutral-0;
     border-radius: 4px;
     box-shadow: 0px 0.6px 1.8px rgba(0, 0, 0, 0.108), 0px 3.2px 7.2px rgba(0, 0, 0, 0.132);
-    display: block;
+    display: table;
+    width: -webkit-fill-available;
```
smoralesd: should we care about other browsers here? found a stack overflow about it [here](https://stackoverflow.com/questions/26275982/putting-moz-available-and-webkit-fill-available-in-one-width-css-property)

PR #780, opened 2019-06-12
refactor: created/moved module specific styles for test status choice group

waabid:
#### Description of changes

Just moved styles from general details view scss file to module specific scss file.

image

#### Pull request checklist

- [n/a] Ran precheckin (`yarn precheckin`)

### DIFF
--- a/src/DetailsView/components/test-status-choice-group.scss
+++ b/src/DetailsView/components/test-status-choice-group.scss
@@ -1,0 +1,53 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../common/styles/colors.scss';
+
+.radio-button-group {
+    float: left;
+    line-height: 24px;
+
+    .radio-label {
+        font-size: 14px;
+        padding: 0 0 0 26px;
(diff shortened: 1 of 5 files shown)


src/DetailsView/components/test-status-choice-group.scss
```diff
@@ -0,0 +1,53 @@
...
+    float: left;
+    line-height: 24px;
+
+    .radio-label {
+        font-size: 14px;
+        padding: 0 0 0 26px;
+        display: inline-block;
+    }
+
+    :global {
```
smoralesd: out of curiosity: what `:global` stand for?

waabid: This is to say that the following classes are not declared in this module. As a result, these are not modified by css-loader. This is beneficial in the scenarios we depend on certain class names being X or Y from components we don't control (Office Fabric).

smoralesd: I see... so you can't import them on your components neither, right?

waabid: Correct!

PR #802, opened 2019-06-14
fix: forcing to inject immediately when we try to inject files

### LINKED ISSUE #786: Target page with placeholder iframe causes visualization enablement to hang
## Describe the bug

If a target page contains certain types of placeholder iframe elements (eg, an element like `<iframe src="javascript:false;">`, [apparently a common technique](https://stackoverflow.com/a/4284645)), actions that involve injecting visualizations into the target page will end up spinning forever without making progress:

* enabling any ad-hoc tool
* performing a fast pass scan
* etc

**To Reproduce**

1. Navigate to https://codepen.io/dbjorge/pen/Kjpmja
2. Enable any visualization (eg, by attempting to enable any ad-hoc tool or running a FastPass)
3. Observe that the "enable" action hangs/remains in progress forever.

## Expected behavior

Shouldn't hang

## Screenshots

image

- OS: Windows
- Version: Version 2.4.1
- Browser Version: Chrome Stable Version 75.0.3770.80 (Official Build) (64-bit)
- Target Page: https://codepen.io/dbjorge/pen/Kjpmja

If asked, will you be willing to submit a PR to fix this bug? Yes

Yes

devaradhanm:
#### Description of changes

We need to test this. This change could mean, if the page is still loading, we may inject it before it completes loading. You can use the latest playground extension to validate this. (Wait for few min to see the today's version in chrome store) I am meanwhile using that to validate on more sites.
I have validated this on https://codepen.io/dbjorge/pen/Kjpmja , slow loading pages. The user experience is same in the slow loading pages as well.

more info on the property can be found here - https://developer.chrome.com/extensions/content_scripts

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #786

### DIFF
--- a/src/background/browser-adapters/chrome-adapter.ts
+++ b/src/background/browser-adapters/chrome-adapter.ts
@@ -65,6 +65,7 @@ export class ChromeAdapter extends ClientChromeAdapter implements BrowserAdapter
             {
                 allFrames: true,
                 file: file,
+                runAt: 'document_start',
             },
             callback,
         );


devaradhanm:
@dbjorge , I was waiting for the release to get completed. I have updated the comment now.

devaradhanm:
@ferBonnin, Playground extension is updated.
The scenario where this change could break is:
if the user tries to enable visualization when the page is being loaded, we may start scanning when the dom tree is not rendered completely. But, even currently, you will be getting unreliable result if you don't wait for the page to stabilize before enabling visualization. So, I don't think this is a blocking issue. Let me know if you feel otherwise.

dbjorge:
I verified that:

* Deva's proposed fix does fix scanning in the full motivating example from the partner team
* The suggested alternatives to the `src="javascript:false;"` that use `about:blank` or similar instead of `javascript:` do not seem to be impacted by the original issue
* Other random prefixes with no special meaning also aren't impacted
* Chrome treats `javascript:` and `JaVaScRiPt:` prefixes the same

I feel a little uneasy about the document_start vs document_idle thing but can't really articulate a specific instance that would motivate why. It's true that a user could slightly more easily scan a page with components that are still loading, but there are already other non-iframe-based cases of this; it's not a problem unique to the iframe situation.

Maybe a compromise solution could be to test for something like `frame.src && /^javascript:/i.test(frame.src)` and use document_start only for the `javascript:` case?

devaradhanm:
@dbjorge , we could use chrome.webNavigation.getAllFrames({tabId: 587}, console.log). this will return all valid frames. This doesn't seem to return placeholder frames. So, we could call chrome.tabs.executeScript on each frame this returns. But, am concerned that we may have some pages, that keeps document not to be idle at all. refer this - https://stackoverflow.com/questions/42509273/code-injection-using-executescript-never-call-callback-in-chrome-extension#

PR #817, opened 2019-06-17
Blob storage edits

annikakouhia:
#### Description of changes

A few organizational/specification changes for the increase in download size using blobs that I merged last week. First, giving more meaningful variable names (in blob-provider and window-utils). Second, abstracting the creation of a blob and extraction of a URL from that blob into a class named file-url-provider.

Mocked the action of this class for all Export Dialog tests.

Removed WindowUtils from deps of Export Dialog since their utility was now handled by the File URL Provider.

#### Pull request checklist

- [X] Addresses an existing issue: Addition to Issue #730 (PR #766)

### DIFF
--- a/src/common/file-url-provider.ts
+++ b/src/common/file-url-provider.ts
@@ -1,13 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { provideBlob } from './blob-provider';
 import { WindowUtils } from './window-utils';
 
 export class FileURLProvider {
-    constructor(private windowUtils: WindowUtils) {}
+    constructor(private windowUtils: WindowUtils, private provideBlob: (blobParts?: any[], mimeType?: string) => Blob) {}
 
     public provideURL = (blobParts?: any[], mimeType?: string): string => {
-        const blob = provideBlob(blobParts, mimeType);
+        const blob = this.provideBlob(blobParts, mimeType);
         const blobURL = this.windowUtils.createObjectURL(blob);
         return blobURL;
     };
--- a/src/common/file-url-provider.ts
+++ b/src/common/file-url-provider.ts
@@ -3,9 +3,9 @@
 import { WindowUtils } from './window-utils';
 
 export class FileURLProvider {
-    constructor(private windowUtils: WindowUtils, private provideBlob: (blobParts?: any[], mimeType?: string) => Blob) {}
+    constructor(private windowUtils: WindowUtils, private provideBlob: (blobParts: any[], mimeType: string) => Blob) {}
 
-    public provideURL = (blobParts?: any[], mimeType?: string): string => {
+    public provideURL = (blobParts: any[], mimeType: string): string => {
         const blob = this.provideBlob(blobParts, mimeType);
         const blobURL = this.windowUtils.createObjectURL(blob);
         return blobURL;
--- a/src/common/file-url-provider.ts
+++ b/src/common/file-url-provider.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { WindowUtils } from './window-utils';
 import { provideBlob } from './blob-provider';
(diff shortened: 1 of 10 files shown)


src/tests/unit/tests/DetailsView/components/export-dialog.test.tsx
```diff
@@ -52,35 +49,30 @@ describe('ExportDialog', () => {
...
         it.each(isOpenOptions)('with open %p', isOpen => {
-            provideBlobMock
-                .setup(p => p(It.isAny(), It.isAnyString()))
-                .returns(() => blobStub)
-                .verifiable(Times.once());
-            windowUtilsMock
-                .setup(w => w.createObjectURL(blobStub))
+            props.isOpen = isOpen;
+            fileProviderMock
+                .setup(f => f.provideURL(It.isAny(), It.isAnyString()))
```
Shobhit1: nit: you can probably  use `blobStub` here to do a strick typecheck than `It.isAny()`

src/common/file-url-provider.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { provideBlob } from './blob-provider';
+import { WindowUtils } from './window-utils';
+
+export class FileURLProvider {
+    constructor(private windowUtils: WindowUtils) {}
```
Shobhit1: This needs tests?

smoralesd: Agreed... you'll need to pass `provideBlob` through the constructor if you want to mock it on the test too.

src/tests/unit/tests/DetailsView/components/export-dialog.test.tsx
```diff
@@ -52,35 +49,30 @@ describe('ExportDialog', () => {
...
         it('closes the dialog onDismiss', () => {
             onCloseMock.setup(oc => oc()).verifiable(Times.once());
-            provideBlobMock
-                .setup(p => p(It.isAny(), It.isAnyString()))
-                .returns(() => blobStub)
-                .verifiable(Times.once());
-            windowUtilsMock
-                .setup(w => w.createObjectURL(blobStub))
+            fileProviderMock
+                .setup(f => f.provideURL(It.isAny(), It.isAnyString()))
```
smoralesd: `f` inside the `setup` call is a little short for a name, how does `provider` sounds?

src/common/file-url-provider.ts
```diff
@@ -0,0 +1,14 @@
...
+
+    public provideURL = (blobParts?: any[], mimeType?: string): string => {
```
haonliu: Will this cause error if both params are not provided? i.e. calling provideBlob()

Shobhit1: This is a good idea for one of the test cases that @annikakouhia can add for this file.

haonliu: Also I think a test case should be added in blob provider.

annikakouhia: Made them both mandatory parameters--blobs can be created without both, but for our purposes we want both values

src/tests/unit/common/file-url-provider.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+        const returnedURL = 'returned url';
+        const blobStub = {} as Blob;
+
+        provideBlobMock
+            .setup(pbm => pbm(content, mimeType))
+            .returns(() => blobStub)
+            .verifiable(Times.once());
+
+        windowUtilsMock
+            .setup(w => w.createObjectURL(blobStub))
```
Shobhit1: nit: You can rename `w` inside the `setup` function to a better name. `windowsUtilProvider` maybe?

src/tests/unit/tests/DetailsView/components/export-dialog.test.tsx
Shobhit1: nit: same as above. `f` can be renamed to something better!

Issue #699, opened 2019-05-23
Result description should maintain its content


mmafe:
## Is your feature request related to a problem? Please describe.
If I fill in the result description (after selecting Export result), and I accidentally close the window (either by clicking outside the window or selecting the close button), the content I typed in disappears.

Going forward, I've been keeping track of my summary findings in a separate document and then pasting that into the Results Description window when I'm ready to export. However, new users would not know to do this and could easily lose their work.

## Describe the desired outcome
It would be wonderful if the content in the Result Description was maintained even if you close it. I would be able to come back to my report and the Result Description would still be there. 

Even better would be to move the Result Description into the Overview page, so you can continually edit it while working on the report. Then once you click Export result, it would generate your report.

ferBonnin:
Hi @mmafe thanks for your suggestion and the interest to Accessibility Insights for Web.

Per conversation with @juliestuart-ms makes sense to keep the information if possible since the user can always clear the information entered. We will we will look into the timelines and effort for making this available and follow up.

ferBonnin:
Hi @mmafe this feature request has now been completed and should be available in [Canary ](https://chrome.google.com/webstore/detail/accessibility-insights-fo/hbcplehnakffdldhldncjlnbpfgogbem) now :)
Please let us know what you think!

PR #826, opened 2019-06-18
Persistent Export Description

annikakouhia:
#### Description of changes

Making the description textbox's text persist when users navigate away from the export dialog. Will persist even if Chrome is completely closed as long as user elects to continue previous assessment. Persists in Assessment, but not in FastPass.

New field "ResultDescription" to the AssessmentStoreData type. Populates upon description change following FLUX framework. When assessment is continued and export dialog box is opened, will populate description field based on the store's data.

Relevant tests added for the ExportDialog tests, ReportExportComponent tests (parent of ExportDialog), and InitialAssessmentStoreDataGenerator tests. Tests modified otherwise with mocks or dummy values for new parameters (including ResultDescription added in AssessmentStoreData and two new functions used by ExportDialog to retrieve and update the result description).

#### Pull request checklist

- [X] Addresses an existing issue: Addresses #699 

persistExample


haonliu:
Is this PR still a draft?

src/DetailsView/actions/details-view-action-message-creator.ts
```diff
@@ -345,6 +346,17 @@ export class DetailsViewActionMessageCreator extends DevToolActionMessageCreator
         });
     }
 
+    public addResultDescription(description: string): void {
```
waabid: I don't think I see tests for this.

src/DetailsView/components/issues-table.tsx
```diff
@@ -95,6 +95,12 @@ export class IssuesTable extends React.Component<IssuesTableProps> {
         );
     }
 
+    private nullUpdatePersistedDescription(value: string): void {}
+
+    private nullGetExportDescription = () => {
```
waabid: I think since you're returning empty string here, you should probably refer to this as 'getEmptyExportDescription'.

Additionally, you can simplify it by just doing:

private getEmptyExportDescription= () => '';

src/background/actions/assessment-action-creator.ts
```diff
@@ -98,6 +100,13 @@ export class AssessmentActionCreator {
         this.assessmentActions.addFailureInstance.invoke(payload);
     }
 
+    @autobind
+    private onAddResultDescription(payload: AddResultDescriptionPayload): void {
```
waabid: Seems like this may be missing tests as well.

PR #832, opened 2019-06-19
refactor: simplify new report components

smoralesd:
#### Description of changes

Main objective of this refactor: remove some conditional behavior from some components on the report by breaking the components into smaller components (without the 'if-else' this time).

This is part of fixing the heading level 3 on the report (right now screen readers announce heading level 3 several times while navigating through the links and button inside the rule group)

#### Pull request checklist

   - n/a
   - n/a
   - n/a


smoralesd:
I'm aware the we may want to change the name of some of the components (new and modified) to clarify what they do. I'm open to suggestions.

src/DetailsView/reports/components/report-sections/rules.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export type RulesDeps = RuleDetailDeps;
+
+export type RulesProps = {
+    deps: RulesDeps;
+    rules: RuleResult[];
+    outcomeType: InstanceOutcomeType;
+};
+
+export const Rules = NamedSFC<RulesProps>('RuleDetailsGroup', ({ rules, outcomeType, deps }) => {
```
Shobhit1: nit: What do you think about renaming it to `RulesGroup`? Rules alone doesn't provide me enough info about the component tbh.

smoralesd: We already have a `RuleDetailsGroup`. Not sure if renaming this to `RulesGroup` will actually clarify the usage and difference.
Don't get me wrong. I also think `Rules` is not the best name...

waabid: This is still named 'RuleDetailsGroup'

src/DetailsView/reports/components/report-sections/rules.tsx
```diff
@@ -0,0 +1,26 @@
...
+    return (
+        <div className="rule-details-group">
```
waabid: Should this class name be different considering the name (and the fact we have 'RuleDetailsGroup' somewhere else as a component)

smoralesd: Yes!!! that's part of the feedback I'm looking for. I don't have a good name for this component and `Rules` doesn't sound exactly right. I'll update the string name when I get to the final naming for the component itself.

Issue #829, opened 2019-06-19
Failure instance dialog is missing spacing & icons


karanbirsingh:
## Describe the bug

In prod version 2.5.4, the failure instance dialog is missing spacing and icons:
![image](https://user-images.githubusercontent.com/7775527/59799198-b58b4b80-9298-11e9-891a-71aa3af518bc.png)

**To Reproduce**
- Run automated checks on bing.com
- select any exclamation point to open the failure instance dialog

## Expected behavior

We expect more spacing between the buttons. Here is a screenshot:

![image](https://user-images.githubusercontent.com/7775527/59799320-f2574280-9298-11e9-8ebe-2708444a0f1e.png)

## Extension (please complete the following information)

- OS: Windows
- Version: 2.5.4
- Browser Version: Chrome 75.0.3770.100 (Official Build) (64-bit)
- Target Page: www.bing.com

waabid:
This seems to be due to an Office Fabric update.

PR #833, opened 2019-06-20
Restore instance dialog UI by moving css class assignment

karanbirsingh:
#### Description of changes

It seems the recent OfficeFabric update #782 may have led to the UI issues described in #829 . I created this [codepen ](https://codepen.io/anon/pen/zVoLJv) to show that using a css class in `layerProps` to target buttons leads to different behavior in OF versions `6.142.0` and `6.189.4`. OF is now on major version 7.

I'm not sure if the PR change is the best way to address this issue; it aligns with the code in this component & unblocks the UI regression in production: https://github.com/microsoft/accessibility-insights-web/blob/9baf98a369968f5e59182c37ec6072e8702e1960/src/popup/components/telemetry-permission-dialog.tsx#L52

image

#### Pull request checklist

- [x] Addresses an existing issue: #829 
- [NA] Added relevant unit test for your changes. (`yarn test`)
- [NA] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [unchanged] (UI changes only) Verified usability with NVDA/JAWS

### DIFF
--- a/src/injected/components/details-dialog.tsx
+++ b/src/injected/components/details-dialog.tsx
@@ -342,10 +342,9 @@ export class DetailsDialog extends React.Component<DetailsDialogProps, DetailsDi
                 }}
                 modalProps={{
                     isBlocking: false,
-                    containerClassName: 'insights-dialog-main-container',
+                    containerClassName: 'insights-dialog-main-override insights-dialog-main-container',
                     layerProps: {
                         onLayerDidMount: this.onLayoutDidMount,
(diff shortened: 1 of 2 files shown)


Shobhit1:
Do we plan to do a prod release for this fix, after this?

karanbirsingh:
> Do we plan to do a prod release for this fix, after its merged in master?

I think it makes sense, we can check via standup/teams as well

PR #836, opened 2019-06-20
fix: target change dialog not showing up for file urls

### LINKED ISSUE #591: URL change dialog doesn't show up for change in file urls
## Describe the bug

URL change dialog box doesn't show up when we go from one file url to another.

**To Reproduce**

1. Open a HTML file in chrome to test for accessibility issues
2. Run assessment by clicking on Assessment from extension's popup page.
3. Verify some assessments so that it makes an entry in indexed DB
4. Change the file url to another url by opening another HTML file.

## Expected behavior

The change in files should trigger the URL change dialog box to show up as it does when we change target page urls.

## GIF

persistence-no-dialog

OS: Windows 10
Version: 10.0.17763 Build 17763
Chrome Version: 74.0.3729.108 (Official Build) beta (64-bit)

Yes/No

Yes

Shobhit1:
#### Description of changes

Changes the way we compare urls for showing `target-change-dialog` especially for `file://` urls.

Compares protocols and hostnames!

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #591

##### Web urls
normal-urls

##### File Urls
file-urls


src/common/url-parser.ts
```diff
@@ -10,6 +10,26 @@ export class UrlParser {
     public areURLHostNamesEqual(urlA: string, urlB: string): boolean {
```
peterdur: Seems like what this method is doing has evolved, and its name should be updated to match, if only to describe the scenario where we want file URLs to compare hrefs but other URLs to compare hosts.

src/common/url-parser.ts
```diff
@@ -10,6 +10,26 @@ export class UrlParser {
...
+
+    public areFileUrlsEqual(urlA: string, urlB: string): boolean {
+        const urlAObj = new URL(urlA);
+        const urlBObj = new URL(urlB);
+
+        return urlAObj.href === urlBObj.href;
+    }
+
+    private areBothFileUrls(urlAObj: URL, urlBObj: URL): boolean {
+        return this.areUrlProtocolsEqual(urlAObj, urlBObj) && urlAObj.protocol === 'file:' && urlBObj.protocol === 'file:';
```
peterdur: We only need 2 of these 3 checks (if A=B and B=C then we know A=C).

src/common/url-parser.ts
```diff
@@ -10,6 +10,26 @@ export class UrlParser {
...
-        return urlAObj.hostname === urlBObj.hostname;
+
+        if (this.areBothFileUrls(urlAObj, urlBObj)) {
+            return this.areFileUrlsEqual(urlA, urlB);
+        }
+
+        return this.areUrlProtocolsEqual(urlAObj, urlBObj) && urlAObj.hostname === urlBObj.hostname;
+    }
+
+    public areFileUrlsEqual(urlA: string, urlB: string): boolean {
```
peterdur: Will `areFileUrlsEqual` need to be public? If not, can we refactor so we don't create a `URL` twice for each string?

src/common/url-parser.ts
Shobhit1: Yes. I added the `A=B` check later to make it more tight but forgot to remove `A=C` check. Will do.

src/common/url-parser.ts
Shobhit1: The idea was to give `URLParser` a function that can be used in the future to support this kind of functionality. But for the issue that I am trying to fix here; it doesn't need to be a `public` function; so will make the changes that you suggested. 👍

src/common/url-parser.ts
Shobhit1: Agreed. Will update the name. Any suggestions for it though? `equateURLs` ? or `areURLsEqual`?

Shobhit1:
Discussed with @ferBonnin. She thinks the solution is okay for file urls to compare the whole urls.

PR #838, opened 2019-06-21
React-based visualization components for highlight drawer

waabid:
#### Description of changes

I would like to use more React components for our injected content scripts. As such, this is a start for that effort; **please refer to highlight-box-drawer.ts** to see the re-factor work done here. These components are not being used anywhere (future PR for that).

#### Pull request checklist

### DIFF
--- a/src/injected/visualization/element-highlight.tsx
+++ b/src/injected/visualization/element-highlight.tsx
@@ -29,7 +29,7 @@ export interface ElementHighlightProps {
     getBoundingRect: GetBoundingRect;
 }
 
-export const ElementHighlight = NamedSFC<ElementHighlightProps>('HighlightBox', props => {
+export const ElementHighlight = NamedSFC<ElementHighlightProps>('ElementHighlight', props => {
     const { deps, drawerConfig, featureFlagStoreData, dialogRender, element, elementResult, bodyStyle, docStyle, getBoundingRect } = props;
     const { clientUtils, drawerUtils } = deps;
 
--- a/src/injected/visualization/element-highlight.tsx
+++ b/src/injected/visualization/element-highlight.tsx
@@ -30,7 +30,7 @@ export interface ElementHighlightProps {
 }
 
 export const ElementHighlight = NamedSFC<ElementHighlightProps>('HighlightBox', props => {
-    const { deps, drawerConfig, dialogRender, element, elementResult, bodyStyle, docStyle, getBoundingRect } = props;
+    const { deps, drawerConfig, featureFlagStoreData, dialogRender, element, elementResult, bodyStyle, docStyle, getBoundingRect } = props;
     const { clientUtils, drawerUtils } = deps;
 
     if (!drawerConfig.showVisualization) {
@@ -45,7 +45,7 @@ export const ElementHighlight = NamedSFC<ElementHighlightProps>('HighlightBox', 
     }
 
     const onClickFailureBoxHandler = () => {
-        dialogRender(elementResult, this.featureFlagStoreData);
+        dialogRender(elementResult, featureFlagStoreData);
     };
 
     const offset = clientUtils.getOffset(element);
--- a/src/injected/visualization/element-highlight.tsx
+++ b/src/injected/visualization/element-highlight.tsx
@@ -1,0 +1,73 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { CSSProperties } from 'react';
+import * as React from 'react';
+
+import { NamedSFC } from '../../common/react/named-sfc';
+import { FeatureFlagStoreData } from '../../common/types/store-data/feature-flag-store-data';
+import { ClientUtils } from '../client-utils';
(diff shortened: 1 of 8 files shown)


src/injected/visualization/element-highlight.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export interface ElementHighlightProps {
+    deps: ElementHighlightDeps;
+    element: Element;
+    elementResult: HtmlElementAxeResults;
+    bodyStyle: CSSStyleDeclaration;
+    docStyle: CSSStyleDeclaration;
+    drawerConfig: DrawerConfiguration;
+    featureFlagStoreData: FeatureFlagStoreData;
+    dialogRender: RenderDialog;
```
smoralesd: should `dialogRender` and `getBoundingRect` be part of the deps instead?

src/injected/visualization/element-highlight.tsx
```diff
@@ -0,0 +1,73 @@
...
+    getBoundingRect: GetBoundingRect;
+}
+
+export const ElementHighlight = NamedSFC<ElementHighlightProps>('HighlightBox', props => {
```
smoralesd: `displayName` doesn't match component name, is that on purpose?

src/injected/visualization/highlight-box.tsx
```diff
@@ -0,0 +1,38 @@
...
+    }
+    const styles: CSSProperties = {
+        background: boxConfig.background,
+        color: boxConfig.fontColor,
+        width: boxConfig.boxWidth,
+        cursor: drawerConfig.cursor,
+        textAlign: drawerConfig.textAlign,
+    };
+
+    const combinedClass = 'insights-highlight-text ' + className;
```
smoralesd: can we use `css` from `'@uifabric/utilities'` to handle the class names?

src/injected/visualization/element-highlight.tsx
```diff
@@ -0,0 +1,73 @@
...
+        outlineColor: drawerConfig.borderColor,
+        top: drawerUtils.getContainerTopOffset(offset),
+        left: drawerUtils.getContainerLeftOffset(offset),
+        minWidth: drawerUtils.getContainerWidth(offset, currentDom, elementBoundingClientRect.width, bodyStyle, docStyle),
+        minHeight: drawerUtils.getContainerHeight(offset, currentDom, elementBoundingClientRect.height, bodyStyle, docStyle),
+    };
+
+    return (
+        <div title={drawerConfig.toolTip} className={'insights-highlight-box'} style={styles}>
+            <HightlightBox deps={deps} drawerConfig={drawerConfig} boxConfig={drawerConfig.textBoxConfig} className={'text-label'} />
```
smoralesd: I'm a little confused with the 2 `<HighlightBox>` here. From looking at the divs of the visualizations on master I can see automated checks visualization uses `failure-label` class name for the inner div that contains the exclamation mark `"!"` but not the `text-label`, while the headings visualization does the opposite (uses `text-label` but no `failure-label`. Is this right? or am I missing something.

If it is right (we use one or the other but not both at the same time), then it seems better to reflect that in this component by using only one `HighlightBox`.
If it is not right (there are case we want both) then maybe we can explore other approaches like using `children` in this  component and let the user class to decide which and how many `HighlihtBox`'s does it needs.

Thoughts?

src/injected/visualization/element-highlight.tsx
waabid: Good catch!

src/injected/visualization/element-highlight.tsx
waabid: Unfortunately not; these are decided/created just prior to this component.

dialogRender uses data from the selected item.

getBoundingRect uses the element OR the drawerUtils IIRC.

src/injected/visualization/element-highlight.tsx
waabid: I would re-visit what `createHighlightElement` is doing in `highlight-box-drawer` currently as this is just reproducing what that is doing but with react components.

> If it is not right (there are case we want both) then maybe we can explore other approaches like using children in this component and let the user class to decide which and how many HighlihtBox's does it needs.

Yes, we would want both.

Regarding responsibility: I am not sure I would want to tackle that change right now. From what I understand, this would be a change to handling this in the formatter or something in that vein, since that is how the highlight boxes are configured (so they would seem a far better fit; they all use the same Drawer as of now). I would much rather target moving towards React drawers that behave/maintain design consistency with how our current drawers behave, such that we can move to React drawers, and then start looking at how to better implement responsibility of behavioral differences.

Not sure if that makes sense but the basic gist is I'd like to not change formatters right now. Formatters are the ones responsible for differentiating between visualization differences **implicitly** through drawer configurations. I do think that ideally we should/could move in that direction but I'd like to do that later as my priority would be to move to using react in drawers with our current code.

Let me know if you want to talk more about this offline. 😄

smoralesd: It does make sense to apply this changes on a separated PR

PR #839, opened 2019-06-21
Reuse summary bar in overview and assessment report

haonliu:
#### Description of changes
Reuse summary bar in overview and assessment report.
Create a separate scss file for summary bar.
Removed old summary bar component.
before: 
image

after: 
image

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1556404


smoralesd:
There are scss changes. Can you add screenshots of the results? Before and after would be great if that's possible.

src/DetailsView/reports/components/outcome-summary-bar.tsx
```diff
@@ -1,33 +1,37 @@
...
-
-const outcomeText = {
-    pass: 'Passed',
-    incomplete: 'Incomplete',
-    fail: 'Failed',
+export type OutcomeSummaryBarProps = {
+    outcomeStats: Partial<OutcomeStats>;
+    allOutcomeTypes: OutcomeType[];
+    units?: OutcomeUnits;
+    inverted?: boolean;
```
smoralesd: may we rename this to 'iconStyle' or 'invertedIconStyle'... inverted made me think of order (inverted => reverse order) which is not the case.

src/DetailsView/reports/components/outcome-summary-bar.tsx
```diff
@@ -1,33 +1,37 @@
...
+export const OutcomeSummaryBar = NamedSFC<OutcomeSummaryBarProps>('OutcomeSummaryBar', props => {
+    return (
+        <div className="outcome-summary-bar">
+            {props.allOutcomeTypes.map(outcomeType => {
+                const { units, inverted } = props;
+                const text = outcomeTypeSemantics[outcomeType].pastTense;
+                const iconMap = inverted === true ? outcomeIconMapInverted : outcomeIconMap;
+                const outcomeIcon = iconMap[outcomeType];
+                const count = props.outcomeStats[outcomeType];
+                const suffix = units === 'percentage' ? '% ' : '';
```
smoralesd: It seems we can promote the `suffix` to be a prop instead of having this tertiary operator here, that way, if we need a different, new suffix, it'll be up the the user class to pass the proper suffix (this align with open/close principle by the way).
For the name of the property, maybe `countSuffix1` is good enough

src/DetailsView/reports/components/outcome-summary-bar.tsx
```diff
@@ -1,33 +1,37 @@
...
+    inverted?: boolean;
 };
 
-export type OutcomeSummaryBarProps = RequirementOutcomeStats & { units?: OutcomeUnits };
+export const OutcomeSummaryBar = NamedSFC<OutcomeSummaryBarProps>('OutcomeSummaryBar', props => {
+    return (
+        <div className="outcome-summary-bar">
+            {props.allOutcomeTypes.map(outcomeType => {
+                const { units, inverted } = props;
+                const text = outcomeTypeSemantics[outcomeType].pastTense;
```
smoralesd: `outcomeTypeSemantics` is a hard dependency here (we're importing it from a different file and using it directly on the component).
I wonder if it make sense to receive this map as a prop instead...

src/DetailsView/reports/components/outcome-type.tsx
```diff
@@ -6,9 +6,15 @@ import { InapplicableIcon, InapplicableIconInverted } from '../../../common/icon
 import { CheckIcon, CheckIconInverted } from './../../../common/icons/check-icon';
 import { CircleIcon } from './../../../common/icons/circle-icon';
 import { CrossIcon, CrossIconInverted } from './../../../common/icons/cross-icon';
-import { InstanceOutcomeType } from './report-sections/outcome-summary-bar';
 import { RequirementOutcomeType } from './requirement-outcome-type';
 
+export type OutcomeUnits = 'percentage' | 'requirements';
+
+export type InstanceOutcomeType = 'pass' | 'fail' | 'inapplicable';
+export const InstanceOutcomeTypes: OutcomeType[] = ['fail', 'pass', 'inapplicable'];
```
smoralesd: Can this be of type `InstanceOutcomeType[]` instead?

src/DetailsView/reports/components/outcome-type.tsx
```diff
@@ -6,9 +6,15 @@ import { InapplicableIcon, InapplicableIconInverted } from '../../../common/icon
...
+export const RequirementOutcomeTypes: OutcomeType[] = ['pass', 'incomplete', 'fail'];
```
smoralesd: This one caught my attention:
1. can this be of type `RequirementOutcomeTypes[]` instead?
2. Also, there is a difference on where we define types for requirement and instances. While instance is all here (`InstanceOutcomeType` and `InstanceOutcomeTypes`), that's not the case for requirement (notice the import on line 9: `import { RequirementOutcomeType } from './requirement-outcome-type';`; I'm curious, why not follow the same pattern?

src/DetailsView/reports/components/outcome-summary-bar.tsx
haonliu: Do you want to put it in the deps? I feel this is fine here since it is just a static object. But I'm open for changes.

smoralesd: It's an static object we're using as a global... not sure that's a pattern we want to follow. Also, we need to consider that right now, the object has definition for 2 sets of slightly different things, requirements and instances. Breaking this object apart will enable us to update/modified this concepts separately and the I guess one of the first step to do that in the near future is to not have a hard dependency on the object. Up to you.

haonliu: Sounds good. I'll leave this PR as it is since there is already 20+ file changes.

PR #845, opened 2019-06-24
refactor: clean up rule detail component

smoralesd:
#### Description of changes

Clean up `RuleDetail` component
- remove `children` rendering as is not being used
- inline the use of `InstanceListGroupHeader` as is only being wrapped on a `<div>` with a hardcoded `className`
- Remove `InstanceListGroupHeader` as is not being used anymore

#### Pull request checklist

- [x] Addresses an existing issue: part of WI # 1558756
   - no changes
   - no changes


src/DetailsView/reports/components/report-sections/rule-detail.tsx
```diff
@@ -17,16 +25,64 @@ export type RuleDetailProps = {
 };
 
 export const RuleDetail = NamedSFC<RuleDetailProps>('RuleDetails', props => {
```
haonliu: Seems like all this component renders is the group header, do we want to consider renaming this component?

smoralesd: I'm not sure... but, this is just part of the changes I'm making for fix the collapsible control so... I can rename later. Not to worried about it.

PR #859, opened 2019-06-25
fix: inspect HTML button should not be disabled and show warning error when devtools are not open

### LINKED ISSUE #547: Inspect HTML button should not be disabled

Shobhit1:
#### Description of changes

Change the way `Inspect HTML Element` button behaves now. It is not disabled anymore and shows a Red error message when pressed (if devtools are not open).

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #547

#### GIF
inspect-html-button

#### Behavior with the other error message:

image


src/tests/unit/tests/injected/details-dialog-handler.test.ts
```diff
@@ -105,16 +115,54 @@ describe('DetailsDialogHandlerTest', () => {
...
+        const devToolActionMessageCreatorMock = Mock.ofType<DevToolActionMessageCreator>(undefined, MockBehavior.Strict);
+        const eventFactory = new EventStubFactory();
+        const event = eventFactory.createMouseClickEvent();
+
+        devToolStoreMock
+            .setup(store => store.getState())
+            .returns(() => {
+                return {
+                    isOpen: false,
+                } as any;
```
waabid: you can set this as the state interface instead of as any.

src/tests/unit/tests/injected/details-dialog-handler.test.ts
```diff
@@ -105,16 +115,54 @@ describe('DetailsDialogHandlerTest', () => {
...
+            })
+            .verifiable(Times.once());
+
+        detailsDialogMock
+            .setup(dialog => dialog.props)
+            .returns(() => {
+                return {
+                    devToolActionMessageCreator: devToolActionMessageCreatorMock.object,
+                    devToolStore: devToolStoreMock.object,
+                } as any;
```
waabid: Same here; use the props interface instead of as any.

Shobhit1:
> LGTM except for the minor comments

Updated. Thanks for the feedback 👍

smoralesd:
Can you have @ferBonnin have a look at the UX here? I'm not sure this is the UX we want. 
Also, did you check how this play with the other message we have on the same area (issue filing when there is no issue filing service set up?

src/injected/components/details-dialog.tsx
```diff
@@ -70,6 +71,7 @@ export class DetailsDialog extends React.Component<DetailsDialogProps, DetailsDi
     public isBackButtonDisabled: () => boolean;
     public isNextButtonDisabled: () => boolean;
     public isInspectButtonDisabled: () => boolean;
+    public showInspectButtonMessage: () => boolean;
```
smoralesd: I don't see any usage of `showInspectButtonMessage` in this class. Is this for the only purpose to call it from a test?
I can see we have this pattern on this class already but it would be nice if we can get off it.
It looks like it make sense to move the buttons out of this component, make a new small component with them. It also looks like this issue is a great opportunity to do that refactor.
Thoughts?

ferBonnin:
@Shobhit1 agree with Sebastian comment. The message should only appear when the user selects the button and dev tols are closed. I also recommend testing with a screen reader to make sure the announcement is made

Shobhit1:
> @Shobhit1 agree with Sebastian comment. The message should only appear when the user selects the button and dev tols are closed. I also recommend testing with a screen reader to make sure the announcement is made

It shows up only when we click on the button. I will update the GIF for that case.

src/injected/components/details-dialog.tsx
Shobhit1: I agree with what you are saying regarding the refactor; but I would rather get this issue done with  minimal changes first and then take refactoring the `details-dialog` as a later PR.
For `showInspectButtonMessage` this is used to check if initially we want to show the `inspect html error` or not in function `renderInspectMessage`

Shobhit1:
> Can you have @ferBonnin have a look at the UX here? I'm not sure this is the UX we want.
> Also, did you check how this play with the other message we have on the same area (issue filing when there is no issue filing service set up?

Updated GIF and screenshot for better understanding at UX.

ferBonnin:
> > @Shobhit1 agree with Sebastian comment. The message should only appear when the user selects the button and dev tols are closed. I also recommend testing with a screen reader to make sure the announcement is made
> 
> @ferBonnin It shows up only when we click on the button. I will update the GIF for that case.

@Shobhit1 new gif is clearer, thanks! looks good

src/injected/components/details-dialog.tsx
smoralesd: So are you fine putting more tech debt on an already dirty component? I think I'll do the refactor first and only then do the fix. That should be our normal mode of operation.

As per `showInspectButtonMessage`... it's not being used on 'renderInspectMessage'... what we're using there on this PR is:
```typescript
if (this.props.dialogHandler.shouldShowInspectButtonMessage(this)) {
```
basically, using the `dialogHandler` directly, and thus, more clean up to do in the future.

Shobhit1: I never said that I am fine with adding more tech debt. I will take care of it but I was willing to delay it after we get this in. On `showInspectButtonMessage` i followed whatever pattern we were already using in this class.

smoralesd: `showInspectButtonMessage` is not being called, only assign!

Shobhit1: will make the changes.

Shobhit1: Removed the `showInspectButtonMessage`.

src/injected/components/details-dialog.tsx
```diff
@@ -200,7 +201,7 @@ export class DetailsDialog extends React.Component<DetailsDialogProps, DetailsDi
     }
 
     private renderInspectMessage(): JSX.Element {
-        if (this.props.dialogHandler.isInspectButtonDisabled(this)) {
+        if (this.props.dialogHandler.shouldShowInspectButtonMessage(this)) {
             return (
                 <div className="insights-dialog-inspect-disabled">
                     {`To enable the Inspect HTML button, open the developer tools (${this.props.devToolsShortcut}).`}
```
peterdur: Seems like this text should be updated (since the button is actually enabled all the time).
Maybe change to "To **use** the Inspect HTML button., **first** open..."?

Shobhit1: updated.

Issue #860, opened 2019-06-26
Accessibility Insights for Web crashing in Linux


ferBonnin:
## Describe the bug

Per support comments in Chrome store, Version 2.5.5 is broken in Linux
https://chrome.google.com/webstore/detail/accessibility-insights-fo/pbjjkligggfmakdaogkfomddhfmpjeni?pli=1 

**To Reproduce**
Use Linux Mint 17

## Expected behavior

Accessibility Insights for Web should work

## Extension (please complete the following information)

- OS: [e.g. Windows/Mac] Windows 
- Version [e.g. 1.140.1] 2.5.5 
- Browser Version [e.g Chrome 71.0] 75.0.3770.100
- Target Page [e.g www.bing.com] any

## Additional context

reported in Chrome store

Shobhit1:
Resolved with 2.6.0 release.

PR #864, opened 2019-06-26
fix: correct typos to make extension not crash on Linux

Shobhit1:
#### Description of changes

Fix typo in a path for devtools.html in manifest.
This is an ongoing step to fix the crash issue in Linux.

#### Pull request checklist

- [x] Addresses an existing issue:  #860 

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -4,7 +4,7 @@ jobs:
     - job: 'unit_tests_and_lints'
 
       pool:
-          vmImage: 'macOS-10.13'
+          vmImage: ${{ parameters.vmImage }}
 
       steps:
           - template: pipeline/prerequisites.yaml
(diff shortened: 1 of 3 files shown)


Shobhit1:
> Changes look good.
> Although, we have seen this issues on the past. This PR will only fix this instance.
> I'm wondering if we should add some more robust validation before calling the issue closed.

I am preparing to write at least a one note page on how to debug this. One of the way that we can do this is by running E2E on ubuntu but since we are using only one `build.yaml`; I dont know how we can split our PR and CI builds for that. Will need to investigate further.
That said; we should merge this and release so that we can unblock Linux users.

Shobhit1:
Verified the playground build on Ubuntu Hyper-V.

PR #865, opened 2019-06-27
chore: remove usage of @autobind to unblock office fabric 7.x upgrade

dbjorge:
#### Description of changes

Replaces all usages of autobind (mostly with arrow functions). This helps move us towards upgrading to office fabric v7.x (which removes the previously deprecated autobind utility)

Most changes in this PR are unremarkable, mechanical conversions. The points that require a little more review are:

1. `src/injected/main-window-initializer.ts` and `src/injected/window-initializer.ts` use an explicit `.bind(this)` instead of an arrow function because the way the former uses `super.methodCall()` cannot be directly replicated in ES5 when `methodCall` is changed from a method to a property.
1. Some tests used `typemoq` in a way that specifically relied on mocked methods being methods, rather than properties that are functions. These tests were rewritten in terms of verifying the act of calling the things under test, rather than verifying the identity of the things under test. The impacted cases are primarily in `src/tests/unit/tests/DetailsView/components/preview-features-toggle-list.test.tsx` and `src/tests/unit/tests/common/components/gear-options-button-component.test.tsx`

#### Pull request checklist

### DIFF
--- a/src/injected/main-window-initializer.ts
+++ b/src/injected/main-window-initializer.ts
@@ -1,7 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { autobind } from '@uifabric/utilities';
-
 import { Assessments } from '../assessments/assessments';
 import { AxeInfo } from '../common/axe-info';
 import { InspectConfigurationFactory } from '../common/configs/inspect-configuration-factory';
@@ -206,13 +204,12 @@ export class MainWindowInitializer extends WindowInitializer {
         await Promise.all(asyncInitializationSteps);
     }
 
-    @autobind
-    protected dispose(): void {
+    protected dispose = (): void => {
         super.dispose();
         this.frameCommunicator.dispose();
         this.tabStoreProxy.dispose();
         this.visualizationScanResultStoreProxy.dispose();
         this.visualizationStoreProxy.dispose();
         this.devToolStoreProxy.dispose();
-    }
+    };
 }
--- a/src/injected/main-window-initializer.ts
+++ b/src/injected/main-window-initializer.ts
@@ -204,12 +204,16 @@ export class MainWindowInitializer extends WindowInitializer {
         await Promise.all(asyncInitializationSteps);
     }
 
-    protected dispose = (): void => {
-        super.dispose();
+    protected dispose(): void {
+        // Using .bind(this) instead of an arrow function because ES5 only supports accessing methods (not properties) of super
+        // When we move to target ES6, this can be simplified to super.dispose()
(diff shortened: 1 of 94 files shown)


src/injected/main-window-initializer.ts
```diff
@@ -206,9 +204,12 @@ export class MainWindowInitializer extends WindowInitializer {
         await Promise.all(asyncInitializationSteps);
     }
 
-    @autobind
     protected dispose(): void {
-        super.dispose();
+        // Using .bind(this) instead of an arrow function because ES5 only supports accessing methods (not properties) of super
+        // When we move to target ES6, this can be simplified to super.dispose()
```
devaradhanm: >/ When we move to target ES6, this can be simpl [](start = 9, length = 47)

can we just target to es6? chrome anyway supports es6 rite?

src/injected/main-window-initializer.ts
```diff
@@ -206,9 +204,12 @@ export class MainWindowInitializer extends WindowInitializer {
...
+        // See https://github.com/microsoft/TypeScript/issues/338
+        super.dispose.bind(this)();
```
devaradhanm: > super.dispose.bind(this)(); [](start = 7, length = 28)

when u bind, super.dispose does it also bind this.dispose?

src/injected/main-window-initializer.ts
dbjorge: I think this would be a good idea, but out of scope from this PR

src/tests/unit/tests/DetailsView/components/preview-features-toggle-list.test.tsx
```diff
@@ -32,37 +33,13 @@ describe('PreviewFeaturesToggleListTest', () => {
...
-                    enabled={props.displayedFeatureFlags[1].enabled}
-                    onClick={setFeatureFlagStub}
-                    id={props.displayedFeatureFlags[1].id}
-                    key={`preview_feature_toggle${props.displayedFeatureFlags[1].id}`}
-                />
-            </div>
-        );
-
-        expect(testSubject.render()).toEqual(expectedComponent);
+        const testSubject = shallow(<PreviewFeaturesToggleList {...props} />);
```
devaradhanm: > c [](start = 7, length = 2)

with this change, we don't validate if onclick is setup properly rite?

before this change:
actionMessageCreator.setFeatureFlag is tested (in another ut file)
onClick=actionMessageCreator.setFeatureFlag

src/injected/main-window-initializer.ts
dbjorge: No, good catch. The correct place to do this is actually at the callsite of the child. Fixed this.

PR #874, opened 2019-06-28
refactor: extract feature flag action creator

smoralesd:
#### Description of changes

Extract feature flag related code from global action creator to new feature flag action creator

#### Pull request checklist

   - no changes
   - no changes

### DIFF
--- a/src/background/global-action-creators/feature-flags-action-creator.ts
+++ b/src/background/global-action-creators/feature-flags-action-creator.ts
@@ -13,7 +13,7 @@ export class FeatureFlagsActionCreator {
         private readonly telemetryEventHandler: TelemetryEventHandler,
     ) {}
 
-    public registerCallback(): void {
+    public registerCallbacks(): void {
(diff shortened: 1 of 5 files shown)


src/background/global-action-creators/feature-flags-action-creator.ts
```diff
@@ -0,0 +1,34 @@
...
+import { TelemetryEventHandler } from '../telemetry/telemetry-event-handler';
+
+export class FeatureFlagsActionCreator {
+    constructor(
+        private readonly interpreter: Interpreter,
+        private readonly featureFlagActions: FeatureFlagActions,
+        private readonly telemetryEventHandler: TelemetryEventHandler,
+    ) {}
+
+    public registerCallback(): void {
```
lisli1: Kind of a nit, but should we have this be `registerCallbacks` (with an 's' at the end) to be consistent with the `GlobalActionCreator`?

smoralesd: Good catch! We do want them to be consistent.

PR #875, opened 2019-06-28
Basic UI for Path/Snippet Addition to Failures

annikakouhia:
#### Description of changes

Adding fields for the path/snippet in the Failure Instance Panel *only* when Manual Instance Detail Feature Flag is flipped to "on." Lots of edits to get the Feature Flag Store Data all the way to the Failure Instance Panel. 
Most test changes are to reflect this new prop added to the flow. One unit test added to Failure Instance Panel Test to check that UI reflects classic version when flag is off and new version when flag is on.
Finally added snippet and path to the state of Failure Instance Panel (where comment was/is as well).

#### Pull request checklist

- [X] Addresses an existing issue: First PR for Feature #1555720

FailureInstanceDetails

### DIFF
--- a/src/DetailsView/components/failure-instance-panel-control.tsx
+++ b/src/DetailsView/components/failure-instance-panel-control.tsx
@@ -48,7 +48,7 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
             isPanelOpen: false,
             failureDescription: this.props.originalText || '',
             selector: '',
-            snippet: 'Code snippet will auto-populate based on the CSS selector input.',
+            snippet: '',
         };
     }
 
@@ -125,7 +125,7 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
                 </div>
                 <div>
                     <label>Code Snippet</label>
-                    <div> {this.state.snippet} </div>
+                    <div>{this.state.snippet}</div>
                 </div>
                 <TextField
                     className="observed-failure-textfield"
--- a/src/DetailsView/components/failure-instance-panel-control.tsx
+++ b/src/DetailsView/components/failure-instance-panel-control.tsx
@@ -130,7 +130,7 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
         );
     }
 
-    protected getSelectorAndSnippet = (): JSX.Element => {
+    private getSelectorAndSnippet = (): JSX.Element => {
         return (
             <div>
                 <a className="learn-more"> Learn more about adding failure instances </a>
@@ -156,43 +156,15 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
         );
     };
 
-    protected panelWithoutSelectorAndSnippet = (testStepConfig: Readonly<Requirement>, panelProps: GenericPanelProps): JSX.Element => {
-        return (
-            <GenericPanel {...panelProps}>
-                {testStepConfig.addFailureInstruction}
-                <TextField
-                    className="observed-failure-textfield"
-                    label="Observed failure"
-                    multiline={true}
-                    rows={8}
-                    value={this.state.failureDescription}
-                    onChange={this.onFailureDescriptionChange}
-                    resizable={false}
-                />
-                <ActionAndCancelButtonsComponent
-                    isHidden={false}
-                    primaryButtonDisabled={this.state.failureDescription === ''}
-                    primaryButtonText={this.props.actionType === CapturedInstanceActionType.CREATE ? 'Add' : 'Save'}
-                    primaryButtonOnClick={
-                        this.props.actionType === CapturedInstanceActionType.CREATE
-                            ? this.onAddFailureInstance
-                            : this.onSaveEditedFailureInstance
-                    }
-                    cancelButtonOnClick={this.closeFailureInstancePanel}
-                />
-            </GenericPanel>
-        );
-    };
-
     protected onFailureDescriptionChange = (event, value: string): void => {
         this.setState({ failureDescription: value });
     };
 
-    protected onSelectorChange = (event, value: string): void => {
+    private onSelectorChange = (event, value: string): void => {
         this.setState({ selector: value });
     };
 
-    protected onValidateSelector = (event): void => {
+    private onValidateSelector = (event): void => {
         const currSelector = this.state.selector;
         this.setState({ snippet: 'snippet for ' + currSelector });
     };
--- a/src/DetailsView/components/failure-instance-panel-control.tsx
+++ b/src/DetailsView/components/failure-instance-panel-control.tsx
@@ -1,12 +1,16 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { ActionButton } from 'office-ui-fabric-react/lib/Button';
+import { ActionButton, DefaultButton } from 'office-ui-fabric-react/lib/Button';
 import { Icon } from 'office-ui-fabric-react/lib/Icon';
 import { Link } from 'office-ui-fabric-react/lib/Link';
 import { TextField } from 'office-ui-fabric-react/lib/TextField';
 import * as React from 'react';
 
 import { AssessmentsProvider } from '../../assessments/types/assessments-provider';
+import { Requirement } from '../../assessments/types/requirement';
+import { BaseStore } from '../../common/base-store';
+import { FeatureFlags } from '../../common/feature-flags';
+import { FeatureFlagStoreData } from '../../common/types/store-data/feature-flag-store-data';
 import { VisualizationType } from '../../common/types/visualization-type';
 import { ActionAndCancelButtonsComponent } from './action-and-cancel-buttons-component';
 import { GenericPanel, GenericPanelProps } from './generic-panel';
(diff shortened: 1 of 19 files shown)


src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -88,6 +97,62 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
         };
 
+        if (this.props.featureFlagStoreData && this.props.featureFlagStoreData[FeatureFlags.manualInstanceDetails]) {
+            return this.panelWithSelectorAndSnippet(testStepConfig, panelProps);
+        } else {
+            return this.panelWithoutSelectorAndSnippet(testStepConfig, panelProps);
+        }
+    }
+
+    protected panelWithSelectorAndSnippet = (testStepConfig: Readonly<Requirement>, panelProps: GenericPanelProps): JSX.Element => {
```
waabid: This seems like it could be it's own component.

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -88,6 +97,62 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
             closeButtonAriaLabel: null,
         };
 
+        if (this.props.featureFlagStoreData && this.props.featureFlagStoreData[FeatureFlags.manualInstanceDetails]) {
```
waabid: This is the right idea but we generally use `FlaggedComponent` in these scenarios. Look into examples of that in the code base and you should be able to use that instead.

src/tests/unit/tests/DetailsView/components/failure-instance-panel-control.test.tsx
```diff
@@ -140,12 +157,28 @@ describe('FailureInstancePanelControlTest', () => {
...
+            step: 'missingHeadings',
+            test: VisualizationType.HeadingsAssessment,
+            addFailureInstance: addInstanceMock.object,
+            actionType: actionType,
+            assessmentsProvider: Assessments,
+            featureFlagStoreData: featureFlagStoreData,
+        };
+    }
+
+    function createPropsWithTypeAndData(
```
smoralesd: To minimize duplicated code between this 2 functions, you may want to have a look at our `BaseDataBuilder` (src\tests\unit\common\base-data-builder.ts) which was create for this purposes. ;)

src/DetailsView/components/failure-instance-panel-control.tsx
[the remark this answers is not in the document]
smoralesd: I think I prefer @waabid idea to use `FlaggedComponent` here!

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -88,6 +97,62 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+                <div>
+                    <DefaultButton text="Validate CSS selector" onClick={this.onValidateSelector} disabled={this.state.selector === ''} />
+                </div>
+                <div>
+                    <label>Code Snippet</label>
+                    <div>{this.state.snippet}</div>
+                </div>
+                <TextField
+                    className="observed-failure-textfield"
+                    label="Observed failure"
```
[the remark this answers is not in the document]
annikakouhia: I've left it as Observed Failure for now just because that's what it currently is online and since I'm using the Flagged Component I didn't want to change the UI until the feature is released entirely. Let me know if you think I should still change now!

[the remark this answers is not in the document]
smoralesd: I understand your argument here and I may be fine with it. I just see on small issue with it, the general idea of the using feature flag to hide "under development" feature is to make sure we can develop the new UI/UX without altering the "old" one. By not updating the usage of this`<TextField>` (which contains the `"Observed failure"` text) we're not fully following that design principle. Now, I don't think is a big issue, it's just, any time we demo this we need to make a note for the audience that this text is "fixed" and not behind the feature flag, which, again, I guess it's fine but it's kind of weird. If you really want to be crisp about the feature flag use, it seems better to move this behind the feature flag too. 
Up to you at this point.

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -115,10 +130,73 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
         );
     }
 
+    protected getSelectorAndSnippet = (): JSX.Element => {
```
smoralesd: any particular reason this function is `protected`??? are we extending this class somewhere?

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -115,10 +130,73 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+                </div>
+                <div>
+                    <label>Code Snippet</label>
+                    <div>{this.state.snippet}</div>
+                </div>
+            </div>
+        );
+    };
+
+    protected panelWithoutSelectorAndSnippet = (testStepConfig: Readonly<Requirement>, panelProps: GenericPanelProps): JSX.Element => {
```
smoralesd: I didn't find any usage of this function (vscode shows "0 references" and I commented out the code and it still builds). 
If we're not using it, let's remove it!

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -115,10 +130,73 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+                />
+            </GenericPanel>
+        );
+    };
+
     protected onFailureDescriptionChange = (event, value: string): void => {
         this.setState({ failureDescription: value });
     };
 
+    protected onSelectorChange = (event, value: string): void => {
```
smoralesd: I see only one reference to this function (in this same class). Any special reason we need it to be `protected` instead of `private`?
Also, it looks like we're not testing this function. You may need to ask @karanbirsingh for more guidance on how to test this.

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -115,10 +130,73 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+        this.setState({ selector: value });
+    };
+
+    protected onValidateSelector = (event): void => {
```
smoralesd: Same here as with `onSelectorChange`
I see only one reference to this function (in this same class). Any special reason we need it to be `protected` instead of `private`?
Also, it looks like we're not testing this function. You may need to ask @karanbirsingh for more guidance on how to test this.

src/DetailsView/components/failure-instance-panel-control.tsx
annikakouhia: whoops! forgot to delete!

src/DetailsView/components/failure-instance-panel-control.tsx
annikakouhia: great point, switching to private since it's internal use only

src/DetailsView/components/failure-instance-panel-control.tsx
smoralesd: I still see this function as `protected` and the comment was resolved without comment.

annikakouhia: I originally changed it back to private, but after restructuring was passing it down to the Failure Instance Panel Details file through props--so thought it had to be protected again to be used outside of Failure Instance Panel Control? Maybe not since we're passing it directly through props and Failure Instance Panel Details isn't extending the Failure Instance Panel Control--is that correct?

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -88,6 +97,62 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+                />
+                <div>Note: If the CSS selector maps to multiple snippets, the first will be selected.</div>
+                <div>
+                    <DefaultButton text="Validate CSS selector" onClick={this.onValidateSelector} disabled={this.state.selector === ''} />
+                </div>
+                <div>
+                    <label>Code Snippet</label>
+                    <div>{this.state.snippet}</div>
+                </div>
+                <TextField
```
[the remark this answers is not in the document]
annikakouhia: Going to finalize to match Figma in a separate PR. For now planning on just getting fields to exist for testing purposes!

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -88,6 +97,62 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+        if (this.props.featureFlagStoreData && this.props.featureFlagStoreData[FeatureFlags.manualInstanceDetails]) {
+            return this.panelWithSelectorAndSnippet(testStepConfig, panelProps);
+        } else {
+            return this.panelWithoutSelectorAndSnippet(testStepConfig, panelProps);
+        }
+    }
+
+    protected panelWithSelectorAndSnippet = (testStepConfig: Readonly<Requirement>, panelProps: GenericPanelProps): JSX.Element => {
+        return (
+            <GenericPanel {...panelProps}>
```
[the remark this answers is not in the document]
annikakouhia: Done!

src/DetailsView/components/failure-instance-panel-control.tsx
annikakouhia: Ran tests with private vs. protected and all passed so pushed with private functions!

annikakouhia:
> LGTM
> nitpick: onValidateSelector has an event param which is not use so it should be possible to remove.

Will add this in with the next PR! Want to get this one in quick for the sake of potential merge conflicts. Thank you!

PR #907, opened 2019-07-03
chore: run e2e tests as part of PR build process on ubuntu

### LINKED ISSUE #898: Add parallel job to run E2E tests on Ubuntu build agent
Right now, our E2E tests only run on MacOS in our CI/PR builds. We'd like them to run in parallel on all of our supported OSes.

As part of this task, investigate limits on parallel build job count and consider making some of the jobs CI-only if necessary

Shobhit1:
#### Description of changes

run parallel PR/CI build jobs on windows & linux for build & E2E

- Runs parallel `build:all` job on `linux & windows`
- Runs `unit tests and lints` on `macOS`
- Runs `e2e tests` on `macOS` and `windows`

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #898 


smoralesd:
e2e_tests linux is "succeeding with issues" https://dev.azure.com/ms/accessibility-insights-web/_build/results?buildId=23815&view=logs&jobId=e7985f60-ceef-5226-255e-cc9034d7b4f4

Shobhit1:
> e2e_tests linux is "succeeding with issues" https://dev.azure.com/ms/accessibility-insights-web/_build/results?buildId=23815&view=logs&jobId=e7985f60-ceef-5226-255e-cc9034d7b4f4

Yes. I know. Its by design. That step will be deprecated in `linux` run!

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /app
+
+COPY package.json /app
+COPY yarn.lock /app
+
+RUN yarn install --frozen-lockfile
+
+COPY . /app
+
+# Start script on Xvfb
```
dbjorge: I don't think the comment is adding any information. Maybe better would be a comment explaining *why* xvfb is necessary?

Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# reference: https://stackoverflow.com/a/51683309/3711475
```
dbjorge: Probably better to reference the official docs, https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker

Shobhit1: Probably will add both, since it is a mixture from both places.

pipeline/e2e-test.yaml
```diff
@@ -5,5 +5,24 @@ steps:
     - script: yarn build:all
       displayName: build:all
 
+    - script: docker build -t app .
+      condition: eq( variables['Agent.OS'], 'Linux' )
```
dbjorge: Would it be more clear about why the environment checks are happening to make the "docker or not" decision part of the strategy matrix, rather than adding the environment check to each individual task? ie, in strategy,

```
      strategy:
          matrix:
              windows:
                  imageName: 'vs2017-win2016'
                  e2eEnvironment: 'agent'
              linux:
                  imageName: 'ubuntu-16.04'
                  e2eEnvironment: 'docker'
```

...and then in the tasks, `condition: eq( variables['e2eEnvrionment'], 'docker' )`/etc?

Shobhit1: I have removed the `strategy` and `environment checks` in favor of a separate `ubuntu` only job that uses a `docker` yaml file. So we dont need all of those checks anymore.

Dockerfile
Shobhit1: done

Dockerfile
Shobhit1: done.

Dockerfile
Shobhit1: done

build.yaml
```diff
@@ -73,6 +73,37 @@ jobs:
...
+    # e2e run on linux - using docker.
+    - job: 'e2e_tests_linux'
+
+      pool:
+          vmImage: 'ubuntu-16.04'
+
+      steps:
+          - template: pipeline/e2e-test-linux-docker.yaml
+
+    - job: 'build_electron_app'
```
dbjorge: The separate build_electron_app step was reverted in master since you started this PR, so I think for this PR's purposes it probably makes sense to omit the separate job for it

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# reference: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker
+FROM node:10
+
+RUN apt-get update && apt-get install -yq gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 libdbus-1-3 libexpat1 libfontconfig1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 libgtk-3-0 libnspr4 libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 ca-certificates fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils wget x11vnc x11-xkb-utils xfonts-100dpi xfonts-75dpi xfonts-scalable xfonts-cyrillic x11-apps xvfb
+
+WORKDIR /app
+
+COPY package.json /app
```
dbjorge: Minor optimization: consolidating the 2 COPY lines into one `COPY yarn.lock package.json /app` line will save on the production of one build image

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN yarn install --frozen-lockfile
+
+COPY . /app
+
+RUN yarn build:dev
+
+# since we need our chromium to run in 'headful' mode (for testing chrome extension)
+# we need a fake display (to run headful chromium), which we create by starting a Virtualized X server environment using xvfb-run
+# man page for command: https://manpages.ubuntu.com/manpages/xenial/man1/xvfb-run.1.html
+CMD xvfb-run --server-args="-screen 0 1024x768x24" yarn test:e2e
```
dbjorge: Consider using "ENTRYPOINT" instead of "CMD" here; if you do that, it will make additional arguments in the `docker run` command be appended to the command rather than replacing the command, so it will enable passing arguments to jest like `docker run -t e2e --testPathPattern first-time-dialog.test.ts`

pipeline/e2e-test-linux-docker.yaml
```diff
@@ -0,0 +1,32 @@
...
+# Licensed under the MIT License.
+
+steps:
+    - script: docker build -t app .
+      displayName: setup docker
+
+    - script: docker run --network=host -i app
+      displayName: run e2e tests on docker
+
+    - bash: |
```
dbjorge: I hit my timebox on the `-v` parameter, so let's keep this way for the time being

Dockerfile
Shobhit1: My `docker` extension on vscode doesn't like that syntax. Gives me `When using COPY with more than one source file, the destination must be a directory and end with a / or a \dockerfile-utils(23)
` error

Dockerfile
Shobhit1: Is it fine if I explore `ENTRYPOINT` when we come back on improving the perf for this?

Shobhit1:
> ould be user-impactful enough to warrant a callout in the release notes and a minor version bump for the extension; this is a feature from our dev team's perspective but from an end user's perspective

Updated.

Dockerfile
dbjorge: Yes, you should do what the error says (end the /app/ directory with a /)

Shobhit1: I think making the list comma-separated for `src files` works. Trying it now.

Shobhit1: Worked after adding `/` at the end of directory path.

PR #913, opened 2019-07-07
fix: a11y issue on target page link

smoralesd:
#### Description of changes

On the new automated checks report we have the target page link twice, one on the top level header of the report, where we use the target page name as the link text; and the second on the scan details section (right after the globe icon) where we use the url as the link text. 

This links are failing `assessment -> links -> link purpose` as they do not have the same **accessible name**.

**Edit:**
To fix this: 
- use `title` attribute on the anchor tag to match the one on the page header (basically the page title)
- update the hidden label to `"target page url"` to clarify what the value is (i.e. the url itself)

This two fixes are needed because we have to modes of operations here:
- **tab key navigation**: using the tab key, the user will navigate through interactive elements. The second tab from the top of the page will make the focus to land directly on the link which the AT will read as: <link text> <visited|not visited> link <link title>. Here **link text** is the url itself while **link title** comes from the title attribute
- **arrow key navigation**: using down (or up) arrow key, the user will navigate through screen reader visible content. On the **Scan details** section, this will move from the title to a list of 2 items (the target page url and the scan date being those items). Another down arrow key and the user will land on the first element which is compose of 2 things: a visually hidden label which the screen reader will announce: "page title url" and then (after another down arrow key) the link itself (without the title in this case). 

**End of the Edit**

#### Pull request checklist

- [x] Addresses an existing issue: Part of WI #1558756
   - no changes

### DIFF
--- a/src/DetailsView/reports/components/report-sections/details-section.tsx
+++ b/src/DetailsView/reports/components/report-sections/details-section.tsx
@@ -10,10 +10,10 @@ import { UrlIcon } from '../../../../common/icons/url-icon';
 import { NamedSFC } from '../../../../common/react/named-sfc';
 import { SectionProps } from './report-section-factory';
 
-export type DetailsSectionProps = Pick<SectionProps, 'pageUrl' | 'description' | 'scanDate' | 'toUtcString'>;
+export type DetailsSectionProps = Pick<SectionProps, 'pageTitle' | 'pageUrl' | 'description' | 'scanDate' | 'toUtcString'>;
 
 export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', props => {
-    const { pageUrl, description, scanDate, toUtcString } = props;
+    const { pageTitle, pageUrl, description, scanDate, toUtcString } = props;
 
     const createListItem = (icon: JSX.Element, label: string, content: string | JSX.Element, contentClassName?: string) => (
         <li>
@@ -32,7 +32,13 @@ export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', pr
         <div className="scan-details-section">
             <h2>Scan details</h2>
             <ul className="details-section-list">
-                {createListItem(<UrlIcon />, 'target page:', <NewTabLink href={pageUrl}>{pageUrl}</NewTabLink>)}
+                {createListItem(
+                    <UrlIcon />,
+                    'target page:',
+                    <NewTabLink href={pageUrl} aria-label={pageTitle}>
+                        {pageUrl}
+                    </NewTabLink>,
+                )}
                 {createListItem(<DateIcon />, 'scan date:', scanDateUTC)}
                 {showCommentRow && createListItem(<CommentIcon />, 'comment:', description, 'description-text')}
             </ul>
--- a/src/DetailsView/reports/components/report-sections/details-section.tsx
+++ b/src/DetailsView/reports/components/report-sections/details-section.tsx
@@ -34,8 +34,8 @@ export const DetailsSection = NamedSFC<DetailsSectionProps>('DetailsSection', pr
             <ul className="details-section-list">
                 {createListItem(
                     <UrlIcon />,
-                    'target page:',
-                    <NewTabLink href={pageUrl} aria-label={pageTitle}>
+                    'target page url:',
+                    <NewTabLink href={pageUrl} title={pageTitle}>
                         {pageUrl}
                     </NewTabLink>,
                 )}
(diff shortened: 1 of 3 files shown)


dbjorge:
(removing auto-merge while I type up fuller response)

dbjorge:
I'm hesitant about updating the aria-label to cover this case; I'm not sure this is in keeping with the spirit of [the corresponding WCAG 3.2.4 requirement](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification), which is not strictly about screen reader users but also describes itself as being in the interest of lessening cognitive load for people with cognitive disabilities.

I think we need to decide whether or not the 2 links in question truly represent one "functional component that appears repeatedly within a set of Web pages". If we do, then I think we should apply the WCAG guidance to the visual text too, not just the aria-label, and if we don't, I'd expect we should probably keep the accessible label similar to the distinct visual text.

@ferBonnin , thoughts? Am I misinterpreting the WCAG guidance?

smoralesd:
> I'm hesitant about updating the aria-label to cover this case; I'm not sure this is in keeping with the spirit of [the corresponding WCAG 3.2.4 requirement](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification), which is not strictly about screen reader users but also describes itself as being in the interest of lessening cognitive load for people with cognitive disabilities.
> 
> I think we need to decide whether or not the 2 links in question truly represent one "functional component that appears repeatedly within a set of Web pages". If we do, then I think we should apply the WCAG guidance to the visual text too, not just the aria-label, and if we don't, I'd expect we should probably keep the accessible label similar to the distinct visual text.
> 
> @ferBonnin , thoughts? Am I misinterpreting the WCAG guidance?

I think I can see your point but let me rephrase to make sure. What I get is this: we have 2 links that have the same url but since they have 2 different texts, we may consider them different functional components. Is that right?

@ferBonnin was concern about **link purpose** requirement of the **links** test (related: [WCAG 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)).

My first though was "it would be nice to not have the url as the link text here", as listening the screen reader going h-t-t-p-:-/-/ seems unnecessary and annoying really) but then there is the issue of having the same link (text and href) duplicated. I guess the second issue (duplicated link) is more of a design issue than usability/accessibility issue so I'll leave this to Fer.

dbjorge:
> I think I can see your point but let me rephrase to make sure. What I get is this: we have 2 links that have the same url but since they have 2 different texts, we may consider them different functional components. Is that right?

Yes, essentially. I think it's probably a little more broad than just using different text; I think looking at the 2 components in context, I would say it's reasonable to conclude that they are different functional components because they are visually distinct, one occurs in a different place/format than all the other (consistent) uses of that Target Page banner throughout the rest of the UI, and they have a slightly different purpose (I would say that the primary use of the Target Page banner is typically navigational, where I think the scan details link is primarily about documenting context of the scan). I think it's a subjective distinction that reasonable people could disagree about, though.

> ferBonnin was concern about **link purpose** requirement of the **links** test (related: [WCAG 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)).

Yes, apologies, I edited in my reasoning for considering 3.2.4 vs 2.4.4 after you'd written this, I should have included it in the original :)

ferBonnin:
> > I'm hesitant about updating the aria-label to cover this case; I'm not sure this is in keeping with the spirit of [the corresponding WCAG 3.2.4 requirement](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification), which is not strictly about screen reader users but also describes itself as being in the interest of lessening cognitive load for people with cognitive disabilities.
> > I think we need to decide whether or not the 2 links in question truly represent one "functional component that appears repeatedly within a set of Web pages". If we do, then I think we should apply the WCAG guidance to the visual text too, not just the aria-label, and if we don't, I'd expect we should probably keep the accessible label similar to the distinct visual text.
> > @ferBonnin , thoughts? Am I misinterpreting the WCAG guidance?
> 
> I think I can see your point but let me rephrase to make sure. What I get is this: we have 2 links that have the same url but since they have 2 different texts, we may consider them different functional components. Is that right?
> 
> @ferBonnin was concern about **link purpose** requirement of the **links** test (related: [WCAG 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)).
> 
> My first though was "it would be nice to not have the url as the link text here", as listening the screen reader going h-t-t-p-:-/-/ seems unnecessary and annoying really) but then there is the issue of having the same link (text and href) duplicated. I guess the second issue (duplicated link) is more of a design issue than usability/accessibility issue so I'll leave this to Fer.

There are two things here:
1) we need to have the link purpose described by either the link or an accessible name/description. (WCAG 2.4.4)
2) Links with different destinations should have different descriptions; links with the same destination should have the same description.
Per the key terms in 3.2.4: they are not different functional components, since you get the same results when you use them.

smoralesd:
@dbjorge @ferBonnin 
Update the PR to use title attribute on the link and update the visually hidden label from "target page" to "target page url". Please note the changes are handling 2 different modes of operation: tab navigation vs arrow key navigation (more details on the PR description)

Issue #895, opened 2019-07-03
Test case: verify automated checks results from major target pages


dbjorge:
Verify that automated checks can successfully scan microsoft.com and bing.com.

Shobhit1:
The basic idea is to verify that our automated checks is not broken when we run on complex websites. The website can be anything like cnn etc.

dbjorge:
Suborder here will be:
1. Implement the meat of the test case against a checked in test page
2. Implement `it.skip` versions against key external properties
3. Go do #896 and incorporate the `it.skip`'d cases there

dbjorge:
Cutting subparts 2 and 3 due to feature timebox

PR #925, opened 2019-07-09
test: add e2e test case pinning target page visualizations for adhoc tests

dbjorge:
#### Description of changes

Adds snapshot tests that pin the contents of our target page visualizer shadow root when enabling the different ad hoc tools.

There are several related changes that we'll be including later in separate PRs, but which I'm leaving out of scope of this PR:
* tab stops (requires extra work on target page)
* updating target page to contain anything impacted by landmark/color visualizers
* adding test case logic to do fuzzy testing of highlight box positions

As part of this, I needed to be able to format a child of an element for snapshotting, rather than a page element, which didn't work with the original factoring of the formatting function as a Page method, so this also refactors the different "formatForSnapshot" functions as separate free functions.

I also tried out using a `TargetPageController` to see what that would look like, per conversation with @smoralesd. I think this PR makes me like it enough to want to keep it, but switching other page automation tasks to use the same strategy is out of scope for this PR.

#### Pull request checklist

- [x] Addresses an existing issue: Contributes to #895


dbjorge:
The new test cases are failing on Windows due to sub-pixel differences between platforms in where the highlight boxes are absolutely positioned on the page. Will consider how to address this.

src/tests/end-to-end/common/element-snapshot-formatter.ts
```diff
@@ -0,0 +1,55 @@
...
+    });
+}
+
+// in some cases (eg, stylesheet links), HTML can contain absolute chrome-extension://{generated-id} paths
+// which differ between builds of the extension. This normalizes those.
+function normalizeExtensionUrls(htmlString: string): string {
+    return htmlString.replace(/chrome-extension:\/\/\w+\//g, '{{EXTENSION_URL_BASE}}/');
+}
+
+// Certain injected target page elements are rendered at absolute positions on the page based on elements from
```
Shobhit1: Love this.

src/tests/end-to-end/tests/popup/adhoc-panel.test.ts
```diff
@@ -65,6 +67,34 @@ describe('Ad hoc tools', () => {
...
+
+    async function enableToggleByAriaLabel(ariaLabel: string): Promise<void> {
+        const toggleSelector = `button[aria-label="${ariaLabel}"]`;
+        const enabledToggleSelector = `${toggleSelector}[aria-checked=true]`;
+        const disabledToggleSelector = `${toggleSelector}[aria-checked=false]`;
+
+        await popupPage.clickSelector(disabledToggleSelector);
+
+        // The toggles will go through a state where they are removed and replaced with a spinner, then re-added to the page
+        // We intentionally omit looking for the loading spinner because it can be fast enough to not be seen by Puppeteer
```
Shobhit1: Is this the reason for the inconsistency in the `tab-stop` test as well? I added an alternative selector and wait for it to avoid this in my PR.

src/tests/end-to-end/tests/popup/adhoc-panel.test.ts
```diff
@@ -65,6 +67,34 @@ describe('Ad hoc tools', () => {
...
     });
 
+    it.each(['Automated checks', 'Landmarks', 'Headings', 'Color'])(
+        'should display the pinned target page visualizations when enabling the "%s" toggle',
+        async (toggleAriaLabel: string) => {
+            await gotoAdhocPanel();
+
+            await enableToggleByAriaLabel(toggleAriaLabel);
+
+            expect(await targetPage.getShadowRootHtmlSnapshot()).toMatchSnapshot();
```
Shobhit1: Once this PR is merged, I can use this piece of code to check tabstop visualizations for my test as well. Thanks for doing this. 👍

src/tests/end-to-end/tests/popup/adhoc-panel.test.ts
dbjorge: Possibly, but I suspect not; my bet would be that for your issue (and for the flakiness in the high contrast toggle click() working), it's probably going to be that we need to do a better job of making sure the button is actionable before clicking on it. [Cypress has a good overview of the prerequisites it checks for before attempting a click event](https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Actionability); it's probably a good guide for us to follow.

Issue #923, opened 2019-07-09
E2E test fails when font fails to load


lisli1:
## Describe the bug

When writing/running E2E tests, there is a `requestfailed` error from attempting to load a font, causing a `forceTestFailure` call in `page.ts`. (See sample error message below.) This causes the whole test suite to fail.

```
Test suite failed to run

    expect(received).toBeNull()

    Received: "request failed - net::ERR_ABORTED, https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-semilight.woff2"

      2 | // Licensed under the MIT License.
      3 | export function forceTestFailure(errorMessage: string): void {
    > 4 |     expect(errorMessage).toBeNull();
        |                          ^
      5 | }
      6 |
```

A similar error is seen in the browser console when running an Assessment: 
```
Refused to run the JavaScript URL because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval' https://az416426.vo.msecnd.net". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution.
```

**To Reproduce**

For first error message described (test suite failed to run):
1. Go to https://dev.azure.com/ms/accessibility-insights-web/_build/results?buildId=24268
2. Navigate to "e2e_tests_windows" > "run e2e tests"
3. See errors

For second error message described (in browser console):
1. Go to www.bing.com
2. Open extension and run Assessment
3. Inspect the resulting details view page
4. Open the browser console
5. See error message

## Expected behavior

Test suite should not fail when there is a font loading failure.

## Extension (please complete the following information)

- OS: Windows
- Version 10.0.18362
- Browser Version Chrome 75.0.3770.100
- Target Page www.bing.com

## Are you willing to submit a PR?

 Yes

## Did you search for similar existing issues?

Yes

## Additional context

**Note: Temporary fix in #927 where the error handler doesn't do anything if the URL contains 'fonts' or 'icons'. Must update/replace error handler when resolving this issue.**

lamaks:
Please provide the URI to repro the issue

ferBonnin:
Reviewed with Peter, closing since this is old and there is a workaround

PR #927, opened 2019-07-10
chore: add checks for fonts and icons in request failed error handler from e2e base page.ts

lisli1:
#### Description of changes

- Added checks for 'fonts' and 'icons' in error handler in `page.ts` that was causing e2e test suites to fail when fonts failed to load

#### Pull request checklist

- [x] Addresses an existing issue: Temporary workaround for #923
- [x] (UI changes only) Added screenshots/GIFs to description above **N/A**
- [x] (UI changes only) Verified usability with NVDA/JAWS **N/A**


src/tests/end-to-end/common/page.ts
```diff
@@ -18,9 +18,10 @@ export class Page {
         underlyingPage.on('pageerror', error => {
             forceEventFailure(`'pageerror' (console.error) with stack: ${error.stack}`);
         });
-        underlyingPage.on('requestfailed', request => {
-            forceEventFailure(`'requestfailed' from '${request.url()}' with errorText: ${request.failure().errorText}`);
-        });
+        // Removing this error handler as a temporary workaround for #923
+        // underlyingPage.on('requestfailed', request => {
```
smoralesd: In general, we do not put comment code on master! 
If anything, I guess you should remove this altogether as git history will preserve this code anyway.

src/tests/end-to-end/common/page.ts
```diff
@@ -18,9 +18,10 @@ export class Page {
...
+        //     forceTestFailure(`request failed - ${request.failure().errorText}, ${request.url()}`);
```
peterdur: Consider checking the URL (maybe for prefix `https://static2.sharepointonline.com`, or suffix `.woff2`) and not calling forceTestFailure if it matches.

lisli1: Updated to check for `fonts` and `icons` in the URL

src/tests/end-to-end/common/page.ts
```diff
@@ -18,8 +19,12 @@ export class Page {
         underlyingPage.on('pageerror', error => {
             forceEventFailure(`'pageerror' (console.error) with stack: ${error.stack}`);
         });
+        // Modifying this error handler as a workaround for #923
```
Shobhit1: nit: I think it makes sense to not have this comment with issue reference here.

dbjorge: I'm not sure if you mean "you think it belongs next to the if statement instead" or "you don't think it's necessary at all". I agree with the former and strongly disagree with the latter.

dbjorge:
@lisli1 Can you use the `test:` or `chore:` prefix for the PR/commit? `fix:` is for fixes to user-facing bugs (something that would motivate a new release, which we would want to include in release notes/release validation)

PR #928, opened 2019-07-10
adding snippet and path to data store and the exported asssessment

annikakouhia:
#### Description of changes

Puts in the backend logic to add path/snippet to the Assessment Store when user clicks "Add Failure Instance" or updates a Failed Instance ("Edit Failure Instance" pathway). Also ensures exported Assessment Report includes not only description, but path and snippet below each documented Failure Instance.

Finally collects Failure Instance comment/path/snippet into its own data structure to minimize number of props passed into Failure Instance Panel Control. Uses this data structure in the onAddFailureInstance and onEditFailureInstance paths to simplify props there as wel.

Currently, if a failure instance doesn't have a field (for instance, only having a comment but no path or snippet), the field names will not show up in the report (see screenshot). This choice hasn't yet been confirmed by Chelsea, but it means we can merge this PR to master without hiding any logic beneath the Feature Flag (since users cannot add a path or snippet unless enabling the feature flag, so will always have an empty path/snippet).

Manually tested and tested with test files.

PRexample

#### Pull request checklist


src/background/assessment-data-converter.ts
```diff
@@ -183,11 +183,20 @@ export class AssessmentDataConverter {
         return null;
     }
 
-    public generateFailureInstance(description: string): UserCapturedInstance {
-        return {
+    public generateFailureInstance(description: string, path: string, snippet: string): UserCapturedInstance {
+        const instance = {
```
Shobhit1: nit: we can avoid all the `if` statements just using js shortcuts:

```js
        const instance = {
            id: this.generateUID(),
            description: description !== '' && description,
            selector: path !== '' && path,
            html: snippet !== '' && snippet,
        };
```
For me, it is much more readable but totally upto you for this.

annikakouhia: Oh awesome I was wondering if there was a more concise way to do this. Thanks!

Shobhit1: Never mind. Wrong approach.

Shobhit1: This works, but upto you if you want to use spread operator or not here.
```js
function generateObject(path, snippet, description) {
    const instance = {
        ...(description !== '') ? { description } : {},
        ...(path !== '') ? { selector: path } : {},
        ...(snippet !== '') ? { html : snippet } : {},
        id: 1,
    };
    console.log(instance);
}

generateObject('', '', ''); // { id: 1 }

generateObject('1', '2', '3'); // { description: '3', selector: '1', html: '2', id: 1 }
```

src/tests/unit/tests/DetailsView/components/failure-instance-panel-control.test.tsx
```diff
@@ -94,7 +94,7 @@ describe('FailureInstancePanelControlTest', () => {
         props.instanceId = '1';
         props.editFailureInstance = editInstanceMock.object;
 
-        editInstanceMock.setup(handler => handler(description, props.test, props.step, props.instanceId)).verifiable(Times.once());
+        editInstanceMock.setup(handler => handler(description, '', '', props.test, props.step, props.instanceId)).verifiable(Times.once());
```
Shobhit1: Do you want to test by passing the value of `path` & `snippet` through `props`; so that if we move the `props` generation out into a separate function; this wont break this code?

src/background/assessment-data-converter.ts
annikakouhia: added it in and things seem to be working! Much more concise

src/tests/unit/tests/DetailsView/components/failure-instance-panel-control.test.tsx
annikakouhia: Discussed offline and since part of state instead of props in the Failure Instance Panel Control, ok to leave as hardcoded (though will abstract to a const)

src/DetailsView/actions/details-view-action-message-creator.ts
```diff
@@ -365,13 +367,22 @@ export class DetailsViewActionMessageCreator extends DevToolActionMessageCreator
         this.dispatcher.sendTelemetry(TelemetryEvents.DETAILS_VIEW_OPEN, telemetryData);
     }
 
-    public editFailureInstance = (description: string, test: VisualizationType, requirement: string, id: string): void => {
+    public editFailureInstance = (
+        description: string,
```
waabid: Does it instead seem better to pass an object which models what the failure instance contains rather than individual properties?

src/DetailsView/components/assessment-instance-edit-and-remove-control.tsx
```diff
@@ -13,9 +13,11 @@ export interface AssessmentInstanceEditAndRemoveControlProps {
     test: VisualizationType;
     step: string;
     id: string;
-    description: string;
+    description?: string;
```
waabid: Similarly, this component could instead consume that model instead of these multiple properties.

src/DetailsView/actions/details-view-action-message-creator.ts
annikakouhia: done! little tricky figuring out logic how it affected failure instance panel control under different circumstances but I think it works

src/DetailsView/components/assessment-instance-edit-and-remove-control.tsx
annikakouhia: propagated up

src/DetailsView/components/assessment-instance-edit-and-remove-control.tsx
```diff
@@ -12,10 +12,14 @@ import { CapturedInstanceActionType, FailureInstancePanelControl } from './failu
 export interface AssessmentInstanceEditAndRemoveControlProps {
     test: VisualizationType;
     step: string;
-    id: string;
-    description: string;
+    currentInstance: {
```
waabid: You can extract this out to a type of it's own and use that type everywhere instead.

src/DetailsView/actions/details-view-action-message-creator.ts
annikakouhia: Simplified a bit more--longer description on below comment. Thanks!

src/DetailsView/components/assessment-instance-edit-and-remove-control.tsx
annikakouhia: Abstracted to its own type (Failure Instance Data) and now propagates that down in props to create the Failure Instance Panel Control! In a separate PR I will simplify the onAddFailureInstance and onEditFailureInstance pathways to use this data structure as well--just wanted to keep this PR as simple as possible for the time being. Thanks!

annikakouhia: Have extracted out to type and propagated up so this type is used in the creation of the panel itself AND propagated down so its used in the store changes prompted by AddFailureInstance and EditFailureInstance

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -45,11 +49,24 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
 
     constructor(props) {
         super(props);
+        let filledOriginalInstance;
```
waabid: I think "filledOriginalInstance" might not be such a great name for this; it might not be an instance handed to it.

I think current instance is fine.

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -45,11 +49,24 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+        } else {
+            filledOriginalInstance = { failureDescription: '', path: '', snippet: '' };
+        }
+
         this.state = {
             isPanelOpen: false,
-            failureDescription: this.props.originalText || '',
-            path: '',
-            snippet: '',
+            currentInstance: {
```
waabid: This doesn't seem necessary; we can just use the instance you have from line 52.

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -133,39 +150,73 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+            path: this.state.currentInstance.path,
+            snippet: this.state.currentInstance.snippet,
+        };
+
+        return updatedInstance;
+    };
+
     protected onFailureDescriptionChange = (event, value: string): void => {
-        this.setState({ failureDescription: value });
+        const updatedInstance = this.getCurrentInstance();
```
waabid: Why not just use the "this.state.currentInstance" here? Alternatively, if we can't use the state object directly, we can simply use a "clone" utility function here instead.

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -133,39 +150,73 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+        updatedInstance.path = value;
+        this.setState({ currentInstance: updatedInstance });
     };
 
     private onValidateSelector = (event): void => {
-        const currSelector = this.state.path;
-        this.setState({ snippet: 'snippet for ' + currSelector });
+        const currSelector = this.state.currentInstance.path;
+        const currSnippet = 'snippet for ' + currSelector;
```
waabid: I know this is from an earlier implementation and not this PR but we shouldn't be adding UI elements (like 'snippet for') in strings like this. Something we might want to change in a future PR.

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -133,39 +150,73 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
 
     protected onSaveEditedFailureInstance = (): void => {
-        this.props.editFailureInstance(this.state.failureDescription, this.props.test, this.props.step, this.props.instanceId);
+        this.props.editFailureInstance(this.state.currentInstance, this.props.test, this.props.step, this.props.instanceId);
         this.setState({ isPanelOpen: false });
     };
 
     protected openFailureInstancePanel = (): void => {
-        this.setState({ isPanelOpen: true, failureDescription: this.props.originalText || '' });
+        let propsDescription = '';
```
waabid: This function seems a bit convoluted to me.

You are currently creating properties and then structuring them to be an object. You can actually just create a default object and then just combine the instance from the props.

```
const defaultInstance = getDefaultInstance();
const currentInstance = {
    ...defaultInstance,
    ...this.props.originalInstance,
}
this.setState({
            isPanelOpen: true,
            currentInstance: currentInstance,
});
```

**Some of this** could also simplify your constructor.

src/background/assessment-data-converter.ts
```diff
@@ -183,11 +183,14 @@ export class AssessmentDataConverter {
...
             id: this.generateUID(),
-            description: description,
-        };
+            ...(description !== '' && { description }),
+            ...(path !== '' && { selector: path }),
+            ...(snippet !== '' && { html: snippet }),
+        } as UserCapturedInstance;
```
waabid: What's the purpose of doing the empty string checks here? Just curious as I'm not sure I'm following. 😄

src/DetailsView/components/failure-instance-panel-control.tsx
annikakouhia: I was thinking "filledOriginalInstance" because the one handed to it may be null, in which case we fill it with empty (but valid) parameters. I get what you're saying though--I'll change it! Thanks!

src/DetailsView/components/failure-instance-panel-control.tsx
annikakouhia: I did it this way because currently we could have had a valid Original Instance which had a comment, but didn't have a path or snippet. If we don't check the validity of each field I've run into errors where a field is null instead of ''. Do you think it's better to do this check during the creation of currentInstance above? Just let me know!

src/DetailsView/components/failure-instance-panel-control.tsx
annikakouhia: I put it in currently just to make sure it updated based on the selector when i clicked validate--is it okay to leave for testing purposes until the end?

src/background/assessment-data-converter.ts
annikakouhia: Currently, if a failure instance is added without a path or snippet, checking for the empty string ensure we don't add {path: ''} {snippet: ''} to our failure instance, and instead only add {comment: 'the comment'}.
This was to ensure that when users export reports, they dont see:
Comment            the comment
Path
Snippet
and instead just see the comment line. There are definitely other places in the code to put this check if you think it would be clearer elsewhere! Just wanted to make sure no work slipped out from beneath the feature flag before we were ready to release :)

src/DetailsView/components/failure-instance-panel-control.tsx
annikakouhia: Great! This definitely cleaned things up--worked in the constructor as well!

src/DetailsView/components/failure-instance-panel-control.tsx
waabid: Since it's already in master, I'm fine with moving forward but it's something we should come back to sooner than later? 😄

annikakouhia: Sounds good! I'll try to get this cleaned up in the next PR so we don't leave it floating

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -19,20 +20,24 @@ import { GenericPanel, GenericPanelProps } from './generic-panel';
 export interface FailureInstancePanelControlProps {
     step: string;
     test: VisualizationType;
-    addFailureInstance?: (description, test, step) => void;
-    editFailureInstance?: (description, test, step, id) => void;
+    addFailureInstance?: (instanceData, test, step) => void;
+    editFailureInstance?: (instanceData, test, step, id) => void;
     actionType: CapturedInstanceActionType;
+    originalInstance?: FailureInstanceData;
```
waabid: I'm not sure if `originalInstance` is a good name here. Original implies there some context to how this property must be used for this control but I don't think that's the case.

A better name would be 'failureInstance' or even 'instance' since that implies no linking with any sort of context of how this control is used.

annikakouhia: I was thinking this parameter is only filled if we open the panel with the intention of "editing" so was trying to make it clear that the passed in instance was the one we were going to be editing. Do you think it's best to just assume it's clear when we use it to fill the "current instance" field?

annikakouhia: Nvm--discussed offline and changing to preserve the ambiguity of the panel!

PR #941, opened 2019-07-11
Adding new PathSnippet Store

annikakouhia:
#### Description of changes

Adding a new store ("PathSnippetStore") which holds a path and snippet. To be used for when users give a path during Failure Instance Creation and we need to retrieve the corresponding snippet from the Target Page.

Added manual testing functionality by making the "Validate Selector" button on the Failure Instance Panel (under the Feature Flag) send the given selector to the Path Snippet Store.

Testing mainly carried out in the Path Snippet Store test file, Action Creator test file, and Details View Action Message Creator test file.

Named PathSnippetStore instead of FailureInstanceStore in the hopes that it could be reused more generally at a later point in time. Also believed FailureInstanceStore would be a bit misleading as most of our Failure Instance data goes straight to the Assessment Store without interacting at all with this new store.

#### Pull request checklist


src/background/stores/path-snippet-store.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { StoreNames } from '../../common/stores/store-names';
+import { PathSnippetData } from '../../common/types/store-data/path-snippet-data';
+import { PathSnippetActions } from '../actions/path-snippet-actions';
+import { BaseStoreImpl } from './base-store-impl';
+
+export class PathSnippetStore extends BaseStoreImpl<PathSnippetData> {
+    private pathSnippetActions: PathSnippetActions;
```
smoralesd: You can declutter this class a little bit by using [parameter properties](https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties). It would look like this
```typescript
...
// no need to declare the property here if using parameter properties on the constructor
// private pathSnippetActions: PathSnippetActions;

 constructor(private readonly pathSnippetActions: PathSnippetActions) {
        super(StoreNames.PathSnippetStore);

         // no need for this assignment anymore neither
         // this.pathSnippetActions = pathSnippetActions;
    }
```

Not all our classes follow this pattern because it was not introduce to our code base from the beginning but it's now the standard strategy to declare class properties.
I'm also adding the `readonly` modifier as the actions should not be change by this class.

src/background/stores/path-snippet-store.ts
```diff
@@ -0,0 +1,40 @@
...
+        this.pathSnippetActions.onAddPath.addListener(this.onAddPath);
+        this.pathSnippetActions.onAddSnippet.addListener(this.onAddSnippet);
+    }
+
+    private onAddPath = (payload: string): void => {
+        this.state.path = payload;
+        this.emitChanged();
+    };
+
+    private onAddSnippet = (payload: string): void => {
```
smoralesd: Having this 2 functions (`onAddPath` and `onAddSnippet`) is fine here but consider to have a more generic version of them. Something around this:
```typescript
    protected addActionListeners(): void {
        this.pathSnippetActions.onAddPath.addListener(payload => this.onChangeProperty('path', payload));
        this.pathSnippetActions.onAddSnippet.addListener(payload => this.onChangeProperty('snippet', payload));
    }

    private onChangeProperty = (property: keyof PathSnippetData, payload: string) => {
        this.state[property] = payload;
    }
```

This is only possible because there is not complex, different logic between updating the path and the snippet. It's less code and, if for any reason, we need to add a new string property (it should even work with other types) this will minimize future changes to the class

src/background/actions/action-creator.ts
```diff
@@ -113,6 +116,8 @@ export class ActionCreator {
         this.registerTypeToPayloadCallback(Messages.Assessment.EnableVisualHelperWithoutScan, this.onEnableVisualHelperWithoutScan);
 
         this.registerTypeToPayloadCallback(Messages.Inspect.SetHoveredOverSelector, this.onSetHoveredOverSelector);
+
+        this.registerTypeToPayloadCallback(Messages.PathSnippet.AddPathForValidation, this.onAddPathForValidation);
```
smoralesd: It seems that this new logic is not related to anything the else the `ActionCreator` currently holds. Do you mind creating a new action creator to separate the concerns here? In general, is good to follow Single Responsibility Principle, and one of the many ways to check for that is looking for subsets of functions that use disjunct subsets of class properties (like `pathSnippetActions` here). If you are able to find this subsets (of functions and properties) that points to the fact the class have more than one responsibility, as in this case.
I know that this class is already breaking the principle but it has been refactored on the pass to reduce this issue and it would be nice to not add to the tech debt here.

src/common/types/store-data/path-snippet-data.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export interface PathSnippetData {
```
smoralesd: It looks like the convention for this data types is to use the suffix `StoreData` so this should be renamed to `PathSnippetStoreData`

src/background/stores/path-snippet-store.ts
annikakouhia: Great thank you!! Will implement!

src/background/stores/path-snippet-store.ts
annikakouhia: Ah got it. Thanks!

src/background/actions/action-creator.ts
annikakouhia: Yes! That makes sense. I based the PathSnippet class off of the Inspect class because I figured they touched similar parts of the codebase, but it makes sense that we don't want to clutter the ActionCreator too much. I'll try abstracting out to its own ActionCreator. Thank you!

src/common/types/store-data/path-snippet-data.ts
annikakouhia: Got it, renaming!

src/tests/unit/tests/background/actions/action-creator.test.ts
```diff
@@ -16,6 +16,7 @@ import {
 import { DetailsViewActions } from '../../../../../background/actions/details-view-actions';
 import { DevToolActions } from '../../../../../background/actions/dev-tools-actions';
 import { InspectActions } from '../../../../../background/actions/inspect-actions';
+import { PathSnippetActions } from '../../../../../background/actions/path-snippet-actions';
```
smoralesd: Since you created a new action creator specifically for the path snippet work, it seems we don't need this changes on this test class now.

src/background/actions/path-snippet-action-creator.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { RegisterTypeToPayloadCallback } from '../../common/message';
+import { Messages } from '../../common/messages';
+import { PathSnippetActions } from './path-snippet-actions';
+
+export class PathSnippetActionCreator {
+    private pathSnippetActions: PathSnippetActions;
```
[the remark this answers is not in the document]
annikakouhia: Oh yes! That makes sense. Changing it now! Thanks :)

src/tests/unit/tests/background/actions/action-creator.test.ts
annikakouhia: Gotcha, yes! All the testing should be in the separate path-snippet-action-creator test. In the PR I'm about to push the PathSnippetActions are only referenced in this file when set to null in the creation of the ActionHub object. Let me know what you think! Thanks!

PR #948, opened 2019-07-12
fix: update links styles, match style on the report

smoralesd:
#### Description of changes

- Updating default links style to:
   - have no underline normally
   - have underline when hover
- Remove link style override on the automated checks report (so we pick up the default links style
- Update font color and weight for the header target page text on the automated checks report

**_Details view_**
**Target page url**
01 - target page normal and hover

**Rule and WCAG links**
02 - rule and wcag links normal and hover

**_Automated checks report_**
**Header target page url**
03 - report target page normal and hover

**Scan details page url**
04 - scan details page url normal and hover

**Rule and WCAG links**
05 - rule resources normal and hover

#### Pull request checklist

   - found during insider validation
   - no code changes
   - no code changes
   - not affected

### DIFF
--- a/src/DetailsView/reports/components/report-sections/header-section.scss
+++ b/src/DetailsView/reports/components/report-sections/header-section.scss
@@ -29,10 +29,11 @@
         margin: 0px 0px 0px 16px;
         display: inherit;
 
-        color: $neutral-60;
+        color: $primary-text;
         font-family: $fontFamily;
         font-size: 14px;
         line-height: 20px;
+        font-weight: 600;
 
         a {
             @include ellipsedText();
--- a/src/DetailsView/reports/components/report-sections/header-section.scss
+++ b/src/DetailsView/reports/components/report-sections/header-section.scss
@@ -33,7 +33,7 @@
         font-family: $fontFamily;
         font-size: 14px;
         line-height: 20px;
-        font-weight: 600;
+        font-weight: normal;
 
         a {
             @include ellipsedText();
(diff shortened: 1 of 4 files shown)


lisli1:
Do we want the automated checks report styles to look like the details view page styles? The links on details view page don't underline on hover (also in prod)

smoralesd:
@lisli1 
This change properties on `insights-link` css class which is being used on `NewTabLink` component. This component is the main way we are using to have links on the extension.
Thus, this change affects the details view, the popup, the guidance content and the automated checks report. Basically, anywhere we are using `NewTabLink` and not overriding the properties from `insights-link`.

src/DetailsView/reports/components/report-sections/header-section.scss
```diff
@@ -29,10 +29,11 @@
         margin: 0px 0px 0px 16px;
         display: inherit;
 
-        color: $neutral-60;
+        color: $primary-text;
```
lisli1: It looks like the color used for the header text in details view is `neutral-100` (inherited from `body` in `detailsview.scss`) but I don't know if that matters.

smoralesd: Good catch. I'll have a closer look, make the color on both places the same.

smoralesd: checked with @cheade and we want `$primary-text` on both places. Also, we want to use `font-weight: normal;` on both place.

Issue #867, opened 2019-06-27
Flaky E2E test


DaveTryon:
## Describe the bug

We have a flaky E2E test that shows up about once a week

**To Reproduce**

1. Create a PR
2. Look at the test output

Sample output (from buildId 22334):
```
##[section]Starting: run e2e tests
==============================================================================
Task         : Command line
Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
Version      : 2.151.1
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
==============================================================================
Generating script.
Script contents:
yarn test:e2e -- --ci
========================== Starting Command Output ===========================
[command]/bin/bash --noprofile --norc /Users/vsts/agent/2.153.2/work/_temp/59db13e1-f3e1-4a58-ad0f-a81a9b872d47.sh
yarn run v1.16.0
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
$ jest --projects src/tests/end-to-end --runInBand --ci
FAIL e2e tests src/tests/end-to-end/tests/common/settings-drop-down.test.ts (52.989s)
  ● Console

    console.log src/tests/end-to-end/common/generate-screenshot.ts:17
      Screenshot file is located at: /Users/vsts/agent/2.153.2/work/1/s/test-results/e2e/failure-screenshots/245b4dc3-bccd-4df7-967c-baf53c4e9e10.png

  ● Settings Dropdown › High contrast mode › should pass accessibility validation in popup dropdown

    TimeoutError: Navigation Timeout Exceeded: 5000ms exceeded

      27 | 
      28 |     public async goto(url: string): Promise<void> {
    > 29 |         await this.screenshotOnError(async () => await this.underlyingPage.goto(url, { timeout: DEFAULT_NEW_PAGE_WAIT_TIMEOUT_MS }));
         |                                                                            ^
      30 |     }
      31 | 
      32 |     public async close(ignoreIfAlreadyClosed: boolean = false): Promise<void> {

      at Promise.then (node_modules/puppeteer/lib/LifecycleWatcher.js:142:21)
        -- ASYNC --
      at Frame.<anonymous> (node_modules/puppeteer/lib/helper.js:111:15)
      at Page.goto (node_modules/puppeteer/lib/Page.js:629:49)
[...]

PR #955, opened 2019-07-16
test: use latest windows hosted agent version in e2e-reliability runs

dbjorge:
#### Description of changes

We tried updating to the latest agent version to see if it had any impact on reducing page crash rates compared to the older agents. It didn't impact page crash rate, but it improved performance 5-10%, so we'll take it anyway as we continue to test other options for fixing the page crashes.

#### Pull request checklist

- [x] Addresses an existing issue: Attempt to address 1 bucket of #867

### DIFF
--- a/pipeline/e2e-reliability.yaml
+++ b/pipeline/e2e-reliability.yaml
@@ -11,7 +11,7 @@ pr:
 jobs:
     - job: 'e2e_tests_windows_1'
       pool:
-          vmImage: 'vs2017-win2016'
+          vmImage: 'windows-2019'
       steps:
           - template: ./install-node-prerequisites.yaml
           - template: ./e2e-test-from-agent.yaml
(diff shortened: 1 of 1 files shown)


dbjorge:
Doesn't fix the windows reliability issues, but it also doesn't regress them and it improves performance slightly, so I'd propose we merge it anyway as we continue investigating flakiness issues.

pipeline/e2e-reliability.yaml
```diff
@@ -11,39 +11,39 @@ pr:
 jobs:
     - job: 'e2e_tests_windows_1'
       pool:
-          vmImage: 'vs2017-win2016'
+          vmImage: 'windows-2019'
```
smoralesd: sounds like we could use a variable for this, right?

dbjorge: We could use a template for the whole block, but going to leave it out of scope til a separate PR :)

PR #956, opened 2019-07-17
chore: improve screenshot method for better debugging

Shobhit1:
#### Description of changes

Adds better logs for screenshot method and checks if the screenshot is generated before printing the screenshot generated message.

#### Pull request checklist

### DIFF
--- a/src/tests/end-to-end/common/generate-screenshot.ts
+++ b/src/tests/end-to-end/common/generate-screenshot.ts
@@ -14,9 +14,13 @@ export async function takeScreenshot(pageInstance: Puppeteer.Page): Promise<Buff
     await makeDir(screenshotsPath);
     const screenshotName = generateUID();
     const filePath = path.join(screenshotsPath, toFilename(`${screenshotName}.png`));
-    console.log(`Screenshot file is located at: ${filePath}`);
-    return await pageInstance.screenshot({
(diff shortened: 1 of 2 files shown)


src/tests/end-to-end/common/generate-screenshot.ts
```diff
@@ -14,9 +14,13 @@ export async function takeScreenshot(pageInstance: Puppeteer.Page): Promise<Buff
...
     const screenshotName = generateUID();
     const filePath = path.join(screenshotsPath, toFilename(`${screenshotName}.png`));
-    console.log(`Screenshot file is located at: ${filePath}`);
-    return await pageInstance.screenshot({
+    const screenshotBuffer = await pageInstance.screenshot({
         path: filePath,
         fullPage: true,
     });
+
+    if (screenshotBuffer.length !== 0) {
```
dbjorge: I'm not sure this is going to give us the better debugging behavior that we want; I thought the problematic behavior was with `await pageInstance.screenshot()` throwing an exception, not with it returning with an empty buffer. I think we'd want to improve that case by adding a try-catch around the `await takeScreenshot` call in page.ts L143, rather than with a buffer length check here, no?

dbjorge:
Since Shobhit is out today and we'd like to continue on this PR to unblock debugging of other underlying issues, closing this one without merging and continuing work in #960

PR #957, opened 2019-07-17
Add Snippet to Store on Path Change

annikakouhia:
#### Description of changes

Currently, when adding a failure instance under our feature flag the following occurs:
1. User manually inputs comment and path
2. User clicks "validate path" button
3. Path is added to Path Snippet store on button click
4. Snippet is hardcoded in from the path just in the failure instance panel component (clicking the button sends the path to the store, and locally changes the component's state to hold a new snippet value simultaneously)
5. Upon clicking "Add", the comment, path, and hardcoded snippet are sent to the assessment store

This PR handles half of the snippet's flow. It covers adding the snippet to the Path Snippet store automatically whenever the path is updated. The value that is added is a hardcoded value based off the path ("Retrieved Snippet from Store for Path: "+ pathValue). A following PR will handle returning this new snippet value to the user and populating the failure instance panel with this new value.

#### Pull request checklist

### DIFF
--- a/src/common/message-creators/path-snippet-action-message-creator.ts
+++ b/src/common/message-creators/path-snippet-action-message-creator.ts
@@ -10,8 +10,8 @@ export class PathSnippetActionMessageCreator {
         const messageType = Messages.PathSnippet.AddCorrespondingSnippet;
         const payload = snippet;
         this.dispatcher.dispatchMessage({
-            messageType: messageType,
-            payload: payload,
+            messageType,
+            payload,
         });
     };
 }
--- a/src/common/message-creators/path-snippet-action-message-creator.ts
+++ b/src/common/message-creators/path-snippet-action-message-creator.ts
@@ -1,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { Messages } from '../messages';
+import { ActionMessageDispatcher } from './action-message-dispatcher';
+
+export class PathSnippetActionMessageCreator {
+    constructor(private readonly dispatcher: ActionMessageDispatcher) {}
+
+    public addCorrespondingSnippet = (snippet: string): void => {
+        const messageType = Messages.PathSnippet.AddCorrespondingSnippet;
+        const payload = snippet;
+        this.dispatcher.dispatchMessage({
+            messageType: messageType,
+            payload: payload,
+        });
+    };
+}
--- a/src/injected/path-snippet-controller.ts
+++ b/src/injected/path-snippet-controller.ts
@@ -21,13 +21,13 @@ export class PathSnippetController {
 
         const pathSnippetStoreState = this.pathSnippetStore.getState();
 
-        if (pathSnippetStoreState.path !== '' && pathSnippetStoreState.snippet === '') {
+        if (pathSnippetStoreState.path !== '') {
             const retrievedSnippet = this.getElementFromPath(pathSnippetStoreState.path);
             this.addCorrespondingSnippet(retrievedSnippet);
         }
     };
 
     private getElementFromPath = (path: string): string => {
-        return 'Retrieved Path from Store';
+        return 'Retrieved Snippet from Store for Path: ' + path;
     };
 }
--- a/src/injected/path-snippet-controller.ts
+++ b/src/injected/path-snippet-controller.ts
@@ -15,11 +15,11 @@ export class PathSnippetController {
     }
 
     private onChangedState = (): void => {
-        if (this.pathSnippetStore.getState() == null) {
+        const pathSnippetStoreState = this.pathSnippetStore.getState();
+
+        if (pathSnippetStoreState == null) {
             return;
         }
-
-        const pathSnippetStoreState = this.pathSnippetStore.getState();
 
         if (pathSnippetStoreState.path !== '') {
             const retrievedSnippet = this.getElementFromPath(pathSnippetStoreState.path);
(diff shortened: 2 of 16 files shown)


src/common/message-creators/path-snippet-action-message-creator.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Messages } from '../messages';
+import { ActionMessageDispatcher } from './action-message-dispatcher';
+
+export class PathSnippetActionMessageCreator {
+    constructor(private readonly dispatcher: ActionMessageDispatcher) {}
+
+    public addCorrespondingSnippet = (snippet: string): void => {
+        const messageType = Messages.PathSnippet.AddCorrespondingSnippet;
+        const payload = snippet;
+        this.dispatcher.dispatchMessage({
```
smoralesd: this can be declutter like this:
```typescript
this.dispatcher.dispatchMessage({
    messageType,
    payload,
});
```

src/injected/path-snippet-controller.ts
```diff
@@ -0,0 +1,33 @@
...
+        this.pathSnippetStore.addChangedListener(this.onChangedState);
+        this.onChangedState();
+    }
+
+    private onChangedState = (): void => {
+        if (this.pathSnippetStore.getState() == null) {
+            return;
+        }
+
+        const pathSnippetStoreState = this.pathSnippetStore.getState();
```
smoralesd: you may move this before the `if`, avoid duplicating `this.pathSnippetStore.getState()`

src/injected/path-snippet-controller.ts
```diff
@@ -0,0 +1,33 @@
...
+
+        if (pathSnippetStoreState.path !== '') {
+            const retrievedSnippet = this.getElementFromPath(pathSnippetStoreState.path);
+            this.addCorrespondingSnippet(retrievedSnippet);
+        }
+    };
+
+    private getElementFromPath = (path: string): string => {
+        return 'Retrieved Snippet from Store for Path: ' + path;
```
smoralesd: How are we using the text `'Retrieved Snippet from Store for Path:'`??? Is it for UI purposes? if that's so... then it should not be stored in the model layer. The suffix can be added on the component itself. That way, if for any reason we need to use the path (and only the path) on other places, we don't have to worry about the prefix.

src/tests/unit/tests/injected/path-snippet-controller.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    let addCorrespondingSnippetMock: IMock<(snippet: string) => void>;
+    let testObject: PathSnippetController;
+
+    beforeEach(() => {
+        pathSnippetStoreMock = Mock.ofType(PathSnippetStore);
+        addCorrespondingSnippetMock = Mock.ofInstance((snippet: string) => {});
+
+        testObject = new PathSnippetController(pathSnippetStoreMock.object, addCorrespondingSnippetMock.object);
+
+        pathSnippetStoreMock.setup(sm => sm.addChangedListener(It.isAny()));
```
smoralesd: You may use a stronger parameter assertion here:
```typescript
import { isFunction } from 'lodash';
...
... sm.addChangedListener(It.is(isFunction)) ...
```

Also, by default, typemoq mocks use "Loose" behavior, which means, if you make a call you have not previously setup on the mock, there is no error being thrown at that point. 

In this case, you're setting up for a function that doesn't return anything but you probably want to make sure that it's being called. You can achieve this by adding the following function call after the `setup` call:
```typescript
.verifiable()
```

and then call this at the end of each related test:
```typescript
pathSnippetStoreMock.verifyAll();
```

src/injected/path-snippet-controller.ts
annikakouhia: I was putting it in as a sanity check that the snippet was being stored via the flow I thought it would be! I'm planning to just store the snippet itself with no pre/post text once we've gotten things working smoothly

src/common/message-creators/path-snippet-action-message-creator.ts
annikakouhia: Got it! Thanks!

src/injected/path-snippet-controller.ts
annikakouhia: Great, yes that cleans things up :) Thanks!

src/tests/unit/tests/injected/path-snippet-controller.test.ts
annikakouhia: Gotcha! isFunction has been added in. I've been calling the verifyAll(); in the afterEach() method since I saw we were calling it with every test--does that accomplish the same thing? Thanks for the quick feedback!

src/injected/path-snippet-controller.ts
annikakouhia: ^But just let me know if you think we should take it out during these middle phases!

src/tests/unit/tests/injected/path-snippet-controller.test.ts
smoralesd: mmm `afterEach()` should be for clean up after the test, but `.verifyAll()` is part of the **AAA** test structure (**Arrange-Act-Assert**) so it should be part of the test function itself. If you happen to have more than one `.verifyAll()` at the end of each and every test, I'd recommend create a private function to run all those assertions there and then call said function from the tests.

I do know we still have `.verifyAll()` calls inside some `afterEach()` but after an engineering discussion we decided this was not the way we wanted to do the **Assert** part of the tests.

annikakouhia: oh got it! That makes sense I didn't realize afterEach() was more about cleanup. It looks like we call the two verifyAll() methods after each test, so I'll wrap them into a private function. Thanks for letting me know!

annikakouhia: I added the testObject.listenToStore() call to the private method as well because I saw we called it for every test. Let me know if you think it would be better left in the tests themselves for the sake of clarity!

PR #958, opened 2019-07-17
test: fix Windows-hosted-agent-specific "Page crashed!" puppeteer errors

dbjorge:
#### Description of changes

We were seeing our Windows hosted Pipelines agents hit "Page crashed!" errors somewhat reliably (~50% of e2e runs). I added functionality to collect detailed chrome logs and crash dumps and collect them as part of our published e2e results, but adding this actually fixed the issue. After some experimentation, I think the key to fixing it was in specifying a custom userDataDir launch parameter to chromium that uses a temporary userDataDir under our test-results folder, rather than deferring to Puppeteer's default behavior of putting it in the system %TEMP% directory. I'm not sure *why* this fixes the problem; maybe some sort of permissions issue, maybe some sort of overaggressive cleanup of the temp folder while the test is still running.

Validated that 15x runs on Windows with the userDataDir in place never saw the "Page crashed!" issue. Prior to this fix, we see it in about 50% of runs.

Out of scope: other buckets of e2e flakiness issues. In particular, it's expected that there will be some e2e-reliability failures still in the checks section of this PR (but none of them should be "Page crashed!" errors)

#### Pull request checklist

- [x] Addresses an existing issue: Partially addresses #867


src/tests/end-to-end/common/browser.ts
```diff
@@ -128,11 +128,16 @@ export class Browser {
...
 
-        this.memoizedBackgroundPage = new Page(await backgroundPageTarget.page());
+        this.memoizedBackgroundPage = new Page(await backgroundPageTarget.page(), { onPageCrash: this.onPageCrash });
 
         return this.memoizedBackgroundPage;
     }
 
+    private onPageCrash = () => {
+        const errorMessage = `!!! Browser.onPageCrashed: see detailed chrome logs '${browserLogPath(this.browserInstanceId)}'`;
+        console.log(errorMessage);
```
smoralesd: should we use `console.error` here??

dbjorge: Practically speaking, I think Jest outputs them the same way, but I think it's more descriptive so I agree with this idea :)

src/tests/end-to-end/common/browser-factory.ts
```diff
@@ -58,6 +68,12 @@ async function launchNewBrowser(): Promise<Puppeteer.Browser> {
     // end shows up as a generic timeout error with no meaningful logging".
     await verifyExtensionIsBuilt(extensionPath);
 
+    const browserLogDir = browserLogPath(browserInstanceId);
+    const userDataDir = path.join(browserLogDir, 'userDataDir');
+
+    // If this isn't present when chrome launches, it'll silently fail to log anything
+    await util.promisify(fs.mkdir)(userDataDir, { recursive: true });
```
lisli1: Does this comment mean that Chrome will silently fail to log anything if the directory doesn't exist? It was a little confusing when I first read it, but maybe adding an explanation for what happens when it is present (i.e. "Create directory so Chrome will log errors").

src/tests/end-to-end/common/browser.ts
dbjorge: updated

src/tests/end-to-end/common/browser-factory.ts
dbjorge: Clarified the comment

src/tests/end-to-end/common/browser.ts
```diff
@@ -12,7 +12,7 @@ export class Browser {
     private memoizedBackgroundPage: Page;
     private pages: Array<Page> = [];
 
-    constructor(private readonly underlyingBrowser: Puppeteer.Browser) {
+    constructor(public readonly browserInstanceId: string, private readonly underlyingBrowser: Puppeteer.Browser) {
```
lisli1: Just curious, what are the reasons for making `browserInstanceId` public? I can't find an instance where this member variable is accessed outside the `Browser` class.

dbjorge: Oops, that was leftover from when I was using a different design for the onPageCrash handler. You're right, this can be private :)

PR #960, opened 2019-07-18
test: prevent errors in takeScreenshot from suppressing the error that motivated taking the screenshot

dbjorge:
#### Description of changes

Previously, in our e2e action wrappers that used `screenshotOnError`, an error case that prevents both the original error and the screenshot-taking from working (eg, underlying browser/page closed prematurely) would only show an error text/stack for the takeScreenshot error, even though the original error is usually more valuable. This causes it to print both errors and use the original as the primary.

Replaces/builds on @Shobhit1 's #956 , since he's out today 

#### Pull request checklist

- [x] Addresses an existing issue: Contributes to #867 

### DIFF
--- a/src/tests/end-to-end/common/generate-screenshot.ts
+++ b/src/tests/end-to-end/common/generate-screenshot.ts
@@ -14,9 +14,13 @@ export async function takeScreenshot(pageInstance: Puppeteer.Page): Promise<Buff
     await makeDir(screenshotsPath);
     const screenshotName = generateUID();
     const filePath = path.join(screenshotsPath, toFilename(`${screenshotName}.png`));
-    console.log(`Screenshot file is located at: ${filePath}`);
-    return await pageInstance.screenshot({
+    const screenshotBuffer = await pageInstance.screenshot({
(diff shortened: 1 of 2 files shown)


src/tests/end-to-end/common/generate-screenshot.ts
```diff
@@ -14,9 +14,13 @@ export async function takeScreenshot(pageInstance: Puppeteer.Page): Promise<Buff
...
     const filePath = path.join(screenshotsPath, toFilename(`${screenshotName}.png`));
-    console.log(`Screenshot file is located at: ${filePath}`);
-    return await pageInstance.screenshot({
+    const screenshotBuffer = await pageInstance.screenshot({
         path: filePath,
         fullPage: true,
     });
+
+    // We don't log this until after screenshot() succeeds to avoid misleading logs if it throws
+    console.log(`Screenshot file is located at: ${filePath}`);
```
lisli1: This is great!

lisli1:
I like that the error message is descriptive and easy to follow :)

PR #963, opened 2019-07-22
refactor: extract details dialog buttons section to new command bar

smoralesd:
#### Description of changes

Currently, the `DetailsDialog` component (used on the target page to show failure details) has a lot of responsibilities, has some serious semantic level problems and is tightly couple to the `DetailsDialogHandler`. This PR is one small step into cleaning up the component by first breaking it into smaller classes/components.

**Out of scope:**
- I did add test for check functions being pass to the `CommandBar` component from the `DetailsDialog`. I did not add test for any of functions being pass/used to/by other components as doing so will increase this PR size and it looks like a pretty deep rabbit hole.
- This is the result of direct, straight forward extraction of a subset of the logic into the new `CommandBar`. The new component can use some more refactor but I see it out of scope as it may involve some refactor on the components being used in here.

#### Pull request checklist

   - does not exist
   - not affected
   - not affected


src/injected/components/details-dialog.tsx
```diff
@@ -38,7 +34,7 @@ export type DetailsDialogDeps = {
     targetPageActionMessageCreator: TargetPageActionMessageCreator;
     browserAdapter: BrowserAdapter;
 } & CopyIssueDetailsButtonDeps &
```
dbjorge: >CopyIssueDetailsButtonDeps [](start = 4, length = 26)

`CommandBarDeps` includes this already, no?

src/tests/unit/tests/injected/components/command-bar.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { DecoratedAxeNodeResult } from '../../../../../injected/scanner-utils';
+import { BaseDataBuilder } from '../../../common/base-data-builder';
+import { EventStubFactory } from '../../../common/event-stub-factory';
+
+describe('CommandBar', () => {
+    describe('renders', () => {
+        const showInspectButtonMessage = [true, false];
+
+        it.each(showInspectButtonMessage)('renders, shows inspect button message: %s', show => {
+            const props = defaultCommandBarPropsBuilder()
```
dbjorge: >defaultCommandBarPropsBuilder [](start = 26, length = 29)

I'm not sure I understand the need for the builder function here; it feels like an extra layer of overhead compared to:

```ts
const props: CommandBarProps = {
    ...defaultCommandBarProps, // or ...defaultCommandBarProps(), if you'd like to keep the separate-object-per-test guarantee
    shouldShowInspectButtonMessage: () => show
}
```

the object expansion version also plays nicer with F2 renaming of the property being overridden.

src/injected/components/details-dialog.tsx
smoralesd: It does... bad merge on my side.

src/tests/unit/tests/injected/components/command-bar.test.tsx
smoralesd: make sense

PR #969, opened 2019-07-23
refactor: extract file url unsupported message panel

smoralesd:
#### Description of changes

`PopupView` knows a lot of lower level details. In this PR I'm extracting one of those concerns out to a new component. 

#### Pull request checklist

   - does not exist
   - not affected
   - not affected

### DIFF
--- a/src/tests/unit/tests/popup/components/__snapshots__/popup-view.test.tsx.snap
+++ b/src/tests/unit/tests/popup/components/__snapshots__/popup-view.test.tsx.snap
@@ -20,6 +20,15 @@ exports[`PopupView render actual content render toggles view: launch pad 1`] = `
 </Fragment>"
 `;
 
+exports[`PopupView render actual content render toggles view: launch pad: subtitle 1`] = `
+"<Fragment>
+  <NewTabLink href=\\"https://go.microsoft.com/fwlink/?linkid=2082374\\" aria-label=\\"demo video\\" title=\\"watch the 3 minute video introduction\\" onClick={[Function: onClickTutorialLink]}>
+    Watch 3-minute video introduction
+  </NewTabLink>
+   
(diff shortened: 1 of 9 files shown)


src/tests/unit/tests/popup/components/__snapshots__/popup-view.test.tsx.snap
```diff
@@ -70,44 +70,65 @@ exports[`PopupView renderFailureMsgPanelForChromeUrl 1`] = `
...
-        >
-          Allow Access to file URLs
-        </span>
-        .
-      </div>
-    </div>
-  </div>
-</div>
+<FileUrlUnsupportedMessagePanel
+  deps={
```
dbjorge: Would it make sense to suppress the deps value from the snapshot, since the intention of deps is that the thing under test here shouldn't necessarily care about the specific dependencies its children need? If you wanted to do it, I think it would look amount to changing L269 of `popup-view.test.tsx` to the following:

```ts
        expect(wrapped.getElement()).toMatchSnapshot({ props: { deps: expect.any(Object) } });
```

smoralesd: TL;DR: changing out we handle snapshots is out of scope for this PR.

As we talked yesterday, it seems we need to run some small investigation around jest snapshot serializers. That looks promising as it will allow us to simplify snapshots (by declutering deps serialization) without having to change every test in order to do so.
That being said, the current practice on our snapshot testing is to include everything we get from calling `.getElemennt()` on the enzyme shallow wrapper, including deps so I'll be hesitant to include this change as part of this PR as it seems out of scope (the scope being extract a component that, from the `PopupView` point of view, is an implementation detail and don't really belong in the original component).

PR #972, opened 2019-07-24
test: update target-page-changed-view test cases to match actual use cases

dbjorge:
#### Description of changes

Noticed a few different issues with `target-page-changed-view.test.tsx`:

* It was testing several cases that aren't allowed to occur in practice (title and toggleLabel are non-optional properties of DisplayableVisualizationTypeData)
* Its tests of subtitle differences were incomplete (the test cases included different subtitles, but didn't actually pass subtitle to the thing under test)
* The test's snapshot behavior wasn't consistent between node 10 and node 12 (the way the testCase objects serialized into test names differed on the 2 platforms)

This PR fixes all of these issues by partially rewriting the tests. This fixes the only case where our unit tests didn't work consistently in node 12.

#### Pull request checklist

### DIFF
--- a/src/tests/unit/tests/DetailsView/components/target-page-changed-view.test.tsx
+++ b/src/tests/unit/tests/DetailsView/components/target-page-changed-view.test.tsx
@@ -6,38 +6,32 @@ import { DisplayableVisualizationTypeData } from '../../../../../common/configs/
 import { VisualizationType } from '../../../../../common/types/visualization-type';
 import { TargetPageChangedView, TargetPageChangedViewProps } from '../../../../../DetailsView/components/target-page-changed-view';
 
-type RenderTestCases = {
-    title?: string;
-    toggleLabel?: string;
-    subtitle?: JSX.Element;
-};
-
 describe('TargetPageChangedView', () => {
-    describe('renders', () => {
-        const clickHandlerStub: () => void = () => {};
-        const testCases: RenderTestCases[] = [
-            {},
-            { title: 'title' },
-            { toggleLabel: 'toggle-label' },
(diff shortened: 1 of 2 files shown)


src/tests/unit/tests/DetailsView/components/target-page-changed-view.test.tsx
```diff
@@ -6,38 +6,32 @@ import { DisplayableVisualizationTypeData } from '../../../../../common/configs/
...
-            { title: 'title', toggleLabel: 'toggle-label', subtitle: <span>subtitle</span> },
-        ];
+    it('renders without optional subtitle', () => {
+        testRenderWithOptionalSubtitle(undefined);
+    });
 
-        it.each(testCases)('case %o', testCase => {
-            const { title = '', toggleLabel = '' } = testCase;
+    it('renders with optional subtitle', () => {
+        testRenderWithOptionalSubtitle(<span>test subtitle content</span>);
```
smoralesd: why not use `it.each` for this 2 cases?

dbjorge: The rendering of the JSX.Element in the test title was the thing causing the node 10/12 incompatibility; there didn't seem to be a good option for pretty-printing it reasonably, it was either "print it as an object in a way that is platform specific" or "print it as '[Object]'".

I could have just made the content text be the test case discriminator and had the test body use a conditional to wrap it (`const subtitle = testCaseSubtitleText ? <span>{testCaseSubtitleText}</span> : undefined;`), but my default instinct is always to avoid conditionals in test cases where it's straightforward to do so :)

PR #973, opened 2019-07-24
Returning Stored Snippet to UI

annikakouhia:
#### Description of changes

Upon click of "validate path" button, the path was previously added to the PathSnippetStore and a corresponding (*hardcoded) snippet was automatically added to the PathSnippetStore as well. Now, the DetailsView page listens to the PathSnippetStore, and propagates the path and snippet stored there all the way to the Failure Instance Panel Control where it is displayed appropriately to the user.

We can now add, edit and remove failure instances which contain an auto-populated/displayed/stored snippet.

We clear the PathSnippetStore on close of the Failure Instance panel. We use the store to decide what FailureInstance object will be passed to the Failure Instance Panel and then used to populate the panel's fields. There are four scenarios the FailureInstance object can be created in:
1. When adding, if we are rendering the panel for the first time we want a FailureInstance with null values so no values are displayed.
2. When adding, if we are re-rendering after clicking the button to validate the path, we want to pass a FailureInstance object with the path and snippet filled from the PathSnippetStore.
3. When editing, if we are rendering the panel for the first time we want a FailureInstance with values from the retrieved instance we are trying to edit (saved in the AssessmentStore).
4. When editing, if we are re-rendering after 'validating' a new selector we retrieve the selector and corresponding snippet from the PathSnippetStore instead of the instance saved in the AssessmentStore (because this instance holds the old, unedited information)

This PR handles all four cases above, propagates information from the background to UI, and tests all functionality.

In coming PRs I will also:
a) Clear the PathSnippetStore on tab refresh
b) Change the PathSnippetStore to an InstanceStore and save the failure instance description alongside the path and snippet to make the logic of the Failure Instance Panel clearer.

addsnippet

#### Pull request checklist


src/tests/unit/tests/DetailsView/components/failure-instance-panel-control.test.tsx
```diff
@@ -87,26 +91,28 @@ describe('FailureInstancePanelControlTest', () => {
     });
 
     test('onValidateSelector ', () => {
-        const selector = 'some selector';
-        const snippet = 'snippet for ' + selector;
+        const path = 'some selector';
```
karanbirsingh: Might be able to move this definition into line 100 similar to how `failureDescription` is defined inline

src/tests/unit/tests/DetailsView/components/failure-instance-panel-control.test.tsx
```diff
@@ -148,10 +154,13 @@ describe('FailureInstancePanelControlTest', () => {
             .props()
             .onDismiss();
 
+        clearPathSnippetDataMock.setup(handler => handler()).verifiable(Times.once());
```
karanbirsingh: These tests might read more naturally if the setup were done before the action to verify - e.g. calling `dismiss` presumably invokes `clearPathSnippetDataMock` but the only lines between setup/verify here are `expects`. I'm somewhat surprised the `Times.once` verification passes

src/tests/unit/tests/DetailsView/handlers/assessment-instance-table-handler.test.tsx
```diff
@@ -121,15 +121,77 @@ describe('AssessmentInstanceTableHandlerTest', () => {
...
+        const expectedRows: CapturedInstanceRowData[] = [
+            {
+                instance: instance,
+                instanceActionButtons: instanceActionButtons,
+            },
+        ];
+        expect(expectedRows).toEqual(rows);
+    });
+
+    test('createCapturedInstanceTableItems withoutPathSnippetStoreData', () => {
```
karanbirsingh: missing a space
```suggestion
    test('createCapturedInstanceTableItems without PathSnippetStoreData', () => {
```

src/tests/unit/tests/DetailsView/handlers/assessment-instance-table-handler.test.tsx
```diff
@@ -220,6 +283,11 @@ describe('AssessmentInstanceTableHandlerTest', () => {
         testSubject.addPathForValidation(path);
     });
 
+    test('clearPathSnippetData', () => {
+        actionMessageCreatorMock.setup(a => a.clearPathSnippetData()).verifiable(Times.once());
```
karanbirsingh: If you remove the call to clearPathSnippetData in your code, does this test fail? It seems like we'd need a `verifyAll` somewhere but I don't see one

src/tests/unit/tests/DetailsView/components/failure-instance-panel-control.test.tsx
annikakouhia: Got it! Good point--not using that variable anywhere else

src/tests/unit/tests/DetailsView/handlers/assessment-instance-table-handler.test.tsx
annikakouhia: Great point--just changed these for my two added tests. The concerning thing is the rest of the functions don't include a call to verifyAll (like the addFailureInstance test right below mine) so they're all passing without verification too... Should I try and refactor the entire file?

src/tests/unit/tests/DetailsView/components/failure-instance-panel-control.test.tsx
annikakouhia: Gotcha yes--I think I was thinking about it as if I were putting the clearPathSnippetDataMock where it would be called in the real order, but I realize now that the *set up* of it makes more sense before, and that's what we're really doing in that line (setting up, not calling the function)

src/tests/unit/tests/DetailsView/components/failure-instance-panel-control.test.tsx
```diff
@@ -209,25 +220,57 @@ describe('FailureInstancePanelControlTest', () => {
...
+
+    test('componentDidUpdate reassigns state', () => {
+        const prevProps = createPropsWithType(CapturedInstanceActionType.CREATE);
+        const newProps = createPropsWithType(CapturedInstanceActionType.CREATE);
+        const newFailureInstance = {
+            failureDescription: null,
+            path: 'inputed path',
+            snippet: 'snippet for path',
+        };
+        newProps.failureInstance = newFailureInstance;
```
karanbirsingh: For these cases where we modify a property after creation, we can sometimes use the spread operator (ellipses, e.g. …) to declare everything up front. For example, this might look like: 
```
const props = {
   failureInstance: YOUR_OBJECT_HERE,
   ...createPropsWithType(CapturedInstanceActionType.EDIT
};
``` 
It's nice when this works because the reader doesn't need to update their understanding of the object on a later line

src/tests/unit/tests/DetailsView/handlers/assessment-instance-table-handler.test.tsx
karanbirsingh: I would check that we need this change, either by forcing any of these tests to fail & seeing if they still pass (like the earlier suggestion), or seeing if an absurd expectation (`Times.exactly(49)` instead of `Times.once()`) still passes. If so, it might be doable to just add `actionMessageCreatorMock.verifyAll()` in each test that has a `setup` for it (I count 5, 3 of which are about path/snippets)

annikakouhia: Cool--I checked the previous suggestion of taking out the call to the function on the testSubject and they failed with the verifyAll() included. I'll add them in!

annikakouhia: I've found 5 that I've added to (addPathForValidation, clearPathSnippetData, addFailureInstance, passUnmarkedInstance, and updateFocusedInstance)

PR #975, opened 2019-07-24
refactor: move frames+tab coordination responsibility out of BrowserAdapter

dbjorge:
#### Description of changes

Make BrowserAdapter slightly thinner by moving the multi-method coordination in BrowserAdapter.sendMessageToFramesAndTab out of the adapter and into its only consumer

#### Pull request checklist

### DIFF
--- a/src/tests/unit/tests/background/tab-context-broadcaster.test.ts
+++ b/src/tests/unit/tests/background/tab-context-broadcaster.test.ts
@@ -1,8 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { Mock } from 'typemoq';
+import { Mock, MockBehavior, Times } from 'typemoq';
 
 import { TabContextBroadcaster } from '../../../../background/tab-context-broadcaster';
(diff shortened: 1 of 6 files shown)


src/tests/unit/tests/background/tab-context-broadcaster.test.ts
```diff
@@ -11,13 +12,14 @@ describe('TabContextBroadcasterTest', () => {
         const testMessage = { someData: 1 } as any;
         const expectedMessage = { tabId: testTabId, ...testMessage } as StoreUpdateMessage<any>;
 
-        const sendMessageToFramesAndTabMock = Mock.ofInstance((tabId, message) => {});
+        const browserAdapterMock = Mock.ofType<BrowserAdapter>(null, MockBehavior.Strict);
```
smoralesd: I think there is a slight difference between using `null` and `undefined` on the mocks... If I remember correctly, using `undefined` makes typemoq to use es6 proxies. Not sure if you're intentionally using `null`. I just thought that I'd mention it.

dbjorge: Tried both, no impact on this particular test. Codebase uses both in existing tests; null is slightly more common so sticking with it here.

PR #977, opened 2019-07-25
refactor: move js/css injection logic out of BrowserAdapter and under test

dbjorge:
#### Description of changes

The BrowserAdapter css/js injection methods had some business logic in them that wasn't tested. This moves those methods to their only consumer (`ContentScriptInjector`), leaves thinner wrappers in BrowserAdapter/ChromeAdapter, and revamps the `ContentScriptInjector` tests to include tests of the moved logic (plus some other paths in the injector that weren't previously well tested)

#### Pull request checklist

### DIFF
--- a/src/common/browser-adapters/browser-adapter.ts
+++ b/src/common/browser-adapters/browser-adapter.ts
@@ -19,8 +19,8 @@ export interface BrowserAdapter {
     sendMessageToTab(tabId: number, message: any): void;
     sendMessageToFrames(message: any): void;
     sendMessageToAllFramesAndTabs(message: any): void;
-    injectJs(tabId, file: string, callback: Function): void;
-    injectCss(tabId, file: string, callback: Function): void;
+    executeScriptInTab(tabId: number, details: chrome.tabs.InjectDetails, callback?: (result: any[]) => void): void;
+    insertCSSInTab(tabId: number, details: chrome.tabs.InjectDetails, callback?: Function): void;
     getRunTimeId(): string;
     createNotification(options: chrome.notifications.NotificationOptions): void;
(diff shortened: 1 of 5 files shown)


src/common/browser-adapters/browser-adapter.ts
```diff
@@ -19,8 +19,8 @@ export interface BrowserAdapter {
     sendMessageToTab(tabId: number, message: any): void;
     sendMessageToFrames(message: any): void;
     sendMessageToAllFramesAndTabs(message: any): void;
-    injectJs(tabId, file: string, callback: Function): void;
-    injectCss(tabId, file: string, callback: Function): void;
+    executeScriptInTab(tabId: number, details: chrome.tabs.InjectDetails, callback?: (result: any[]) => void): void;
```
smoralesd: I'm wondering if having `chrome.tabs.InjectDetails` for the type is a good idea in terms of decoupling from chrome as much as we can... just a thought, I don't think you need to change it on this PR is just something we should consider.

dbjorge: the `chrome.*` namespace is a little misleadingly named; Firefox also implements it. I'd be interested in switching to use the [`browser.*` namespace](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Build_a_cross_browser_extension) throughout instead (Mozilla provides an [officially supported polyfill](https://github.com/mozilla/webextension-polyfill/) for chrome/edge support of it), where all of these could switch to using the `browser.*` equivalents, but would rather do that in more targetted PRs for it; for now, using `chrome.tabs.InjectDetails` is consistent with the type usage throughout the other methods on this adapter

PR #980, opened 2019-07-26
test: add e2e page controllers for each major type of page involved in tests

dbjorge:
#### Description of changes

Adds in a "controller" class for each major type of page our e2e tests automate (`BackgroundPage`, `DetailsViewPage`, `TargetPage`, etc), so future PRs can start putting common logical operations required to test those pages in a common place instead of spreading them through multiple tests. 

Out of scope: pulling all the consolidatable helpers out of every test into the controllers. Want to leave this for separate PRs, this one is already big.

#### Pull request checklist


smoralesd:
LGTM!!! I'll wait for you to change the status to "Ready for review" to have a second look.

dbjorge:
The remaining issues in the e2e-reliability build seem to be incompatibilities between office fabric's Dropdown component and the way we're disabling animations; I think we'll probably end up having to revert the animation disabling to work around it, but leaving that for a separate PR

src/tests/end-to-end/common/browser.ts
```diff
@@ -21,67 +23,70 @@ export class Browser {
...
-    }
-
-    public async setupNewTargetPage(): Promise<TargetPageController> {
-        const targetPage = await this.newTestResourcePage('all.html');
-
-        await targetPage.bringToFront();
-        const targetPageTabId = await this.getActivePageTabId();
-        return new TargetPageController(targetPage, targetPageTabId);
-    }
+    public async backgroundPage(): Promise<BackgroundPage> {
```
smoralesd: I just though of this: while other "creational" functions here (like ` newPage` or ' newTargetPage'), `backgroundPage` is the only one that cached the result to return it on a subsequent call. I'm wondering if we should make that clear by renaming the function (maybe all the creational functions need to be renamed)

dbjorge: This is the reason I named it `backgroundPage()` instead of `newBackgroundPage()` like the rest of them; I figured this was probably sufficient since I'd expect a typical maintainer of our code would probably be aware already that a background page is something that hangs around persistently, rather than being launched explictly. Do you have a specific verb in mind that might be more clear? `get` and `find` are the ones that came to my mind but I didn't think they were significantly more clear than `backgroundPage()` on its own...

smoralesd: I thought about `get`, not sure if that's strong enough though. I guess the current difference may be enough.

PR #982, opened 2019-07-29
refactor: setup relative path for imports

smoralesd:
#### Description of changes

Currently, moving files around force us to update the import path for our own code. To make this kind of refactor simpler, we should be using relative paths so instead of using this:
```typescript
import { NamedSFC } from '../../../../common/react/named-sfc';
```
we could use this:
```typescript
import { NamedSFC } from 'common/react/named-sfc';
```
That means that moving that file will not affect how do we import `NamedSFC` and thus, less file changes when moving files around.

In order to achieve this, we need to:
- update tsconfig's `baseUrl` value
   - this will make vscode work properly
- update how webpack resolve modules from our own source code
   - this will make "building" work properly

So only 2 changes are necessary to enable us to use relative path imports. As an example of how this will look, I update all components under `src\DetailsView\reports\components\report-sections` to use relative path imports.

**NOTES**
- One of the nice thing about how tsc and webpack handle this is we can have relative and absolute path imports, i.e. we don't need to update all of our imports right away.
- I started with `report-sections` folder because my most immediate plan is to move the whole `reports` folder out of the `DetailsView` folder.

#### Pull request checklist

   - does not exist
   - not affected
   - not affected
   - not affected
   - not affected

### DIFF
--- a/src/DetailsView/reports/components/report-sections/body-section.tsx
+++ b/src/DetailsView/reports/components/report-sections/body-section.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import * as React from 'react';
 import { NamedSFC } from 'common/react/named-sfc';
+import * as React from 'react';
 
 export const BodySection = NamedSFC('BodySection', ({ children }) => {
     return <body>{children}</body>;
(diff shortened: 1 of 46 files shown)


smoralesd:
> I like this and think the implementation looks fine, but would like a second reviewer since it's essentially a team coding style change

Make sense. Thanks for adding the tag.

PR #990, opened 2019-07-29
test: add end-to-end test to verify tab stops visualization works

Shobhit1:
#### Description of changes
PR adds tab-stops end-to-end test and tries to verify that the visualization works when tabstop is enabled from fastpass page.

#### Pull request checklist

- [x] Addresses an existing issue: Contributes to #990 


src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,81 @@
...
+describe('Tabstop tests', () => {
+    describe('tabstop from fastpass', () => {
+        let browser: Browser;
+        let targetPage: TargetPage;
+        let fastPassPage: Page;
+
+        beforeAll(async () => {
+            browser = await launchBrowser({ suppressFirstTimeDialog: true });
+            targetPage = await browser.newTargetPage();
+            fastPassPage = await gotoFastPass(browser, targetPage.tabId);
```
smoralesd: should we be using `DetailsViewPage` (`src\tests\end-to-end\common\page-controllers\details-view-page.ts`) which has `switchToFastPass()`???

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                browserLocal.waitForDetailsViewPage(targetPage).then(page => (fastPass = page)),
+                popupPage.clickSelector(popupPageElementIdentifiers.fastPass),
+            ]);
+
+            return fastPass;
+        }
+
+        async function goToTabStopTest(): Promise<void> {
+            await fastPassPage.clickSelector(fastPassSelectors.tabstopNavButtonSelector);
+            // await fastPassPage.disableAnimations();
```
smoralesd: commented code

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,81 @@
...
+            await fastPassPage.clickSelector(disabledToggleSelector);
+
+            const EXTRA_TOGGLE_OPERATION_TIMEOUT_MS = 5000;
+
+            await fastPassPage.waitForSelector(enabledToggleSelector, {
+                timeout: DEFAULT_PAGE_ELEMENT_WAIT_TIMEOUT_MS + EXTRA_TOGGLE_OPERATION_TIMEOUT_MS,
+            });
+        }
+
+        async function verifyTabStopPageLoaded(): Promise<void> {
```
smoralesd: should we rename this to `waitFor...` instead of using `verify`??? On a test context, the latter kinds of implies some sort of more explicit assertion of some thing we want as a result or side effect of the test itself, but in this case, this is seems to be use on the **Arrange** part of the test not on the **Assert** part.

src/tests/end-to-end/common/element-identifiers/common-selectors.ts
```diff
@@ -11,3 +11,9 @@ export const CommonSelectors = {
 export const GuidanceContentSelectors = {
     mainContentContainer: '.content-container',
 };
+
+export const fastPassSelectors = {
+    tabstopNavButtonSelector: 'a[title="Tab stops"]',
```
smoralesd: should it be "tab**S**topNavButtonSelector"? to make it match the other names case styles?

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,81 @@
...
+            // await fastPassPage.disableAnimations();
+            await verifyTabStopPageLoaded();
+        }
+
+        async function enableToggleByAriaLabel(ariaLabel: string): Promise<void> {
+            const toggleSelector = `button[aria-label="${ariaLabel}"]`;
+            const enabledToggleSelector = `${toggleSelector}[aria-checked=true]`;
+            const disabledToggleSelector = `${toggleSelector}[aria-checked=false]`;
+
+            await fastPassPage.waitForDuration(6000);
```
smoralesd: It doesn't seem to be clear by reading this function why we start by waiting for 6000 ms (I presume). Shouldn't we interact with the page and then wait for "something" to happen?

Also, `waitForDuration` seems brittle. `6000` seems like an arbitrary value which may or may not work base on the system the test is being run on. Can we not wait for a selector? 

If I'm getting the purpose of this function correctly from it's name (`enableToggleByAriaLabel`), the intention is to actually enable tab stops visualization, is that right? 

On thing about enabling tab stops: when we do enable tab stops (or any other visualization I think) we create an specific `div` (more precisely `<div id="accessibility-insights-root-container">`). If the target page is created fresh for this test every time, then there is no such `div` until we enable tab stops visualization. Is this not possible?

src/tests/end-to-end/tests/details-view/__snapshots__/tabstop.test.ts.snap
```diff
@@ -0,0 +1,60 @@
...
+      rel="stylesheet"
+      type="text/css"
+    />
+    <link
+      href="{{EXTENSION_URL_BASE}}/bundle/injected.css"
+      rel="stylesheet"
+      type="text/css"
+    />
+    <div
+      class="insights-container insights-highlight-container insights-issues"
```
smoralesd: This looks off.... by enabling tab stops visualization I can see this on the target page: 
```html
<div class="insights-container insights-highlight-container insights-tab-stops">
```
but the snapshot doesn't have this div... it has one with `insights-issues` for class name which shows up when enabling automated checks... there is something off here for sure.

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                popupPage.clickSelector(popupPageElementIdentifiers.fastPass),
+            ]);
+
+            return fastPass;
+        }
+
+        async function goToTabStopTest(): Promise<void> {
+            await fastPassPage.clickSelector(fastPassSelectors.tabstopNavButtonSelector);
+            // await fastPassPage.disableAnimations();
+            await verifyTabStopPageLoaded();
```
smoralesd: `verifyTabStopPageLoaded` is a thin wrapper around an specific `waitForSelector` call... the abstraction level here seems off between lines 57 and 59 (57 has more details than 59) I'm wondering if we actually need to have `verifyTabStopPageLoaded` in the first place

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+            // press tabs 3 times. todo: DRY this code
+            await targetPage.keyPress('Tab');
+            await targetPage.keyPress('Tab');
+            await targetPage.keyPress('Tab');
+
+            expect(await targetPage.getShadowRootHtmlSnapshot()).toMatchSnapshot();
+        });
+
+        async function gotoFastPass(browserLocal: Browser, targetTabId: number): Promise<Page> {
```
smoralesd: should this be "go**T**oFastPass" so it match with other go to function casing?

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,81 @@
...
+            const popupPage = await browserLocal.newPopupPage(targetPage);
+            let fastPass: Page;
+            await Promise.all([
+                browserLocal.waitForDetailsViewPage(targetPage).then(page => (fastPass = page)),
```
smoralesd: do we need to use the details view page??? why can we not enable tab stops from the popup page, ad hoc tools view?

src/tests/end-to-end/common/element-identifiers/common-selectors.ts
```diff
@@ -11,3 +11,9 @@ export const CommonSelectors = {
...
+    tabStopToggle: 'button[aria-label="Tab stops"]',
+    tabStopToggleAlternativeSelector: 'button[role="switch"]',
```
smoralesd: Didn't find any reference to this new selector on this PR... if we're not using it I guess we shouldn't add it

src/tests/end-to-end/common/element-identifiers/common-selectors.ts
Shobhit1: Will update.

src/tests/end-to-end/common/element-identifiers/common-selectors.ts
Shobhit1: I have some refactor that I need to do for this portion of the PR. Will update all the comments after that. But agreed. There are certain selectors that I added when I was experimenting with selection of the radio button but not using them anymore, so will clean that up!

src/tests/end-to-end/tests/details-view/tabstop.test.ts
Shobhit1: Will go through the new implementation and update this!

src/tests/end-to-end/tests/details-view/tabstop.test.ts
Shobhit1: I wanted to create one test with the flow from details page and then add similar test cases for other scenarios like popup-> adhoc tools view; which will be in the followup PRs on this

src/tests/end-to-end/tests/details-view/tabstop.test.ts
Shobhit1: We do need it, just to wait for the radio button to load and do its animation stuff. Without a couple of waitForSelectors, I found that the click event was very flaky!

src/tests/end-to-end/tests/details-view/tabstop.test.ts
Shobhit1: updated.

Shobhit1:
Pending:

- Change the `shadowDomHTMLSnapshot` function to sanitize the `height` and `width` of the target page visualization. This is causing flakiness because this is very environment based. So, need to work on that with @dbjorge or try it out myself.

src/tests/end-to-end/tests/details-view/__snapshots__/tabstop.test.ts.snap
Shobhit1: Updated.

src/tests/end-to-end/common/element-identifiers/common-selectors.ts
```diff
@@ -11,3 +11,16 @@ export const CommonSelectors = {
...
+    tabStopNavButtonSelector: 'a[title="Tab stops"]',
+    tabStopToggleCheckedSelector: 'button[aria-label="Tab stops"][aria-checked=true]',
+    tabStopToggleUncheckedSelector: 'button[aria-label="Tab stops"][aria-checked=false]',
+    tabStopVisulizationStart: '.insights-tab-stops',
+};
+
+export const targetPageSelectors = {
+    targetPageNativeWidgetIFrameName: 'native-widgets',
+    targetPageNativeWidgetFirstRadio: '#input-radio-1',
+    targetPageNativeWidgetH1: '#h1-native-widget',
```
lisli1: I added a `target-page-selectors.ts` file in my PR (#993) that might be a more relevant place for these selectors.

src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -86,7 +86,11 @@ export class Page {
         return await this.screenshotOnError(async () => await this.underlyingPage.evaluate(fn, ...args));
     }
 
-    public async getMatchingElements<T>(selector: string, elementProperty: keyof Element): Promise<T[]> {
+    public async $(selector: string): Promise<any> {
+        return await this.screenshotOnError(async () => await this.underlyingPage.$(selector));
```
lisli1: This looks really similar to `getSelectorElement` below. Should we pick one/combine them?

Shobhit1: Done. Removed `$` and used `getSelectorElement` for my purpose. Good catch. Thanks.

src/tests/end-to-end/common/element-identifiers/common-selectors.ts
Shobhit1: Moved the selectors that I was using to `target-page-selectors.ts`

src/tests/end-to-end/common/element-identifiers/common-selectors.ts
```diff
@@ -11,3 +11,10 @@ export const CommonSelectors = {
 export const GuidanceContentSelectors = {
     mainContentContainer: '.content-container',
 };
+
+export const fastPassSelectors = {
```
dbjorge: I think if these are specific to fast pass, it might be better to put them in a fastpass-specific selectors file, rather than common-selectors.ts

src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -246,6 +260,10 @@ export class Page {
         });
     }
 
+    public async takePagesScreenshot(): Promise<void> {
```
dbjorge: `takeScreenshot()` or `screenshot()` are probably fine here; this is within the `Page` class, so all of the methods are going to be assumed to be acting on the `Page` by default.

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dbjorge: I think this is more of a test of the target page visualizations than the details-view, since you have a test case that doesn't involve the details view at all. Maybe `target-page/visualization-tab-stops.test.ts` would be more meaningful than `details-view/tabstop.test.ts`?

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,104 @@
...
+            await targetPage.waitForDescendentSelector(shadowRoot, fastPassSelectors.tabStopVisulizationStart, { visible: true });
+
+            for (let i = 0; i < 3; i++) {
+                await targetPage.keyPress('Tab');
+            }
+
+            expect(await targetPage.getShadowRootHtmlSnapshot()).toMatchSnapshot();
+        });
+
+        async function goToTabStopTest(): Promise<void> {
```
dbjorge: I think these 2 helpers (maybe slightly more general versions of them) would probably be more appropriate on the `DetailsViewPage`

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,104 @@
...
+                await browser.close();
+            }
+        });
+
+        test('matches snapshot when tabstop is used from adhoc panel', async () => {
+            await popupPage.enableToggleByAriaLabel('Tab stops');
+
+            await targetPage.bringToFront();
+
+            const nativeWidgetHeadingTargetPage = await targetPage.getSelectorElement(TargetPageElementSelectors.targetPageNativeWidgetH1);
```
dbjorge: This can be collapsed to just `await targetPage.clickSelector(TargetPageElementSelectors.targetPageNativeWidgetH1);`

src/tests/end-to-end/tests/details-view/tabstop.test.ts
```diff
@@ -0,0 +1,104 @@
...
+        afterAll(async () => {
+            if (browser) {
+                await browser.close();
+            }
+        });
+
+        test('matches snapshot when tabstop is used from adhoc panel', async () => {
+            await popupPage.enableToggleByAriaLabel('Tab stops');
+
+            await targetPage.bringToFront();
```
dbjorge: Were you able to determine why explicitly `bringToFront()`ing and clicking the elements on the page were necessary? These are steps I would expect should not be required by this test, so if we really need to keep them in, it would be nice to see a comment explaining why (ideally referencing an issue)

src/tests/end-to-end/tests/details-view/__snapshots__/tabstop.test.ts.snap
```diff
@@ -0,0 +1,201 @@
...
+    />
+    <link
+      href="{{EXTENSION_URL_BASE}}/bundle/injected.css"
+      rel="stylesheet"
+      type="text/css"
+    />
+    <div
+      class="insights-container insights-highlight-container insights-tab-stops"
+    >
+      <svg
```
dbjorge: That the <svg> element here contains no `<text>`, `<line>`, or `<ellipse>` elements suggests very strongly that these tests aren't validating what I suspect you intended.

src/tests/end-to-end/tests/details-view/tabstop.test.ts
dbjorge: The "validate in the target page" steps are duplicated between the 2 test cases; consider extracting a helper for it.

src/tests/end-to-end/common/element-identifiers/common-selectors.ts
Shobhit1: Done.

src/tests/end-to-end/tests/details-view/tabstop.test.ts
Shobhit1: Moved.

src/tests/end-to-end/tests/details-view/__snapshots__/tabstop.test.ts.snap
Shobhit1: Tried to do this, but now i need to write more functions to sanitize the visualizations!

.gitignore
```diff
@@ -7,3 +7,5 @@
 *.scss.d.ts
 # Dependency directories
 node_modules/
+.DS_STORE
```
Shobhit1: I just figured them out with Mac as my primary machine. This is something that mac machines produce if you interact with `Finder` on terminal.

src/tests/end-to-end/tests/target-page/tabstop.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+    });
+
+    test('works when tabstop is triggered from adhoc panel', async () => {
+        popupPage = await browser.newPopupPage(targetPage);
+        await popupPage.gotoAdhocPanel();
+        await popupPage.enableToggleByAriaLabel('Tab stops');
+
+        await targetPage.bringToFront();
+        await targetPage.waitForSelector(TargetPageElementSelectors.targetPageNativeWidgetFirstRadio);
```
dbjorge: Is this really necessary? This doesn't perform any interaction with the selector, it just ensures it's present on the page.

src/tests/end-to-end/tests/target-page/tabstop.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        const shadowRoot = await getShadowRootAfterTabStopActivation();
+        await validateTabStopVisualizationOnTargetPage(shadowRoot);
+    });
+
+    async function getShadowRootAfterTabStopActivation(): Promise<ElementHandle<Element>> {
```
dbjorge: I think this helper is combining operations that don't form a logical unit - the name of the method including a conjunction in the middle is a good sign that this is happening. I'd just inline this helper into the test body, now that there are no longer 2 different tests trying to share it.

src/tests/end-to-end/tests/target-page/tabstop.test.ts
Shobhit1: This is basically waiting for the selector to be in focus. I can change it to a `click` and it will mostly increase the reliability of the test.

src/tests/end-to-end/tests/target-page/tabstop.test.ts
Shobhit1: Ok. Makes sense. Will update the test!

src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -125,6 +131,10 @@ export class Page {
         return this.waitForSelector(`#${id}`);
     }
 
+    public async waitForDuration(duration: number): Promise<void> {
```
dbjorge: These tests are no longer using any of these new methods defined on `Page`; waitForNavigation isn't contentious, but `waitForDuration` and `frames` are probably methods we'd prefer tests avoid the temptation to call, so I'd be inclined to leave them out if they aren't necessary, and `takeScreenshot` seems redundant

src/tests/end-to-end/common/element-identifiers/target-page-selectors.ts
```diff
@@ -6,3 +6,14 @@ export const TargetPageInjectedComponentSelectors = {
     issueDialog: '.insights-dialog-main-container',
     insightsVisualizationBox: '.insights-highlight-container',
 };
+
+export const TargetPageElementSelectors = {
+    targetPageNativeWidgetFirstRadio: '#input-radio-1',
+};
+
+export const tabStopShadowDomSelectors = {
```
dbjorge: inconsistent casing between `const`s here

src/tests/end-to-end/common/element-identifiers/popup-page-element-identifiers.ts
```diff
@@ -10,4 +10,5 @@ export const popupPageElementIdentifiers = {
     backToLaunchPadLink: '#back-to-launchpad-link',
     hamburgerMenuButton: '#feedback-collapse-menu-button',
     hamburgerMenu: '.popup-menu',
+    fastPass: '#fast-pass',
```
dbjorge: We aren't using any of the fastpass selectors anymore (here or in `fastpass-selectors.ts`), so I'd be inclined to remove them from the PR; every selector we're encoding in the E2E tests is an opportunity to become out of date with the main body of code, especially if there isn't an active test using them.

src/tests/end-to-end/common/element-identifiers/target-page-selectors.ts
Shobhit1: removed the first one.

src/tests/end-to-end/common/element-identifiers/fastpass-selectors.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export const fastPassSelectors = {
+    tabStopVisulizationStart: '.insights-tab-stops',
```
dbjorge: I think this belongs in `target-page-selectors` - that's where you're looking for it in the test

src/tests/end-to-end/tests/target-page/tabstop.test.ts
Shobhit1: Removed. The test still works fine. Thanks for making me try and remove this. 👍

src/tests/end-to-end/common/page-controllers/page.ts
Shobhit1: removed the extra ones.

src/tests/end-to-end/common/element-identifiers/popup-page-element-identifiers.ts
Shobhit1: removed.

src/tests/end-to-end/tests/target-page/tabstop.test.ts
dbjorge: Are you sure? That doesn't match [the description of `page.waitForSelector` in the Puppeteer API docs](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagewaitforselectorselector-options); it doesn't specify that it scrolls to the element or focuses the element, it just waits for it to be present in the DOM. I think this is probably a no-op as is and I would be surprised if adding a click changed the reliability of the test.

src/tests/end-to-end/common/element-identifiers/target-page-selectors.ts
dbjorge: ...but this new one you've added is still inconsistent with the only other one in the file

src/tests/end-to-end/common/element-identifiers/fastpass-selectors.ts
Shobhit1: agreed;

src/tests/end-to-end/common/element-identifiers/target-page-selectors.ts
Shobhit1: changed.

dbjorge:
Code looks good now, just want to see a few clean e2e-reliability runs and I think this is good to go! :)

PR #993, opened 2019-07-30
test: a11y scans of injected ui components from target page

### LINKED ISSUE #888: Test case: a11y scans of injected UI components from target page
Add axe-core scans run in the context of the target page for
* Issue dialog
* visualization boxes

lisli1:
#### Description of changes

- Added target page selector constants to `TargetPageInjectedComponentSelectors` in `target-page-selectors.ts`
- Added functions `waitForDescendantSelector` and `waitForDescendantSelectorToDisappear` to `page.ts` page controller to wait for a selector (to appear/disappear) within a specific element, such as within shadow root
- Added various functions to `page.ts` to wrap direct `underlyingPage`/`element` calls in `screenshotOnError` for more error details
- Added function `waitForVisualizationBoxesToDisappear` in `target-page.ts` to ensure the visualization boxes are not present before each test case in `visualization-boxes.test.ts`
- Added function `injectScriptFile` to `page.ts` page controller so axe can be run on the target page
- Moved popup page functions for opening adhoc panel and enabling toggles from `adhoc-panel.test.ts` to `popup-page.ts` to reuse them in target page visualization tests (`issue-dialog.test.ts` and `visualization-boxes.test.ts`)
- Injected `axe` into the page when it is undefined, which is necessary for running `scanForAccessibilityIssues` on the target page (axe is defined in the JavaScript context of the extension but not the target page, which are sandboxed by the browser for security)
- Created test to scan for accessibility issues in issue dialog box on target page in `issue-dialog.test.ts`
- Created test to scan for accessibility issues in `Automated checks`, `Headings`, and `Landmarks` visualization boxes on the target page

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #888 
- [x] **N/A** (UI changes only) Added screenshots/GIFs to description above
- [x] **N/A** (UI changes only) Verified usability with NVDA/JAWS


src/tests/end-to-end/common/scan-for-accessibility-issues.ts
```diff
@@ -18,6 +19,14 @@ export interface PrintableResult {
 }
 
 export async function scanForAccessibilityIssues(page: Page, selector: string): Promise<PrintableResult[]> {
+    const axeIsUndefined = await page.evaluate(() => {
+        return (window as any).axe === undefined;
+    });
+
+    if (axeIsUndefined) {
+        await page.injectScriptFile(path.join(__dirname, '../../../../node_modules/axe-core/axe.min.js'));
```
smoralesd: Not sure if the hardcoded path can be improved or not (like in... not have the `../../../../` in front of it) but also not too concerned neither. I guess it's up to you if you want to improve this or not

src/tests/end-to-end/tests/target-page/issue-dialog.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    it('should pass accessibility validation', async () => {
+        await popupPage.gotoAdhocPanel();
+        await popupPage.enableToggleByAriaLabel('Automated checks');
+
+        await targetPage.bringToFront();
+
+        const shadowRoot = await targetPage.getShadowRoot();
+        const issueHighlightLabel = await shadowRoot.$('.failure-label');
```
dbjorge: I don't think this is waiting for the `.failure-label` to appear; I think it's assuming it's already appeared. Probably safer to use a wait-based operation.

`ElementHandle` doesn't expose a `waitForSelector` directly like `Page` does, but you can emulate one with `waitForFunction`; probably would be a good thing to add to our `Page` wrapper, something like:

```ts
async function waitForDescendentSelector(parentElement: ElementHandle<Element>, descendentSelector: string, options?: WaitOptions) {
    options = {
        timeout: DEFAULT_TIMEOUT_CONSTANT,
        ...options
    };
    return await this.underlyingPage.waitForFunction((parent, selector) => parent.querySelector(selector), options, parentElement, descendentSelector);
}
```

src/tests/end-to-end/tests/target-page/issue-dialog.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        await issueHighlightLabel.click();
+
+        await targetPage.waitForSelector('.insights-dialog-main-container');
+
+        const results = await scanForAccessibilityIssues(targetPage, '#accessibility-insights-root-container');
```
dbjorge: Probably more consistent to have this scan the dialog container

src/tests/end-to-end/tests/target-page/issue-dialog.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        await popupPage.gotoAdhocPanel();
+        await popupPage.enableToggleByAriaLabel('Automated checks');
+
+        await targetPage.bringToFront();
+
+        const shadowRoot = await targetPage.getShadowRoot();
+        const issueHighlightLabel = await shadowRoot.$('.failure-label');
+        await issueHighlightLabel.click();
+
+        await targetPage.waitForSelector('.insights-dialog-main-container');
```
dbjorge: We've been keeping the actual selector constants in a separate file so far; I'm kind of on the fence about whether that's still necessary now that we have the controller classes, but we should probably be consistent

src/tests/end-to-end/tests/target-page/visualization-boxes.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    beforeAll(async () => {
+        browser = await launchBrowser({ suppressFirstTimeDialog: true });
+        targetPage = await browser.newTargetPage();
+        popupPage = await browser.newPopupPage(targetPage);
+        await popupPage.gotoAdhocPanel();
+    });
+
+    beforeEach(async () => {
+        await popupPage.disableAllToggles();
```
dbjorge: I'm not sure this is guaranteed to wait for the old visualizations to actually disappear from the page. Consider adding a `TargetPage.waitForVisualizationsToDisappear()` or something (it'll probably want to accept either of "the root container is not present" or "the shadow root is present but empty")

src/tests/end-to-end/tests/target-page/visualization-boxes.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    afterAll(async () => {
+        if (browser) {
+            await browser.close();
+            browser = undefined;
+        }
+    });
+
+    const adhocTools = ['Automated checks', 'Headings', 'Landmarks'];
+
+    it.each(adhocTools)('should pass accessibility validation', async adhocTool => {
```
dbjorge: When you use `.each`, it's good to include some sort of placeholder for the test parameter in the test name to make sure the tests end up with more actionable/unique names; here, if one of these cases failed, the test name would read as `should pass accessibility validation`, `should pass accessibility validation #1`, or `should pass accessibility validation #2`, respectively, but if you used a test name parameter like `'for ad-hoc tool "%s" should pass accessibility validation'`, you'd instead get test names for the individual cases like `for ad-hoc tool "Automated checks" should pass accessibility validation`.

https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout has documentation about how the formatting works.

src/tests/end-to-end/common/page-controllers/popup-page.ts
```diff
@@ -1,12 +1,51 @@
...
+
+        const EXTRA_TOGGLE_OPERATION_TIMEOUT_MS = 5000;
+
+        await this.waitForSelector(enabledToggleSelector, {
+            timeout: DEFAULT_PAGE_ELEMENT_WAIT_TIMEOUT_MS + EXTRA_TOGGLE_OPERATION_TIMEOUT_MS,
+        });
+    }
+
+    public async disableAllToggles(): Promise<void> {
+        const enabledToggles = await this.underlyingPage.$$('button[aria-checked="true"]');
```
dbjorge: When you're making direct underlyingPage/element calls, they should be wrapped in screenshotOnError

src/tests/end-to-end/common/page-controllers/popup-page.ts
```diff
@@ -1,12 +1,51 @@
...
+        const toggleSelector = `button[aria-label="${ariaLabel}"]`;
+        const enabledToggleSelector = `${toggleSelector}[aria-checked=true]`;
+        const disabledToggleSelector = `${toggleSelector}[aria-checked=false]`;
+
+        await this.clickSelector(disabledToggleSelector);
+
+        // The toggles will go through a state where they are removed and replaced with a spinner, then re-added to the page
+        // We intentionally omit looking for the loading spinner because it can be fast enough to not be seen by Puppeteer
+
+        const EXTRA_TOGGLE_OPERATION_TIMEOUT_MS = 5000;
```
dbjorge: `timeouts.ts` has a `DEFAULT_TARGET_PAGE_SCAN_TIMEOUT_MS`; we can probably use that here for the waitForSelector timeout in place of `DEFAULT_PAGE_ELEMENT_WAIT_TIMEOUT_MS + EXTRA_TOGGLE_OPERATION_TIMEOUT_MS`

src/tests/end-to-end/tests/target-page/issue-dialog.test.ts
lisli1: I'm debating if I should put the target page selectors in `common-selectors.ts` or create a new file `target-page-selectors.ts` for them. What are your thoughts?

src/tests/end-to-end/tests/target-page/issue-dialog.test.ts
lisli1: I was trying to get the element that I would click, but I agree that it is safer to wait for it to appear before doing so.

src/tests/end-to-end/tests/target-page/issue-dialog.test.ts
lisli1: Went with `target-page-selectors.ts`

src/tests/end-to-end/common/page-controllers/target-page.ts
```diff
@@ -38,4 +39,11 @@ export class TargetPage extends Page {
         const shadowRoot = await this.getShadowRoot();
         return await formatChildElementForSnapshot(shadowRoot, '#insights-shadow-container');
     }
+
+    public async waitForVisualizationBoxesToDisappear(): Promise<void> {
+        if (document.querySelector(TargetPageInjectedComponentSelectors.insightsRootContainer)) {
```
dbjorge: >document [](start = 12, length = 8)

This is the jsdom document provided by node, not the document of the target page

src/tests/end-to-end/common/scan-for-accessibility-issues.ts
lisli1: Did some research and couldn't find a way to do this for now

PR #998, opened 2019-07-31
Retrieve Snippet from Element

annikakouhia:
#### Description of changes

Up until now, the feature can receive a path, place it in the PathSnippet store, trigger the call to fill the snippet with a hardcoded value, and return the snippet to the UI. This PR replaces the hardcoded value with the logic which goes into the Target Page, finds the element described by the path, and returns the *real* snippet to the store.

Created new file called "Element Finder By Path" which mirrors the work done in "Element Finder By Position", but for finding a snippet given a path versus a path given a location. It checks for iFrames and will use the frame communicator to move a frame deeper if the current frame is an iFrame.

Note: Changed the "await-promise" rule in the TSLint file. Originally didn't let you await anything but a Promise, but now also can await PromiseLike. We can work on changing the test file to a) use a .then() instead of await, or wrap the "testSubject.processRequest(messageStub)" in a Promise.resolve() method. Let me know if you think changing the TSLint file was the wrong move. Research: 
https://palantir.github.io/tslint/rules/await-promise/
https://github.com/Microsoft/TypeScript/issues/8310
https://github.com/search?q=%22await-promise%22%3A+%5Btrue%2C+%22PromiseLike%22%5D&type=Code

#### Pull request checklist


src/injected/element-finder-by-path.ts
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import Q from '../../node_modules/@types/q';
```
Shobhit1: Is this intentional? This doesn't look correct. Can you please check?

src/injected/element-finder-by-path.ts
```diff
@@ -0,0 +1,83 @@
...
+        private readonly htmlElementUtils: HTMLElementUtils,
+        private readonly frameCommunicator: FrameCommunicator,
+        private readonly q: typeof Q,
+    ) {}
+
+    public initialize(): void {
+        this.frameCommunicator.subscribe(ElementFinderByPath.findElementByPathCommand, this.onfindElementByPath);
+    }
+
+    protected onfindElementByPath = (
```
Shobhit1:
```suggestion
    protected onFindElementByPath = (
```

src/injected/element-finder-by-path.ts
```diff
@@ -0,0 +1,83 @@
...
+            result => {
+                responder(result, null, sourceWin);
+            },
+            err => {
+                responder(null, err, sourceWin);
+            },
+        );
+    };
+
+    public processRequest = (message: ElementFinderByPathMessage): Q.IPromise<string> => {
```
Shobhit1: I am just curious if we can avoid using `Q` in these functions? Will `Promise` API not be sufficient for your use case?

annikakouhia: Q officially removed! Thank you!

src/injected/element-finder-by-path.ts
annikakouhia: Whoops--good catch, got it!

src/injected/element-finder-by-path.ts
annikakouhia: Q officially removed! Thank you

PR #1001, opened 2019-08-01
Add Spectron E2E tests for Electron

jalkire:
#### Description of changes
This PR adds (basic) E2E tests for our existing (basic) Electron experience. It uses [Spectron](https://electronjs.org/spectron), Electron's official E2E testing framework.

I am quite interested in alternative suggestions for where these tests should reside if anyone has differing views.

#### Pull request checklist


src/tests/electron/main.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        const electronPath = `${(global as any).rootDir}/drop/electron/extension/bundle/main.bundle.js`;
+        app = new Application({
+            path: Electron as any,
+            args: [electronPath],
+        });
+
+        return app.start();
+    });
+
+    test('test that app opened & loaded site', () => {
```
dbjorge: Jest nowadays has support for `async` test methods that contain `await` statements; you can avoid a lot of the `.then` chaining boilerplate and instead do something like this:

```ts
test('test name', async () => {
    expect(app.browserWindow.isVisible()).toBe(true);
    expect(await app.client.getWindowCount()).toBe(2);
    expect(await app.webContents.getTitle()).toBe('the title');
});
```

src/tests/electron/jest.config.js
```diff
@@ -0,0 +1,19 @@
...
+const rootDir = '../../../';
+const currentDir = '<rootDir>/src/tests/electron';
+
+module.exports = {
+    ...common,
+    displayName: 'electron tests',
+    moduleFileExtensions: ['ts', 'js'],
+    rootDir: rootDir,
+    roots: [currentDir],
+    reporters: ['default', ['jest-junit', { outputDirectory: '.', outputName: '<rootDir>/test-results/e2e/junit-e2e.xml' }]],
```
dbjorge: Should probably use a separate results file

jalkire: Oops, meant to do that. Fixed now.

src/tests/electron/main.test.ts
jalkire: This is a big improvement, thanks! Unfortunately, making the change revealed an issue regarding spectron's typings (or lack thereof). I had to disable a tslint rule to make this work--I added a comment explaining why. Let me know if you think there's an improvement I can make

src/tests/electron/main.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as Electron from 'electron';
+import { Application } from 'spectron';
+
+describe('Electron E2E', () => {
+    let app: Application;
+
+    beforeAll(() => {
+        const electronPath = `${(global as any).rootDir}/drop/electron/extension/bundle/main.bundle.js`;
+        app = new Application({
+            path: Electron as any,
```
waabid: Why the "as any" here?

jalkire: From Spectron's [docs](https://github.com/electron-userland/spectron#application-api):

> If you want to invoke electron directly with your app's main script then you should specify path as electron via electron-prebuilt and specify your app's main script path as the first argument in the args array.

electron-prebuilt is now just electron. The typings don't line up because in all other use cases, you would actually specify a string path. I felt like following the official guidance was worth the `as any` in this case. It's possible our future packaging work will change this if we use a different binary.

PR #1017, opened 2019-08-06
Build mac executables in ADO

karanbirsingh:
#### Description of changes

We build a mac executable in ADO and introduce an electron-builder YML file that allows us to specify the app directory and mac target. This also removes unnecessary files from the resulting application, so we're in the 100-200 MB range.

There are opportunities to share agents between different jobs - it seems it might be easier to do this after our web e2e tests are back in.

#### Pull request checklist

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -64,40 +64,6 @@ jobs:
                 artifactName: 'drop'
             displayName: publish drop
 
-    # CI build only
-    - job: 'publish_electron_windows_linux'
-      #condition: and(always(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI'))
-      pool:
-          vmImage: 'windows-latest'
-      steps:
-          - template: pipeline/install-node-prerequisites.yaml
-
-          - script: yarn dist:electron -wl
-            displayName: create electron distributables
-
-          - task: PublishBuildArtifacts@1
-            inputs:
-                pathtoPublish: '$(System.DefaultWorkingDirectory)/dist'
-                artifactName: $(Agent.JobName)
-            displayName: publish dist
-
-    # CI build only
-    - job: 'publish_electron_mac'
-      #condition: and(always(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI'))
-      pool:
-          vmImage: 'macOS-10.14'
-      steps:
-          - template: pipeline/install-node-prerequisites.yaml
-
-          - script: yarn dist:electron -m
-            displayName: create electron distributables
-
-          - task: PublishBuildArtifacts@1
-            inputs:
-                pathtoPublish: '$(System.DefaultWorkingDirectory)/dist'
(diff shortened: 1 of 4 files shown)


build.yaml
```diff
@@ -104,3 +87,23 @@ jobs:
           - template: pipeline/install-node-prerequisites.yaml
           - template: pipeline/electron/electron-e2e-test-interactive.yaml
           - template: pipeline/electron/electron-e2e-publish-results.yaml
+
+    # CI build only
+    - job: 'publish_electron_windows_linux'
+      condition: and(always(), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI'))
+      pool:
+          vmImage: 'windows-latest'
```
Shobhit1: nit: Do we want to use a fixed `vmImage` rather than `windows-latest` for this?

pipeline/electron/distribute-electron.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+parameters:
+    platforms: '' # electron-builder platforms, e.g. -wl for win/linux, -m for mac, etc
```
Shobhit1: question: can we remove the `platforms` property when we are leaving it `blank('')` or is this required for this `yaml` to be valid?

karanbirsingh: I don't think so based on the docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops

build.yaml
karanbirsingh: I could go either way. One thing to keep in mind is that windows agents change very slowly. Here's the list: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent A change here would imply a major release.

pipeline/electron/distribute-electron.yaml
Shobhit1: Maybe I didn't frame the question that well. My question is whether it is required for this particular `yaml` to have a property called `platforms`, to be valid? If thats the case what is the behavior it exhibits when its blank. 
Sorry if I wasn't clear first time. 👍

build.yaml
Shobhit1: Cool. I dont have a strong preference for either option; just wanted to make sure that we are consistent with our usage of vmImage through out our build pipelines.

pipeline/electron/distribute-electron.yaml
karanbirsingh: If the YAML is blank then the platform defaults to match the job agent. I think this is nice behavior - for example, running it on a Mac agent will create a Mac installer.

Shobhit1: Awesome. Good to know.

PR #1018, opened 2019-08-06
Comma error

annikakouhia:
#### Description of changes

The HTML QuerySelector function uses commas to separate multiple selectors it should consider. Usually, if it cannot find an element it will return null, but in the case that the path *begins* with a comma, the function breaks and the program stops without throwing an error. I've put in a check that will throw the appropriate error if a comma (or semicolon since we use those to separate iframes per user instructions) begins the given path. This will keep the attempt from stopping early without notifying users.

#### Pull request checklist

### DIFF
--- a/src/injected/element-finder-by-path.ts
+++ b/src/injected/element-finder-by-path.ts
@@ -35,6 +35,10 @@ export class ElementFinderByPath {
     };
 
     public processRequest = (message: ElementFinderByPathMessage): PromiseLike<string> => {
+        if (!this.checkSyntax(message.path[0])) {
+            return Promise.resolve('error');
+        }
+
         const element = this.htmlElementUtils.querySelector(message.path[0]) as HTMLElement;
 
         if (element == null) {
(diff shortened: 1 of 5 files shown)


src/injected/element-finder-by-path.ts
```diff
@@ -35,6 +35,10 @@ export class ElementFinderByPath {
     };
 
     public processRequest = (message: ElementFinderByPathMessage): PromiseLike<string> => {
+        if (!this.checkSyntax(message.path[0])) {
+            return Promise.resolve('error');
```
dbjorge: I think it would be a little tough for someone seeing this message to understand why it's happening. `Promise`s support a specific mechanism (`.reject`) for reporting errors; I'd recommend using that, since that's what an async caller will generally be expecting it to work like. That also has the advantage of letting you use an `Error` instead of a `string` to represent the error message, which is nice because it captures the stack that the error occurred at and gives someone looking at the error message file/line information about where the error occurred.

Using it might look something like:

```ts
return Promise.reject(new Error(`Syntax error in selector path "${message.path[0]}"`));
```

annikakouhia: This makes a ton of sense^ I was following a pattern which just resolved with null upon error because there are nested promises and it made it a bit easier to tell what error was coming from what, but giving the error a significant message seems much better. I'll put this in now! Thank you (and thank you for the great explanation!)

annikakouhia: After talking to Wahaj we left out the Error specification message because we aren't going to give users different errors based on the situation (quite yet!)

annikakouhia: But resolves() calls have been transitioned to rejects()

PR #1038, opened 2019-08-07
Failure Instances: First round of CSS

annikakouhia:
#### Description of changes

First round of CSS fixes to match Figma. Changes occur on the Failure Instance Panel as well as the list of Failure Instances that grows on the main page as you add new Failure Instances.

To Note:
--Few tricky testing situations (shallow within a shallow in the Failure Instance Panel Control Test)
--Still working on best way to handle longer snippets (before expand/collapse functionality)

#### Pull request checklist

addendum
Addendum2


src/DetailsView/Styles/detailsview.scss
```diff
@@ -228,6 +231,74 @@ div.insights-file-issue-details-dialog-container {
         margin-top: 30px;
         font-size: 21px;
     }
+    .footer {
+        margin-bottom: 16px;
+        margin-top: 8px;
+        padding-top: 16px;
+        padding-right: 16px;
+        border-top-style: solid;
```
waabid: A couple things:
1. You can simplify this to a single property of: border-top.
2. Can we find the value that maps to this color rather than using the color directly?

src/DetailsView/Styles/detailsview.scss
```diff
@@ -217,6 +217,9 @@ div.insights-file-issue-details-dialog-container {
 
 .failure-instance-panel {
     .observed-failure-textfield {
+        .ms-Label {
```
waabid: Here and the other case below:

can we figure out a way to get this resulting styling without using "ms-Label" as the class?

src/DetailsView/Styles/detailsview.scss
```diff
@@ -228,6 +231,74 @@ div.insights-file-issue-details-dialog-container {
...
+
+.learn-more {
+    color: $communication-primary;
+    text-decoration: none;
+    display: inline-block;
+    margin-bottom: 24px;
+}
+
+.failure-instance-selector {
+    .ms-Label {
```
waabid: As above:

can we get this styling without using internal office fabric classes? while these may be unlikely to change, since they have no obligation to keep that consistent, we open up our CSS to being broken in that scenario. I would only do this if we were certain there is no other way.

src/DetailsView/Styles/detailsview.scss
annikakouhia: Got it! I couldn't find this color in the Colors.scss--Should I try and add it?

src/DetailsView/Styles/detailsview.scss
annikakouhia: Changed it to ":first-child".... let me know if this is any better than using "ms-Label"!

src/DetailsView/Styles/detailsview.scss
annikakouhia: Looked like this one in particular wasn't actually doing anything so I removed it/checked for other instances

src/DetailsView/components/failure-instance-panel-details.tsx
```diff
@@ -13,26 +14,47 @@ export type FailureInstancePanelDetailsProps = {
 };
 
 export const FailureInstancePanelDetails = NamedSFC<FailureInstancePanelDetailsProps>('FailureInstancePanelDetails', props => {
+    const getSnippetInfo = (): JSX.Element => {
+        if (!props.snippet) {
+            return (
+                <div className="failure-instance-snippet-empty-body"> Code snippet will auto-populate based on the CSS selector input.</div>
+            );
+        } else if (props.snippet.startsWith('No code snippet is mapped to:')) {
```
karanbirsingh: We should explore defining the "did I get an error" state separately from the UX string we'd like to use. That way the `path-snippet-controller` and this line both don't know about the UX string so changing it is easier in the future.

src/injected/path-snippet-controller.ts
```diff
@@ -36,19 +36,11 @@ export class PathSnippetController {
 
         this.elementFinderByPath.processRequest(message).then(
             result => {
-                this.sendBackElementFromPath(result, path);
+                this.addCorrespondingSnippet(false, result);
             },
             err => {
-                this.sendBackElementFromPath('error', path);
+                this.addCorrespondingSnippet(true, 'No code snippet is mapped to: ' + path);
```
karanbirsingh: Discussed offline, this is much better than earlier when we had three instances of 'no code snippet...'. We can still improve it further by moving the UX string to the JSX layer in an upcoming PR.

src/DetailsView/Styles/detailsview.scss
annikakouhia: Following up with Chelsea--updates to come in next PR!

src/DetailsView/Styles/detailsview.scss
annikakouhia: Changing to using CSS styling functions found on the MS Fabric site which tell us how to safely do this---In the next PR

src/DetailsView/Styles/detailsview.scss
annikakouhia: Same as above^

src/DetailsView/Styles/detailsview.scss
annikakouhia: Also collapsing margins and padding to single lines in future PR

src/DetailsView/components/failure-instance-panel-details.tsx
```diff
@@ -1,38 +1,61 @@
...
     path: string;
-    snippet: string;
+    snippetValue: string;
+    snippetError: boolean;
     onSelectorChange: (event, value) => void;
     onValidateSelector: (event) => void;
 };
 
 export const FailureInstancePanelDetails = NamedSFC<FailureInstancePanelDetailsProps>('FailureInstancePanelDetails', props => {
+    const getSnippetInfo = (): JSX.Element => {
```
waabid: You can improve this in terms of readability a bit by structuring it differently:

```
if (props.snippetError) {
    // return error snippet here
}

if (!props.snippetValue) {
    // return default auto populate message here

}

return 'whatever the default component is here';
```

waabid: This can also be it's own component; should simplify it but not necessary in this PR 😄

src/DetailsView/Styles/detailsview.scss
```diff
@@ -228,6 +231,70 @@ div.insights-file-issue-details-dialog-container {
...
+.failure-instance-snippet-filled-body {
+    margin-top: 8px;
+    margin-bottom: 24px;
+    padding-top: 12px;
+    padding-bottom: 12px;
+    padding-left: 16px;
+    padding-right: 16px;
+    max-height: 200px;
+    color: $primary-text;
+    background-color: #f3f3f3;
```
Shobhit1: Can you please try and use `background-color` from the theme palette in the `colors.scss`. This avoid us having any contrast accessibility issues when the user switch to `high-contrast` theme.

src/DetailsView/Styles/detailsview.scss
```diff
@@ -228,6 +231,70 @@ div.insights-file-issue-details-dialog-container {
         margin-top: 30px;
         font-size: 21px;
     }
+    .footer {
+        margin-bottom: 16px;
+        margin-top: 8px;
+        padding-top: 16px;
+        padding-right: 16px;
+        border-top: #edebe9 1px solid;
```
Shobhit1: This `border-top` color can be used from the common colors as well. Please check `colors.scss` to find a color that has a `high-contrast` color to go with that.

src/DetailsView/components/assessment-instance-details-column.tsx
```diff
@@ -28,7 +30,10 @@ export class AssessmentInstanceDetailsColumn extends React.Component<AssessmentI
                 ) : null}
                 <div>
                     <TooltipHost content={textContent} calloutProps={{ gapSpace: 0 }}>
-                        <div className="assessment-instance-textContent">{textContent}</div>
+                        <div className="all-content">
+                            {showHeader ? <strong className="instance-header">{this.props.headerText} </strong> : null}
```
Shobhit1: is the `space` after `{this.props.headerText}` intentional?

src/DetailsView/components/failure-instance-panel-details.tsx
```diff
@@ -1,38 +1,61 @@
...
+        if (!props.snippetError && !props.snippetValue) {
+            return (
+                <div className="failure-instance-snippet-empty-body"> Code snippet will auto-populate based on the CSS selector input.</div>
```
Shobhit1: nit: Is the space before `Code ..... ` intentional?

src/common/message-creators/path-snippet-action-message-creator.ts
```diff
@@ -6,9 +6,9 @@ import { ActionMessageDispatcher } from './action-message-dispatcher';
 export class PathSnippetActionMessageCreator {
     constructor(private readonly dispatcher: ActionMessageDispatcher) {}
 
-    public addCorrespondingSnippet = (snippet: string): void => {
+    public addCorrespondingSnippet = (showError: boolean, snippet?: string): void => {
```
Shobhit1: curious: what is the advantage of adding a new parameter to a function as `first` parameter? Just curious. not a blocker; just trying to understand. 😸

PR #1063, opened 2019-08-13
Adding CSS to Failure Instance Panel

annikakouhia:
#### Description of changes

CSS fixes to match Figma. Changes occur on the Failure Instance Panel as well as the list of Failure Instances that grows on the main page as you add new Failure Instances.

To Note:
-- Duplicated JS functions to set the style of two TextFields (Comment in FIPC and CSS Selector in FIPD)
-- Still working with Fer and Chelsea to decide if we want to change the wording on the Validate Selector button
-- Path Snippet Store changes are coming in separate PR
-- Changes on the screen where you click to open the panel (with the list of added failure instances) is NOT behind feature flag (approved by Fer as long as I implemented her thoughts correctly)
-- Footer should have a bar above it, but that's for a Panel and we use a GenericPanel (may be another Fabric version issue as well)
-- Can't add placeholder text to TextFields until we update Fabric Version

#### Pull request checklist

### DIFF
--- a/src/DetailsView/components/failure-instance-panel-control.tsx
+++ b/src/DetailsView/components/failure-instance-panel-control.tsx
@@ -150,6 +150,7 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
     }
 
     private getActionCancelButtons = (): JSX.Element => {
+        const createMode = this.props.actionType === CapturedInstanceActionType.CREATE ? true : false;
         return (
             <div className="footer">
                 <ActionAndCancelButtonsComponent
@@ -157,12 +158,8 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
                     primaryButtonDisabled={
                         this.state.currentInstance.failureDescription === null && this.state.currentInstance.path === null
                     }
-                    primaryButtonText={this.props.actionType === CapturedInstanceActionType.CREATE ? 'Add failed instance' : 'Save'}
-                    primaryButtonOnClick={
-                        this.props.actionType === CapturedInstanceActionType.CREATE
-                            ? this.onAddFailureInstance
-                            : this.onSaveEditedFailureInstance
-                    }
+                    primaryButtonText={createMode ? 'Add failed instance' : 'Save'}
+                    primaryButtonOnClick={createMode ? this.onAddFailureInstance : this.onSaveEditedFailureInstance}
                     cancelButtonOnClick={this.closeFailureInstancePanel}
                 />
             </div>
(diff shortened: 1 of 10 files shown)


annikakouhia:
@waabid I've simplified multiple line styles down to a single line as you suggested (i.e. putting all padding values on one line). I also talked to Chelsea and got colors to be ones from our Colors file. Finally, I put in JS functions to safely set the styling of subcomponents of TextFields as suggested in Fabric documentation. Let me know if you think any of these issues are still actionable!

annikakouhia:
@Shobhit1 
1. Chelsea had accidentally used a few new colors in the Figma, but we spoke and got it figured out so they're all now cleared by Chelsea and from the Colors.scss file!
2. In the Assessment-instance-details-column file, the space after {this.props.headerText} was intentional (I put in a CSS attribute to make sure it wouldn't collapse) so we would have a space between the header ("Comment:") and the value afterward. I could also add the space to the header itself? Making it "Comment: " instead of "Comment:". What do you think?
3. Whoops the other space you caught (before "Code..." in the Failure-Instance-Panel-Details file) was unintentional. I've taken it out.
4. In the Path-snippet-action-message-creator file I added a new parameter first in the method because I originally was thinking it would be required, but the (now second) parameter would be optional so I wanted to order them. I think they'll both be mandatory so less important now :)

Let me know if I should address these changes in any other way! Thank you!

src/DetailsView/components/failure-instance-panel-control.tsx
```diff
@@ -135,29 +137,37 @@ export class FailureInstancePanelControl extends React.Component<FailureInstance
...
+    private getActionCancelButtons = (): JSX.Element => {
+        return (
+            <div className="footer">
                 <ActionAndCancelButtonsComponent
                     isHidden={false}
                     primaryButtonDisabled={
                         this.state.currentInstance.failureDescription === null && this.state.currentInstance.path === null
                     }
-                    primaryButtonText={this.props.actionType === CapturedInstanceActionType.CREATE ? 'Add' : 'Save'}
+                    primaryButtonText={this.props.actionType === CapturedInstanceActionType.CREATE ? 'Add failed instance' : 'Save'}
```
waabid: I've noticed you're checking for this "Create" action type twice; you could isolate this, at least visually, by putting them in local variables (that you reference here) in an if statement before the return statement.

annikakouhia: Got it! Created a new Boolean variable called createMode

annikakouhia: Boolean in and tests are passing! Thank you!

PR #1065, opened 2019-08-13
refactor: replace NodeSelector with Document in preparation for Typescript upgrade

dbjorge:
#### Description of changes

Updating from Typescript 2.9 to 3.5 results in a variety of errors that look like this:

```
ERROR in Q:/repos/accessibility-insights-web/src/common/document-manipulator.ts(4,42):
TS2304: Cannot find name 'NodeSelector'.
```

It looks like https://github.com/microsoft/TSJS-lib-generator/pull/647 was the point where it was removed, merged into a separate `ParentNode` interface. We considered using `ParentNode` instead here, but `Document` seemed closer to the intended usage in most cases and led to the tests working more similarly to the product code than anything else, so that's what I went with.

Replacing the usages with Document is a no-op in Typescript 2.9 (so nice to do as a separate PR, to make the eventual Typescript upgrade PR smaller). It also forces the affected tests to be a little bit better (they create test documents for the cases where the product passes real documents, rather than creating test nodes within a global document and hoping that they act similarly and that no other test touches global document state).

#### Pull request checklist

### DIFF
--- a/src/DetailsView/details-view-renderer.tsx
+++ b/src/DetailsView/details-view-renderer.tsx
@@ -19,7 +19,7 @@ import { PreviewFeatureFlagsHandler } from './handlers/preview-feature-flags-han
 
 export class DetailsViewRenderer {
     private renderer: typeof ReactDOM.render;
-    private dom: NodeSelector & Node;
+    private dom: Document;
     private scopingActionMessageCreator: ScopingActionMessageCreator;
     private inspectActionMessageCreator: InspectActionMessageCreator;
     private issuesSelection: ISelection;
@@ -34,7 +34,7 @@ export class DetailsViewRenderer {
 
     constructor(
         private readonly deps: DetailsViewContainerDeps,
-        dom: NodeSelector & Node,
+        dom: Document,
         renderer: typeof ReactDOM.render,
         scopingActionMessageCreator: ScopingActionMessageCreator,
         inspectActionMessageCreator: InspectActionMessageCreator,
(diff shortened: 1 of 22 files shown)


Shobhit1:
Just out of curiosity and to learn; can you please link the resource or a github issue/document where we can read more about the `NodeSelector` being replaced by `Document` in Typescript 3.0 or NodeSelector being replaced in the new version.

dbjorge:
@Shobhit1:

> Just out of curiosity and to learn; can you please link the resource or a github issue/document where we can read more about the `NodeSelector` being replaced by `Document` in Typescript 3.0 or NodeSelector being replaced in the new version.

Updated PR description accordingly, good suggestion!

waabid:
>  rather than creating test nodes within a global document and hoping that they act similarly and that no other test touches global document state)

Irrelevant to this PR but ... if I'm understanding you correctly: I think this is not true. document.createElement does not modify the actual _global_ document. It's simply a function create elements and does not create it within the document (you need to attach/append items you've created yourself).

dbjorge:
@waabid I agree/understand that `document.createElement` doesn't add the element as a child element of the global document implicitly, but it does still maintain a link to the global document (via `ownerDocument`) which some of our product code does traverse. I don't think any of our current tests are broken because of this behavior, but I think on principle it makes more sense to try to give the tests isolated test data from one another where there is an easy mechanism to do so

waabid:
> @waabid I agree/understand that `document.createElement` doesn't add the element as a child element of the global document implicitly, but it does still maintain a link to the global document (via `ownerDocument`) which some of our product code does traverse. I don't think any of our current tests are broken because of this behavior, but I think on principle it makes more sense to try to give the tests isolated test data from one another where there is an easy mechanism to do so

Ooooh good point!

PR #1066, opened 2019-08-13
Add code to extract Axe.Android results from a mapped port

DaveTryon:
#### Description of changes

Add an API to fetch scan results from a port-attached Android device. For now, the ScanResults class exposes only the information that is needed for the connection dialog. We still need to determine exactly how we'll convert the contents to the data layer.

#### Pull request checklist


src/electron/platform/android/fetch-scan-results.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import axios from 'axios';
+import { ScanResults } from './scan-results';
+
+export function fetchScanResults(port: number): Promise<ScanResults> {
+    return new Promise<any>((resolve, reject) => {
```
Shobhit1: You can simplify this as following:

```ts
export function fetchScanResults(port: number): Promise<ScanResults> {
    return axios
        .get(`http://localhost:${port}/axe/result`)
        .then(response => {
            return new ScanResults(response.data);
        })
        .catch(error => {
            return error;
        });
}
```
It will still return a promise, you dont need to wrap this in a `new Promise` to return a `Promise`.
When you call `fetchScanResults`; at that moment, you can handle the error properly.

Shobhit1: You can also leave out the `catch` part here and just let it propagate to the  `checkPort` function; you are already handling the `Promise` error in that function.

DaveTryon: The suggested code (with the catch) caused the returned promise's reject method to not get called correctly. It worked if I removed the catch completely, but then it's dependent upon the existence of an upstream catch handler, which we can't guarantee. To retain the error handling without making assumptions about the upstream code (remembering that `checkPort` is just for testing), I'm going to stick with returning a `new Promise`

package.json
```diff
@@ -111,6 +111,7 @@
     "dependencies": {
         "applicationinsights-js": "^1.0.20",
         "axe-core": "3.2.2",
+        "axios": "^0.19.0",
```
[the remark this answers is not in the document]
DaveTryon: yarn.lock is now updated

src/electron/platform/android/scan-results.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    constructor(rawData: any) {
+        this.rawData = rawData;
+    }
+
+    public get deviceName(): string {
+        try {
+            return this.rawData.axeContext.axeDevice.name;
+        } catch {
+            return null;
```
DaveTryon: I chose to return null here since it seemed like a bad idea to leak exceptions caused by the private implementation. Happy to throw here if that's the general preference

src/tests/unit/electron/platform/android/scan-results.test.tsx
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jalkire: I don't think this file should be .tsx. Those are usual react/ui related files.

DaveTryon: Renamed to scan-results.test.ts

DaveTryon:
This PR has somehow picked up a bunch of unexpected changes. Closing and creating a new one

DaveTryon:
New PR (identical fork and checkpoint) is at https://github.com/microsoft/accessibility-insights-web/pull/1077

PR #1072, opened 2019-08-14
changing to a try catch

annikakouhia:
#### Description of changes

QuerySelector() is an HTML Element Utils method that takes a path and finds the corresponding element. We use it to retrieve a snippet based on a path. The method returns null if the path doesn't map to a snippet, but we recently realized it will just hang if given a path which follows incorrect syntax. I at first thought it was only in the case that the path started with a comma, but soon realized there were more cases (like when the path started with a number). For this reason, I modified the code to not try and test the path for all potentially failing edge cases (as is in the code base currently to fix commas specifically), but instead wrap the call in a try/catch statement.

#### Pull request checklist

### DIFF
--- a/src/injected/element-finder-by-path.ts
+++ b/src/injected/element-finder-by-path.ts
@@ -35,11 +35,13 @@ export class ElementFinderByPath {
     };
 
     public processRequest = (message: ElementFinderByPathMessage): PromiseLike<string> => {
-        if (!this.checkSyntax(message.path[0])) {
+        let element = null;
+
+        try {
+            element = this.htmlElementUtils.querySelector(message.path[0]) as HTMLElement;
+        } catch (e) {
             return Promise.reject();
         }
-
-        const element = this.htmlElementUtils.querySelector(message.path[0]) as HTMLElement;
 
         if (element == null) {
             return Promise.reject();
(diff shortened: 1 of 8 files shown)


src/injected/element-finder-by-path.ts
```diff
@@ -35,12 +35,14 @@ export class ElementFinderByPath {
     };
 
     public processRequest = (message: ElementFinderByPathMessage): PromiseLike<string> => {
-        if (!this.checkSyntax(message.path[0])) {
+        let element = null;
+
+        try {
+            element = this.htmlElementUtils.querySelector(message.path[0]) as HTMLElement;
+        } catch (e) {
             return Promise.reject();
```
dbjorge: `Promise.reject(e)` would give the caller a little more context if something does go wrong. I think you could actually simplify this and automatically get that behavior if you wrote `processRequest` as an async function, like this:

```ts
    public processRequest = async (message: ElementFinderByPathMessage): Promise<string> => {
        // if this throws an error, it's propogated correctly without you handling it specifically
        const element = this.htmlElementUtils.querySelector(message.path[0]) as HTMLElement;

        if (element == null) {
            // You can throw new Error here with more context than the original Promise.reject()s were giving
            throw new Error(`Element with selector ${firstSelector} not found`);
        }

        // ... etc

        if (element.tagName.toLocaleLowerCase() !== 'iframe' && message.path.length === 1) {
            // In the happy path, you can simplify and avoid the `Promise.resolve` wrapper
            return element.outerHTML;
        }

        // You can still await a `PromiseLike`, so you shouldn't have to refactor frame-communicator to enable using plain Promises here
        return await this.iterateDeeperOnIframe(element, message);
    };
```

PR #1073, opened 2019-08-14
Update electron device-connect-view title bar (and more)

jalkire:
#### Description of changes
Currently, the electron app displays some placeholder text. This PR is the first step on the way towards the app actually resembling our design. It removes the default title bar and frame and adds a new logo, title, and cancel button.
image

#### Pull request checklist

### DIFF
--- a/src/electron/device-connect-view/Styles/deviceconnectview.scss
+++ b/src/electron/device-connect-view/Styles/deviceconnectview.scss
@@ -7,6 +7,16 @@ body {
     margin: 0px;
 }
 
+button {
+    -webkit-app-region: no-drag;
+}
+
+footer {
+    position: fixed;
+    bottom: 0;
+    padding: 32px;
+}
+
 .window-title {
     padding: 6px 9px;
 
--- a/src/electron/device-connect-view/Styles/deviceconnectview.scss
+++ b/src/electron/device-connect-view/Styles/deviceconnectview.scss
@@ -1,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../../common/styles/common.scss';
+
+body {
+    -webkit-app-region: drag;
+    margin: 0px;
+}
+
+.window-title {
+    padding: 6px 9px;
+
+    .header-text {
+        color: $ada-brand-color;
+        font-family: $fontFamily;
+        font-size: 14.51px;
+        font-weight: $fontWeightSemiBold;
+        line-height: 19.3px;
+    }
+
+    .header-icon {
+        float: left;
+        padding: 0px 3px;
+        height: 17px;
+    }
+}
--- a/src/electron/device-connect-view/components/window-footer.tsx
+++ b/src/electron/device-connect-view/components/window-footer.tsx
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { css } from '@uifabric/utilities';
 import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
 import * as React from 'react';
 import { NamedSFC } from '../../../common/react/named-sfc';
@@ -12,10 +11,8 @@ export interface WindowFooterProps {
 export const WindowFooter = NamedSFC<WindowFooterProps>('WindowFooter', (props: WindowFooterProps) => {
     const onCancelClick = () => props.cancelClick();
     return (
-        <footer className="ms-Grid window-footer">
-            <div className={css('ms-Grid-row')}>
-                <DefaultButton className={css('window-footer-button-cancel')} onClick={onCancelClick} text="Cancel"></DefaultButton>
-            </div>
+        <footer className="window-footer">
+            <DefaultButton className="window-footer-button-cancel" onClick={onCancelClick} text="Cancel"></DefaultButton>
         </footer>
     );
 });
--- a/src/electron/device-connect-view/components/window-footer.tsx
+++ b/src/electron/device-connect-view/components/window-footer.tsx
@@ -1,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { css } from '@uifabric/utilities';
+import { remote } from 'electron';
+import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
+import * as React from 'react';
+
+export class WindowFooter extends React.Component {
+    public render(): JSX.Element {
+        return (
+            <footer className="ms-Grid window-footer">
+                <div className={css('ms-Grid-row')}>
+                    <DefaultButton onClick={this.cancelClick} text="Cancel"></DefaultButton>
+                </div>
+            </footer>
+        );
+    }
+
+    private cancelClick(): void {
+        remote.getCurrentWindow().close();
+    }
+}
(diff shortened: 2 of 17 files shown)


src/tests/electron/main.test.ts
```diff
@@ -21,7 +21,7 @@ describe('Electron E2E', () => {
         // so tslint thinks some of the methods do not return promises.
         // tslint:disable: await-promise
         expect(await app.browserWindow.isVisible()).toBe(true);
-        expect(await app.client.getWindowCount()).toBe(2);
+        expect(await app.client.getWindowCount()).toBe(1);
```
jalkire: This PR stops opening the electron devtools. Those were the second window.

src/electron/device-connect-view/Styles/deviceconnectview.scss
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../../common/styles/common.scss';
+
+body {
+    -webkit-app-region: drag;
```
jalkire: This lets you click and drag the window around since the system title bar is disabled.

src/electron/device-connect-view/components/window-footer.tsx
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT License.
+import { css } from '@uifabric/utilities';
+import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
+import * as React from 'react';
+
+export interface WindowFooterProps {
+    cancelClick: () => void;
+}
+
+export class WindowFooter extends React.Component<WindowFooterProps> {
```
Shobhit1: You can make this a pure component rather than a class, since the only thing that this class does is `render()`.
e.g: look at  [blocking-dialog.tsx](https://github.com/microsoft/accessibility-insights-web/blob/master/src/common/components/blocking-dialog.tsx) as an example

jalkire: Changed this and `window-header`. Do you know of any good reading/guidance on the benefits/trade offs of pure components vs new component classes?

Shobhit1: I just figured out that these are called `function components` rather than `pure components`. I apologize for my mistake in my first comment there. The example of `blocking-dialog` is also not a pure component but a function component.
That said, `Pure Components` is another way of defining a React component. Read more about that at: https://reactjs.org/docs/react-api.html#reactpurecomponent.

Internally, React treats `class component` and `function component` as same. In previous versions of react, the react stateless components (or function component) were seen as pure but that has since been changed by React.

For reading about what kind of advantages it provides, you can take a look at: https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc?source=post_page-----231e3fbd7108----------------------

Please let me know if you want to discuss this further. I will also put in something for our engineering discussion next Thursday.

jalkire:
@msft-github-bot delay merging this pull request for 2 hours 30 mins

src/electron/device-connect-view/Styles/deviceconnectview.scss
```diff
@@ -0,0 +1,36 @@
...
+    margin: 0px;
+}
+
+button {
+    -webkit-app-region: no-drag;
+}
+
+footer {
```
dbjorge: Consider using per-component styles for the styles that are specific to a given react component, rather than one big style document for the whole page; that's the model we're trying to move to for newer components. `/src/reports/components/` has a lot of examples of this to take a look at

src/electron/device-connect-view/Styles/deviceconnectview.scss
dbjorge: Is there a technical limitation that requires we put this on `body` and then disable it per-child like the `button` CSS does? I'd have expected to want to put this on `.window-title` instead...

src/electron/device-connect-view/components/window-footer.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export interface WindowFooterProps {
+    cancelClick: () => void;
+}
+
+export const WindowFooter = NamedSFC<WindowFooterProps>('WindowFooter', (props: WindowFooterProps) => {
+    const onCancelClick = () => props.cancelClick();
+    return (
+        <footer className="ms-Grid window-footer">
+            <div className={css('ms-Grid-row')}>
```
dbjorge: You don't have to use the `css` wrapper unless you're combining a collection of styles together

src/electron/device-connect-view/components/window-footer.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { NamedSFC } from '../../../common/react/named-sfc';
+
+export interface WindowFooterProps {
+    cancelClick: () => void;
+}
+
+export const WindowFooter = NamedSFC<WindowFooterProps>('WindowFooter', (props: WindowFooterProps) => {
+    const onCancelClick = () => props.cancelClick();
+    return (
+        <footer className="ms-Grid window-footer">
```
dbjorge: Is it necessary for this component to use the `ms-Grid`/`ms-Grid-row` classes? My understanding is that we usually try to treat the `ms-*` classes as office fabric implementation details; I think ideally we'd avoid using their classes unless there was some very compelling reason for it

dbjorge: (same throughout for other components)

src/electron/device-connect-view/components/window-title.tsx
```diff
@@ -0,0 +1,23 @@
...
+    title: string;
+    children?: JSX.Element;
+}
+
+export const WindowTitle = NamedSFC<WindowTitleProps>('WindowTitle', (props: WindowTitleProps) => {
+    return (
+        <header className="ms-Grid window-title">
+            <div className={css('ms-Grid-row')}>
+                {props.children}
+                <div role="heading" aria-level={1} className="ms-Grid-col header-text">
```
dbjorge: Why a `<div role="heading" aria-level={1}>` rather than just an `<h1>`?

src/electron/device-connect-view/Styles/deviceconnectview.scss
jalkire: Not a technical limitation, just my assumption about what the behavior should be. I will follow up with design about the intended behavior.

src/electron/device-connect-view/components/window-title.tsx
jalkire: No particular reason--modeled this after [header.tsx](https://github.com/microsoft/accessibility-insights-web/blob/master/src/popup/components/header.tsx)

src/electron/device-connect-view/Styles/deviceconnectview.scss
jalkire: Word back is, "When there's a lack of an app header bar, the common pattern is to have the whole screen area dragable." So I will plan to leave this as-is, unless you still have a concern.

src/electron/device-connect-view/components/window-footer.tsx
jalkire: Thanks!

src/electron/device-connect-view/Styles/deviceconnectview.scss
jalkire: Updated to use per-component styles

src/electron/device-connect-view/components/window-footer.tsx
jalkire: Removed wrapper (and class)

src/electron/device-connect-view/components/window-footer.tsx
jalkire: Nope, not necessary. Removed.

src/electron/device-connect-view/components/window-title.tsx
jalkire: Now an h1

PR #1074, opened 2019-08-15
chore: Update typescript to 3.5.3

dbjorge:
#### Description of changes

Updates typescript to the latest stable version.

The notable ancillary changes this required were:

* This update increased the amount of memory used by webpack-all to a tipping point where it began occasionally crashing without finishing. Uses `grunt-concurrent` to have build:all run 3 instances of webpack rather than 1 instance of webpack building 3 configs
* Document.getElementsByTagName was updated to return `HTMLCollectionOf<Element>` instead of `NodeListOf<Element>`; there was a cascading series of type and test updates in response to this
* `MessageEvent`s were generalized to allow events from more than just `Window`s as sources; updated `window-message-handler` with an `e.source as Window`, since it knows it will only register to receive messages from `Window`s.
* Removed a few obsolete `tsconfig.json` entries that were tripping up the new version (in particular, `extendedDiagnostics: true` *actually* causes tsc to emit diagnostics way more verbose than we want in the updated version

To test build reliability with the changes, also introduces a new `build-reliability` pipeline similar to the existing `e2e-reliability` one.

#### Pull request checklist

### DIFF
--- a/src/views/content/content-page.tsx
+++ b/src/views/content/content-page.tsx
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { flatten, toPairs } from 'lodash';
+import { flatten, forEach, toPairs } from 'lodash';
 import * as React from 'react';
 
 import { GuidanceTag } from 'content/guidance-tags';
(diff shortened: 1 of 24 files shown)


Shobhit1:
Ran a `e2e-reliability` build for this branch just to verify once. Waiting for its result.

dbjorge:
Reverted the .snapshot changes; they were caused by accidentally over-updating yarn.lock in an earlier iteration of this PR

Shobhit1:
It looks good to me but if we want to look ahead of the `build` and `e2e` reliability; then my question is how much manual validation does it need?
Maybe a second reviewer is required for this. So, adding that label on this PR.

src/views/content/content-page.tsx
```diff
@@ -36,11 +36,9 @@ export function ContentCreator<M extends HyperlinkDefinitionMap>(
     function mapLinks(markup: Markup): HyperlinkComponentMap<M> {
         const map: Partial<HyperlinkComponentMap<M>> = {};
 
-        if (linkMap) {
-            toPairs(linkMap).forEach(([key, { href, text }]) => {
-                map[key] = ({ children }) => <markup.HyperLink href={href}>{children || text}</markup.HyperLink>;
-            });
-        }
+        forEach(linkMap, (hyperlink: HyperlinkDefinition, key: keyof M) => {
```
waabid: nit: previous version had destructuring hyperlink, might want to re-add that.

PR #1085, opened 2019-08-16
Updating CSS of instance failure details pane

karanbirsingh:
#### Description of changes

The default failure instance comment view in Canary was inadvertently touched in PR #1063 - this PR reverts some of those changes. We will reintroduce the styling updates so they are fully contained inside the feature-flag in a separate PR.

The side effect of this PR is that the feature-flag-on experience no longer has an absolute footer for the add/cancel buttons. We will re-add this soon.

Before:
screenshot of comment field before this PR

After:
screenshot of comment field after this PR, better styled

#### Pull request checklist

- [no] Addresses an existing issue: Fixes #0000

### DIFF
--- a/src/DetailsView/Styles/detailsview.scss
+++ b/src/DetailsView/Styles/detailsview.scss
@@ -224,11 +224,7 @@ div.insights-file-issue-details-dialog-container {
         margin: 20px 0px 20px 0px;
     }
     .header-text {
-        margin-top: 30px;
         font-size: 21px;
-    }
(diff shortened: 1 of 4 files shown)


karanbirsingh:
> There is a little bit of vertical positioning difference between these changes and current prod, and the "Add" button is now called "Add failed instance", but I don't think these are blocking issues.

@peterdur good point, I should mention I ran these by Fer and she was alright

PR #1086, opened 2019-08-16
feature: introducing interfaces for unified data model

Shobhit1:
#### Description of changes

This PR introduces interfaces that are the base for our new Unified data model. The interfaces will house the unified rule instance later.

Some properties, such as `platformInfo` and `toolInfo` in `ScanResults` are currently optional because we are not expecting to populate them as of now but will do so eventually. Some information are also currently not found in all data sources (for example, platform and tool information are not included in results returned from axe-windows automation).

#### Pull request checklist

- [x] Addresses an existing issue: Feat # 1567501
- [ ] Added relevant unit test for your changes. (`yarn test`) - n/a
- [x] Ran precheckin (`yarn precheckin`) - n/a
- [ ] (UI changes only) Added screenshots/GIFs to description above - n/a
- [ ] (UI changes only) Verified usability with NVDA/JAWS - n/a


src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+    version: string;
+}
+
+export interface OSData {
+    name: string;
+    osVersion?: string;
+    userAgent?: string;
+}
+
+export interface ViewPortData {
```
RobGallo: Could we change this to something simple like ScreenProperties: { height: number, width: number}?

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+
+// this is similar to `TestEngine` interface from axe-core
+export interface ScanEngine {
+    name: string;
+    version: string;
+}
+
+export interface OSData {
+    name: string;
+    osVersion?: string;
```
RobGallo: The interface already has "Os" in it; so "os" in "osVersion" is redundant.

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+    userAgent?: string;
```
RobGallo: What type of data is userAgent meant to indicate? Is it the browser? What are the possible multiple meanings?

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface ScanResults {
+    results: UnifiedResultInstance[];
+    platformInfo?: PlatformData;
+    toolInfo?: ToolData;
+}
+
+export interface UnifiedRule {
+    id: string;
+    ruleDescription: string;
```
waabid: Should just be description.

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// this is similar to `TestEngine` interface from axe-core
+export interface ScanEngine {
```
RobGallo: This name doesn't indicate what type of object this is. For reference, OsData indicates that the interface holds data about the OS. Personally, I prefer "properties", but that's up to you.
This is important because without the distinction, it would be easy to look at this interface and expect to see functions which operate the scan engine.

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface UnifiedResultInstance {
+    id: string;
+    evaluation: StoredInstancePropertyBag;
+    identifiers: StoredInstancePropertyBag;
+    descriptors: StoredInstancePropertyBag;
+    resolution: StoredInstancePropertyBag;
+}
+
+export type ResultInstanceStatus = 'pass' | 'fail' | 'incomplete' | 'inapplicable';
```
waabid: I am not sure if a result instance would be "inapplicable" (although a rule may be).

src/common/types/store-data/unified-data-interface.ts
RobGallo: Are we expecting an OS without a version?

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export interface InstancePropertyBag<T> {
+    [property: string]: T;
+}
+
+export type StoredInstancePropertyBag = InstancePropertyBag<any>;
+
+export interface UnifiedResultInstance {
+    id: string;
```
waabid: The result instance should have a status property

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+    platformHeight?: number;
+    platformWidth?: number;
+}
+
+export interface PlatformData {
+    osInfo: OSData;
+    viewPortInfo: ViewPortData;
+}
+
+export interface ToolData {
```
RobGallo: This appears to be a wrapper for the ScanEngine interface. Why not just use the ScanEngine interface?

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+    scanEngine: ScanEngine;
+}
+
+export interface ScanResults {
+    results: UnifiedResultInstance[];
+    platformInfo?: PlatformData;
+    toolInfo?: ToolData;
```
RobGallo: Is it ever possible not to know the platform or the tool?

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export interface UnifiedRule {
```
RobGallo: No how-to-fix? How about a field indicating the standard the rule is based on?

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+    id: string;
+    ruleDescription: string;
+}
+
+export interface InstancePropertyBag<T> {
```
RobGallo: Informational: is the template pattern here following a JavaScript/TypeScript convention? For example, if this is a properly bag, why not make the value type any, rather than creating a separate symbol?

src/common/types/store-data/unified-data-interface.ts
lisli1: I modeled this after the possible arrays in axe-core, but I can remove it if it doesn't make sense to have it here. On that note, can you give an example of when a result instance would be "incomplete"?

src/common/types/store-data/unified-data-interface.ts
RobGallo: What does id indicate?

src/common/types/store-data/unified-data-interface.ts
RobGallo: How is this intended to be used?

waabid: I am not sure to be honest; I think maybe going with unknown might be more appropriate as axe-windows definitely has that.

waabid: There should be a value on the instance that uses this.

src/common/types/store-data/unified-data-interface.ts
waabid: This is to uniquely identify this result instance; I expect us to generate a guid to fill this.

src/common/types/store-data/unified-data-interface.ts
waabid: How to fix is stored at the result instance level (under resolution) since it can be dynamic even under the same rule.

Agreed on having something that links this to guidance/a standard (we already do this for axe-core specific rules in code base anyways)

src/common/types/store-data/unified-data-interface.ts
lisli1: They are optional because we don't get this data from all sources as of now. For example, in [axe-windows/docs/AutomationReference](https://github.com/microsoft/axe-windows/blob/master/docs/AutomationReference.md), we currently don't receive tool or platform information. However, we expect to get this information in the future.

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,55 @@
...
+    osInfo: OSData;
+    viewPortInfo: ViewPortData;
+}
+
+export interface ToolData {
+    scanEngine: ScanEngine;
+}
+
+export interface ScanResults {
+    results: UnifiedResultInstance[];
```
RobGallo: How about UnifiedResults and UnifiedResult and names? I think "instance" is unnecessary. And the results container and the result instances will match.

src/common/types/store-data/unified-data-interface.ts
waabid: Initially I figured that the types might be useful for where we implement them but you're correct, I don't think this is necessary anymore.

We should remove it.

src/common/types/store-data/unified-data-interface.ts
RobGallo: "rule" is redundant.

src/common/types/store-data/unified-data-interface.ts
lisli1: Replaced "incomplete" and "inapplicable" with "unknown"

src/common/types/store-data/unified-data-interface.ts
lisli1: `userAgent` is a string property that contains both the browser name and version. We use this in [AI Web navigator utils](https://github.com/microsoft/accessibility-insights-web/blob/master/src/common/navigator-utils.ts) to get the browser name and version. Technically, I think we could get the name and version from the `userAgent` and populate those properties instead, thus removing the need to store the whole `userAgent`.

src/common/types/store-data/unified-data-interface.ts
lisli1: Thanks for the feedback on "OS" :) I will just remove that prefix from "version"

src/common/types/store-data/unified-data-interface.ts
lisli1: We weren't sure if that made sense for web (and maybe axe-windows), since our "viewport" would actually be a "window." It makes more sense for axe-android to have "ScreenProperties."

lisli1: I will remove "platform" from "platformHeight" and "platformWidth" for now.

src/common/types/store-data/unified-data-interface.ts
lisli1: Would it be directly under `UnifiedResultInstance`, or as a property in `evaluation`?

waabid: directly, I think makes sense

src/common/types/store-data/unified-data-interface.ts
lisli1: Updated to say "properties"

src/common/types/store-data/unified-data-interface.ts
lisli1: I like this clarification and will update it accordingly, unless anyone has any objections

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -0,0 +1,56 @@
...
+export interface UnifiedResult {
+    id: string;
+    status: ResultStatus;
+    evaluation: StoredInstancePropertyBag;
+    identifiers: StoredInstancePropertyBag;
+    descriptors: StoredInstancePropertyBag;
+    resolution: StoredInstancePropertyBag;
+}
+
+export type ResultStatus = 'pass' | 'fail' | 'unknown';
```
karanbirsingh: I think pass/fail/unknown can work for our UI now (if we didn't need to map this to UI constructs we could just take a status at runtime) - it means anything that's not pass/fail from axe-android will map to unknown. axe-windows has warnings, for example. I think it makes sense to keep this list small initially and expand if needed.

src/common/types/store-data/unified-data-interface.ts
Shobhit1: Just as an example: `userAgent` for my current browser is

`"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36"`

src/common/types/store-data/unified-data-interface.ts
Shobhit1: This was to basically have more information about `Tool` which right now is limited to `ScanEngine` information; but might increase to include more properties.
Basically `ToolData` needs to be a much bigger umbrella than `ScanEngine`.

src/common/types/store-data/unified-data-interface.ts
Shobhit1: Thats the intention here, `unknown` is bigger in terms of what it can include; if we want to segregate this further; we can do it  (Expansion of these interfaces in future is the whole idea)

src/common/types/store-data/unified-data-interface.ts
Shobhit1: updated to make it non-optional;

PR #1087, opened 2019-08-19
Add E2E code to check for cancel button

DaveTryon:
#### Description of changes

Add an extra check to the the E2E tests for the electron API so far. We needed access to the WebdriverIO types (version 4.8.0), which undergird the spectron types. The changes to package.json and yarn.lock were both tool-driven in response to the following command:

yarn add webdriverio@4.8.0 -D

#### Pull request checklist

### DIFF
--- a/src/tests/electron/main.test.ts
+++ b/src/tests/electron/main.test.ts
@@ -3,6 +3,7 @@
 import * as Electron from 'electron';
 import { Application } from 'spectron';
 import * as WebdriverIO from 'webdriverio';
+import { CommonSelectors } from './common/element-identifiers/common-selectors';
 
 describe('Electron E2E', () => {
     let app: Application;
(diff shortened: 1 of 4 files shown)


src/tests/electron/main.test.ts
```diff
@@ -23,6 +24,11 @@ describe('Electron E2E', () => {
         expect(await app.browserWindow.isVisible()).toBe(true);
         expect(await app.client.getWindowCount()).toBe(1);
         expect(await app.webContents.getTitle()).toBe('Accessibility Insights for Mobile');
+
+        const cancelButton = '.window-footer-button-cancel';
```
Shobhit1: We do have a `selectors` file to save these selectors to maintain one common place for all the selectors. Not sure about electron, but for web; look at: [common-selectors.ts](https://github.com/microsoft/accessibility-insights-web/blob/master/src/tests/end-to-end/common/element-identifiers/common-selectors.ts).
I believe, we can follow the same pattern for electron.

DaveTryon: Added common-selectors.ts in iteration 2

Issue #1088, opened 2019-08-19
Overview in Assessment is misaligned


ferBonnin:
## Describe the bug

The overview section in assessment is misaligned and the region with the list of tests is not all scrollable

**To Reproduce**

1. Go to assessment
2. Click on overview
3. Scroll down

## Expected behavior

List of tests details should be well aligned, user should be able to fully scroll the list of tests in the left nav

## Screenshots

![Screenshot overview section](https://user-images.githubusercontent.com/9062104/63286345-a3179900-c26c-11e9-82c1-d54de2d031f7.png)

## Extension (please complete the following information)

- OS: [e.g. Windows/Mac] Windows
- Version [e.g. 1.140.1] 2.8
- Browser Version [e.g Chrome 71.0] 76
- Target Page [e.g www.bing.com] any

dbjorge:
#1096 and #1098 fix the issues mentioned here and are merged into master. Canary release with fixes is pending.

dbjorge:
Fixes are available in [Insiders](https://chrome.google.com/webstore/detail/accessibility-insights-fo/nnmjfbmebeckhpejobgjjjnchlljiagp) v2.8.1 - @ferBonnin, can you double check the resolution? :)

ferBonnin:
tested in high contrast mode, zoom 200% different sizes, looks good :)

PR #1098, opened 2019-08-20
fix: prevent redundant assessment overview page scroll bars 

dbjorge:
#### Description of changes

Second half of fixing #1088. A regression between chrome 75 and 76 broke the scrolling behavior of the assessment overview page, resulting in multiple redundant scrollbars and a malformed size on the left nav area. This addresses both issues.

This does *not* address the test details rows being misaligned; see separate PR #1096 for that.

Before:
image

After:
image

#### Pull request checklist

- [x] Addresses an existing issue: Partially addresses #1088

### DIFF
--- a/src/DetailsView/Styles/detailsview.scss
+++ b/src/DetailsView/Styles/detailsview.scss
@@ -343,6 +343,7 @@ div.insights-file-issue-details-dialog-container {
         width: 100%;
     }
     .details-view-command-bar {
+        height: $detailsViewCommandBarHeight;
         width: 100%;
         display: flex;
         justify-content: space-between;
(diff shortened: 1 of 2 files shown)


src/DetailsView/Styles/detailsview.scss
```diff
@@ -351,11 +352,11 @@ div.insights-file-issue-details-dialog-container {
         background-color: $neutral-0;
         font-size: 14px;
         font-weight: normal;
-        border-bottom: 1px solid $neutral-alpha-8;
+        border-bottom: $detailsViewCommandBarBorderHeight solid $neutral-alpha-8;
 
         .details-view-target-page {
             margin-left: 12px;
-            padding: 13px 8px;
+            padding: 8px 8px;
```
dbjorge: This looks unrelated but is important; previously, this padding was expanding the command bar to exceed its height variable, but we need the height to be statically determined for the max-height trickery in the content panes to work.

PR #1104, opened 2019-08-20
fix: issue title generated contains the whole path for elements in iframe

### LINKED ISSUE #548: issue title generated contains the whole path for elements in iframe
## Describe the bug

**To Reproduce**

1. Go to any page with issues under iframes
2. Run automated checks
3. Try to file github issue for an element under iframe

## Expected behavior

The title should only contain the last selector part. for eg: #item1"

## Actual behavior

The title should the whole selector path for eg: "#iframe1;#item1"

The issue is caused by a bug in getSelectorLastPart method in [](https://github.com/Microsoft/accessibility-insights-web/blob/9f2b53c62105967e3b3b27f05d1d27b8e0fd42ba/src/background/issue-details-text-generator.ts ) (we are not handling iframe case)

- OS: [e.g. Windows/Mac]
- Version [e.g. 1.140.1]
- Chrome Version [e.g 71.0]
- Target Page [e.g www.bing.com]

If asked, will you be willing to submit a PR to fix this bug? Yes/No

Did you search for similar issues in our github issues or on [stackoverflow](https://stackoverflow.com/questions/tagged/accessibility-insights)

smoralesd:
#### Description of changes

- fix logic to get last part of the selector to consider iframes
   - basically split the selector using ```';'```
   - apply the original logic to the last element of the splited selector
- update the class that generate the text for the "copy issue details" button to use the ```IssueUrlCreationUtils.getSelectorLastPart```
   - as a result of this, I was able to remove duplicated code

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #548
   - does not apply
   - does not apply

### DIFF
--- a/src/issue-filing/common/issue-filing-url-string-utils.ts
+++ b/src/issue-filing/common/issue-filing-url-string-utils.ts
@@ -21,10 +21,15 @@ const getTitle = (data: CreateIssueDetailsTextData): string => {
 };
 
 const getSelectorLastPart = (selector: string): string => {
-    let selectorLastPart = selector;
-    if (selector.lastIndexOf(' > ') > 0) {
-        selectorLastPart = selector.substr(selector.lastIndexOf(' > ') + 3);
+    const splitedSelector = selector.split(';');
+    let selectorLastPart = splitedSelector[splitedSelector.length - 1];
+
+    const childCombinator = ' > ';
+
+    if (selector.lastIndexOf(childCombinator) > 0) {
+        selectorLastPart = selector.substr(selector.lastIndexOf(childCombinator) + childCombinator.length);
     }
+
     return selectorLastPart;
 };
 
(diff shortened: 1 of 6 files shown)


src/issue-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -21,10 +21,15 @@ const getTitle = (data: CreateIssueDetailsTextData): string => {
...
-    let selectorLastPart = selector;
-    if (selector.lastIndexOf(' > ') > 0) {
-        selectorLastPart = selector.substr(selector.lastIndexOf(' > ') + 3);
+    const splitedSelector = selector.split(';');
+    let selectorLastPart = splitedSelector[splitedSelector.length - 1];
+
+    const childCombinator = ' > ';
+
+    if (selector.lastIndexOf(childCombinator) > 0) {
+        selectorLastPart = selector.substr(selector.lastIndexOf(childCombinator) + childCombinator.length);
```
Shobhit1: nit: instead of manipulating/changing an existing string; does it make sense to return the new string here directly and keep `selectorLastPart` as `const`?
Also does it make sense to store the condition in a variable that gives us some indication on what the condition is checking for? For example:
` const childCombinatorExists = selector.lastIndexOf(childCombinator) > 0`

smoralesd: Changed from `let selectorLastPart` to `const selectorLastPart`.

As per the condition, it is a simple condition, as in, there is only one term, no AND nor OR operations. moving it to a const seems a little overkill.

PR #1109, opened 2019-08-20
Add feature flag for universal cards UI

RobGallo:
#### Description of changes

- Add feature flag for universal cards UI.
- Change AdhocIssuesTestView to return a flagged component

This change is just to create a flag and a place to put the new UI that will be coming soon for the cards view.

When the flag is on, the component returns the new CardsView. When the flag is off, the component returns the IssuesView, which is essentially the old AdhocIssuesTestView.

IssuesView (the old AdhocIssuesTestView) was created to make it super easy to remove it once the feature flag is no longer necessary. 

#### Pull request checklist


src/DetailsView/components/cards-view.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { NamedSFC } from '../../common/react/named-sfc';
+
+export interface CardsViewProps {}
+
+export const CardsView = NamedSFC<CardsViewProps>('CardsView', props => {
+    return (
+        <>
+            <h2>Cards view</h2>
+            <p>Hello, World!</p>
+</>
```
waabid: nit: spacing here seems to be off. Do you have the suggested extensions installed? Prettier should format on file save.

src/DetailsView/components/issues-view.tsx
```diff
@@ -0,0 +1,66 @@
...
+    visualizationScanResultData: VisualizationScanResultData;
+    visualizationConfigurationFactory: VisualizationConfigurationFactory;
+    clickHandlerFactory: DetailsViewToggleClickHandlerFactory;
+    issuesSelection: ISelection;
+    issuesTableHandler: IssuesTableHandler;
+    configuration: VisualizationConfiguration;
+    userConfigurationStoreData: UserConfigurationStoreData;
+}
+
+export const IssuesView = NamedSFC<IssuesViewProps>('IssuesView', ({ children, ...props }) => {
```
waabid: Not sure I like the name 'IssuesView' as it's broad and could still even encompass card semantically.

I'd like to call this something like DetailsListIssuesView or TableIssuesView, which might be more relevant.

PR #1121, opened 2019-08-21
Introduce new actions for platform-agnostic scan result store

karanbirsingh:
#### Description of changes

In order to ensure our automated checks cards UI is platform-agnostic, we will be introducing a new store (separate PR) with types found in `unified-data-interface`: https://github.com/microsoft/accessibility-insights-web/blob/d868fed8c97d0ddc0dff4e82973cc6429c9b7a70/src/common/types/store-data/unified-data-interface.ts#L30

Our overarching goal is to send platform-agnostic data (`null` for now) to the store in parallel with existing scan data already sent by analyzers. This PR creates actions for the upcoming store.

We originally said this, but will defer the analyzer change until later:

> It seems the right way to do this now is to create a new analyzer (`unified-rule-analyzer`). Our analyzers use inheritance & therefore the tests are a bit duplicated with those from `rule-analyzer`. 

#### Pull request checklist


src/tests/unit/tests/injected/analyzers/unified-rule-analyzer.test.ts
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
karanbirsingh: advice appreciated on how to / whether to share code between this and rule-analyzer. Would be nice to use composition instead but the methods in question are protected

src/injected/analyzers/unified-rule-analyzer.ts
```diff
@@ -0,0 +1,36 @@
...
+        protected sendMessageDelegate: (message) => void,
+        protected dateGetter: () => Date,
+        protected telemetryFactory: TelemetryDataFactory,
+        protected readonly visualizationConfigFactory: VisualizationConfigurationFactory,
+    ) {
+        super(config, scanner, scopingStore, sendMessageDelegate, dateGetter, telemetryFactory, visualizationConfigFactory);
+    }
+
+    protected onResolve = (analyzerResult: AxeAnalyzerResult): void => {
+        this.sendScanCompleteResolveMessage(analyzerResult, this.config);
```
karanbirsingh: would like to use `super` here but `onResolve` is an arrow function instead of an instance member function

src/injected/analyzers/unified-rule-analyzer.ts
```diff
@@ -0,0 +1,36 @@
...
+    constructor(
+        protected config: RuleAnalyzerConfiguration,
+        protected scanner: ScannerUtils,
+        protected scopingStore: BaseStore<ScopingStoreData>,
+        protected sendMessageDelegate: (message) => void,
+        protected dateGetter: () => Date,
+        protected telemetryFactory: TelemetryDataFactory,
+        protected readonly visualizationConfigFactory: VisualizationConfigurationFactory,
+    ) {
+        super(config, scanner, scopingStore, sendMessageDelegate, dateGetter, telemetryFactory, visualizationConfigFactory);
```
waabid: Is the constructor needed when you aren't doing anything other than calling the parent constructor?

src/injected/analyzers/unified-rule-analyzer.ts
waabid: This is actually a good indicator of poor design in the past, I think. 😭

I'm fine with this for now but, irrespective of this PR (hence why I'm fine with this), we really should go back and re-factor the analyzers.

karanbirsingh:
looking into accomplishing this via a `postOnResolve` callback in the `rule-analyzer` instead

PR #1123, opened 2019-08-21
Revert device connect state if port number changes

jalkire:
#### Description of changes
Currently in the device connect screen, if a user enters a port, validates it, and then changes the port, the state remains unchanged. That behavior makes it seem like the user has validated their newly entered port, which is not the case. This PR updates the behavior such that whenever the user changes the port value, the state is restored to the original, pre-validation state. There are no actual UI/visual changes.

GIF of state change (same effect for connection success and failure):
electron

#### Pull request checklist

- [-] (UI changes only) Added screenshots/GIFs to description above
- [-] (UI changes only) Verified usability with NVDA/JAWS

### DIFF
--- a/src/electron/device-connect-view/components/device-connect-body.tsx
+++ b/src/electron/device-connect-view/components/device-connect-body.tsx
@@ -8,18 +8,20 @@ import { DeviceConnectFooter } from './device-connect-footer';
 import { DeviceConnectHeader } from './device-connect-header';
 import { DeviceConnectPortEntry } from './device-connect-port-entry';
 
-export type OnConnectedCallback = (isConnected: boolean, hasFailedConnecting: boolean, deviceName?: string) => void;
-export type OnConnectingCallback = () => void;
+export type UpdateStateCallback = (newState: DeviceConnectState, deviceName?: string) => void;
 
+export enum DeviceConnectState {
+    Default,
+    Connecting,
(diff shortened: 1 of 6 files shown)


src/electron/device-connect-view/components/device-connect-body.tsx
```diff
@@ -67,11 +67,11 @@ export class DeviceConnectBody extends React.Component<DeviceConnectBodyProps, D
         });
     };
 
-    private OnConnectedCallback: OnConnectedCallback = (isConnected: boolean, deviceName?: string) => {
+    private OnConnectedCallback: OnConnectedCallback = (isConnected: boolean, hasFailedConnecting: boolean, deviceName?: string) => {
```
DaveTryon: Would an enum for connectionState be clearer than 2 bools? Something like:
connectionUnknown  (equivalent to isConnected = false, hasFailedConnecting = false)
connectionFailed (equivalent to isConnected = false, hasFailedConnection = true)
connectionValidated (equivalent to isConnected = true, hasFailedConnecting = false)

There's no state where isConnected = true and hasFailedConnecting=true

jalkire: Switched to an enum. This allowed me to eliminate an extra callback(!)

PR #1128, opened 2019-08-22
Update UnifiedResult interface fields

RobGallo:
#### Description of changes

- Add 'ruleId' as a first-class citizen of UnifiedResult because all results are expected to have rule ids
- rename 'id' to 'uid' because the latter more clearly describes the fields purpose

#### Pull request checklist

### DIFF
--- a/src/common/types/store-data/unified-data-interface.ts
+++ b/src/common/types/store-data/unified-data-interface.ts
@@ -48,7 +48,6 @@ export interface UnifiedResult {
     uid: string;
     status: ResultStatus;
     ruleId: string;
-    evaluation: StoredInstancePropertyBag;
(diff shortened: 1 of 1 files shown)


src/common/types/store-data/unified-data-interface.ts
```diff
@@ -45,8 +45,9 @@ export interface InstancePropertyBag {
 export type StoredInstancePropertyBag = InstancePropertyBag;
 
 export interface UnifiedResult {
-    id: string;
+    uid: string;
     status: ResultStatus;
+    ruleId: string;
     evaluation: StoredInstancePropertyBag;
```
waabid: We should remove the evaluation property as it will now be redundant/not necessary.

PR #1129, opened 2019-08-22
Send additional message to unified scan results store upon scans

karanbirsingh:
#### Description of changes

This PR sends an additional message to the unified-scan-results store from automated checks in FastPass. The results payload in the message is null for now. In a later PR, the axe-core results will be converted to a platform-agnostic structure consumed by the unified-scan-results store. We also considered creating a new subclass of rule-analyzer to accomplish this but would like to rely less on inheritance in this area. The original message to the visualization-scan-results store has not been changed. 

#### Pull request checklist


src/injected/analyzers/rule-analyzer.ts
```diff
@@ -65,6 +69,7 @@ export class RuleAnalyzer extends BaseAnalyzer {
 
     protected onResolve = (analyzerResult: AxeAnalyzerResult): void => {
         this.sendScanCompleteResolveMessage(analyzerResult, this.config);
+        this.postOnResolve(analyzerResult, this.sendMessageDelegate);
```
waabid: This looks good for the most part but I have a small concern;

It seems odd to me to pipe the send message delegate through this way; ideally, the analyzer shouldn't really know anything about how "postOnResolve" is implemented but this dependency strongly implies that it will be sending a message (and if it isn't, why are we passing the dependency?).

The way to avoid it, in my opinion, would be to make the "sendConvertedResults" into a class object that is given the send message delegate (or you could use bind as well) and then send the object to analyzerProvider as parameter on creation (which can then send it to the specific analyzer).

karanbirsingh: makes sense - updated with a new class whose send function is passed through

PR #1182, opened 2019-08-29
Update heading for collapsible component cards

lisli1:
#### Description of changes

Added heading div for collapsible component cards around ActionButton

#### Pull request checklist

### DIFF
--- a/src/DetailsView/components/cards/collapsible-component-cards.tsx
+++ b/src/DetailsView/components/cards/collapsible-component-cards.tsx
@@ -9,6 +9,7 @@ import { collapsibleContainer, collapsibleContent, collapsibleControl, collapsib
 export interface CollapsibleComponentCardsProps {
     header: JSX.Element;
     content: JSX.Element;
+    headingLevel: number;
     contentClassName?: string;
     containerClassName?: string;
     buttonAriaLabel?: string;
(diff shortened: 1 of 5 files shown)


src/DetailsView/components/cards/collapsible-component-cards.tsx
```diff
@@ -31,6 +32,7 @@ export class CollapsibleComponentCards extends React.Component<CollapsibleCompon
 
     public render(): JSX.Element {
         const showContent = this.state.showContent;
+        const titleContainerProps = this.props.headingLevel ? { role: 'heading', 'aria-level': this.props.headingLevel } : undefined;
```
waabid: We should make a few changes here:

1. Make heading level optional because that is how it behaves (specifying heading level 0 to get that behavior is odd).
2. Make this a strict null check (i.e. ```this.props.headingLevel != null```
3. Make the test case test for when we do not provide a heading level at all.

I think this would make for a more appropriate consumable component. My guess is that the intention of the original collapsible-container was similar.

src/DetailsView/components/cards/collapsible-component-cards.tsx
```diff
@@ -9,6 +9,7 @@ import { collapsibleContainer, collapsibleContent, collapsibleControl, collapsib
 export interface CollapsibleComponentCardsProps {
     header: JSX.Element;
     content: JSX.Element;
+    headingLevel?: number;
```
waabid: You can make it non-optional.

PR #1197, opened 2019-08-30
refactor: store data to follow deps pattern

smoralesd:
#### Description of changes

Looking around some recent changes around the manual instance details panel I notice we still suffer from wiring up store data (in this case, it was the feature flags store data, to show the old/new failure instance panel).

This PR is an attempt to follow a similar pattern to what we have setup with the deps.

Obviously, I didn't refactor for all stores as that could probably change every single component we use on the details view. That means there is a lot of small refactors we need to do to clean this up.

**Notes for the reviewer**: It may help to review the changes commit by commit instead of all of them at the same time

#### Pull request checklist

   - there is not an issue filed for this
   - not affected
   - not affected


waabid:
I'm fine with this change; but I'd like to get some other eyes on it as well. I've not thought about this problem and I'm assuming someone else might have (and may have other potential solutions as well). 😄

dbjorge:
I wonder if the feature flag state, specifically, is sufficiently cross-cutting that even this design becomes more overhead than is ideal to pass them around whenever we need a new flag in a new place that didn't start out using store data otherwise. Perhaps this is one of the few cases sufficiently cross-cutting to consider using a React Context for?

PR #1199, opened 2019-08-30
Improved CSS for cards components

waabid:
#### Description of changes

This adds css from either reports or modifies CSS in details view css to improve the cards UI.

Changes to details view were required in a few cases to adjust the margins correctly (and also to improve the UI otherwise).

image

#### Pull request checklist

### DIFF
--- a/src/DetailsView/Styles/detailsview.scss
+++ b/src/DetailsView/Styles/detailsview.scss
@@ -756,6 +756,7 @@ div.insights-file-issue-details-dialog-container {
         }
         .left-nav,
         .test-step-nav {
+            width: 200px;
             z-index: 100;
             background-color: $neutral-0;
             .index-circle {
@@ -888,16 +889,13 @@ div.insights-file-issue-details-dialog-container {
                 }
                 .issues-table {
                     height: 100%;
-                    width: 100%;
+                    width: calc(100% - 48px);
                     display: flex;
                     flex-direction: column;
                     margin-top: 24px;
-                    > h1,
-                    .issues-table-subtitle,
-                    .details-view-toggle,
-                    .details-disabled-message {
-                        margin-left: 24px;
-                    }
+
+                    margin-left: 24px;
+                    margin-right: 24px;
                     .issues-table-subtitle {
                         color: $secondary-text;
                         max-width: 570px;
@@ -962,7 +960,6 @@ div.insights-file-issue-details-dialog-container {
                     min-height: 0;
                 }
                 .issues-details-list {
-                    margin-left: 24px;
                     display: flex;
                     flex-direction: column;
                     height: 60%;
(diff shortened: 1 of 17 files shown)


src/DetailsView/Styles/detailsview.scss
```diff
@@ -888,16 +889,13 @@ div.insights-file-issue-details-dialog-container {
...
                     display: flex;
                     flex-direction: column;
                     margin-top: 24px;
-                    > h1,
-                    .issues-table-subtitle,
-                    .details-view-toggle,
-                    .details-disabled-message {
-                        margin-left: 24px;
-                    }
+
```
Shobhit1: Any specific reason, not to use single `margin` property like `margin: 24px 24px auto 24px` rather than use separate properties? Just curious.

waabid: Good catch; will change.

waabid:
> I played with the styling and it looks much better ☀️ Can we also add `background-color: $neutral-0;` to `.rule-more-resources` in `rule-resources.scss`? That will improve high contrast usability

Good catch! Updated!

src/DetailsView/Styles/detailsview.scss
```diff
@@ -888,16 +889,10 @@ div.insights-file-issue-details-dialog-container {
                 }
                 .issues-table {
                     height: 100%;
-                    width: 100%;
+                    width: calc(100% - 48px);
```
dbjorge: Can this use a comment/var to indicate where the 48px is coming from?

waabid: Sure; let me look into doing that.

waabid: Done! 😄

src/DetailsView/Styles/detailsview.scss
```diff
@@ -867,7 +867,7 @@ div.insights-file-issue-details-dialog-container {
                 min-height: 0;
                 .target-page-changed,
                 .static-content-details-view {
-                    margin: 24px 24px 0px 24px;
+                    margin: $fastPassRightPanelMarginTop $fastPassRightPanelMarginLeft auto $fastPassRightPanelMarginLeft;
```
dbjorge: Was the second term in this value meant to be $fastPassRightPanelMargin**Right**?

src/DetailsView/Styles/detailsview.scss
dbjorge: Personally, I think the split versions of compound properties like this are more readable; if they're split, I think it's much harder to overlook two of them being transposed and much less likely that a future will need to look up documentation to understand the setting

src/DetailsView/Styles/detailsview.scss
waabid: Yes! Updated!

src/DetailsView/Styles/detailsview.scss
```diff
@@ -866,7 +867,7 @@ div.insights-file-issue-details-dialog-container {
                 min-height: 0;
                 .target-page-changed,
                 .static-content-details-view {
-                    margin: 24px 24px 0px 24px;
+                    margin: $fastPassRightPanelMarginTop $fastPassRightPanelMarginLeft auto $fastPassRightPanelMarginRight;
```
peterdur: Should this be 
```suggestion
                    margin: $fastPassRightPanelMarginTop $fastPassRightPanelMarginRight auto $fastPassRightPanelMarginLeft;
```
instead (top, right, bottom, left)? (also at line 895 below)

(I'm thinking Dan might be right below about preferring the split versions of these properties so the matching is explicit...)

waabid: Updated;

PR #523, opened 2019-04-10
replace underlying indexedDB library with open source idb-keyval package

Shobhit1:
#### Description of changes

Replacing the `idb.ts` with `idb-keyval` package. Its identical in its usage and now we dont have to worry about testing `idb.ts` file (which did not have any unit tests).

#### Pull request checklist

- [n/a] Addresses an existing issue: Fixes #0000
- [n/a] Added relevant unit test for your changes. (`npm run test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] Ran precheckin (`npm run precheckin`)
- [x] Added screenshots/GIFs for UI changes.

PR #1209, opened 2019-09-04
chore: replace underlying idb (indexed DB) library with open source idb-keyval package

Shobhit1:
#### Description of changes

This PR replaces underlying indexed DB implementation with actual `idb-keyval` package.
We are using very basic Indexed DB features i.e. using the DB as a key value store, which the `idb-keyval` package does a great job of.
The current implementation was inspired from the library itself, so it makes sense to use it now that we are not using `uglifyjs` and dont have any conflicting issues with our bundling and the package.

Previous PR #523 was closed because of the issue with Uglify, but now we are able to do this.

#### Pull request checklist


src/common/indexedDB/indexedDB.ts
```diff
@@ -49,12 +49,14 @@ export interface IndexedDBAPI {
...
     private store;
+    private readonly dbName: string = 'default-db';
+    private readonly storeName: string = 'default-store';
 
     constructor() {
         if (!window.indexedDB) {
             throw 'IndexedDB is not available';
         }
-        this.store = new Store();
+        this.store = new Store(this.dbName, this.storeName);
```
waabid: We should be injecting this dependency rather than making this component responsible for it.

Shobhit1: Good suggestion. Changed.

PR #1211, opened 2019-09-04
chore: removed duplicate result-section-title component

waabid:
#### Description of changes

We created this component as a part of the cards feature and needed it's css to be the same. Now that css-modules can exist in our report UI and this component is shared, we can remove the old component that is no longer being referenced and rename the new component to not mention v2.

#### Pull request checklist


src/reports/components/report-sections/collapsible-result-section.tsx
```diff
@@ -3,7 +3,7 @@
 import { NamedSFC } from 'common/react/named-sfc';
 import * as React from 'react';
 
-import { ResultSectionTitleV2 } from '../../../DetailsView/components/cards/result-section-title-v2';
+import { ResultSectionTitle } from '../../../DetailsView/components/cards/result-section-title';
```
peterdur: I think we can remove the .. chain after Sebastian's #982 (here and in two other places below).

PR #1212, opened 2019-09-04
fix:  Part - 3 reduce the number of systems in e2e-reliability builds

Shobhit1:
##### Part 3: Reduce number of systems

Some of the things that this PR does:
- reduce the number of runs from 5 to 3 (to see if memory allocated per build run in ADO plays a role in flakiness on linux machines or not).

#### Pull request checklist


src/tests/end-to-end/tests/target-page/visualization-boxes.test.ts
```diff
@@ -19,21 +19,18 @@ describe('Target Page visualization boxes', () => {
...
-
     afterAll(async () => {
         if (browser) {
             await browser.close();
             browser = undefined;
         }
     });
 
-    const adhocTools = ['Automated checks', 'Headings', 'Landmarks'];
+    const adhocTools = ['Automated checks', 'Headings'];
```
dbjorge: Did you try reordering these and testing for the same class of flakiness? I'm suspicious about whether it's actually an issue with Landmarks as opposed to being "the third one to run" or "the one after Headings".

dbjorge: It would be good to leave a comment here describing why we're only testing some of the tools

src/tests/end-to-end/tests/target-page/visualization-boxes.test.ts
```diff
@@ -19,21 +19,18 @@ describe('Target Page visualization boxes', () => {
         await popupPage.gotoAdhocPanel();
     });
 
-    beforeEach(async () => {
```
dbjorge: What motivated this change?

src/tests/end-to-end/tests/target-page/visualization-boxes.test.ts
Shobhit1: I did not, but thats a good point.

pipeline/e2e-reliability.yaml
```diff
@@ -17,9 +17,3 @@ jobs:
 
     - template: ./e2e-job-per-environment.yaml
       parameters: { jobNameSuffix: _3 }
-
-    - template: ./e2e-job-per-environment.yaml
```
dbjorge: Unless this actually reduces the probability with which a given job fails like the PR description suggests, I would prefer not to make this change. The e2e-reliability build exists for us to diagnose the reliability of the blocking CI/PR builds, not for its own sake; reducing the number of jobs that run doesn't make our end to end tests more reliable, it just makes the e2e-reliability build a worse signal of reliability

src/tests/end-to-end/tests/target-page/visualization-boxes.test.ts
Shobhit1: jest-circus and puppeteer have some inconsistent behavior with `beforeEach` and so I wanted to try and see if removing the code from `beforeEach` in our `e2e` will help or not.

pipeline/e2e-reliability.yaml
Shobhit1: In my opinion, 3 was a good number to run builds/system(OS); it was still testing for reliability but instead of 15, it just makes 9 runs now. tbh I dont think i am making it a worse single of reliability; because if the e2e-reliability with 5 systems always fails, then there is no reliability anyway. This change was also helping me to have a shorter test feedback loop. I can reverse this one to see how other changes perform on 5 systems.

src/tests/end-to-end/tests/target-page/visualization-boxes.test.ts
Shobhit1: Also, I took out `Landmarks` because it showed to be the most failed test in past 15 days. I am still going to experiment with the changes that you mentioned to see whats the reason behind it.

Shobhit1:
Closing this PR as this might not be the best change to make in terms of what it gives (green checks for a build) vs the cost of what it takes away (the worsening of the reliability)

pipeline/e2e-reliability.yaml
dbjorge: I agree that I don't this will make the reliability of individual jobs better or worse.

I hear you that using 9 instead of 15 would make *the e2e-reliability run* pass more often. But what I'm disagreeing with is the premise that *making the e2e-reliability run pass more often* has any value on its own. The e2e-reliability run's purpose is to *detect reliability issues in our CI/PR builds*; reducing from 15 to 9 means that a flakiness issue that only repros 10% of the time will go from "79% chance to be detected" to 61%.

I am confused by the assertion that the change helps to have a shorter test feedback loop; the jobs run in parallel, so I'd expect the exact number to have minimal impact on feedback loop time. Have you hit cases recently where the jobs haven't been able to parallelize for some reason?

PR #1215, opened 2019-09-05
test: Part - 2 high contrast e2e - wait for toggle state to be changed to checked before proceeding further when toggling high contrast mode

Shobhit1:
##### Part - 2: Toggling high contrast mode should wait for selector to be actually toggled/checked.

This PR' is part of PRs which try to reduce some of the inconsistency that we are getting with our e2e reliability build.

PR tries to improve e2e reliability by waiting for selector (checked=true) when we are trying to toggle high contrast on.
This test was chosen because the high contrast mode validation selector (`body.high-contrast-theme`) was not being found during a lot of test failures (as per ADO test analytics). This changes makes sure that we look for that selector only after toggle is checked to true.

#### Pull request checklist


dbjorge:
Updated the title from `fix:` to `test:`; fix implies that it's a user-facing bugfix that someone would validate during a release, test/chore implies that it's not user-facing

PR #1222, opened 2019-09-09
refactor: minor clean up on components/tests

smoralesd:
#### Description of changes

While looking at the electron app code to get familiar before starting a new feature, @JGibson2019 and I did some small clean up refactoring on components and tests.

One thing that caught our attention: `DeviceConnectPortEntry` uses an `async` `onClick` function for the primary `Button` it uses. I think this is the first time I see and async onClick function on our code base. Looking at the tests, we found a call to `setInmediate`. Again, this was odd as is the first time I see this on our code base. Playing around with the test code, we figure out this two things were related (async onClick function and setInmediate on the test). Because the onClick function is async, you need to wait (somehow) for it to finish before asserting on the tests (e.g calling `verifyAll` on a mock) but enzyme doesn't expose onClick handlers in a way that we can directly use `await` or `then`, hence, the use of `setInmediate`. We decided to change this a little bit (after some discussion with @dbjorge) to match our code style better. We introduce a `tick` function to wrap the waiting code (the call to `setInmediate` although, we end up using `setTimeout` with 0 ms delay). This way we can write `await tick();` and not have to either inline a function inside `setInmediate` (which kind of looks odd because of indentation, etc) or create a private function of sort to pass to `setInmediate`.

#### Pull request checklist

   - does not exist
   - not affected
   - not affected

### DIFF
--- a/src/electron/device-connect-view/components/device-connect-port-entry.tsx
+++ b/src/electron/device-connect-view/components/device-connect-port-entry.tsx
@@ -24,17 +24,12 @@ export class DeviceConnectPortEntry extends React.Component<DeviceConnectPortEnt
     }
 
     public render(): JSX.Element {
-        const onPortTextChanged = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
-            this.setState({ isValidateButtonDisabled: !newValue || newValue === '', port: newValue });
-            this.props.updateStateCallback(DeviceConnectState.Default);
-        };
-
         return (
             <div className="device-connect-port-entry">
                 <h3>Android device port number</h3>
                 <MaskedTextField
                     ariaLabel="Port number"
-                    onChange={onPortTextChanged}
+                    onChange={this.onPortTextChanged}
                     placeholder="12345"
                     className="port-number-field"
                     maskChar=""
@@ -51,6 +46,11 @@ export class DeviceConnectPortEntry extends React.Component<DeviceConnectPortEnt
             </div>
         );
     }
+
+    private onPortTextChanged = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
+        this.setState({ isValidateButtonDisabled: !newValue || newValue === '', port: newValue });
+        this.props.updateStateCallback(DeviceConnectState.Default);
+    };
 
     private onValidateClick = async (event: React.MouseEvent<HTMLButtonElement>): Promise<void> => {
         this.setState({ isValidateButtonDisabled: true });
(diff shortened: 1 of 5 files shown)


src/tests/unit/tests/electron/device-connect-view/components/device-connect-port-entry.test.tsx
```diff
@@ -8,92 +8,128 @@ import { shallow } from 'enzyme';
...
+                deviceName: string;
+                description: string;
+            };
+
+            const testCases: ValidatePortTestCase[] = [
+                {
+                    fetch: () => Promise.resolve({ deviceName: 'dev', appIdentifier: 'app' } as ScanResults),
+                    connectState: DeviceConnectState.Connected,
+                    deviceName: 'dev - app',
+                    description: 'fetch succeed',
```
dbjorge: Is this leftover from an earlier iteration?

src/electron/device-connect-view/components/device-connect-port-entry.tsx
```diff
@@ -52,6 +47,11 @@ export class DeviceConnectPortEntry extends React.Component<DeviceConnectPortEnt
         );
     }
 
+    private onPortTextChanged = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
```
Shobhit1: Curious to know the reason/thought process for moving the `onPortTextChanged` function as a private member of the `class` rather than a `closure` in the `render` function.

smoralesd: Consistency within the file and with the rest of the code base. I guess we do have functions inside functions but is not the general practice, it's more the exception.

src/tests/unit/tests/electron/device-connect-view/components/device-connect-port-entry.test.tsx
smoralesd: It's for the "name" of the test... I'm not 100% happy with how does this look in the code.. but I didn't find a better way to handle the test name from the object. I'm open to suggestions.

src/electron/device-connect-view/components/device-connect-port-entry.tsx
Shobhit1: General practice in our code base. Do we want to even check the pros/cons of that approach? I mean, if we are doing something in a particular way and a new way come along; doesn't it give us an opportunity to analyze the earlier choice? I dont have a problem with this approach; but just stating what I think of this change.

smoralesd: Does that sound like something you could bring up on eng discussions?

src/tests/unit/tests/electron/device-connect-view/components/device-connect-port-entry.test.tsx
smoralesd: Updated. I was able to use a table to pass the parameters to the test function.

PR #1225, opened 2019-09-10
refactor: electron external link props

lisli1:
#### Description of changes

Updated `ElectronExternalLinkProps` to contain 'children' property instead of 'text' so its usage is similar to that of `NewTabLink`. This is one of the steps to allow for the reuse of the `TelementryPermissionDialog`, which contains the `telemetryPopupNotice` that uses `NewTabLink`. Will update `telemetryPopupNotice` to be able to use `NewTabLink` or `ElectronExternalLink` in a future PR.

#### Pull request checklist

- [x] Addresses an existing issue: Partially addresses WI # 1596167

### DIFF
--- a/src/tests/unit/tests/electron/device-connect-view/components/__snapshots__/electron-external-link.test.tsx.snap
+++ b/src/tests/unit/tests/electron/device-connect-view/components/__snapshots__/electron-external-link.test.tsx.snap
@@ -2,15 +2,8 @@
 
 exports[`ElectronExternalLinkTest render 1`] = `
 <StyledLinkBase
-  0="t"
-  1="e"
-  2="s"
-  3="t"
-  4="-"
-  5="t"
-  6="e"
-  7="x"
-  8="t"
   onClick={[Function]}
-/>
+>
+  test-text
+</StyledLinkBase>
 `;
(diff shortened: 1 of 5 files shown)


src/tests/unit/tests/electron/device-connect-view/components/__snapshots__/electron-external-link.test.tsx.snap
```diff
@@ -2,8 +2,15 @@
 
 exports[`ElectronExternalLinkTest render 1`] = `
 <StyledLinkBase
+  0="t"
```
Shobhit1: This doesn't look correct. Can you please check this again?

smoralesd: @lisli1 this snapshot tells me we were a little of on how we use the `children` props. You'd probably want to update `ElectronExternalLink` to render something like this:
```typescript
return <Link onClick={onClick}>{props.children}</Link>;
```
(before, we were passing `{...props.children}` inside the first `Link` tag... the `...` is the responsible for this odd-looking snapshot.

PR #1228, opened 2019-09-10
chore: change popup panel selector to an id selector

Shobhit1:
#### Description of changes

We do a `verifyAdhocPanelLoaded` function to verify that adhoc panel loaded; that function was testing for a `.ad-hoc-tools-panel` classname selector and it was causing some errors. 
This is a change to see if changing that selector to the id : `#popup-container` makes any stability improvements.

#### Pull request checklist

### DIFF
--- a/src/tests/end-to-end/common/element-identifiers/popup-page-element-identifiers.ts
+++ b/src/tests/end-to-end/common/element-identifiers/popup-page-element-identifiers.ts
@@ -5,7 +5,7 @@ export const popupPageElementIdentifiers = {
     startUsingProductButton: 'button.start-using-product-button',
     launchPad: '#new-launch-pad',
     launchPadItemTitle: '.launch-pad-item-title',
-    adhocPanel: '.ad-hoc-tools-panel',
+    adhocPanel: '#popup-container',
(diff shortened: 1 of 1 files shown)


src/tests/end-to-end/common/element-identifiers/popup-page-element-identifiers.ts
```diff
@@ -5,7 +5,7 @@ export const popupPageElementIdentifiers = {
     startUsingProductButton: 'button.start-using-product-button',
     launchPad: '#new-launch-pad',
     launchPadItemTitle: '.launch-pad-item-title',
-    adhocPanel: '.ad-hoc-tools-panel',
+    adhocPanel: '#popup-container',
```
smoralesd: This match the div id of `src\popup\popup.html`, i.e. it will match any content inside the popup (Launch Pad, Ad Hoc Panel...)
In other words, finding a div with this id doesn't mean that the ad hoc panel is up. Is that the intention?

Shobhit1: You are right. That was not the intention.

PR #1246, opened 2019-09-13
refactor: using Interpreter across all action creators

smoralesd:
#### Description of changes

While working on implementing the flux architecture for the electron app, I found we should be able to reuse the background-side action creators from the browser extension, but in order to do so, it is needed that all action creators are "created equal" (in the sense of using the `Interpreter` to register callbacks for messages instead of directly using `registerTypeToPayloadCallback` to do so).

This PR update all action creators (and tests) to use `Interpreter` directly.

#### Pull request checklist

- [x] Addresses an existing issue: part of WI # 1597954
   - not affected
   - not affected

### DIFF
--- a/src/tests/unit/tests/background/global-action-creators/user-configuration-action-creator.test.ts
+++ b/src/tests/unit/tests/background/global-action-creators/user-configuration-action-creator.test.ts
@@ -9,14 +9,12 @@ import {
 } from 'background/actions/action-payloads';
 import { UserConfigurationActions } from 'background/actions/user-configuration-actions';
 import { UserConfigurationActionCreator } from 'background/global-action-creators/user-configuration-action-creator';
-import { Interpreter } from 'background/interpreter';
-import { isFunction } from 'lodash';
-import { IMock, It, Mock, Times } from 'typemoq';
+import { IMock, Mock } from 'typemoq';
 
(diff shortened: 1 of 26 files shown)


src/tests/unit/tests/background/global-action-creators/user-configuration-action-creator.test.ts
```diff
@@ -83,14 +81,14 @@ describe('UserConfigurationActionCreator', () => {
...
-        const actionsMock = createActionsMock('setIssueFilingServiceProperty', setIssuFilingServicePropertyMock.object);
+        const setIssueFilingServicePropertyMock = createActionMock(payload);
+        const actionsMock = createActionsMock('setIssueFilingServiceProperty', setIssueFilingServicePropertyMock.object);
         const interpreterMock = createInterpreterMock(Messages.UserConfig.SetIssueFilingServiceProperty, payload);
         const testSubject = new UserConfigurationActionCreator(interpreterMock.object, actionsMock.object);
 
         testSubject.registerCallback();
 
-        setIssuFilingServicePropertyMock.verifyAll();
+        setIssueFilingServicePropertyMock.verifyAll();
```
lisli1: Good catch on the misspelled variable!

PR #1249, opened 2019-09-16
feature: create UserConfigStore in electron view initializer

Shobhit1:
#### Description of changes

Create UserConfigStore in `device-connect-view-initializer.ts` for electron initialization with UserConfig persisted data.

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes AD #1597951

### DIFF
--- a/src/background/background-init.ts
+++ b/src/background/background-init.ts
@@ -7,6 +7,7 @@ import { Store } from 'idb-keyval';
 import { AxeInfo } from '../common/axe-info';
 import { ChromeAdapter } from '../common/browser-adapters/chrome-adapter';
 import { VisualizationConfigurationFactory } from '../common/configs/visualization-configuration-factory';
+import { COMMON_CONSTANTS } from '../common/constants';
 import { EnvironmentInfoProvider } from '../common/environment-info-provider';
 import { IndexedDBAPI, IndexedDBUtil } from '../common/indexedDB/indexedDB';
 import { InsightsFeatureFlags } from '../common/insights-feature-flags';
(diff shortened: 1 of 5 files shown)


src/background/background-init.ts
```diff
@@ -39,13 +40,10 @@ import { UserStoredDataCleaner } from './user-stored-data-cleaner';
...
 declare var window: Window & InsightsFeatureFlags;
 
-const defaultIndexedDBName: string = 'default-db';
-const defaultIndexedDBStoreName: string = 'default-store';
-
 const browserAdapter = new ChromeAdapter();
 const urlValidator = new UrlValidator(browserAdapter);
 const backgroundInitCleaner = new UserStoredDataCleaner(browserAdapter);
-const store = new Store(defaultIndexedDBName, defaultIndexedDBStoreName);
+const store = new Store(COMMON_CONSTANTS.defaultIndexedDBName, COMMON_CONSTANTS.defaultIndexedDBStoreName);
```
peterdur: I was confused for a bit between `Store` in idb-keyval and `Store` in src\common\flux. Maybe we could always import the one from idb-keyval with an alias as done below in device-connect-view-initializer.ts?

src/common/constants.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export const COMMON_CONSTANTS = {
```
peterdur: Should it be the responsibility of the initializers to provide the database and store names? This seems like an internal detail, especially if we're always passing the same values. Would it make sense instead to hide them inside a function that wraps the constructor?

PR #1252, opened 2019-09-16
feature: create UserConfigStore in electron view initializer

Shobhit1:
#### Description of changes

Create UserConfigStore in device-connect-view-initializer.ts for electron initialization with UserConfig persisted data.

#### Pull request checklist

### DIFF
--- a/src/background/background-init.ts
+++ b/src/background/background-init.ts
@@ -7,6 +7,7 @@ import { Store } from 'idb-keyval';
 import { AxeInfo } from '../common/axe-info';
 import { ChromeAdapter } from '../common/browser-adapters/chrome-adapter';
 import { VisualizationConfigurationFactory } from '../common/configs/visualization-configuration-factory';
+import { COMMON_CONSTANTS } from '../common/constants';
 import { EnvironmentInfoProvider } from '../common/environment-info-provider';
 import { IndexedDBAPI, IndexedDBUtil } from '../common/indexedDB/indexedDB';
 import { InsightsFeatureFlags } from '../common/insights-feature-flags';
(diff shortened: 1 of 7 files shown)


src/background/get-persisted-data.ts
```diff
@@ -9,21 +9,19 @@ export interface PersistedData {
...
-            persistedData.assessmentStoreData = assessmentData;
-        }),
-    );
-    promises.push(
-        indexedDBInstance.getItem(IndexedDBDataKeys.userConfiguration).then(userConfig => {
-            persistedData.userConfigurationData = userConfig;
-        }),
-    );
+    const promises: Array<Promise<any>> = dataKeysToFetch.map(key => {
+        return indexedDBInstance.getItem(key).then(assessmentData => {
```
smoralesd: `then(assessmentData)...` the name should be updated, right?

Shobhit1: Updated. Good catch.

src/background/background-init.ts
```diff
@@ -39,19 +41,17 @@ import { UserStoredDataCleaner } from './user-stored-data-cleaner';
...
 declare var window: Window & InsightsFeatureFlags;
 
-const defaultIndexedDBName: string = 'default-db';
-const defaultIndexedDBStoreName: string = 'default-store';
-
 const browserAdapter = new ChromeAdapter();
 const urlValidator = new UrlValidator(browserAdapter);
 const backgroundInitCleaner = new UserStoredDataCleaner(browserAdapter);
-const store = new Store(defaultIndexedDBName, defaultIndexedDBStoreName);
+const store = new Store(COMMON_CONSTANTS.defaultIndexedDBName, COMMON_CONSTANTS.defaultIndexedDBStoreName);
```
peterdur: For a moment I thought this was `Store` from src\common\flux\store.ts rather than from idb-keyval. Below we import with the alias `IndexedDBStore`; would it make sense to do the same here?

src/common/constants.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export const COMMON_CONSTANTS = {
```
peterdur: Does it make sense for initializers to need to provide the database and store names? Seems like this is an implementation detail that could be hidden from consumers, perhaps in a function that wraps the `Store` constructor?

src/background/background-init.ts
Shobhit1: Updated.

src/common/constants.ts
Shobhit1: Updated.

PR #1258, opened 2019-09-17
feat: wiring up telemetry opt in dialog

smoralesd:
#### Description of changes

Wire up store, actions and action creator and render the telemetry opt in dialog base on user configuration from the persisted local storage.

01 - telemetry opt in dialog

#### Pull request checklist

- [x] Addresses an existing issue: completes WI # 1597959

### DIFF
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -11,7 +11,7 @@ const createWindow = () => {
         webPreferences: { nodeIntegration: true },
         frame: false,
         width: 600,
-        height: 500,
+        height: 391,
     });
 
     mainWindow
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -11,7 +11,7 @@ const createWindow = () => {
         webPreferences: { nodeIntegration: true },
         frame: false,
         width: 600,
-        height: 391,
+        height: 500,
     });
 
     mainWindow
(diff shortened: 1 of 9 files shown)


src/electron/main/main.ts
```diff
@@ -11,7 +11,7 @@ const createWindow = () => {
         webPreferences: { nodeIntegration: true },
         frame: false,
         width: 600,
-        height: 391,
+        height: 500,
```
jalkire: 500 seems a little tall for the window. Is there a way we could make the telemetry popup wider rather than make the window taller? That would look closer to the figma. Feel free to ignore this if you're already planning on tuning the style in a future PR.

smoralesd: That's a good point... I'd look into how much work is it to make it wider without affecting the size on the extension. If it's easy, I'll do it as part of this PR, if it's not, I'll have a second PR to fix this.

smoralesd: @jalkire I was able to override the dialog size without modifying the windows size or affecting how the dialog looks like on the extension. Thanks for the feedback

smoralesd:
Marking this PR as **do not merge**
I just found a bug on the links of the telemetry opt in dialog. I'm working on a fix on a different branch and will update this PR to use the fix.

smoralesd:
#1262 is on master now so this PR is ready for re-review. Only change after #1262 is the use of `ElectronLink` instead of `ElectronExternalLink`

PR #1261, opened 2019-09-18
feature: wire up telemetry initialization for electron application

Shobhit1:
#### Description of changes

This PR wires up the telemetry initialization and initial telemetry send (only after the user has a consented/agreed to send telemetry). 

The changes that this PR introduces to the electron application:
- save/persist initial `installationData` in the IndexedDB (this is different from the web extension that uses `chrome's` own local storage. 
- Also make sure that we are only storing `installationData`; ai-web stores other stuff in there as well, which is not required at the moment for electron app.
- introduce `electron-storage-adapter` and `electron-appdata-adapter` to get the required information for telemetry setup
- Link all of the other stuff to initialize telemetry in our electron `initializer` to initialize telemetry

#### Pull request checklist

- [x] Addresses an existing issue: Fixes AD #1600256 


src/background/IndexedDBDataKeys.ts
```diff
@@ -3,4 +3,5 @@
 export class IndexedDBDataKeys {
     public static readonly assessmentStore: string = 'assessmentStoreData';
     public static readonly userConfiguration: string = 'userConfiguration';
+    public static readonly installation: string = 'installation';
```
smoralesd: is possible to make the value `'installationData'` instead? I'm thinking about the code we found that may be able to generate this keys from, let's say, an associated store. Ensure that we follow the same pattern on the string may simplify doing the above in the future, in case we need to, and it's not a big change on this PR neither

src/background/telemetry/telemetry-client-provider.ts
```diff
@@ -7,15 +7,15 @@ import { DateProvider } from '../../common/date-provider';
...
+import { InstallationData } from '../installation-data';
 import { AppInsightsTelemetryClient } from './app-insights-telemetry-client';
 import { ApplicationTelemetryDataFactory } from './application-telemetry-data-factory';
 import { NullTelemetryClient } from './null-telemetry-client';
 import { TelemetryClient } from './telemetry-client';
 import { TelemetryLogger } from './telemetry-logger';
 
 export const getTelemetryClient = (
-    userData: LocalStorageData,
+    installationData: InstallationData,
```
smoralesd: nitpick: it's probably not worth to actually do this but... the signature change of this function is self-contain and makes sense in it's own... so this change could be it's own small PR.

src/electron/adapters/electron-app-data-adapter.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { app } from 'electron';
+import { AppDataAdapter } from '../../common/browser-adapters/app-data-adapter';
+
+export class ElectronAppDataAdapter implements AppDataAdapter {
+    public getVersion(): string {
+        return app.getVersion();
```
smoralesd: what's the value we're actually getting here while running the app? is the value we have on `package.json`. I'm not sure we actually want to use that value or something else. I guess my concern is going to be resolve a little down the path, so this doesn't need to be resolve in this PR

src/electron/adapters/electron-storage-adapter.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { IndexedDBDataKeys } from '../../background/IndexedDBDataKeys';
+import { StorageAdapter } from '../../common/browser-adapters/storage-adapter';
+import { IndexedDBAPI } from '../../common/indexedDB/indexedDB';
+
+export class ElectronStorageAdapter implements StorageAdapter {
+    constructor(private readonly indexedDBInstance: IndexedDBAPI) {}
+    public setUserData(items: Object, callback?: () => void): void {
+        // tslint:disable-next-line:no-floating-promises - top-level entry points are intentionally floating promises
```
smoralesd: isn't worth it to at least log the error instead of disabling the tslint rule?

src/electron/adapters/electron-storage-adapter.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { IndexedDBDataKeys } from '../../background/IndexedDBDataKeys';
+import { StorageAdapter } from '../../common/browser-adapters/storage-adapter';
+import { IndexedDBAPI } from '../../common/indexedDB/indexedDB';
+
+export class ElectronStorageAdapter implements StorageAdapter {
+    constructor(private readonly indexedDBInstance: IndexedDBAPI) {}
```
smoralesd: code seems a little cramped here... some new lines between functions may help (at this point, I'm not even sure why we don't have a tslint/prettier setting for this)

src/electron/adapters/electron-storage-adapter.ts
```diff
@@ -0,0 +1,23 @@
...
+    public setUserData(items: Object, callback?: () => void): void {
+        // tslint:disable-next-line:no-floating-promises - top-level entry points are intentionally floating promises
+        this.indexedDBInstance.setItem(IndexedDBDataKeys.installation, items).then(() => callback);
+    }
+    public getUserData(keys: string | Object | string[], callback: (items: { [key: string]: any }) => void): void {
+        // tslint:disable-next-line:no-floating-promises - top-level entry points are intentionally floating promises
```
smoralesd: same as before with at least logging the error

src/electron/adapters/electron-storage-adapter.ts
```diff
@@ -0,0 +1,23 @@
...
+        this.indexedDBInstance.getItem(IndexedDBDataKeys.installation).then(data => {
+            return callback(data);
+        });
+    }
+    public removeUserData(key: string): void {
+        // tslint:disable-next-line:no-floating-promises - top-level entry points are intentionally floating promises
```
smoralesd: same as before with at least logging the error

src/electron/device-connect-view/device-connect-view-initializer.ts
```diff
@@ -1,28 +1,48 @@
...
 
-const indexedDBDataKeysToFetch = [IndexedDBDataKeys.userConfiguration];
+const indexedDBDataKeysToFetch = [IndexedDBDataKeys.userConfiguration, IndexedDBDataKeys.installation];
 
 // tslint:disable-next-line:no-floating-promises - top-level entry points are intentionally floating promises
 getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch).then((persistedData: Partial<PersistedData>) => {
-    // tslint:disable-next-line:no-unused-variable
+    const installationData: InstallationData = persistedData.installationData;
+
+    // wireup telemetry
```
smoralesd: is this comment necessary?

src/electron/main/main.ts
```diff
@@ -22,6 +22,7 @@ const createWindow = () => {
     mainWindow.on('ready-to-show', () => {
         mainWindow.setMenu(null);
         mainWindow.show();
+        mainWindow.webContents.openDevTools();
```
smoralesd: Just making sure you don't forget to remove this line

src/tests/unit/tests/background/get-persisted-data.test.ts
```diff
@@ -44,14 +51,17 @@ describe('GetPersistedDataTest', () => {
         } as PersistedData);
     });
 
-    it('gets only the userconfigData when its called for', async () => {
-        const indexedDataKeysToFetch = [IndexedDBDataKeys.userConfiguration];
+    it('gets electron specific data keys when passed with those keys', async () => {
```
smoralesd: does the tested object really know about electron/chrome extension difference? I'm not sure making the difference in the test name is correct... I actually think it's a little misleading... the important fact the test is proving is: `getPersistedData` uses the `indexedDataKeysToFetch` param to read the persisted data... who and how that data is used is outside the context of the function itself

smoralesd:
This seems to go in the right direction. 

I would advice though to break this in some smaller PRs as there are definitely some changes subsets that make sense on their own.

src/background/IndexedDBDataKeys.ts
Shobhit1: Yeah. I went with `userConfiguration` way to be consistent; but I can see why the other choice is good.

src/electron/adapters/electron-app-data-adapter.ts
Shobhit1: The value that we get from this is the electron version. https://electronjs.org/docs/api/app#appgetversion

src/electron/adapters/electron-storage-adapter.ts
Shobhit1: The tslint error is not for the error but that I am not `awaiting` the promise. But, I will verify and may be use `logger` to log this.

src/electron/adapters/electron-app-data-adapter.ts
smoralesd: We are not actually changing the version on package.json (it's been 1.0.4 since before we move to git hub).
I think this is fine for now, I'm assuming we're going to have an electron app release feature where we actually take care of the app version.

src/background/IndexedDBDataKeys.ts
lisli1: For consistency, do we want all the values in the `IndexedDBDataKeys` to end with `Data`?

src/electron/device-connect-view/device-connect-view-initializer.ts
lisli1: Will remove comment as it doesn't provide any important info

src/tests/unit/tests/background/get-persisted-data.test.ts
lisli1: Updated to "uses specified data keys to read persisted data"

src/background/IndexedDBDataKeys.ts
Shobhit1: I dont think its a good idea to change previously set IndexedDB keys as this can lead to customer losing their assessment data.

smoralesd: +1 to @Shobhit1 We actually cannot change this keys without having some sort of data migration thing going on... and that looks like out of the scope of this PR and feature.

src/electron/adapters/electron-storage-adapter.ts
smoralesd: so I check this:
```// tslint:disable-next-line:no-floating-promises - top-level entry points are intentionally floating promises```
is not needed if you add a `.catch(...)` at the end of the promise. This was my original point... also, the last part of the comment on tslint disable (```top-level entry points are intentionally floating promises```) doesn't apply on this case, basically this is not a top-level entry point

Comment (Shobhit1) [reply src/electron/adapters/electron-storage-adapter.ts]: Agreed. Updated.

File: src/electron/adapters/electron-storage-adapter.ts [hunk 11]
```diff
@@ -0,0 +1,37 @@
...
+            .catch(error => {
+                this.logger.error('Error occured when trying to set user data:', error);
+            });
+    }
+
+    public getUserData(keys: string | Object | string[], callback: (items: { [key: string]: any }) => void): void {
+        this.indexedDBInstance
+            .getItem(IndexedDBDataKeys.installation)
+            .then(data => {
+                return callback(data);
```
smoralesd: should this `then` call:
```
 .then(data => {
                return callback(data);
```
look the same as the previous one?
``` .then(() => callback)```

Comment (Shobhit1) [reply src/electron/adapters/electron-storage-adapter.ts]: Updated.

File: src/electron/adapters/electron-storage-adapter.ts [hunk 12]
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { IndexedDBDataKeys } from '../../background/IndexedDBDataKeys';
+import { StorageAdapter } from '../../common/browser-adapters/storage-adapter';
+import { IndexedDBAPI } from '../../common/indexedDB/indexedDB';
+import { createDefaultLogger } from '../../common/logging/default-logger';
+import { Logger } from '../../common/logging/logger';
+
+export class ElectronStorageAdapter implements StorageAdapter {
```
smoralesd: I just notice this file doesn't have unit tests and it should

src/tests/unit/electron/adapters/electron-storage-adapter.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        };
+
+        indexedDBInstanceMock
+            .setup(async indexedDB => await indexedDB.setItem(IndexedDBDataKeys.installation, It.isValue(expectedData)))
+            .returns(() => Promise.reject('test-error'))
+            .verifiable(Times.once());
+
+        loggerMock.setup(logger => logger.error('Error occurred when trying to set user data: ', 'test-error')).verifiable(Times.once());
+
+        electronStorageAdapter.setUserData(expectedData);
```
smoralesd: we're not testing the use case where we pass a callback to `setUserData`

src/tests/unit/electron/adapters/electron-storage-adapter.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    let indexedDBInstanceMock: IMock<IndexedDBAPI>;
+    let loggerMock: IMock<Logger>;
+
+    beforeEach(() => {
+        indexedDBInstanceMock = Mock.ofType<IndexedDBAPI>();
+        loggerMock = Mock.ofType<Logger>();
+        electronStorageAdapter = new ElectronStorageAdapter(indexedDBInstanceMock.object, loggerMock.object);
+    });
+
+    it('sets user data with input items', async () => {
```
smoralesd: test scenarios look good to me ... it would be nice to group them by the function we're testing though. like:
```typescript
describe('setUserItem', () => {...});

describe('getUserItem', () => {...});

describe('removeUserItem', () => {...});
```

Shobhit1: Yup. I did that. refactoring more.

PR #1267, opened 2019-09-19
refactor: split telemetry event (browser/electron)

smoralesd:
#### Description of changes

As part of adding the Telemetry Opt-In Dialog on AI-Android, we'll need some new telemetry events. We already have a file with telemetry event names const which make sense for the browser extension. We'd probably want to differentiate browser only from electron only telemetry events, so in this PR:

- Renaming existing telemetry events file (so we know they are for the browser extension)
- Add new telemetry events file for AI-Android with only one event name (to be use in the near future)

#### Pull request checklist

- [x] Addresses an existing issue: Part of WI # 1600256
   - not actual code changes
   - not actual code changes
   - not affected
   - not affected


src/common/electron-telemetry-events.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
peterdur: Should this file be in /src/electron rather than /src/common?

smoralesd: Good catch. Fixed.

PR #1268, opened 2019-09-19
refactor: use the new failed instances section in the automated checks report

waabid:
#### Description of changes

We created a lot of duplicate controls while working on the new cards UI due to a few engineering reasons. This is the final change which actually starts using the new failed instances section (**note: does not include deleting old failed instances section and the older components)**.

The main blockage was the collapsible components in the report and the details view are different. As such, I've made changes in the report generator and the details view initializer to start specifying which collapsible control we can utilize and in the new result section start using the given collapsible control.

I also created a duplicate version of the older collapsible component for the moment (called report-collapsible-component) that consumes the same props as the non-report version of the component allowing for us to use common typing. This also resulted in minor changes in the collapsible script provider (fixed collapsed css) and some other minor css changes.

In the future/after this PR, I would like to do the following:
1) Remove all usages of the older collapsible container component; currently the non applicable and passing instances section use it. It should be easy to replace but I don't want to increase the size of this PR.
2) Remove older versions of the v2 controls created for the cards view.

#### Pull request checklist


src/reports/components/report-sections/collapsible-script-provider.tsx
```diff
@@ -1,17 +1,23 @@
...
-    const collapsibles = doc.getElementsByClassName('collapsible-control');
+    const collapsibles = doc.getElementsByClassName('collapsible-container');
 
     for (let index = 0; index < collapsibles.length; index++) {
         const self = collapsibles.item(index);
         self.addEventListener('click', () => {
-            const content = self.parentElement.nextElementSibling as HTMLElement;
+            const button = self.querySelector('.collapsible-control');
+            const content = button.parentElement.nextElementSibling as HTMLElement;
+            const expanded = button.getAttribute('aria-expanded') === 'false' ? false : true;
```
peterdur: I found some of the use of `expanded` a little tricky to follow, since this code is around toggling the whether the section is expanded. I wonder if it would be clearer to have `expanded` be `expandedBefore` and have `expandedAfter` set to `!expandedBefore`. Then we can set the `aria-expanded` attribute below to `expandedAfter`, set `aria-hidden` to `!expandedAfter`, and remove the `collapsed` class if `expandedAfter`. Together these changes might help the code read more naturally.

waabid: Agreed; I've updated it. let me know what you think.

src/reports/components/report-sections/report-collapsible-container.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+    const contentId = `content-container-${id}`;
+
+    const outerDivClassName = css('collapsible-container', containerClassName, 'collapsed');
+
+    const titleContainerProps = { role: 'heading', 'aria-level': headingLevel };
+
+    return (
+        <div className={outerDivClassName}>
+            <div className="title-container" {...titleContainerProps}>
```
peterdur: When I see `titleContainerProps` my expectation is a props for a `TitleContainer` React component, which confused me for a moment here. Also, I was distracted by the difference between unvarying `className="title-container` being specified inline but unvarying `role="heading"` being injected. Could it be simpler to write this as
```suggestion
            <div className="title-container" role="heading" aria-level={headingLevel}>
```
instead?

src/reports/components/report-sections/collapsible-script-provider.tsx
peterdur: Looks good! I like your choice of `wasExpandedBefore` and `isExpandedAfter`.

src/reports/components/report-sections/report-collapsible-container.tsx
waabid: Done

PR #1269, opened 2019-09-19
Update to axe-core 3.3.2

karanbirsingh:
#### Description of changes

This PR consumes axe-core 3.3.2 and includes:
- updates to rule-to-WCAG mappings, including `scrollable-region-focusable` and `label-content-name-mismatch` as `BestPractice` so they aren't run in the standard set
- an update to our `restoreScroll` option type, which [worked internally](https://github.com/dequelabs/axe-core/commit/d55f3cda4598320f8ee496684a01f8b9e1598f8d?short_path=8be2f64#diff-8be2f642524623d5012c7b2cf4fa4eb9) but wasn't included in axe.d.ts [until now](https://github.com/dequelabs/axe-core/pull/1697/files)
- updated launchpad & guidance snapshot
- update to color guidance to include interactive link to improve keyboard accessibility

example new rule result:
role-img-alt screenshot

updated use-of-color guidance to include WCAG link:
use of color screenshot

#### Pull request checklist

### DIFF
--- a/src/scanner/rule-to-links-mappings.ts
+++ b/src/scanner/rule-to-links-mappings.ts
@@ -23,7 +23,7 @@ export const ruleToLinkConfiguration: DictionaryStringTo<HyperlinkDefinition[]> 
     'input-button-name': [link.WCAG_4_1_2],
     'landmark-unique': [BestPractice],
     'role-img-alt': [link.WCAG_1_1_1],
-    'scrollable-region-focusable': [BestPractice],
+    'scrollable-region-focusable': [link.WCAG_2_1_1, BestPractice],
     'area-alt': [link.WCAG_1_1_1],
     'image-alt': [link.WCAG_1_1_1],
     'image-redundant-alt': [BestPractice],
--- a/src/scanner/rule-to-links-mappings.ts
+++ b/src/scanner/rule-to-links-mappings.ts
@@ -17,6 +17,13 @@ const BestPractice: HyperlinkDefinition = {
 };
 
 export const ruleToLinkConfiguration: DictionaryStringTo<HyperlinkDefinition[]> = {
+    'aria-input-field-name': [link.WCAG_4_1_2],
+    'aria-toggle-field-name': [link.WCAG_4_1_2],
+    'avoid-inline-spacing': [link.WCAG_1_4_12],
+    'input-button-name': [link.WCAG_4_1_2],
+    'landmark-unique': [BestPractice],
+    'role-img-alt': [link.WCAG_1_1_1],
+    'scrollable-region-focusable': [BestPractice],
     'area-alt': [link.WCAG_1_1_1],
     'image-alt': [link.WCAG_1_1_1],
     'image-redundant-alt': [BestPractice],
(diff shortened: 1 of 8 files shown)


src/scanner/rule-to-links-mappings.ts
```diff
@@ -17,6 +17,13 @@ const BestPractice: HyperlinkDefinition = {
 };
 
 export const ruleToLinkConfiguration: DictionaryStringTo<HyperlinkDefinition[]> = {
+    'aria-input-field-name': [link.WCAG_4_1_2],
+    'aria-toggle-field-name': [link.WCAG_4_1_2],
+    'avoid-inline-spacing': [link.WCAG_1_4_12],
+    'input-button-name': [link.WCAG_4_1_2],
+    'landmark-unique': [BestPractice],
+    'role-img-alt': [link.WCAG_1_1_1],
+    'scrollable-region-focusable': [BestPractice],
```
karanbirsingh: `scrollable-region-focusable` should actually be mapped to 2_1_1, but it seems like mapping to `BestPractice` is how we exclude rules from being run, regardless of their original designation from `axe`. Am I missing something?

dbjorge: You're correct. I'd recommend mapping to both for clarity (this will still exclude them), similar to how `landmark-complementary-is-top-level` is marked up below

src/scanner/rule-to-links-mappings.ts
```diff
@@ -17,6 +17,13 @@ const BestPractice: HyperlinkDefinition = {
 };
 
 export const ruleToLinkConfiguration: DictionaryStringTo<HyperlinkDefinition[]> = {
```
karanbirsingh: would like to alphabetize these in a different PR

src/scanner/rule-to-links-mappings.ts
karanbirsingh: Good idea, updated

ferBonnin: this rule had a false positive, we dont want to enable it until we consume the fix too

karanbirsingh:
The `axe-core` change looks good in validation. Checks are failing because:
- `scrollable-region-focusable` fails in the remote build (need to introduce a scrollbar locally) for `adhoc/color/guidance`. We exclude this rule in production because of a false positive edge case [here](https://github.com/dequelabs/axe-core/issues/1701). We need to understand whether our failure is a false positive or not, and update our E2E test or E2E ruleset accordingly. It looks like the color guidance page has no links or interactive elements
- `e2e linux tests 1` shows `Hamburger menu Normal mode should have content matching snapshot` failing because [Chrome failed to start](https://dev.azure.com/ms/accessibility-insights-web/_build/results?buildId=37229&view=ms.vss-test-web.build-test-results-tab&runId=1725124&resultId=100015&paneView=debug)
- One of the mac tabstop tests fails [here](https://dev.azure.com/ms/accessibility-insights-web/_build/results?buildId=37229&view=ms.vss-test-web.build-test-results-tab&runId=1725096&resultId=100258&paneView=debug)

PR #1270, opened 2019-09-20
refactor: NullTelemetryClient to use ApplicationTelemetryDataFactory

lisli1:
#### Description of changes

Currently, `AppInsightsTelemetryClient` takes in an instance of `ApplicationTelemetryDataFactory` as part of its constructor. `ApplicationTelemetryDataFactory` has data specific to the application, such as version, build, and installation ID. Refactoring `NullTelemetryClient` to also take in an instance of `ApplicationTelemetryDataFactory`, since the information provided by it could still be useful.

#### Pull request checklist

### DIFF
--- a/src/tests/unit/tests/background/telemetry/null-telemetry-client.test.ts
+++ b/src/tests/unit/tests/background/telemetry/null-telemetry-client.test.ts
@@ -5,12 +5,15 @@ import { NullTelemetryClient } from 'background/telemetry/null-telemetry-client'
 import { TelemetryBaseData } from 'background/telemetry/telemetry-base-data';
 import { TelemetryLogger } from 'background/telemetry/telemetry-logger';
 import { IMock, Mock, Times } from 'typemoq';
+import { ApplicationTelemetryDataFactory } from '../../../../../background/telemetry/application-telemetry-data-factory';
 
 describe('Null telemetry client', () => {
     let loggerMock: IMock<TelemetryLogger>;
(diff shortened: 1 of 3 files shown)


src/tests/unit/tests/background/telemetry/null-telemetry-client.test.ts
```diff
@@ -5,12 +5,15 @@ import { NullTelemetryClient } from 'background/telemetry/null-telemetry-client'
...
 import { IMock, Mock, Times } from 'typemoq';
+import { ApplicationTelemetryDataFactory } from '../../../../../background/telemetry/application-telemetry-data-factory';
 
 describe('Null telemetry client', () => {
     let loggerMock: IMock<TelemetryLogger>;
+    let telemetryDataFactoryMock: IMock<ApplicationTelemetryDataFactory>;
 
     beforeEach(() => {
         loggerMock = Mock.ofType<TelemetryLogger>();
+        telemetryDataFactoryMock = Mock.ofType<ApplicationTelemetryDataFactory>();
```
smoralesd: It would be nice to update `trackEvent` test to make sure we are including what `getData` returns into the `properties` key of the resulting object.
In this case, adding a mock that behave loosely doesn't directly break the test, plus, not updating the test means we are not proving the tested subject is properly using the `ApplicationTelemetryDataFactory`

smoralesd:
Love the branch name by the way

PR #1271, opened 2019-09-20
refactor: ApplicationTelemetryDataFactory

smoralesd:
#### Description of changes

Currently, `ApplicationTelemetryDataFactory` has a couple of issues:
- it uses `title` as a hard dependency for the `applicationName` property. This means we cannot easily reuse this class on AI-Android, where the current value of `title` doesn't work (as it is `"Accessibility Insights for Web"`... note the **Web** in it)
- it expect an instance of a helper class to get the app version... since the app version doesn't change dynamically while running the app (extension or electron) we can change the param to be the version directly
- same goes for the build param

#### Pull request checklist

- [x] Addresses an existing issue: Part of Telemetry Opt In feature
   - not affected
   - not affected

### DIFF
--- a/src/background/telemetry/telemetry-client-provider.ts
+++ b/src/background/telemetry/telemetry-client-provider.ts
@@ -15,6 +15,7 @@ import { TelemetryClient } from './telemetry-client';
 import { TelemetryLogger } from './telemetry-logger';
 
 export const getTelemetryClient = (
+    applicationName: string,
     installationData: InstallationData,
     appDataAdapter: AppDataAdapter,
     logger: TelemetryLogger,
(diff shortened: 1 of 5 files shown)


src/background/telemetry/telemetry-client-provider.ts
```diff
@@ -15,6 +15,7 @@ import { TelemetryClient } from './telemetry-client';
 import { TelemetryLogger } from './telemetry-logger';
 
 export const getTelemetryClient = (
+    applicationName: string,
```
Shobhit1: curiousity question: Is it intentional to make the `applicationName` parameter as the first param that gets passed to the `getTelemetryClient`?

smoralesd: yes... but it's opinionated... `applicationName` is a `string` i.e. a primitive type, while all the other params are instances of some class/type we have define elsewhere

PR #1279, opened 2019-09-23
Add APP_INITIALIZED telemetry event

smoralesd:
#### Description of changes

Sending APP_INITIALIZED telemetry event when the device connect view is loaded.
We only try to send the event if the user configuration flag `isFirstTime` is false. `isFirstTime` will trigger the telemetry opt in dialog, giving the user a change to opt in or out before we actually send any telemetry, thus is specific event uses both `isFirstTime` and `enableTelemetry` to decide.

#### Pull request checklist

- [x] Addresses an existing issue: Completes WI # 1604608
   - not affected
   - not affected

### DIFF
--- a/src/electron/common/send-app-initialized-telemetry-sender.ts
+++ b/src/electron/common/send-app-initialized-telemetry-sender.ts
@@ -1,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { UserConfigurationStoreData } from 'common/types/store-data/user-configuration-store';
+import { TelemetryEventHandler } from '../../background/telemetry/telemetry-event-handler';
+import { TelemetryEventSource, TriggeredByNotApplicable } from '../../common/extension-telemetry-events';
+import { APP_INITIALIZED } from './electron-telemetry-events';
+
+export const sendAppInitializedTelemetryEvent = (
+    userConfigurationData: UserConfigurationStoreData,
+    telemetryEventHandler: TelemetryEventHandler,
+) => {
+    // not sending the telemetry event the first time the user opens the app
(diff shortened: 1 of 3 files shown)


src/electron/common/send-app-initialized-telemetry-sender.ts
```diff
@@ -0,0 +1,24 @@
...
+import { TelemetryEventSource, TriggeredByNotApplicable } from '../../common/extension-telemetry-events';
+import { APP_INITIALIZED } from './electron-telemetry-events';
+
+export const sendAppInitializedTelemetryEvent = (
+    userConfigurationData: UserConfigurationStoreData,
+    telemetryEventHandler: TelemetryEventHandler,
+) => {
+    // not sending the telemetry event the first time the user opens the app
+    // this way, we let the user opt in or out before sending the first event
+    if (userConfigurationData.isFirstTime) {
```
karanbirsingh: How does this approach compare to calling `publishTelemetry` regardless and relying on the handler to know whether it is enabled?

smoralesd: I think I miss understood the default state for user configuration store. It looks like the default state is;
```typescript
{
    isFirstTime: true,
    enableTelemetry: false,
    ...
}
```
It seems like using the `TelemetryEventHandler` directly should work as expected here. I'll do some validation before updating this PR. 

Thanks for catching this @karanbirsingh

smoralesd: @karanbirsingh 
TL;DR we don't need this PR.

I checked this and the default state of the telemetry client is to not send any telemetry. I checked this by hacking the popup initialization process (we send `'PopupInitialized'` telemetry event when the user opens the popup) and all looks good. 

Thanks again for asking the right question!

PR #1291, opened 2019-09-25
refactor: promisifying storage adapter api

smoralesd:
#### Description of changes

As part of the Telemetry Opt In Dialog on AI-Android, we are re-implementing the storage adapter api using IndexedDB api which is promise-based. This makes for some weird-looking production and test code. 
Additionally, we have talked about making all our code to use promises instead of callback so this makes sense and is a first step into that direction.

In thist PR:
- introduce `webextension-polyfill` & `@types/firefox-webext-browser` dependencies
   - to have promise-based version of chrome api's
- convert all 3 functions from `src/common/browser-adapters/storage-adapter.ts` to be promise-based instead of callback-based
- update all the users of the `StorageAdapter`, this includes `background-init`, a util class to delete deprecated keys from user data (e.g. 'alias' is a deprecated key)
- update webpack config to copy the polyfill javascript file to the drop folders
- update `background.html` to include a script tag so we can have access to the polyfill on runtime

#### Pull request checklist

   - does not exist
   - not affected
   - not affected


src/background/install-data-generator.ts
```diff
@@ -40,7 +40,7 @@ export class InstallDataGenerator {
             year: currentDate.getUTCFullYear(),
         };
 
-        this.storageAdapter.setUserData({ [LocalStorageDataKeys.installationData]: this.installationData });
+        this.storageAdapter.setUserData({ [LocalStorageDataKeys.installationData]: this.installationData }).catch(console.log);
```
dbjorge: If you use console.error instead, the e2e tests should pick it up as a test failure should it occur

src/electron/adapters/electron-storage-adapter.ts
```diff
@@ -11,13 +11,13 @@ import { Logger } from '../../common/logging/logger';
...
 
-    public setUserData(items: Object, callback?: () => void): void {
-        this.indexedDBInstance
-            .setItem(IndexedDBDataKeys.installation, items)
-            .then(() => callback && callback())
-            .catch(error => {
-                this.logger.error('Error occurred when trying to set user data: ', error);
-            });
+    public setUserData(items: Object): Promise<void> {
+        return new Promise((resolve, reject) => {
```
dbjorge: This seems redundant; can it not just return `this.indexedDBInstance.setItem(...)`'s promise directly?

src/tests/unit/electron/adapters/electron-storage-adapter.test.ts
```diff
@@ -13,75 +13,50 @@ describe('ElectronStorageAdapter', () => {
...
     beforeEach(() => {
         indexedDBInstanceMock = Mock.ofType<IndexedDBAPI>();
         loggerMock = Mock.ofType<Logger>();
-        electronStorageAdapter = new ElectronStorageAdapter(indexedDBInstanceMock.object, loggerMock.object);
+        testSubject = new ElectronStorageAdapter(indexedDBInstanceMock.object, loggerMock.object);
     });
 
-    describe('setting user data with electron storage adapter works', () => {
-        it('sets user data with input items', async () => {
+    describe('setUserDataP', () => {
```
dbjorge: Typo (extra P)

src/tests/unit/electron/adapters/electron-storage-adapter.test.ts
```diff
@@ -13,75 +13,50 @@ describe('ElectronStorageAdapter', () => {
...
-
-            loggerMock.setup(logger => logger.error(It.isAny())).verifiable(Times.never());
-
-            electronStorageAdapter.setUserData(expectedData, mockCallback);
-
-            await tick();
-            expect(mockCallback).toHaveBeenCalled();
-            indexedDBInstanceMock.verifyAll();
-        });
+        it('throws', async () => {
```
dbjorge: `it throws` doesn't really tell me enough to understand what requirement this is verifying. `it propagates exceptions from indexedDB as-is`? (similar for other test names)

src/background/install-data-generator.ts
smoralesd: oh... that's a good point!

src/electron/adapters/electron-storage-adapter.ts
smoralesd: so... the return type is different.. while chrome api return `Promise<void>`, `IndexedDBAPI` uses `Promise<boolean>'. I tried to cast it but is not possible and this is one way I can "convert" the promise type to match the interface type... I can see this going the other way (we make the interface have a return type of `Promise<boolean>`) but then the "convertion" will happen on the chrome adapter...

src/tests/unit/electron/adapters/electron-storage-adapter.test.ts
smoralesd: Good catch... this came from when I have both callback-based and promise-base set user data functions. Will update

src/tests/unit/electron/adapters/electron-storage-adapter.test.ts
smoralesd: Great suggestion for the name of the test! Will update

src/electron/adapters/electron-storage-adapter.ts
smoralesd: Updated according to our offline conversation... looks much better the new way.

src/electron/adapters/electron-storage-adapter.ts
```diff
@@ -5,45 +5,27 @@ import { pick } from 'lodash';
...
-                    .filter(internalKey => internalKey !== key)
-                    .reduce((obj, k) => {
-                        obj[key] = data[k];
-                        return obj;
-                    }, {});
-                this.indexedDBInstance.setItem(IndexedDBDataKeys.installation, filtered).catch(e => this.logger.error(e));
-            })
-            .catch(error => this.logger.error('Error occurred when trying to remove user data: ', error));
+    public async removeUserData(key: string): Promise<void> {
+        await this.indexedDBInstance.getItem(IndexedDBDataKeys.installation).then(data => {
```
dbjorge: I think it's a little unexpected to see `await` and `.then(...)` syntaxes being mixed together in one function; maybe use await for both cases?

webpack.config.js
```diff
@@ -6,6 +6,7 @@ const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 const TerserWebpackPlugin = require('terser-webpack-plugin');
 const CaseSensitivePathsPlugin = require('case-sensitive-paths-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const CopyWebpackPlugin = require('copy-webpack-plugin');
```
dbjorge: It looks like this is following the package's [Usage with webpack without bundling guidance](https://github.com/mozilla/webextension-polyfill#usage-with-webpack-without-bundling), but I'm not sure this is necessary since we are already bundling; I'd have thought [Basic Setup with module bundlers](https://github.com/mozilla/webextension-polyfill#basic-setup-with-module-bundlers) would be what we'd want to follow (ie, just `import * as browser from 'webextension-polyfill` in chrome-adapter.ts and let webpack deal with bundling like any other package). Did you encounter issues with the usual approach?

smoralesd: I tried that first but typings doesn't play nice with that approach... I can give it a second look though

src/electron/adapters/electron-storage-adapter.ts
smoralesd: I was debating about that... you just tilt the balance into async/await syntax. Updated

webpack.config.js
smoralesd: Got it working!

src/background/background-init.ts
```diff
@@ -37,21 +36,22 @@ import { getTelemetryClient } from './telemetry/telemetry-client-provider';
...
 const indexedDBInstance: IndexedDBAPI = new IndexedDBUtil(getIndexedDBStore());
 const indexedDBDataKeysToFetch = [IndexedDBDataKeys.assessmentStore, IndexedDBDataKeys.userConfiguration];
 
-backgroundInitCleaner.cleanUserData(deprecatedStorageDataKeys);
+cleanKeysFromStorage(browserAdapter, deprecatedStorageDataKeys).catch(reason => console.error('error while cleaning user data: ', reason));
 
-// tslint:disable-next-line:no-floating-promises - top-level entry points are intentionally floating promises
-getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch).then((persistedData: PersistedData) => {
-    browserAdapter.getUserData(storageDataKeys, (userData: LocalStorageData) => {
+const persistedDataPromise = getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch);
```
devaradhanm: do we have a work item to clean up logic in background-init & just use the file for dependency injection 
& not have any logic?

src/background/install-data-generator.ts
```diff
@@ -40,7 +40,7 @@ export class InstallDataGenerator {
             year: currentDate.getUTCFullYear(),
         };
 
-        this.storageAdapter.setUserData({ [LocalStorageDataKeys.installationData]: this.installationData });
+        this.storageAdapter.setUserData({ [LocalStorageDataKeys.installationData]: this.installationData }).catch(console.error);
```
devaradhanm: so, we don't want to crash if failed? same applies for other cases

src/background/background-init.ts
devaradhanm: also, we don't want to make this sequential?

smoralesd: - We don't have work item to clean up background init
- Not sure about the sequential part... it looks like we could but I think that will make the code inside the `then` function more confusing... once you have both `persistedData` and `userData` all the rest of the logic looks simple. Also, using `all` to wait for both promises allow to rearrange initialization steps every way we need/want to.

src/background/install-data-generator.ts
smoralesd: We were not doing it before... we may want to do something about it but I think that's not in scope for promisifying the API

src/background/background-init.ts
```diff
@@ -37,21 +36,22 @@ import { getTelemetryClient } from './telemetry/telemetry-client-provider';
...
+const userDataPromise = browserAdapter.getUserData(storageDataKeys);
```
devaradhanm: should we get user data after the deprecated keys are removed since removing deprecated keys is a async call? (I know you didnt change the behavior. it could be handled in a later pr)

src/common/browser-adapters/chrome-adapter.ts
```diff
@@ -126,16 +127,16 @@ export class ChromeAdapter implements BrowserAdapter, StorageAdapter, CommandsAd
         chrome.runtime.sendMessage(message);
```
devaradhanm: should this be browser?

src/common/browser-adapters/chrome-adapter.ts
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { browser } from 'webextension-polyfill-ts';
```
devaradhanm: have you validated this change on edge? as per their github page, they don't support edge. - https://github.com/mozilla/webextension-polyfill#supported-browsers

devaradhanm: they havn't validated after migrating to chrome based edge

src/background/background-init.ts
smoralesd: That's a good point!
I think is a minor improvement though is I did not find any other reference to the deprecated keys expect for the removal call. I'll leave that for a future PR.

src/common/browser-adapters/chrome-adapter.ts
smoralesd: I think they refer to old edge but I will properly validate and come back with what I find.

smoralesd: Tested on **Microsoft Edge Version 79.0.279.0 (Official build) dev (64-bit)** and it works as expected!

PR #1301, opened 2019-09-26
chore: manually update license-check-and-add to 3.0.3 

Shobhit1:
#### Description of changes

This PR manually updates license-check-and-add to 3.0.3 and makes the required changes for it to work. Thanks to @dbjorge who figured the rootcause and a proper way to handle the breaking changes from the version upgrade.

#### Pull request checklist


.gitignore
```diff
@@ -9,3 +9,4 @@
 node_modules/
 .DS_STORE
 yarn-error.log
+.yarnrc
```
dbjorge: This shouldn't be .gitignored or deleted; it's intentional that it's checked in. See the comment in .yarnrc for the reasoning.

copyright-header.config.json
```diff
@@ -0,0 +1,48 @@
+{
+    "license": "copyright-header.txt",
+    "ignore": [
+        "**/*.svg",
```
dbjorge: Can you sort this list?

copyright-header.config.json
```diff
@@ -0,0 +1,48 @@
...
+        "Dockerfile",
+        "**/*.snap",
+        "./.vscode",
+        "./.git",
+        "./.github",
+        "./dist",
+        "./drop",
+        "./extension",
+        "./node_modules",
+        "./copyright-header.txt",
```
dbjorge: Doesn't need to be here, it'll pass

copyright-header.config.json
```diff
@@ -0,0 +1,48 @@
...
+        "**/*.svg",
+        "**/*.scss.d.ts",
+        "**/.DS_Store",
+        "**/*.gitignore",
+        ".gitattribute",
+        "yarn.lock",
+        ".prettierignore",
+        ".yarnrc",
+        ".gitattributes",
+        "Dockerfile",
```
dbjorge: Dockerfile is interesting enough that it probably should be given a copyright header

copyright-header.config.json
```diff
@@ -0,0 +1,48 @@
...
+        "yaml|npmrc|yml|ps1": {
+            "eachLine": {
+                "prepend": "# "
+            }
+        },
+        "md": {
+            "prepend": "<!--",
+            "append": "-->"
+        },
+        "snap|ts|tsx|d.ts|js|scss|css": {
```
dbjorge: We can probably omit snap, since we're ignoring `**/*.snap` anyway

copyright-header.config.json
Shobhit1: It has a copyright header but the checker cannot check for files without a type; I will have to check what it accepts for files which doesn't have a extension

copyright-header.config.json
Shobhit1: Sure.

.gitignore
Shobhit1: I missed that. Updated to restore it back.

copyright-header.config.json
Shobhit1: Sure.

PR #1305, opened 2019-09-26
chore: removed dead styles in injected.scss

waabid:
#### Description of changes

Same process as before:

Search for the class name across our file base: if it only appears in the scss files, it's highly unlikely to be used.
Remove these styles.
Locally build the app and try to find these classes in html; if none were found, then we can be fairly certain this is dead css.

#### Pull request checklist

- [n/a] Ran precheckin (`yarn precheckin`)

### DIFF
--- a/src/injected/styles/injected.scss
+++ b/src/injected/styles/injected.scss
@@ -169,22 +169,6 @@
         margin-top: 3px !important;
     }
 
-    .insights-dialog-main-override .insights-dialog-rule-link,
-    .insights-dialog-main-override-shadow .insights-dialog-rule-link {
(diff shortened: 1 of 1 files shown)


src/injected/styles/injected.scss
```diff
@@ -223,7 +207,6 @@
                 }
 
                 display: inline;
-                float: right;
```
lisli1: Was this style detected through a different process (not finding all references of a certain class)?

waabid: This was found by VSCode telling me this property would never be applied due to the display property above.

PR #1318, opened 2019-09-30
refactor: fluxify validate port logic

smoralesd:
#### Description of changes

As part of "Automated checks results from axe-android in Electron with Cards UI" we need to convert our current logic to validate the port to follow flux architecture.

- Adding new store
- Adding new actions
- Updating current action creator (to have the core logic to validate the port, i.e. the call to the android service)
- Update current UI to use the new classes
- Wire up at the component root

The gifs shows the different state changes being handle by the UI: 
1. default (no connection has been made) 
1. connecting (user inputs a port and activate the "Validate port" button)
1. connected (successful connection, got a device name + app id)
1. default (user changes the port value)
1. connecting (again)
1. error (did not found any device + a11y service on that port)

01 - connect to device

#### Pull request checklist

- [x] Addresses an existing issue: completes WI # 1604620
   - not affected
   - not affected


src/common/stores/store-names.ts
```diff
@@ -15,4 +15,5 @@ export enum StoreNames {
     UserConfigurationStore,
     PathSnippetStore,
     UnifiedScanResultStore,
+    DeviceStore,
```
devaradhanm: it might be better to group electron store names

src/electron/device-connect-view/components/device-connect-body.tsx
```diff
@@ -1,19 +1,20 @@
...
+import { DeviceConnectState } from './device-connect-state';
 
 export type UpdateStateCallback = (newState: DeviceConnectState, deviceName?: string) => void;
 
-export enum DeviceConnectState {
-    Default,
-    Connecting,
-    Connected,
-    Error,
+export interface DeviceConnectBodyState {
```
devaradhanm: should this be DeviceConnectInfo instead of body state? it is not storing the state of the body, but the state of device connection.

src/electron/device-connect-view/components/device-connect-body.tsx
```diff
@@ -22,52 +23,24 @@ export type DeviceConnectBodyDeps = {
...
+    const canStartTesting = props.viewState.deviceConnectState === DeviceConnectState.Connected;
+
+    return (
+        <div className="device-connect-body">
+            <DeviceConnectHeader />
+            <DeviceConnectPortEntry deps={props.deps} viewState={{ deviceConnectState: props.viewState.deviceConnectState }} />
+            <DeviceConnectConnectedDevice
+                isConnecting={isConnecting}
+                connectedDevice={props.viewState.connectedDevice}
+                hasFailed={hasFailedConnecting}
```
devaradhanm: P2: (i know it is old code) it looks like DeviceConnectConnectedDevice can decide whether it is connecting / failed by reading the state props.viewState.deviceConnectState instead of the parent computing this value.

src/electron/device-connect-view/components/device-connect-port-entry.tsx
```diff
@@ -41,38 +42,35 @@ export class DeviceConnectPortEntry extends React.Component<DeviceConnectPortEnt
...
+        const props = {
+            disabled: this.isValidateButtonDisabled(),
+            className: 'button-validate-port',
+            onClick: this.onValidateClick,
+        };
 
-    private onValidateClick = async (event: React.MouseEvent<HTMLButtonElement>): Promise<void> => {
-        this.setState({ isValidateButtonDisabled: true });
+        if (this.props.viewState.deviceConnectState !== DeviceConnectState.Connected) {
+            return <PrimaryButton {...props}>Validate port number</PrimaryButton>;
```
devaradhanm: what is the aria label value difference when you use Primary vs DefaultButton instead of using Button with primary attribute ?

src/electron/device-connect-view/components/device-connect-port-entry.tsx
```diff
@@ -41,38 +42,35 @@ export class DeviceConnectPortEntry extends React.Component<DeviceConnectPortEnt
...
+    }
 
-        await this.props.deps
-            .fetchScanResults(port)
-            .then(data => {
-                this.props.updateStateCallback(DeviceConnectState.Connected, `${data.deviceName} - ${data.appIdentifier}`);
-            })
-            .catch(err => this.props.updateStateCallback(DeviceConnectState.Error));
+    private onPortTextChanged = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
+        this.setState({ port: newValue });
```
devaradhanm: what happens if emitChanged event is fired when the user is typing?

devaradhanm: never mind. it will only change if the props value changes for this object

src/common/stores/store-names.ts
smoralesd: that sounds like a good idea and I like to implement it but I'll leave it for a different PR, the main reason is the base store implementation constructor receives `storeName: StoreNames`, i.e. we are using this enum as a proper type on some places and that will need a little more than spliting the file: 
```typescript
export abstract class BaseStoreImpl<TState> extends Store implements BaseStore<TState> {
    ...
    constructor(storeName: StoreNames) {
       ...
    }
```

src/electron/device-connect-view/components/device-connect-body.tsx
smoralesd: Agreed, but:
- as you mention this PR is not changing `DeviceConnectConnectedDevice`
- I'd like to keep this PR as small as I can

src/electron/device-connect-view/components/device-connect-port-entry.tsx
smoralesd: if it helps, react handle state changes as async operations, it tries to merge them before applying the new state values.

src/electron/device-connect-view/components/device-connect-body.tsx
smoralesd: The idea was to follow a similar pattern as the one we have with deps. I updated this type to reflect that.

src/electron/device-connect-view/components/device-connect-view-container.tsx
```diff
@@ -36,18 +42,30 @@ export class DeviceConnectViewContainer extends React.Component<DeviceConnectVie
...
-        this.props.deps.userConfigurationStore.addChangedListener(() => {
-            const newState = this.props.deps.userConfigurationStore.getState();
-            this.setState({
-                userConfigurationStoreData: newState,
-            });
+        this.addChangeListener('deviceStore');
+        this.addChangeListener('userConfigurationStore');
+    }
+
+    private addChangeListener(storeName: keyof DeviceConnectViewContainerStoreDeps): void {
```
devaradhanm: can we leverage BaseClientStoresHub? this code seems to be duplicating some of the functionality of BaseClientStoresHub

smoralesd: Right! We should be able to do that. I'll udpate this PR with those changes.

src/electron/device-connect-view/components/device-connect-port-entry.tsx
```diff
@@ -41,38 +42,35 @@ export class DeviceConnectPortEntry extends React.Component<DeviceConnectPortEnt
...
-                    Validate port number
-                </Button>
+                {this.renderValidationPortButton()}
             </div>
         );
     }
 
-    private onPortTextChanged = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
-        this.setState({ isValidateButtonDisabled: !newValue || newValue === '', port: newValue });
-        this.props.updateStateCallback(DeviceConnectState.Default);
```
jalkire: We lost this state change with the refactor:
1. Connect to device successfully
2. Edit port
3. Expected state: Expected state: start testing disabled, validate port button primary (the default state)
    Actual state: start testing enabled, validate port button not primary (the connected state)

It looks like the reason is because no message is sent back to the store as a part of the new `onPortTextChanged`, so no state update comes back through the props.

devaradhanm: we shouldn't be changing state when the text is being changed rite? may be thats the reason. we update state only when the user clicks validate?

src/electron/device-connect-view/components/device-connect-view-container.tsx
devaradhanm: cool. other than that, LGTM

src/electron/device-connect-view/components/device-connect-port-entry.tsx
jalkire: During the feature to implement this screen, the intentional design was that a change in port number after a validation has successfully occurred should return the state to pre-validation, since the new port number has not been validated.

src/electron/flux/action/device-actions.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { Action } from 'common/flux/action';
+
+import { ConnectingPayload, ConnectionSucceedPayload } from './device-action-payloads';
+
+export class DeviceActions {
+    public readonly connectionSucceed = new Action<ConnectionSucceedPayload>();
```
DaveTryon: >connectionSucceed [](start = 20, length = 17)

The verb tenses here are inconsistent. I'd expect either connectionFailed and connectionSucceeded, or connectionFail and ConnectionSucceed.

smoralesd: Good catch! Decided to go with the past tense.

src/electron/device-connect-view/components/device-connect-port-entry.tsx
smoralesd: Good catch @jalkire I'll update the logic to account for this

src/electron/device-connect-view/components/device-connect-view-container.tsx
smoralesd: Updated

src/electron/device-connect-view/components/device-connect-port-entry.tsx
smoralesd: Updated!

PR #1322, opened 2019-09-30
feat: device disconnected popup

smoralesd:
#### Description of changes

Add the new "Device disconnected" popup.

01 - device disconnected popup

**Notes:**
- The "MY DEVICE" text is only for the screenshot purpose as that text is a prop that should come from the renderer and it should match the actual device name we get from the service.

**Out of scope:**
- actually render the dialog is not part of this PR as we still don't have the view that will use it.

#### Pull request checklist

- [x] Addresses an existing issue: completes WI # 1606858

### DIFF
--- a/src/electron/device-disconnected-popup/device-disconnected-popup.scss
+++ b/src/electron/device-disconnected-popup/device-disconnected-popup.scss
@@ -2,22 +2,26 @@
 // Licensed under the MIT License.
 @import '../../common/styles/fonts.scss';
 
-.title-container {
-    display: flex;
+.device-disconnected-popup {
+    .title-container {
+        display: flex;
 
-    svg {
-        align-self: center;
+        svg {
+            align-self: center;
+        }
     }
-}
 
-.ms-Dialog-header {
-    height: 50px;
-}
+    :global {
+        .ms-Dialog-header {
+            height: 50px;
+        }
+    }
 
-.title-text {
-    margin-left: 8px;
-    font-family: $semiBoldFontFamily;
-    font-size: 20px;
-    font-weight: 600;
-    line-height: 28px;
+    .title-text {
+        margin-left: 8px;
+        font-family: $semiBoldFontFamily;
+        font-size: 20px;
+        font-weight: 600;
+        line-height: 28px;
+    }
 }
--- a/src/electron/device-disconnected-popup/device-disconnected-popup.scss
+++ b/src/electron/device-disconnected-popup/device-disconnected-popup.scss
@@ -1,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../common/styles/fonts.scss';
+
+.title-container {
+    display: flex;
+
+    svg {
+        align-self: center;
+    }
+}
+
+.ms-Dialog-header {
+    height: 50px;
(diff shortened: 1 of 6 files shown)


src/electron/device-disconnected-popup/device-disconnected-popup.scss
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../common/styles/fonts.scss';
+
+.title-container {
```
devaradhanm: the name sounds generic. title-container, ms-Dialog-header & title-text.

Either wrap this under a parent container called say, device-disconnected-container or make the names more specific or move them to a common scss file if you want them to be common for other titles & headers too.

smoralesd: I did 2 things:
- Add a class name for the dialog so we can "group" this class names inside it
- Enabled CSS modules so the final class names are made unique (on prod at least) by the build process.

As per `ms-Dialog-header`, that comes from office fabric, we have no way to directly change it

src/electron/device-disconnected-popup/device-disconnected-popup.scss
```diff
@@ -0,0 +1,27 @@
...
+.device-disconnected-popup {
+    .title-container {
+        display: flex;
+
+        svg {
+            align-self: center;
+        }
+    }
+
+    :global {
```
waabid: I've had some hit or miss usage with:

```
:global {
    ... // css here
}
```

Using:

```
:global(.class-name) { 
    .... // css here
}
```
seems to give more consistent behavior. I'm not fully sure why ...

devaradhanm: @waabid, this also seems to add the :global attribute to css file. will this affect the style of the headers outside device-disconnected-popup as well? If so, then we should avoid using this.

src/electron/device-disconnected-popup/device-disconnected-popup.scss
waabid: Not sure if the change has been made in master but it seems like electron always uses dev config instead of the prod config.

For now that means the 2nd item you mentioned isn't true _yet_. Just something to note.

src/electron/device-disconnected-popup/device-disconnected-popup.scss
smoralesd: Thanks! Will update

waabid: > @waabid, this also seems to add the :global attribute to css file. will this affect the style of the headers outside device-disconnected-popup as well? If so, then we should avoid using this.

I'm not seeing this behavior?

smoralesd: I pulled back from using css modules on this PR as we saw some weird behavior and css module is not the main purpose of the PR

src/electron/device-disconnected-popup/device-disconnected-popup.scss
smoralesd: That's correct... we don't have a prod config for electron. But I saw the config is there for us to use when the time comes.

PR #1333, opened 2019-10-01
Convert android results to unified rule format

DaveTryon:
#### Description of changes

Convert android results to unified rule format. Includes only the conversion code, with none of the plumbing of data in or data out.

#### Pull request checklist

### DIFF
--- a/src/electron/platform/android/rule-information-provider.ts
+++ b/src/electron/platform/android/rule-information-provider.ts
@@ -61,6 +61,6 @@ export class RuleInformationProvider {
 
     public getRuleInformation(ruleId: string): RuleInformation {
         const ruleInfo = this.supportedRules[ruleId];
-        return ruleInfo ? ruleInfo : null;
+        return ruleInfo || null;
     }
 }
--- a/src/electron/platform/android/rule-information-provider.ts
+++ b/src/electron/platform/android/rule-information-provider.ts
@@ -1,0 +1,75 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { RuleInformation } from './rule-information';
+import { RuleResultsData } from './scan-results';
+
+export class RuleInformationProvider {
+    private supportedRules: RuleInformation[];
+
+    constructor() {
+        this.supportedRules = [];
+        this.supportedRules.push(
+            new RuleInformation(
+                'ColorContrast',
+                'Text elements must have sufficient contrast against the background.',
+                this.getColorContrastHowToFix,
+            ),
+        );
+        this.supportedRules.push(
+            new RuleInformation('TouchSizeWcag', 'Touch inputs must have a sufficient target size.', this.getTouchSizeHowToFix),
+        );
+        this.supportedRules.push(
+            new RuleInformation(
+                'ActiveViewName',
+                "Active views must have a name that's available to assistive technologies.",
+                () =>
+                    'The view is active but has no name available to assistive technologies. Provide a name for the view using its contentDescription, hint, labelFor, or text attribute (depending on the view type)',
+            ),
+        );
+        this.supportedRules.push(
+            new RuleInformation(
+                'ImageViewName',
+                'Meaningful images must have alternate text.',
+                () =>
+                    'The image has no alternate text and is not identified as decorative. If the image conveys meaningful content, provide alternate text using the contentDescription attribute. If the image is decorative, give it an empty contentDescription, or set its isImportantForAccessibility attribute to false.',
+            ),
+        );
+        this.supportedRules.push(
+            new RuleInformation(
+                'EditTextValue',
+                'EditText elements must expose their entered text value to assistive technologies',
+                () =>
+                    "The element's contentDescription overrides the text value required by assistive technologies. Remove the element’s contentDescription attribute.",
+            ),
+        );
+    }
+
+    private getColorContrastHowToFix(ruleResultsData: RuleResultsData): string {
+        const ratio = ruleResultsData.props['Color Contrast Ratio'] as string;
+        const foreground = (ruleResultsData.props['Foreground Color'] as string).substring(2);
+        const background = (ruleResultsData.props['Background Color'] as string).substring(2);
+
+        return `The text element has insufficient contrast of ${ratio}. Foreground color: #${foreground}, background color: #${background}). Modify the text foreground and/or background colors to provide a contrast ratio of at least 4.5:1 for regular text, or 3:1 for large text (at least 18pt, or 14pt+bold).`;
+    }
+
+    private getTouchSizeHowToFix(ruleResultsData: RuleResultsData): string {
+        const dpi = ruleResultsData.props['Screen Dots Per Inch'] as number;
+        const physicalWidth = ruleResultsData.props['width'] as number;
+        const physicalHeight = ruleResultsData.props['height'] as number;
+        const logicalWidth = physicalWidth / dpi;
+        const logicalHeight = physicalHeight / dpi;
+
+        return `The element has an insufficient target size (width: ${logicalWidth}dp, height: ${logicalHeight}dp). Set the element's minWidth and minHeight attributes to at least 48dp.`;
+    }
+
+    public getRuleInformation(ruleId: string): RuleInformation {
+        for (const ruleInformation of this.supportedRules) {
+            if (ruleId === ruleInformation.ruleId) {
+                return ruleInformation;
+            }
+        }
+
+        return null;
+    }
+}
--- a/src/electron/platform/android/rule-information-provider.ts
+++ b/src/electron/platform/android/rule-information-provider.ts
@@ -1,48 +1,44 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+import { DictionaryStringTo } from '../../../types/common-types';
 import { RuleInformation } from './rule-information';
 import { RuleResultsData } from './scan-results';
 
 export class RuleInformationProvider {
-    private supportedRules: RuleInformation[];
+    private supportedRules: DictionaryStringTo<RuleInformation>;
 
     constructor() {
-        this.supportedRules = [];
-        this.supportedRules.push(
-            new RuleInformation(
+        this.supportedRules = {
+            ColorContrast: new RuleInformation(
                 'ColorContrast',
                 'Text elements must have sufficient contrast against the background.',
                 this.getColorContrastHowToFix,
             ),
-        );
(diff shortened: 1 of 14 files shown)


src/electron/platform/android/rule-information-provider.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import { RuleInformation } from './rule-information';
+import { RuleResultsData } from './scan-results';
+
+export class RuleInformationProvider {
+    private supportedRules: RuleInformation[];
+
+    constructor() {
+        this.supportedRules = [];
+        this.supportedRules.push(
```
karanbirsingh: You may be able to initialize the variable explicitly without any pushes:
```
rules = [ new X(), new Y(), new Z() ]; // with new lines
```

src/electron/platform/android/rule-information-provider.ts
```diff
@@ -0,0 +1,75 @@
...
+        const physicalWidth = ruleResultsData.props['width'] as number;
+        const physicalHeight = ruleResultsData.props['height'] as number;
+        const logicalWidth = physicalWidth / dpi;
+        const logicalHeight = physicalHeight / dpi;
+
+        return `The element has an insufficient target size (width: ${logicalWidth}dp, height: ${logicalHeight}dp). Set the element's minWidth and minHeight attributes to at least 48dp.`;
+    }
+
+    public getRuleInformation(ruleId: string): RuleInformation {
+        for (const ruleInformation of this.supportedRules) {
```
karanbirsingh: if this were a dictionary keyed on the ruleId, would this be easier? If it needs to be a list, [`find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Syntax) might be helpful

src/electron/platform/android/rule-information.ts
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License.
+
+import { RuleResultsData } from './scan-results';
+
+export type HowToFixDelegate = (ruleResultsData: RuleResultsData) => string;
+
+export class RuleInformation {
+    private RuleId: string;
+    private RuleDescription: string;
+    private HowToFixDelegate: HowToFixDelegate;
```
karanbirsingh: We might be able to initialize these directly in the constructor like the link [here](https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties). We have some examples in our codebase if you look for "private readonly" inside a constructor

src/electron/platform/android/scan-results-to-unified-results.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+function getStatus(status: string): InstanceResultStatus {
+    switch (status) {
+        case 'PASS':
+            return 'pass';
+        case 'FAIL':
+            return 'fail';
+        default:
+            return 'unknown';
```
karanbirsingh: The unit tests give us a [table ](https://dev.azure.com/ms/accessibility-insights-web/_build/results?buildId=39457&view=logs&jobId=3613c2bd-8b62-5a70-8491-3e8459586450&taskId=d1343856-9cd9-5649-cdc6-81bc16053e6f&lineStart=1376&lineEnd=1376&colStart=68&colEnd=127) of code coverage & uncovered lines; you might explore these to see that any we're missing are intentionally skipped. For instance, this line is missed

src/electron/platform/android/rule-information-provider.ts
DaveTryon: Used a dictionary, as we discussed

src/electron/platform/android/rule-information-provider.ts
DaveTryon: Used a dictionary

src/electron/platform/android/rule-information.ts
DaveTryon: Refactored as suggested

src/electron/platform/android/scan-results-to-unified-results.ts
DaveTryon: Added a case for this

src/electron/platform/android/rule-information.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { RuleResultsData } from './scan-results';
+
+export type HowToFixDelegate = (ruleResultsData: RuleResultsData) => string;
+
+export class RuleInformation {
+    constructor(readonly RuleId: string, readonly RuleDescription: string, readonly howToFixDelegate: HowToFixDelegate) {}
```
karanbirsingh: the variables `RuleId` and `RuleDescription` should be camelCase to match other files; we may also consider `private readonly…` instead of `readonly` for these variables

DaveTryon: Fixed as discussed

src/electron/platform/android/scan-results.ts
```diff
@@ -22,4 +25,13 @@ export class ScanResults {
             return null;
         }
     }
+
+    public get ruleResults(): RuleResultsData[] {
+        try {
+            const results = this.rawData.axeRuleResults;
+            return results ? results : [];
```
karanbirsingh: FYI (Not necessary, this pattern is fine), `return results || []` is another common pattern - the `||` operator will return the first 'truthy' value it finds and `null` is not truthy

DaveTryon: Thanks! Will include this in the next iteration if there is one or in next PR if not. Next PR will need to touch this file anyway.

src/electron/platform/android/scan-results-to-unified-results.ts
```diff
@@ -0,0 +1,62 @@
...
+export function convertScanResultsToUnifiedResults(scanResults: ScanResults, uuidGenerator: UUIDGeneratorType): UnifiedResult[] {
+    if (scanResults == null || scanResults.ruleResults == null) {
+        return [];
+    }
+
+    return createUnifiedResultsFromScanResults(scanResults.ruleResults, uuidGenerator);
+}
+
+function createUnifiedResultsFromScanResults(ruleResults: RuleResultsData[], uuidGenerator: UUIDGeneratorType): UnifiedResult[] {
+    const ruleInformationProvider: RuleInformationProvider = new RuleInformationProvider();
```
smoralesd: In general, we don't instanciate dependencies outside the composition root (`src\electron\device-connect-view\device-connect-view-initializer.ts` in the case of AI-Android) or factory classes (which we don't have on AI-Android). This works good enough for classes, which is not the case here.
In the case of functions, we usually follow the pattern of a creator function, where the creator function allows to bind any dependency (`src\issue-filing\services\azure-boards\create-azure-boards-issue-filing-url.ts` has an example of that).

All that said... I'm wondering if it make sense to have more than one instance of `RuleInformationProvider` as the class seems to contain "static" information of the rules we run on an android scan.

I'm not 100% sure how to change this right now, or if it's worth to change right now, my main intention is to share our general approach on how we handle instantiation and dependencies.

src/tests/unit/tests/electron/platform/android/scan-results-helpers.ts
```diff
@@ -0,0 +1,65 @@
...
+    const props = {};
+    // This is based on the output of the Android service
+    props['Screen Dots Per Inch'] = dpi;
+    props['height'] = height;
+    props['width'] = width;
+
+    return buildRuleResultObject('TouchSizeWcag', status, props);
+}
+
+export function buildColorContastRuleResultObject(status: string, ratio: number, foreground: string, background: string): RuleResultsData {
```
smoralesd: typo: buildColorCont**r**astRuleResultObject

src/electron/platform/android/scan-results-to-unified-results.ts
DaveTryon: Karan and I had a similar conversation. Long-term, we definitely only need one, but we were thinking of leaving it this way until we sort out the plumbing around calling the converter (since this will necessarily alter the signature of the converter delegates)

src/tests/unit/tests/electron/platform/android/rule-information.test.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { RuleInformation } from '../../../../../../electron/platform/android/rule-information';
+
+describe('RuleInformation', () => {
+    test('RuleId works correctly', () => {
+        const ruleIds = ['abc', 'xyz', 'this should work'];
```
smoralesd: `ruleIds` is the same on all 3 test. The duplication can be avoided by placing the declaration right below the `describe` (and outside all the `test` functions)

src/tests/unit/tests/electron/platform/android/rule-information.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        for (const ruleDescription of ruleDescriptions) {
+            const ruleInformation = new RuleInformation('a', ruleDescription, () => 'c');
+            expect(ruleDescription === ruleInformation.ruleDescription);
+        }
+    });
+
+    test('HowToFix works correctly', () => {
+        const howToFixStrings = ['abc', 'xyz', 'this should work'];
+        for (const howToFixString of howToFixStrings) {
+            const ruleInformation = new RuleInformation('a', 'b', () => howToFixString);
```
smoralesd: while passing functions as parameters to classes (or other functions) we usually make use of a mocked function to verify it's being call with the expected parameters and the right amount of times. In general, we use `typemoq` for this purpose. You can find an example in this file `src\tests\unit\tests\common\components\copy-issue-details-button.test.tsx` for the `onClickMock`

DaveTryon: In this case, I'm trying to ensure that calling the howToFix getter actually ends up calling the howToFixDelegate property. I intentionally don't want to mock it out. I'll see if it makes sense in the other test cases

src/tests/unit/tests/electron/platform/android/rule-information.test.ts
DaveTryon: Updated as described

src/tests/unit/tests/electron/platform/android/scan-results-helpers.ts
DaveTryon: Fixed

src/electron/platform/android/scan-results.ts
DaveTryon: Fixed

PR #1335, opened 2019-10-01
refactor: store listening and focus change handling in client view controller to separate classes

waabid:
#### Description of changes

Re-factoring client-view-controller as it's a pain to modify right now. This PR specifically targets the store listening aspect and the focus change handling aspect.

#### Pull request checklist


src/tests/unit/tests/injected/client-store-listener.test.ts
```diff
@@ -0,0 +1,61 @@
...
+
+describe('ClientStoreListener', () => {
+    describe('initialize', () => {
+        let testSubject: ClientStoreListener;
+        let storeHubMock: IMock<BaseClientStoresHub<TargetPageStoreData>>;
+        let onReadyToExecuteVisualizationUpdatesMock: IMock<(storeData: TargetPageStoreData) => void>;
+        let triggerOnChange: () => void;
+
+        beforeEach(() => {
+            storeHubMock = Mock.ofType<BaseClientStoresHub<TargetPageStoreData>>(null, MockBehavior.Strict);
```
smoralesd: I think typemoq docs mention `undefined` for the first parameter will make use of es6 `Proxy` class and this is what we've been doing in general. Is there any specific reason why in here is `null`? Not that it matter that much except for the consistency of it

src/tests/unit/tests/injected/client-store-listener.test.ts
```diff
@@ -0,0 +1,61 @@
...
+            onReadyToExecuteVisualizationUpdatesMock = Mock.ofType<(storeData: TargetPageStoreData) => void>();
+
+            storeHubMock
+                .setup(shm => shm.addChangedListenerToAllStores(It.isAny()))
+                .callback(givenCallback => (triggerOnChange = givenCallback));
```
smoralesd: looks like `triggerOnChange` is been called on each tests... does it make sense to call it right here on the setup? like this:
```typescript
  storeHubMock
                .setup(shm => shm.addChangedListenerToAllStores(It.isAny()))
                .callback(givenCallback => givenCallback());
```

src/tests/unit/tests/injected/focus-change-handler.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        targetPageActionMessageCreatorMock.verifyAll();
+        scrollingControllerMock.verifyAll();
+    });
+
+    test('onStoreChange: new target is not null', () => {
+        const storeData: TargetPageStoreData = {
+            visualizationStoreData: {
+                focusedTarget: sampleTarget,
+            } as VisualizationStoreData,
```
smoralesd: do we need this cast (`as VisualizationStoreDatta`)? I think the second one (`as TargetPageStoreData`) should take care of the whole casting.

src/injected/client-store-listener.ts
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { BaseClientStoresHub } from '../common/stores/base-client-stores-hub';
```
smoralesd: should we import using relative paths here? so instead of
```typescript
import { BaseClientStoresHub } from '../common/stores/base-client-stores-hub';
```
we'll have
```typescript
import { BaseClientStoresHub } from 'common/stores/base-client-stores-hub';
```

src/tests/unit/tests/injected/client-store-listener.test.ts
waabid: > This looks good but I left some minor comments.
> I'm also assuming there will be another PR to hook up this new classes as I don't see then being use neither the `ClientViewController` being updated!
> Good refactor.

Yep! Didn't want a massive PR 😄

src/tests/unit/tests/injected/client-store-listener.test.ts
waabid: Good question! Since the setup for the interactions with the storeHubMock are different for each test/scenario, and as such in the test blocks, the function can only be triggered after that setup is complete.

smoralesd: It will be trigger only when you actuate on the test subject though, right?

PR #1345, opened 2019-10-02
Add automated checks command bar

jalkire:
#### Description of changes
The ongoing automated checks results from Android feature calls for the addition of a command bar that will ultimately look like this:
image

For now, the only command with an actual purpose is the rescan button. We could add the rest now, but they will not do anything until future features are complete.

We don't yet have the UI to house this bar, so for now it is just waiting to be used.

#### Pull request checklist

- [-] (UI changes only) Added screenshots/GIFs to description above
- [-] (UI changes only) Verified usability with NVDA/JAWS

### DIFF
--- a/src/electron/automated-checks/components/automated-checks-command-bar.tsx
+++ b/src/electron/automated-checks/components/automated-checks-command-bar.tsx
@@ -22,7 +22,6 @@ export const AutomatedChecksCommandBar = NamedFC<AutomatedChecksCommandBarProps>
         return (
             <div className={automatedChecksCommandBar}>
                 <ActionButton
-                    className="button-rescan"
                     iconProps={{
                         iconName: 'Refresh',
                     }}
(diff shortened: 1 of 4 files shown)


src/electron/automated-checks/components/automated-checks-command-bar.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export const AutomatedChecksCommandBar = NamedFC<AutomatedChecksCommandBarProps>(
+    'AutomatedChecksCommandBar',
+    (props: AutomatedChecksCommandBarProps) => {
+        const onClick = () => props.deps.deviceConnectActionCreator.resetConnection();
+        return (
+            <div className={automatedChecksCommandBar}>
+                <ActionButton
+                    className="button-rescan"
```
smoralesd: I can't find styles for `button-rescan` style. If we're using this only for the sake of testing (like in `.find('.button-rescan')`) I'll suggest using an id or use the text property (like in this test `src\tests\unit\electron\device-disconnected-popup\device-disconnected-popup.test.tsx`)

jalkire: That is the only reason it has a className. Thanks for the tip--will update!

jalkire: Updated to use text

PR #1347, opened 2019-10-02
add tool info to unified scan result store

lisli1:
#### Description of changes

- Added a `getToolData()` function to `EnvironmentDataProvider` because `EnvironmentDataProvider` already has all the environment data needed to populate and return a ToolData object.
- Added `toolInfo` property to `UnifiedScanCompletedPayload` to send tool data to the `UnifiedScanResultStore` as part of the payload in `UnifiedResultSender`
- Created `ApplicationProperties` interface in `unified-data-interface.ts` to be able to specify application data, which is different from scan engine data, as part of the ToolData
- Updated relevant tests

#### Pull request checklist

- [x] Addresses an existing issue: WI # 1584979

### DIFF
--- a/src/injected/analyzers/unified-result-sender.ts
+++ b/src/injected/analyzers/unified-result-sender.ts
@@ -4,7 +4,6 @@ import { UnifiedScanCompletedPayload } from '../../background/actions/action-pay
 import { EnvironmentInfoProvider } from '../../common/environment-info-provider';
 import { Messages } from '../../common/messages';
 import { UUIDGeneratorType } from '../../common/uid-generator';
-import { toolName } from '../../content/strings/application';
 import { ConvertScanResultsToUnifiedResultsDelegate } from '../adapters/scan-results-to-unified-results';
 import { ConvertScanResultsToUnifiedRulesDelegate } from '../adapters/scan-results-to-unified-rules';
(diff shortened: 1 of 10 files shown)


src/injected/analyzers/unified-result-sender.ts
```diff
@@ -13,13 +14,15 @@ export class UnifiedResultSender {
...
         private readonly convertScanResultsToUnifiedRules: ConvertScanResultsToUnifiedRulesDelegate,
+        private readonly environmentInfoProvider: EnvironmentInfoProvider,
         private readonly generateUID: UUIDGeneratorType,
     ) {}
 
     public sendResults: PostResolveCallback = (axeResults: AxeAnalyzerResult) => {
         const payload: UnifiedScanCompletedPayload = {
             scanResult: this.convertScanResultsToUnifiedResults(axeResults.originalResult, this.generateUID),
             rules: this.convertScanResultsToUnifiedRules(axeResults.originalResult),
+            toolInfo: this.environmentInfoProvider.getToolData(),
```
haonliu: Sending tool info along with the scan result every time a scan is completed doesn't feel quite right. Tool info should be sent only once, when the 'main window' is initialized.

PR #1350, opened 2019-10-02
Basic Window State Store Implementation to be able to update the route id between different views

JGibson2019:
#### Description of changes

Created a store for the Window State to be able to update the route id between different views

#### Pull request checklist

- [ N/A] Addresses an existing issue: Fixes #0000


src/electron/flux/action/window-state-actions.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { Action } from 'common/flux/action';
+
+export class WindowStateActions {
+    public readonly setDeviceConnectRoute = new Action<void>();
```
smoralesd: I'm wondering if this should be more generic, like `setRoute = new Action<RoutePayload>()`... not sure if it's a good idea to have a set of valid routes defined on a separated file, but it sounds like each "view" may have it's own configuration (like the route -which can be the id- and the default window size)

src/tests/unit/tests/electron/flux/store/electron-window-state-store.test.ts
```diff
@@ -0,0 +1,46 @@
...
+            expect(testSubject).toBeDefined();
+        });
+    });
+
+    it('returns device connect as the default state', () => {
+        const testSubject = new ElectronWindowStateStore(new WindowStateActions());
+        testSubject.initialize();
+
+        expect(testSubject.getState()).toEqual(testSubject.getDefaultState());
+        expect(testSubject.getState()).toEqual(getDeviceConnectViewState());
```
smoralesd: not sure what the second `expect` is doing here, is it really needed?

src/tests/unit/tests/electron/flux/store/electron-window-state-store.test.ts
```diff
@@ -0,0 +1,46 @@
...
+            const testSubject = createStoreWithNullParams(ElectronWindowStateStore);
+            expect(testSubject).toBeDefined();
+        });
+    });
+
+    it('returns device connect as the default state', () => {
+        const testSubject = new ElectronWindowStateStore(new WindowStateActions());
+        testSubject.initialize();
+
+        expect(testSubject.getState()).toEqual(testSubject.getDefaultState());
```
smoralesd: can the default state be built using `createStoreWithNullParams(ElectronWindowStateStore).getDefaultState();` ??? that way you don't need to call `.initialize`

devaradhanm: we also wanted to validate the initialstate. it is easier to validate the initial state without using store tester & just call initialize.

src/tests/unit/tests/electron/flux/store/electron-window-state-store.test.ts
devaradhanm: we test the default state is device connect view.
& we also test that the initial test is device connect view.

it might have been more clear, if we explicitly validate both states to be same as getDeviceConnectViewState()

src/electron/flux/action/window-state-actions.ts
devaradhanm: RoutePayload  type => { routeId : ElectronRoute}

src/electron/flux/types/electron-window-state-store-data.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export type electronRoutes = 'deviceConnectView' | 'resultsView';
```
devaradhanm: make this caps. ElectronRoutes

src/tests/unit/tests/electron/flux/store/electron-window-state-store.test.ts
smoralesd: not sure what you're trying to test... but I can say this: the `BaseStoreImpl` class contains this function:
```typescript
    public initialize(initialState?: TState): void {
        this.state = initialState || this.getDefaultState();

        this.addActionListeners();
    }
```
and that's being tested here: `src\tests\unit\tests\background\stores\base-store.test.ts`
If you're trying to test the the initial state is being set by using `getDefaultState` that's already being cover elsewhere and there is no need to test that here too

src/tests/unit/tests/electron/flux/store/electron-window-state-store.test.ts
smoralesd: I think this is a good point that we are missing on other store test (the default state) but I'd suggest using a snapshot instead.

JGibson2019: So should I update the expect statements to be 

expect(testSubject.getDefaultState()).toEqual(getDeviceConnectViewState());
expect(testSubject.getState()).toEqual(getDeviceConnectViewState());

src/common/stores/store-names.ts
```diff
@@ -17,4 +17,5 @@ export enum StoreNames {
     UnifiedScanResultStore,
     CardSelectionStore,
     DeviceStore,
+    ElectronWindowStateStore,
```
smoralesd: I know we mention this offline but I'll coment again: I don't think we should use `Electron` in the name of our classes/enum/types. Each store is actually agnostic to where it's being used (extension or electron) and as such we should not "stigmatize" them like this. It should be up to the reader to look at the proper initializer to find out where the store (or class/type/interface/enum for that matter) is being used

src/electron/flux/types/electron-window-state-store-data.ts
smoralesd: not sure if using `electron` here is a good idea (same reason as in my comment on `src/common/stores/store-names.ts`)

src/electron/flux/action/window-state-actions.ts
JGibson2019: updated in most recent commit

src/tests/unit/tests/electron/flux/store/electron-window-state-store.test.ts
JGibson2019: So I pushed a change with the updated expect statements before this comment; do I need to change to snapshots for this PR or add it in my next PR? @smoralesd

src/electron/flux/types/electron-window-state-store-data.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export type ElectronRoutes = 'deviceConnectView' | 'resultsView';
```
JGibson2019: So should I rename ElectronRoutes to "WindowRoutes" @smoralesd @devaradhanm

src/tests/unit/tests/electron/flux/store/electron-window-state-store.test.ts
JGibson2019: added snapshots

src/electron/flux/types/electron-window-state-store-data.ts
JGibson2019: renamed to viewroutes in window-state-store-data

src/common/stores/store-names.ts
JGibson2019: updated to remove Electron from the name in the 5th commit

src/electron/flux/types/electron-window-state-store-data.ts
JGibson2019: renamed to ViewRoutes in 5th commit

src/electron/flux/action/route-payloads.ts
```diff
@@ -0,0 +1,7 @@
+import { ViewRoutes } from '../types/window-state-store-data';
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
smoralesd: file header should be the top-most content of the file

src/electron/flux/store/window-state-store.ts
```diff
@@ -0,0 +1,31 @@
...
+        };
+    }
+
+    protected addActionListeners(): void {
+        this.windowStateActions.changeDeviceRoute.addListener(this.onChangeDeviceRoute);
+    }
+
+    private onChangeDeviceRoute = (payload: RoutePayload) => {
+        if (this.state.routeId === payload.routeId) {
+            return;
```
smoralesd: the "no update" case needs a unit test

JGibson2019: added it in the newest commit

PR #1351, opened 2019-10-03
[spike] refactor issue filing types to enable filing from unified results

karanbirsingh:
This PR shows one possible approach to enable issue filing from both old axe-core data and from new unified-store data. It's not production-ready yet but should help drive conversation

summary:
- `CreateIssueDetailsTextData` only holds the fields issue filing needs to create a bug, no more or less
- Each consumer of issue filing converts their data-type (axe-core data or unified-data) to `CreateIssueDetailsTextData` (the cards UI will soon need to do this with unified-data)
- Some strongly typed fields that we expect to exist in all platforms are added to unified-results to make the conversion easier

pros:
- The issue filing plumbing using `CreateIssueDetailsTextData` can stay mostly intact if we convert earlier. Otherwise, we need to make sure the plumbing can work with both `CreateIssueDetailsTextData` and `UnifiedResult`
- `createIssueDetailsBuilder` remains fairly simple, referencing the `CreateIssueDetailsTextData` fields directly
- we share `createIssueDetailsBuilder` between the old consumers (target dialog, etc) and new consumers (card UI buttons)
- ...feel free to add more

cons:
- We lose some flexibility for each field we make strongly-typed. For example, `how-to-fix` is a string instead of the `all/none/any` fields used in card-rows. The existing issue filing implementation treats `how-to-fix` as a string, so this might be OK in some instances
- ...feel free to add more

### DIFF
--- a/src/DetailsView/components/Issues-details-pane.tsx
+++ b/src/DetailsView/components/Issues-details-pane.tsx
@@ -98,7 +98,7 @@ export class IssuesDetailsPane extends React.Component<IssuesDetailsPaneProps, I
             },
             element: {
                 id: result.id,
-                idDisplayName: 'Element path',
+                identifierName: 'Element path',
                 shortId: 'last selector',
             },
             howToFixSummary: result.failureSummary,
--- a/src/DetailsView/components/Issues-details-pane.tsx
+++ b/src/DetailsView/components/Issues-details-pane.tsx
@@ -87,9 +87,22 @@ export class IssuesDetailsPane extends React.Component<IssuesDetailsPaneProps, I
 
     private renderSingleIssue(result: DecoratedAxeNodeResult): JSX.Element {
         const issueData: CreateIssueDetailsTextData = {
-            pageTitle: this.props.pageTitle,
-            pageUrl: this.props.pageUrl,
-            ruleResult: result,
+            rule: {
+                description: result.help,
+                id: result.ruleId,
+                url: result.helpUrl,
+                guidance: result.guidanceLinks,
+            },
+            targetApp: {
+                name: 'name',
+            },
+            element: {
+                id: result.id,
+                idDisplayName: 'Element path',
+                shortId: 'last selector',
+            },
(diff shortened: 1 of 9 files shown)


src/DetailsView/components/Issues-details-pane.tsx
```diff
@@ -87,9 +87,22 @@ export class IssuesDetailsPane extends React.Component<IssuesDetailsPaneProps, I
...
-            pageUrl: this.props.pageUrl,
-            ruleResult: result,
+            rule: {
+                description: result.help,
+                id: result.ruleId,
+                url: result.helpUrl,
+                guidance: result.guidanceLinks,
+            },
+            targetApp: {
+                name: 'name',
```
karanbirsingh: used to be `pageTitle/pageUrl` & passed in from a parent, will have to do something similar since `UnifiedResult` doesn't have it

src/DetailsView/components/Issues-details-pane.tsx
```diff
@@ -87,9 +87,22 @@ export class IssuesDetailsPane extends React.Component<IssuesDetailsPaneProps, I
...
+            },
+            element: {
+                id: result.id,
+                identifierName: 'Element path',
+                shortId: 'last selector',
```
karanbirsingh: map from data to last part of selector (existing code for this in issue filing string utils)

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -52,11 +52,21 @@ export interface UnifiedResult {
...
+export type UnifiedIdentifiers = {
+    id: string;
+    shortId: string;
+    identifierName: string;
+} & InstancePropertyBag;
+
+export type UnifiedResolution = {
+    howToFixSummary: string;
+} & InstancePropertyBag;
+
```
karanbirsingh: we should be careful which fields we add to ensure this doesn't become a union of 3 different platform property bags

PR #1352, opened 2019-10-03
Card footer, copy issue details menu item action

AhmedAbdoOrtiga:
#### Description of changes

- This PR should handle copy issue menu item action.
- The generated text now is just a place holder till we finish the unified data story.

#### Pull request checklist


src/DetailsView/details-view-initializer.ts
```diff
@@ -314,6 +314,7 @@ if (isNaN(tabId) === false) {
                 getPropertyConfigById: getPropertyConfiguration,
                 collapsibleControl: CardsCollapsibleControl,
                 cardInteractionSupport: allCardInteractionsSupported,
+                navigator: new NavigatorUtils(window.navigator),
```
haonliu: The construction of NavigatorUtils happens in three places in this file (line 202, 233, 317). Maybe we can store it in a variable and reuse it.

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -4,21 +4,39 @@ import { IPoint } from '@uifabric/utilities';
...
+import { NavigatorUtils } from '../../navigator-utils';
+import { WindowUtils } from '../../window-utils';
+import { Toast } from '../toast';
 import { kebabMenuButton } from './card-footer.scss';
 
+export type CardKebabMenuButtonDeps = {
+    windowUtils: WindowUtils;
+    issueDetailsTextGenerator: IssueDetailsTextGenerator;
+    detailsViewActionMessageCreator: DetailsViewActionMessageCreator;
+    navigator: NavigatorUtils;
```
haonliu: call this navigatorUtils might be better.

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -32,10 +50,24 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
...
                 {this.renderContextMenu()}
+                {this.renderToast()}
             </div>
         );
     }
 
+    public renderToast(): JSX.Element {
+        return (
+            <>
+                {' '}
```
haonliu: Why is this here?

src/common/navigator-utils.ts
```diff
@@ -30,4 +30,9 @@ export class NavigatorUtils {
 
         return userAgent.substring(versionOffset + versionPrefix.length + 1).split(' ')[0];
     }
+
+    public copyToClipboard(data: string): void {
```
haonliu: Seems this should return a promise.

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -71,8 +103,14 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
...
 
-    private copyFailureDetails = (): void => {
-        // todo
+    private copyFailureDetails = (event: React.MouseEvent<any>): void => {
+        const onClick = this.props.deps.detailsViewActionMessageCreator.copyIssueDetailsClicked;
+        if (onClick) {
+            onClick(event);
+        }
+        // tslint:disable-next-line: no-floating-promises
+        this.props.deps.navigator.copyToClipboard('The quick brown fox jumps over the lazy dog'); // to be changed when we finish the new data format and builder
```
haonliu: make this a promise call and use await to make sure that the toast appears after the copy action finishes.

src/reports/report-html-generator.tsx
```diff
@@ -50,6 +50,10 @@ export class ReportHtmlGenerator {
                 getGuidanceTagsFromGuidanceLinks: this.getGuidanceTagsFromGuidanceLinks,
                 getPropertyConfigById: this.getPropertyConfiguration,
                 cardInteractionSupport: noCardInteractionsSupported,
+                windowUtils: null,
```
haonliu: I understand why these values are null, but it would be nice if we can come up with a way to avoid this.

src/common/components/cards/card-kebab-menu-button.tsx
AhmedAbdoOrtiga: this can't be promise because onClick should be either boolean or void but I will find another solution for it

src/reports/report-html-generator.tsx
AhmedAbdoOrtiga: I think it's ok, the values should never be used in the report, so we should be fine

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -71,8 +102,15 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
...
     };
 
-    private copyFailureDetails = (): void => {
-        // todo
+    private copyFailureDetails = (event: React.MouseEvent<any>): void => {
+        const onClick = this.props.deps.detailsViewActionMessageCreator.copyIssueDetailsClicked;
+        if (onClick) {
+            onClick(event);
+        }
+        // tslint:disable-next-line: no-floating-promises
```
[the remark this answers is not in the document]
AhmedAbdoOrtiga: That's not the case, we won't be getting errors, the onClick function should return void or boolean
but when we call that async function, we either have to change it to return promise<void> which we can't as I mentioned above, or just disable the tslint.
so there's no error would be thrown, still if you think we can handle it in a better way, let's talk about it

src/common/navigator-utils.ts
```diff
@@ -30,4 +30,8 @@ export class NavigatorUtils {
 
         return userAgent.substring(versionOffset + versionPrefix.length + 1).split(' ')[0];
     }
+
+    public async copyToClipboard(data: string): Promise<void> {
```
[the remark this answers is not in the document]
AhmedAbdoOrtiga: I can change it, but I think promise means it's async function

src/tests/unit/tests/common/components/cards/card-kebab-menu-button.test.tsx
```diff
@@ -1,19 +1,79 @@
...
+
+    it('render ContextualMenu', () => {
+        const rendered = shallow<CardKebabMenuButton>(<CardKebabMenuButton {...defaultProps} />);
+        rendered.find(ActionButton).simulate('click', event);
+        expect(rendered.getElement()).toMatchSnapshot();
+        expect(rendered.state().target).toBe(event.currentTarget);
+    });
+
+    it('should click copy failure details and show the toast', () => {
+        // navigatorUtilsMock.setup((navigator: any) => navigator.then).returns(() => {});
```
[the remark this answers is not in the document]
AhmedAbdoOrtiga: I mentioned in the PR description that I'm still working on this test case, but I'm done, it should be fixed in the next commit.

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -71,8 +102,18 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
...
+        if (onClick) {
+            onClick(event);
+        }
+
+        this.props.deps.navigatorUtils
+            .copyToClipboard('The quick brown fox jumps over the lazy dog') // to be changed when we finish the new data format and builder
+            .then(() => {
+                this.setState({ showingCopyToast: true });
+            })
+            .catch();
```
smoralesd: add `catch()` doesn't fully address the original comment here

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -71,15 +106,27 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
         // todo
     };
 
-    private copyFailureDetails = (): void => {
-        // todo
+    private copyFailureDetails = (event: React.MouseEvent<any>): void => {
+        const onClick = this.props.deps.detailsViewActionMessageCreator.copyIssueDetailsClicked;
```
smoralesd: Why are we checking `detailsViewActionMessageCreator.copyIssueDetailsClicked` is not null here? is that even possible?

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -71,15 +106,27 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
...
+        if (onClick) {
+            onClick(event);
+        }
+
+        this.props.deps.navigatorUtils
+            .copyToClipboard('The quick brown fox jumps over the lazy dog') // to be changed when we finish the new data format and builder
+            .then(() => {
+                this.setState({ showingCopyToast: true });
+            })
+            .catch(() => {
```
smoralesd: this could be further improve like this:
```typescript
.catch(error => this.logger.error("Coudnpt copy failure details!", error));
```

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -4,21 +4,43 @@ import { IPoint } from '@uifabric/utilities';
...
+    issueDetailsTextGenerator: IssueDetailsTextGenerator;
+    detailsViewActionMessageCreator: DetailsViewActionMessageCreator;
+    navigatorUtils: NavigatorUtils;
+};
+
 export interface CardKebabMenuButtonState {
     isContextMenuVisible: boolean;
     target?: HTMLElement | string | MouseEvent | IPoint | null;
+    showingCopyToast: boolean;
+    showNeedsSettingsContent: boolean;
```
smoralesd: `showNeedsSettingsContent` property on the state is only being set on the constructor but never actually used. Should be removed until we actually need it.

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -71,15 +106,27 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
         // todo
     };
 
-    private copyFailureDetails = (): void => {
-        // todo
+    private copyFailureDetails = (event: React.MouseEvent<any>): void => {
```
smoralesd: I can't find usages of `copyFailureDetails `... it's private so only place that I should find usages is this same class but I don't see any... am I missing something?

src/tests/unit/tests/common/components/cards/card-kebab-menu-button.test.tsx
```diff
@@ -1,19 +1,83 @@
...
+
+        rendered
+            .find(ContextualMenu)
+            .prop('items')
+            .find(elem => elem.key === 'copyfailuredetails')
+            .onClick(event);
+
+        await copyToClipboardPromise;
+
+        expect(rendered.state().showingCopyToast).toBe(true);
```
smoralesd: looks like this tests is missing an expectation for a snapshot of the `Toast` component... I had a look at this: `src\tests\unit\tests\common\components\copy-issue-details-button.test.tsx`

src/common/components/cards/card-kebab-menu-button.tsx
AhmedAbdoOrtiga: it's used in getMenuItems(), the copy failure item onClick

smoralesd: Missed that the first time

PR #1359, opened 2019-10-03
Refactor CardSelectionStore interface and add unit tests

RobGallo:
#### Description of changes

- Refactor the store to use a dictionary of rule ids and result instance ids as the keys, rather than using arrays with the ids as fields. This makes it more efficient in code to locate a given rule or card to be expanded or selected.
- Implement the scanCompleted, toggleExpandCollapseRule, and collapseAllRules functions
- create unit tests for the above three functions

The toggleCardSelection function remains undone only to make this PR slightly smaller.

#### Pull request checklist

### DIFF
--- a/src/background/stores/card-selection-store.ts
+++ b/src/background/stores/card-selection-store.ts
@@ -42,7 +42,7 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
     };
 
     private toggleRuleExpandCollapse = (payload: RuleExpandCollapsePayload): void => {
-        if (!payload || !payload.ruleId || !this.state.rules || !this.state.rules[payload.ruleId]) {
+        if (!payload || !payload.ruleId || !this.state.rules[payload.ruleId]) {
             return;
         }
 
--- a/src/background/stores/card-selection-store.ts
+++ b/src/background/stores/card-selection-store.ts
@@ -1,8 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { StoreNames } from '../../common/stores/store-names';
-import { CardSelectionStoreData } from '../../common/types/store-data/card-selection-store-data';
-import { CardSelectionPayload, RuleExpandCollapsePayload, UnifiedScanCompletedPayload } from '../actions/action-payloads';
+import { CardSelectionStoreData, RuleExpandCollapseData } from '../../common/types/store-data/card-selection-store-data';
+import { RuleExpandCollapsePayload, UnifiedScanCompletedPayload } from '../actions/action-payloads';
 import { CardSelectionActions } from '../actions/card-selection-actions';
 import { UnifiedScanResultActions } from '../actions/unified-scan-result-actions';
 import { BaseStoreImpl } from './base-store-impl';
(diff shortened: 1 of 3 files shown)


src/background/stores/card-selection-store.ts
```diff
@@ -24,17 +24,72 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
...
+        }
 
-    public toggleCardSelection(payload: CardSelectionPayload): void {}
+        for (const resultInstanceUid of Object.keys(rule.cards)) {
+            rule.cards[resultInstanceUid] = false;
+        }
+    };
 
-    public CollapseAllRules(): void {}
+    public toggleRuleExpandCollapse = (payload: RuleExpandCollapsePayload): void => {
```
waabid: I don't believe this needs to be a public method.

src/background/stores/card-selection-store.ts
```diff
@@ -24,17 +24,72 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
...
+        if (!payload || !payload.ruleId || !this.state.rules || !this.state.rules[payload.ruleId]) {
```
waabid: The condition for "!this.state.rules" will always be true as the default is not null.

(Not sure if you're aware but an empty object will evaluate to true.)

src/background/stores/card-selection-store.ts
```diff
@@ -24,17 +24,72 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
...
-    private onScanCompleted = (payload: UnifiedScanCompletedPayload): void => {};
+        const rule = this.state.rules[payload.ruleId];
+
+        rule.isExpanded = !rule.isExpanded;
+
+        if (!rule.isExpanded) {
+            this.deselectAllCardsInRule(rule);
+        }
+
+        super.emitChanged();
```
waabid: The pattern we have for emit changed is to call it using "this" rather than "super"; doesn't really functionally matter (as far as I know) but I'd prefer to be consistent across our code.

src/background/stores/card-selection-store.ts
```diff
@@ -24,17 +24,72 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
...
+    };
+
+    public toggleCardSelection = (): void => {};
+
+    public CollapseAllRules = (): void => {
```
waabid: A couple nits:
1. This should not be public.
2. The casing should be camelCase.

src/background/stores/card-selection-store.ts
```diff
@@ -24,17 +24,72 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
...
+        for (const ruleId of Object.keys(this.state.rules)) {
```
waabid: I think this would be a good place to use ```forOwn``` from lodash.

src/background/stores/card-selection-store.ts
```diff
@@ -24,17 +24,72 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
...
+            const rule = this.state.rules[ruleId];
+            rule.isExpanded = false;
+            this.deselectAllCardsInRule(rule);
+        }
+
+        super.emitChanged();
+    };
+
+    private onScanCompleted = (payload: UnifiedScanCompletedPayload): void => {
```
waabid: In previous stores, we don't usually do a whole lot in the stores. If we are changing this pattern up here, which is fine, I'd like to come up with better descriptive names for the functions to explain what exactly they are doing. Since these private anyways, it's not going to matter to any consumers anyways, but would help in readability and understanding.

src/common/types/store-data/card-selection-store-data.ts
```diff
@@ -1,17 +1,11 @@
...
-export interface CardSelectionData {
-    resultInstanceUid: string;
-    isSelected: boolean;
-}
-
 export interface RuleExpandCollapseData {
-    ruleId: string;
     isExpanded: boolean;
-    cards: CardSelectionData[];
+    cards: { [resultInstanceUid: string]: boolean };
```
waabid: You can still declare a type with this and use that here.

```export type CardSelectionData = { [resultInstanceUid: string]: boolean };```

and then reference accordingly. Same goes for below.

File: src/tests/unit/tests/background/stores/card-selection-store.test.ts [hunk 8]
```diff
@@ -0,0 +1,250 @@
...
+            },
+        };
+
+        createStoreForUnifiedScanResultActions('scanCompleted')
+            .withActionParam(payload)
+            .testListenerToBeCalledOnce(initialState, expectedState);
+    });
+
+    test('ToggleRuleExpandCollapse expanded', () => {
+        const initialState: CardSelectionStoreData = {
```
waabid: I'm seeing some duplicate state declarations. Up to you but I think you could maybe create variables in a beforeEach loop instead to minimize that or maybe have a ```getState``` function which adjusts the properties accordingly to it's params. It should make the tests less verbose/remove duplication.

src/background/stores/card-selection-store.ts
RobGallo: Having this check allowed me to find the problem where the functions needed to be declared as arrow functions. So I feel somewhat validated for adding it. Thoughts?

src/background/stores/card-selection-store.ts
RobGallo: Done.

src/background/stores/card-selection-store.ts
RobGallo: Done.

src/background/stores/card-selection-store.ts
RobGallo: Done.

src/background/stores/card-selection-store.ts
RobGallo: I was looking for something like forOwn, but I didn't know what it was. Thanks!

src/background/stores/card-selection-store.ts
waabid: We usually do not put stuff like this, that validates whether something is an arrow function or otherwise, in our code, so I'd err on the side on not including it but up to you.

Might actually make sense to discuss briefly in an engineering discussion if you feel strongly otherwise.

RobGallo: Made the change.

src/background/stores/card-selection-store.ts
RobGallo: I hesitate to make that change because the name is identical to the function in the unified result store. Since they both listen for the same action, I'd like to keep them aligned.

src/common/types/store-data/card-selection-store-data.ts
RobGallo: Done.

src/tests/unit/tests/background/stores/card-selection-store.test.ts
RobGallo: Fixed.

src/background/stores/card-selection-store.ts
waabid: That's fine; I do think that is not fully necessary since we understand due to line 22, what action this function is linked to (scanComplete). 

The only issue is, that as is, this is not easily parse-able as to what this code is doing. Another approach would be to move some of the code to a more appropriately named private function (which can identify what this code is actually doing).

PR #1360, opened 2019-10-03
Add 'formatAsCode' property to HowToFix property bag, update tests

DaveTryon:
#### Description of changes

The cards feature calls for certain portions of the text to be rendered in a style similar to markdown's backtick styling (example: "Set the `isImportantForAccessibility` attribute to false"), but we currently have no way to apply intra-string styling. We discussed various options, including what the web UI currently does for Axe-core results. This PR takes the approach of treating the styling as "hints" that are specified by the data provider (in this case, the converter), and those get applied (as appropriate) at the data consumer (in this case, the cards UI). In essence, the data provider is using the property bag to specify input parameters to the consumer, and the consumer owns the actual application of the parameters. This puts the authoring into one place, and the consuming into one (different) place, so we have loose coupling. It also makes it easy to create per-rule custom formatting.

#### Pull request checklist

### DIFF
--- a/src/electron/platform/android/rule-information-provider.ts
+++ b/src/electron/platform/android/rule-information-provider.ts
@@ -27,11 +27,13 @@ export class RuleInformationProvider {
                 () =>
                     this.buildHowtoFixPropertyBag(
                         'The view is active but has no name available to assistive technologies. Provide a name for the view using its contentDescription, hint, labelFor, or text attribute (depending on the view type)',
+                        ['contentDescription', 'hint', 'labelFor', 'text'],
                     ),
             ),
             ImageViewName: new RuleInformation('ImageViewName', 'Meaningful images must have alternate text.', () =>
                 this.buildHowtoFixPropertyBag(
                     'The image has no alternate text and is not identified as decorative. If the image conveys meaningful content, provide alternate text using the contentDescription attribute. If the image is decorative, give it an empty contentDescription, or set its isImportantForAccessibility attribute to false.',
+                    ['contentDescription', 'isImportantForAccessibility'],
                 ),
             ),
             EditTextValue: new RuleInformation(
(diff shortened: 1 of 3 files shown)


src/electron/platform/android/rule-information-provider.ts
```diff
@@ -27,11 +27,13 @@ export class RuleInformationProvider {
                 () =>
                     this.buildHowtoFixPropertyBag(
                         'The view is active but has no name available to assistive technologies. Provide a name for the view using its contentDescription, hint, labelFor, or text attribute (depending on the view type)',
+                        ['contentDescription', 'hint', 'labelFor', 'text'],
```
DaveTryon: The consumer would do some sort of replacement to apply the appropriate formatting. This specific approach is subject to false positives (for example, if "text" appeared in the string without being a property), but since we control the text and the styling, it's probably OK. If we wanted to, we could go with an approach that used offsets into the string--that would be less clear but potentially safer. I opted to start with the simple approach to start with, but I'm happy to revisit that approach...

devaradhanm: how about using some kind of place holder. like this -> <<contentDescription>>

Now, if you pass codeAsString as contentDescription, it will try to find that string which is under <<>>?

devaradhanm: (It doesnt have to be <<>>. )

devaradhanm: other than this concern, LGTM

DaveTryon: I considered that approach, although it's really creating a pseudo-markup language, complete with escape sequences around whatever characters create the markup. It also means that the simplest consumers (those that just want the plain text) need to do extra work. We may indeed reach a point where this is called for, but I'd rather start simple and add the extra complexity if/when we need it. The keys-based routing that exists makes that transition fairly easy to manage, but this feels like a case of YAGNI.

PR #1370, opened 2019-10-04
refactor: create new classes with smaller scope in responsibilities to eventually replace client-view-controller

waabid:
#### Description of changes

This PR breaks apart the client view controller into another 4 pieces and they are as such:

1. Function isVisualizationEnabled: determine's whether a visualization is enabled or disabled based on the data provided.

2. Function visualizationNeedsUpdate: determine whether a visualization needs to be updated based on data provided.

3. Object VisualizationUpdateChangeHandler: this is the callback that listens to the store hub changes and calls the update visualization callback on all visualization types provided.

4. Object TargetPageVisualizationUpdater: for a visualization, this will update the visualization's state in the target page.

After this PR is complete, I will make a change to start relying on these new classes and remove the client-view-controller.

#### Pull request checklist

### DIFF
--- a/src/injected/client-view-controller.ts
+++ b/src/injected/client-view-controller.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import * as _ from 'lodash';
+import { cloneDeep, isEqual } from 'lodash';
+
 import { BaseStore } from '../common/base-store';
 import { TestMode } from '../common/configs/test-mode';
 import { VisualizationConfiguration } from '../common/configs/visualization-configuration';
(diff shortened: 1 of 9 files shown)


src/injected/client-view-controller.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import * as _ from 'lodash';
+
```
smoralesd: :sunglasses:

src/injected/is-visualization-enabled.ts
```diff
@@ -0,0 +1,33 @@
...
+export const isVisualizationEnabled: IsVisualizationEnabledCallback = (
+    config: VisualizationConfiguration,
+    step: string,
+    visualizationState: VisualizationStoreData,
+    assessmentState: AssessmentStoreData,
+    tabState: TabStoreData,
+) => {
+    const scanData = config.getStoreData(visualizationState.tests);
+    return (
+        config.getTestStatus(scanData, step) &&
```
smoralesd: **not for this PR** but since I haven't work on the visualization/drawing logic in a while... I found `config.getTestStatus` name confusing... is returning a boolean, I'm guessing this has to do with the test being enabled or not, which, if true, points to the proper name for this function.

PR #1371, opened 2019-10-04
Refactor issue filing to enable multiple result data types

karanbirsingh:
#### Description of changes

This PR follows up from #1351 . The current issue filing implementation assumes the issue-details type contains axe-core result data. This needs to change if we want to reuse the plumbing for the upcoming platform-agnostic data.
- `CreateIssueDetailsTextData` only holds the fields issue filing needs to create a bug, no more or less
- Each consumer of issue filing will convert their data-type (axe-core data or unified-data) to `CreateIssueDetailsTextData` (the cards UI will soon need to do this with unified-data)

#### Pull request checklist

### DIFF
--- a/src/issue-filing/common/create-issue-details-builder.ts
+++ b/src/issue-filing/common/create-issue-details-builder.ts
@@ -26,7 +26,7 @@ export const createIssueDetailsBuilder = (markup: MarkupFormatter): IssueDetails
             link(data.targetApp.url, data.targetApp.name),
             sectionSeparator(),
 
-            sectionHeader('Element'),
+            sectionHeader('Element path'),
             sectionHeaderSeparator(),
             data.element.identifier,
             sectionSeparator(),
--- a/src/issue-filing/common/create-issue-details-builder.ts
+++ b/src/issue-filing/common/create-issue-details-builder.ts
@@ -9,34 +9,33 @@ import { MarkupFormatter } from './markup/markup-formatter';
 
 export const createIssueDetailsBuilder = (markup: MarkupFormatter): IssueDetailsBuilder => {
     const getter = (environmentInfo: EnvironmentInfo, data: CreateIssueDetailsTextData): string => {
-        const result = data.ruleResult;
-
         const { howToFixSection, link, sectionHeader, snippet, sectionHeaderSeparator, footerSeparator, sectionSeparator } = markup;
+
+        const snippetSection = data.snippet
+            ? [sectionHeader('Snippet'), sectionHeaderSeparator(), snippet(data.snippet), sectionSeparator()]
+            : null;
 
         const lines = [
             sectionHeader('Issue'),
             sectionHeaderSeparator(),
-            `${snippet(result.help)} (${link(result.helpUrl, result.ruleId)})`,
+            `${snippet(data.rule.description)} (${link(data.rule.url, data.rule.id)})`,
             sectionSeparator(),
 
             sectionHeader('Target application'),
             sectionHeaderSeparator(),
-            link(data.pageUrl, data.pageTitle),
+            link(data.targetApp.url, data.targetApp.name),
             sectionSeparator(),
 
-            sectionHeader('Element path'),
+            sectionHeader('Element'),
             sectionHeaderSeparator(),
-            data.ruleResult.selector,
+            data.element.identifier,
             sectionSeparator(),
 
-            sectionHeader('Snippet'),
-            sectionHeaderSeparator(),
-            snippet(result.snippet),
-            sectionSeparator(),
+            ...snippetSection,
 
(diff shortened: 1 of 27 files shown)


src/issue-filing/common/create-issue-details-builder.ts
```diff
@@ -9,34 +9,33 @@ import { MarkupFormatter } from './markup/markup-formatter';
...
             sectionSeparator(),
 
             sectionHeader('Target application'),
             sectionHeaderSeparator(),
-            link(data.pageUrl, data.pageTitle),
+            link(data.targetApp.url, data.targetApp.name),
             sectionSeparator(),
 
-            sectionHeader('Element path'),
+            sectionHeader('Element'),
```
karanbirsingh: This is the only part of this PR that affects the UI - Fer suggested it as a platform-agnostic label. May move into separate PR

src/issue-filing/common/issue-filing-url-string-utils.ts
```diff
@@ -15,9 +15,7 @@ const getTitle = (data: CreateIssueDetailsTextData): string => {
         prefix = prefix + ': ';
     }
 
-    const selectorLastPart = getSelectorLastPart(data.ruleResult.selector);
-
-    return `${prefix}${data.ruleResult.help} (${selectorLastPart})`;
+    return `${prefix}${data.rule.description} (${data.element.conciseName})`;
 };
 
 const getSelectorLastPart = (selector: string): string => {
```
karanbirsingh: In a different PR we may want to move `getSelectorLastPart` out of this file because the other functions are platform-agnostic whereas this is web-specific.

src/issue-filing/rule-result-to-issue-filing-data.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
haonliu: There seems no unit test for this class?

karanbirsingh: Ah, thanks! mixed up my branches - let me port it over

src/tests/unit/tests/DetailsView/components/issues-details-pane.test.tsx
```diff
@@ -111,6 +112,9 @@ describe('IssuesDetailsPaneTest', () => {
         }
 
         const deps: IssuesDetailsPaneDeps = {
+            axeResultToIssueFilingDataConverter: {
```
haonliu: It would be better to mock this converter to return something meaningful. right now it renders undefined in the snapshot.

haonliu:
LGTM in general, left a few minor comments.

src/tests/unit/tests/background/issue-details-text-generator.test.ts
```diff
@@ -20,19 +20,23 @@ describe('Issue details text builder', () => {
...
-                help: 'RR-help',
-                html: 'RR-html',
-                ruleId: 'RR-rule-id',
-                helpUrl: 'RR-help-url',
-                selector,
-                snippet: 'RR-snippet   space',
-            } as any,
-        };
+            rule: {
+                description: 'RR-help',
```
lisli1: The values were probably kept the same to avoid updating the snapshot, but should we change the values to more closely match the new structure of the data? It might be confusing when looking at this test in the future why it says "RR" when it's not mapped to a `ruleResult` or "failureSummary" when the key is a `howToFixSummary`.

karanbirsingh: Good point! I think we should do it in a different PR. If the snapshot doesn't change in this PR it helps show that the behavior has not changed?

src/tests/unit/tests/common/message-creators/issue-filing-action-message-creator.test.ts
```diff
@@ -77,36 +76,23 @@ describe('IssueFilingActionMessageCreator', () => {
...
-                    messageType: Messages.IssueFiling.FileIssue,
-                    payload: {
-                        service: testService,
-                        issueData: issueDetailsData,
-                        telemetry,
-                    },
-                }),
-            ),
-        );
+        const issueDetailsData: CreateIssueDetailsTextData = {} as any;
```
lisli1: Just wondering, why are we casting `issueDetailsData as any` here but using `issueDetailsData: {} as CreateIssueDetailsTextData` in issue-filing-button.test?

karanbirsingh: I think both are similar & don't know of any meaningful differences - afaik it's preference / what nearby code has already that might determine what one chooses

PR #1375, opened 2019-10-07
Add title bar for electron main window

jalkire:
#### Description of changes
This PR adds (but does not use) the title bar for the electron main window. The design for the title bar has a few more buttons, but they are not used for this feature.
Screenshot:
image

#### Pull request checklist


src/electron/views/automated-checks/components/title-bar.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const TitleBar = NamedFC<TitleBarProps>('TitleBar', (props: TitleBarProps) => {
+    const currrentWindow = props.deps.currentWindow;
+    const minimize = () => currrentWindow.minimize();
+    const maximize = () => (currrentWindow.isMaximized() ? currrentWindow.restore() : currrentWindow.maximize());
+    const close = () => currrentWindow.close();
+
+    return (
+        <div className={titleBar}>
+            <BrandWhite />
+            <ActionButton
```
smoralesd: Looking at the `ActionButton` properties I noticed `ariaHidden={true}` and `tabIndex={-1}`.
I think that's correct, Fernanda & me check edge and teams and this looks like the expected behavior) but I'm worried about keyboard shortcuts for this actions. I'm guessing closing the windows should just work; alt + F4 looks universal enough (on Windows) that I'm not too worried about but I'm wondering about how to minimize and maximize.

src/electron/views/automated-checks/components/title-bar.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+export interface TitleBarProps {
+    deps: TitleBarDeps;
+}
+
+export const TitleBar = NamedFC<TitleBarProps>('TitleBar', (props: TitleBarProps) => {
+    const currrentWindow = props.deps.currentWindow;
+    const minimize = () => currrentWindow.minimize();
+    const maximize = () => (currrentWindow.isMaximized() ? currrentWindow.restore() : currrentWindow.maximize());
```
devaradhanm: how about using store here instead? windowstateStore? & let WindowFrameUpdater deal with window updates?

smoralesd: That's a very good point... not sure if the the current `WindowFrameUpdated` implementation handles restore/maximize logic as we have here...

src/electron/views/automated-checks/components/title-bar.tsx
jalkire: At least on windows, we've got win + down/up arrow which transition between max/restore/min as well as alt + space which gives all the options in a menu.

src/electron/views/automated-checks/components/title-bar.tsx
devaradhanm: we can improve the windowFrameUpdater, so that window state is maintained in one place.

jalkire: Sounds like `WindowFrameUpdater` is the way to go. Do we want to get that into this PR, or bring it in as a followup PR to keep PRs more concise?

smoralesd: I think it should be part of this PR. @devaradhanm ?

jalkire: Updated to use `WindowFrameUpdater` + `WindowStateStore` for changes to window state

src/electron/window-frame-updater.ts
```diff
@@ -2,20 +2,42 @@
...
+            if (this.browserWindow.isMaximized()) {
+                return;
+            }
             this.browserWindow.setSize(600, 391);
         } else {
             this.browserWindow.maximize();
         }
     };
+
+    private onWindowStateChange = async (store: WindowStateStore) => {
```
smoralesd: does this needs to be `async` ?

src/electron/window-frame-updater.ts
```diff
@@ -2,20 +2,42 @@
...
     constructor(private readonly windowStateStore: WindowStateStore, private readonly browserWindow: BrowserWindow) {}
 
     public initialize(): void {
+        this.windowStateStore.addChangedListener(this.onWindowStateChange);
         this.windowStateStore.addChangedListener(this.onRouteChange);
     }
 
-    private onRouteChange = async (state: WindowStateStoreData) => {
-        if (state.routeId === 'deviceConnectView') {
+    private onRouteChange = async (store: WindowStateStore) => {
```
smoralesd: good catch on the `state` vs `store` ... 
seems that this functions doesn't need to be `async` though... is that correct?

src/tests/unit/tests/electron/flux/action-creator/window-state-action-creator.test.ts
```diff
@@ -28,4 +29,17 @@ describe(WindowStateActionCreator, () => {
 
         setRouteActionMock.verifyAll();
     });
+
+    it('calling setRoute invokes setWindowState action with given payload', () => {
```
smoralesd: test name seems off here (it mention calling `setRoute`)

src/tests/unit/tests/electron/flux/store/window-state-store.test.ts
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { RoutePayload } from 'electron/flux/action/route-payloads';
-import { WindowStateActions } from 'electron/flux/action/window-state-actions';
-import { WindowStateStore } from 'electron/flux/store/window-state-store';
-import { WindowStateStoreData } from 'electron/flux/types/window-state-store-data';
-
+import { WindowStatePayload } from 'electron/flux/action/window-state-payload';
+import { RoutePayload } from '../../../../../../electron/flux/action/route-payloads';
```
smoralesd: all imports from the `src/electron` folder should be non-relative (as before the changes)
We change a setting to make this easier for everyone under `.vscode\settings.json`, property `    "typescript.preferences.importModuleSpecifier": "non-relative"`; you can check if you already had that or update from master maybe?

src/electron/window-frame-updater.ts
jalkire: Hmm, it doesn't seem like it does. Will remove.

src/electron/window-frame-updater.ts
jalkire: Same response as above

src/tests/unit/tests/electron/flux/action-creator/window-state-action-creator.test.ts
jalkire: Copy paste strikes again

src/tests/unit/tests/electron/flux/store/window-state-store.test.ts
jalkire: I think this might have been a remnant from a merge from master.

src/electron/window-frame-updater.ts
```diff
@@ -2,20 +2,42 @@
...
     constructor(private readonly windowStateStore: WindowStateStore, private readonly browserWindow: BrowserWindow) {}
 
     public initialize(): void {
+        this.windowStateStore.addChangedListener(this.onWindowStateChange);
         this.windowStateStore.addChangedListener(this.onRouteChange);
     }
 
-    private onRouteChange = async (state: WindowStateStoreData) => {
-        if (state.routeId === 'deviceConnectView') {
+    private onRouteChange = (store: WindowStateStore) => {
```
devaradhanm: to avoid this mistake in future, can you update the signature of addChangedListener? let the parameter type be (store: T) => void instead of function

smoralesd: I'll take this on a separated PR

src/common/flux/store.ts
```diff
@@ -3,14 +3,14 @@
 import { EventHandlerList } from './event-handler-list';
 
 export class Store {
-    private changedHandlers = new EventHandlerList();
+    private changedHandlers = new EventHandlerList<Store, unknown>();
 
-    public addChangedListener(handler: Function): void {
-        this.changedHandlers.subscribe(handler as any);
+    public addChangedListener(handler: (store: Store, args?: unknown) => void): void {
```
devaradhanm: i ended up using args here to avoid fixing tests on listener expectations.

PR #1383, opened 2019-10-07
feature: link card click behavior with card-selection-store and create relevant files

Shobhit1:
#### Description of changes

This PR is part of incremental changes; it creates card-selection-message-creator; action creator and other relevant files that are required to link the click event on the card to update the card-selection-store.

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes AD #1607403


src/common/components/cards/instance-details.tsx
```diff
@@ -44,7 +46,7 @@ export const InstanceDetails = NamedFC<InstanceDetailsProps>('InstanceDetails',
 
     return (
         <>
-            <table className={reportInstanceTable}>
+            <table className={reportInstanceTable} onClick={() => deps.cardSelectionMessageCreator.toggleCardSelection(result.uid)}>
```
waabid: Can we use a named function here? Might make this easier to debug in the future.

Shobhit1: good suggestion. Updated.

PR #1388, opened 2019-10-08
feat: add more components to accessibility checks view

smoralesd:
#### Description of changes

Add more content to the accessibility checks view. 

Not part of this PR:
- hook up the Cards UI
- handle the "scanning" state

#### Pull request checklist


src/electron/views/automated-checks/components/automated-checks-view.tsx
```diff
@@ -1,10 +1,18 @@
...
-export class AutomatedChecksView extends React.Component {
-    public render(): JSX.Element {
-        return <>automated checks</>;
-    }
-}
+import { NamedFC } from 'common/react/named-fc';
+import { CommandBar, CommandBarProps } from './command-bar';
+import { HeaderSection } from './header-section';
+
+export type AutomatedChecksViewProps = CommandBarProps;
```
dbjorge: We usually write props explicitly and leave the implicit pass-through for deps, no? Should we do that here?

smoralesd: Thats a good point but since this is not the final version of this component and we don't have any prop/dep directly use by `AutomatedChecksView`, I'll resolve the comment on a future PR when we add the title bar component and we actually wire up this component

PR #1391, opened 2019-10-08
refactor: remove unused properties (fingerprint and isVisible) from result interfaces in target page

waabid:
#### Description of changes

This removes properties that are not used in the result interfaces that we use in the target page. This should make things a bit easier to transform our new data model into these interfaces.

#### Pull request checklist


src/background/assessment-data-converter.ts
```diff
@@ -167,7 +167,7 @@ export class AssessmentDataConverter {
             isCapturedByUser: false,
             failureSummary: ruleResult.failureSummary,
             isVisualizationEnabled: false,
-            isVisible: elementAxeResult.isVisible,
+            isVisible: true,
```
waabid: I decided to leave the "isVisible" property in the test step result interface as a matter of not increasing the size of this PR even further; previously this would always be true anyways, so this made sense to me.

PR #1400, opened 2019-10-09
Handle copy failure

haonliu:
#### Description of changes
- When copy failed, show toast with text: Failed to copy Failure details. Please try again.
- Removed Contextual Menu and used menu props on action button component as a part of this PR, so that we do not have to handle the state of the contextual menu.
- Updated snapshots and unit tests.

image

#### Pull request checklist

### DIFF
--- a/src/common/components/cards/card-kebab-menu-button.tsx
+++ b/src/common/components/cards/card-kebab-menu-button.tsx
@@ -170,6 +170,6 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
     }
 
     private showToastWithFailureMessage(): void {
-        this.setState({ showingCopyToast: true, toastText: 'Failed to copy Failure details. Please try again.' });
+        this.setState({ showingCopyToast: true, toastText: 'Failed to copy failure details. Please try again.' });
     }
 }
--- a/src/common/components/cards/card-kebab-menu-button.tsx
+++ b/src/common/components/cards/card-kebab-menu-button.tsx
@@ -1,8 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { IPoint } from '@uifabric/utilities';
 import { ActionButton } from 'office-ui-fabric-react/lib/Button';
-import { ContextualMenu, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
+import { DirectionalHint, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
 import * as React from 'react';
 
 import { IssueDetailsTextGenerator } from '../../../background/issue-details-text-generator';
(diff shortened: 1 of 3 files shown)


src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -26,10 +25,9 @@ export type CardKebabMenuButtonDeps = {
 } & IssueFilingButtonDeps;
 
 export interface CardKebabMenuButtonState {
-    isContextMenuVisible: boolean;
     showNeedsSettingsContent: boolean;
-    target?: HTMLElement | string | MouseEvent | IPoint | null;
     showingCopyToast: boolean;
+    toastText: string;
```
AhmedAbdoOrtiga: I think it'd better if we extract the toast component in a separate and toasText should be a prop, that would help even refactoring with the copy-issue-details **button**

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -163,11 +161,15 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
...
+    }
+
+    private hideToast(): void {
+        this.setState({ showingCopyToast: false, toastText: '' });
+    }
 
-    private dismissDropdown(): void {
-        this.setState({ target: null, isContextMenuVisible: false });
+    private showToastWithFailureMessage(): void {
+        this.setState({ showingCopyToast: true, toastText: 'Failed to copy failure details. Please try again.' });
```
AhmedAbdoOrtiga: Can we check with Fer and Chelsea since this is a UI change

lisli1: I am also curious if we have asked Fer and Chelsea what they think of this failure message.

lisli1:
Does the UI look the same without Contextual Menu?

src/common/components/cards/card-kebab-menu-button.tsx
haonliu: yeah, I've checked with Fer yesterday. This is the text she told me yesterday.

src/common/components/cards/card-kebab-menu-button.tsx
haonliu: Agreed, but that itself could be a separate PR.

haonliu:
> Does the UI look the same without Contextual Menu?

Yes. This seems to be the new way to do a contextual menu in office fabric: https://developer.microsoft.com/en-us/fabric#/controls/web/contextualmenu

src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -170,11 +162,15 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
...
 
-    private openDropdown = (event): void => {
-        this.setState({ target: event.currentTarget, isContextMenuVisible: true });
-    };
+    private showToastWithSuccessMessage(): void {
+        this.setState({ showingCopyToast: true, toastText: 'Failure details copied.' });
+    }
+
+    private hideToast(): void {
+        this.setState({ showingCopyToast: false, toastText: '' });
```
dbjorge: Does the toast's disappearance involve a fade out animation, or is it immediate? If it isn't immediate, we might not want to empty the text here

haonliu: I tried remove the text state change here, there is no change to the dismiss animation. So my understanding is that there is no fade out animation. 
Also, compared with master, not seeing fade out animation.

PR #1401, opened 2019-10-09
Updated DeviceConnectView initializer and container to refer to root container

JGibson2019:
#### Description of changes

Moved the DeviceConnectViewInitializer to the views folder's root level and renamed it to renderer initializer; renamed deviceConnectView.html to index.html and moved it to the root level of view's folder. Updated selectors to now refer to #root-container

#### Pull request checklist

- [ N/A] Addresses an existing issue: Fixes #0000
- [ N/A] Added relevant unit test for your changes. (`yarn test`)
- [ X] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [ X] Ran precheckin (`yarn precheckin`)
- [N/A ] (UI changes only) Added screenshots/GIFs to description above
- [ N/A] (UI changes only) Verified usability with NVDA/JAWS


src/electron/main/main.ts
```diff
@@ -15,7 +15,7 @@ const createWindow = () => {
     });
 
     mainWindow
-        .loadFile(path.resolve(__dirname, '../electron/views/device-connect-view/deviceConnectView.html'))
```
smoralesd: any particular reason why we didn't move `index.html` to be directly under `views` folder?

JGibson2019: I did and I forgot to push the commit before opening the PR

webpack.config.js
```diff
@@ -55,7 +55,7 @@ const commonEntryFiles = {
 };
 
 const electronEntryFiles = {
-    deviceConnectView: [path.resolve(__dirname, 'src/electron/views/device-connect-view/device-connect-view-initializer.ts')],
+    deviceConnectView: [path.resolve(__dirname, 'src/electron/views/renderer-initializer.ts')],
```
smoralesd: should we change `deviceConnectView` here? Doing this will requiere to update the `script` tag under `index.html` file

JGibson2019: Done; it's now called renderer and the index.html file is updated

PR #1410, opened 2019-10-10
Responsive congrats component

AhmedAbdoOrtiga:
#### Description of changes

- Responsive congrats component.

**Normal**
image

**Re-flowed**
image

**High Contrast**
image

#### Pull request checklist


src/common/styles/colors.scss
```diff
@@ -233,7 +233,5 @@ $spinner-text: var(--spinner-text);
 // consistent colors
 $always-white: #ffffff;
 $always-black: #161616;
-$always-primary-text: rgba(0, 0, 0, 0.9);
```
smoralesd: was this added as part of your onging work??? I think `colors.scss` holds the "official" color palette and thus, I'm not sure we should remove anything from here...

src/reports/components/report-sections/no-failed-instances-congrats.scss
```diff
@@ -2,35 +2,24 @@
...
-    display: flex;
-    align-items: center;
-    justify-content: center;
-}
-
 .report-congrats-message {
-    width: 364px;
-    text-align: center;
+    word-break: break-all;
+    word-wrap: break-word;
```
peterdur: - Should `word-wrap` be `overflow-wrap` instead? `word-wrap` is an alias per [this MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) and it was hard to look this up until I figured this out.
- In the screenshot, the text is wrapping in the middle of "word", which is not usual behavior. This seems to be due to the use of `break-all`. Removing the `word-break: break-all` seems to restore usual behavior. Is there a scenario where we need the `break-all`?

src/reports/components/report-sections/no-failed-instances-congrats.tsx
```diff
@@ -3,27 +3,13 @@
...
-    reportCongratsMessage,
-    reportCongratsScreen,
-} from './no-failed-instances-congrats.scss';
+import { reportCongratsHead, reportCongratsInfo, reportCongratsMessage } from './no-failed-instances-congrats.scss';
 
 export const NoFailedInstancesCongrats = NamedFC('NoFailedInstancesCongrats', () => {
     return (
-        <div className={reportCongrats} key="report-congrats">
-            <div>
-                <InlineImage imageType={InlineImageType.AdaLaptop} alt="" />
```
smoralesd: did you check the fast pass html report? I think remove this `InlineImage` here will also make the image to not show on the report... is that intended?

src/reports/components/report-sections/no-failed-instances-congrats.scss
AhmedAbdoOrtiga: - sure we can, as I know word-wrap and overflow-wrap are identical.
- Yeah that's what Fer wants, and that's what we do in the report.

src/common/styles/colors.scss
AhmedAbdoOrtiga: this was added in a previous PR to fix the same issue, but then we decided to remove the whole Ada image, so we don't need it anymore.

src/reports/components/report-sections/no-failed-instances-congrats.scss
haonliu: I'm also a bit concerned about breaking in the middle of a word, it would be nice if we can avoid doing that.

AhmedAbdoOrtiga: As I mentioned before replying to the second point 
- Yeah that's what we want, I checked with Fer , and that's what we do in the report.

PR #1412, opened 2019-10-10
feat(android-cards-view): convert scan results and invoke unified scan complete action

smoralesd:
#### Description of changes

- expose `axeVersion` from `ScanResults` class to be used on the unified scan completed payload
- introduce UnifiedScanCompletedPayloadBuilder to convert scan results to proper payload
- introduce application properties getter
- on `ScanController`, when the scan succeed, converting the scan results and invoking the unified scan completed action

#### Pull request checklist

   - not affected
   - not affected

### DIFF
--- a/src/electron/platform/android/unified-result-builder.ts
+++ b/src/electron/platform/android/unified-result-builder.ts
@@ -13,6 +13,8 @@ import {
     ConvertScanResultsToUnifiedRulesDelegate,
 } from 'electron/platform/android/scan-results-to-unified-rules';
 
+export type UnifiedScanCompletedPayloadBuilder = (scanResults: ScanResults) => UnifiedScanCompletedPayload;
+
 export const createBuilder = (
     getUnifiedResults: ConvertScanResultsToUnifiedResultsDelegate,
     getUnifiedRules: ConvertScanResultsToUnifiedRulesDelegate,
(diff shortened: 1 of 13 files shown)


src/electron/platform/android/unified-result-builder.ts
```diff
@@ -0,0 +1,43 @@
...
+    const payload: UnifiedScanCompletedPayload = {
+        scanResult: getUnifiedResults(scanResults, uuidGenerator),
+        rules: getUnifiedRules(scanResults, uuidGenerator),
+        toolInfo: {
+            scanEngineProperties: {
+                name: 'axe-android',
+                version: scanResults.axeVersion,
+            },
+            applicationProperties: applicationPropertiesDelegate(),
+        },
```
jalkire: Is there a place where we might want to separate the tool info generation out to? I see that we use `EnvironmentInfoProvider` to generate similar properties for the `axe-core` results--is there a particular reason we would rather encapsulate that logic in here for android vs following a similar pattern to the 'axe-core' info? Definitely a minor thing, but it just feels a little off to me that we're hardcoding `name: 'axe-android'` into this file, which doesn't seem like it should know/care about the name of the scanning engine in my opinion.

smoralesd: I implement the change in this PR... I was a litte hesitant to do it because we need the `scanResults` object to get the `axeVersion` value but on comment from review seems enough to just do it.

PR #1416, opened 2019-10-10
[cards footer] update aria labels

lisli1:
#### Description of changes

- Added `ariaLabel` "More actions" to kebab menu button
- Replaced `ariaLabel` with `ariaHidden` on eye icon next to label "highlight unavailable" so screen readers don't read out "highlight unavailable" twice

#### Pull request checklist

- [x] Addresses an existing WI: # 1612315
- [x] (UI changes only) Verified usability with NVDA/JAWS **no UI changes but verified current changes with NVDA/JAWS**


src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -51,6 +51,7 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
         return (
             <div className={kebabMenuButton}>
                 <ActionButton
+                    ariaLabel="More actions"
```
dbjorge: We recently updated the label on a different menu button to include the term "menu"; maybe this should be "more actions menu" or similar instead? @ferBonnin, thoughts?

lisli1: JAWS reads it as "More actions button menu" and NVDA reads it as "More actions menu button subMenu". I think NVDA shouldn't read "subMenu" before it has been opened but I'm not sure about this.

ferBonnin: NVDA behaviour is odd, but if it already reads more actions button menu then its fine :)

PR #1424, opened 2019-10-10
feat(card-buttons): add highlight status icons using svgs generated from UI mocks

dbjorge:
#### Description of changes

Adds SVG icons for the different highlight states by exporting them from the figma mockups.

In the live extension right now you can only see the "unavailable" one since it's the only state that's hooked up. Below are screenshots where I've forced the other 2 states so we can see examples:

screenshot of highlight visible status

screenshot of highlight hidden status

screenshot of highlight unavailable status

#### Pull request checklist

- [x] Addresses an existing issue: 1612317
  - omitted tests pinning the icons because they are static/generated blobs, consistent with the other static icons nearby 


karanbirsingh:
The icons seem vertically aligned with the top of the text, and I wonder if we'd rather vertically center them within the text if possible. The current screenshots are aligned with our mocks, so it's as designed

dbjorge:
Chelsea and I agree, they look a little off even though they were exported directly from the mocks. Will investigate before merging.

dbjorge:
@karanbirsingh updated impl and screenshots with better vertical alignment

PR #1429, opened 2019-10-11
Manage Window State changes

devaradhanm:
#### Description of changes

WindowFrameUpdater => responsible for performing window operations via actions.
WindowFrameListener => listens to window state & updates store via action

Main changes are in WindowFrameUpdater, WindowFrameListener, TitleBar & RootContainerRenderer, WindowStateStore
(Other files were modified due to this refactoring + delegating action creators)

#### Pull request checklist

### DIFF
--- a/src/electron/window-frame-listener.ts
+++ b/src/electron/window-frame-listener.ts
@@ -1,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { BrowserWindow } from 'electron';
+import { WindowStateActionCreator } from './flux/action-creator/window-state-action-creator';
+
+export class WindowFrameListener {
+    constructor(private readonly windowStateActionsCreator: WindowStateActionCreator, private readonly browserWindow: BrowserWindow) {}
+
+    public initialize(): void {
+        this.browserWindow.on('minimize', this.onWindowStateChange);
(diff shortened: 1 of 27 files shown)


src/electron/window-frame-listener.ts
```diff
@@ -0,0 +1,24 @@
...
+import { WindowStateActionCreator } from './flux/action-creator/window-state-action-creator';
+
+export class WindowFrameListener {
+    constructor(private readonly windowStateActionsCreator: WindowStateActionCreator, private readonly browserWindow: BrowserWindow) {}
+
+    public initialize(): void {
+        this.browserWindow.on('minimize', this.onWindowStateChange);
+        this.browserWindow.on('maximize', this.onWindowStateChange);
+        this.browserWindow.on('unmaximize', this.onWindowStateChange);
+    }
```
jalkire: Looking over the [possible events](https://electronjs.org/docs/api/browser-window#instance-events), do you think we might want to listen to some of these?
`restored`
`enter-full-screen`
`leave-full-screen`

Full screen would only matter if that is the event that gets thrown on MacOS when the maximize equivalent is clicked.

devaradhanm: i havnt tested in mac. I will test in mac & get back if these events are needed. i don't need these events for windows, for our current usage.

devaradhanm: i will be sending another pr to address mac os.

PR #1431, opened 2019-10-11
Add getHighlightsFromCardSelectionStoreData function

RobGallo:
#### Description of changes

The function returns a list of unified result ids which can be mapped to elements to be highlighted.

The highlighting logic should be as follows:

- If no cards are selected, result uids for all cards in all expanded rules will be returned.
- If cards are selected, the result uids of selected cards will be returned, provided a given card is contained in an expanded rule. It should never happen that cards are selected in a rule that isn't expanded, but the logic of getHighlightsFromCardSelectionStoreData should still handle this possibility gracefully.
- If no rules are expanded, all uids will be returned.

Also, added a selectedCardCount field to CardSelectionStoreData. This makes it much easier to efficiently calculate which highlights to return. Otherwise, the logic of getHighlightsFromCardSelectionStoreData would need to either iterate the store data more than once, first counting the selected cards, or it would need to have more involved logic to analyze the state of the store data before deciding which result uids to return.

#### Pull request checklist

### DIFF
--- a/src/background/stores/card-selection-store.ts
+++ b/src/background/stores/card-selection-store.ts
@@ -26,6 +26,7 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
     public getDefaultState(): CardSelectionStoreData {
         const defaultValue: CardSelectionStoreData = {
             rules: {},
+            selectedCardCount: 0,
         };
 
         return defaultValue;
@@ -37,6 +38,10 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
         }
 
         forOwn(rule.cards, (isSelected, resultInstanceUid, cards) => {
+            if (cards[resultInstanceUid]) {
+                this.state.selectedCardCount -= 1;
+            }
+
             cards[resultInstanceUid] = false;
         });
     };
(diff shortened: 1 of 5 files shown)


src/background/stores/card-selection-store.ts
```diff
@@ -26,6 +26,7 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
     public getDefaultState(): CardSelectionStoreData {
         const defaultValue: CardSelectionStoreData = {
             rules: {},
+            selectedCardCount: 0,
```
dbjorge: This information can already be inferred from `rules`, no? Can `get-highlights-from-card-selection-store-data.ts` infer it rather than having to repeat state, eg, by updating lines 22-26 to

```
    const selectedResultUids = getOnlyResultUidsFromSelectedCards(storeData.rules, expandedRuleIds);
    if (selectedResultUids.length > 0) {
        return selectedResultUids;
    }

    return getAllResultUidsFromRuleIdArray(storeData.rules, expandedRuleIds);
```

RobGallo: Very nice. That's what I was looking for.

dbjorge: This line can probably be removed now

src/common/get-highlights-from-card-selection-store-data.ts
```diff
@@ -0,0 +1,73 @@
...
+        return getAllResultUids(storeData.rules);
+    }
+
+    const selectedResultUids = getOnlyResultUidsFromSelectedCards(storeData.rules, expandedRuleIds);
+    if (selectedResultUids.length > 0) {
+        return selectedResultUids;
+    }
+
+    return getAllResultUidsFromRuleIdArray(storeData.rules, expandedRuleIds);
+    return getOnlyResultUidsFromSelectedCards(storeData.rules, expandedRuleIds);
```
dbjorge: I think line 28 is dead code now with the most recent change

PR #1438, opened 2019-10-11
feat(android-cards-view): hook up scanning dialog

smoralesd:
#### Description of changes

Add scanning dialog and hook it up on the automated checks view.
Additionally, I had to fix how we generate the how to fix information for color contrast to handle the null case.

02 - scanning spinner

**Notes** 
- We change the design (and thus it doesn't match the mocks) after taking with @ferBonnin. There is an issue in using modal dialogs: the window bar is inside the html (it is not a native window bar basically) this means that using a modal make the window bar inaccessible with the mouse. To solve this, we decided to go with a similar design as we have on the browser extension, in the Fast Pass section when rendering automated checks.
- The spinner alerts (so the screen reader read the contents out loud) when the scan start, there is no alert for when the scan is done, we're following the pattern from the browser extension where the "scan finished" alert is the responsability of a different component (the one that shows the results, ergo, the Cards UI)

#### Pull request checklist


src/tests/unit/tests/electron/platform/android/rule-information-provider.test.ts
```diff
@@ -68,6 +68,13 @@ describe('RuleInformationProvider', () => {
         expect(unifiedResolution).toMatchSnapshot();
     });
 
+    test('getRuleInformation handles no foreground/background color values', () => {
+        const testRuleId: string = 'ColorContrast';
+        const ruleResult: RuleResultsData = buildColorContrastRuleResultObject('FAIL', 2.798498811425733, null, null);
```
DaveTryon: When the service has no colors, there's also no contrast ratio
```suggestion
        const ruleResult: RuleResultsData = buildColorContrastRuleResultObject('FAIL', null, null, null);
```

src/electron/platform/android/rule-information-provider.ts
```diff
@@ -50,14 +50,26 @@ export class RuleInformationProvider {
...
+            `The text element has insufficient contrast of ${ratio}. Foreground color: ${foreground}, background color: ${background}). Modify the text foreground and/or background colors to provide a contrast ratio of at least 4.5:1 for regular text, or 3:1 for large text (at least 18pt, or 14pt+bold).`,
         );
     };
 
+    private getColorValue(ruleResultsData: RuleResultsData, propertyName: string): string {
+        let result = 'NO VALUE AVAILABLE';
+        const value = ruleResultsData.props[propertyName] as string;
+
+        if (value) {
+            result = `#${value.substring(2)}`;
```
waabid: ```value.substring(2)``` is a bit unhelpful to read. Why the use of 2 here? Might be simplified with a named variable.

src/tests/unit/tests/electron/views/automated-checks/__snapshots__/automated-checks-view.test.tsx.snap
```diff
@@ -1,5 +1,40 @@
...
 
+exports[`AutomatedChecksView renders scanning spinner: automated checks view 1`] = `
+<React.Fragment>
+  <TitleBar
+    deps={
+      Object {
+        "currentWindow": null,
+        "deviceConnectActionCreator": null,
+        "scanActionCreator": proxy {
+          "___id": "BCDF5CE5-F0DF-40B7-8BA0-69DF395033C8",
```
DaveTryon: Potentially dumb question--where is this Guid coming from? Is it reproducible?

src/tests/unit/tests/electron/views/automated-checks/automated-checks-view.test.tsx
```diff
@@ -1,25 +1,47 @@
...
-                deviceConnectActionCreator: null,
-                currentWindow: null,
-                windowStateActionCreator: null,
-                scanActionCreator: Mock.ofType(ScanActionCreator).object,
-            },
-        } as AutomatedChecksViewProps;
+    describe('renders', () => {
+        it('the automated checks view', () => {
+            const props: AutomatedChecksViewProps = {
+                deps: {
```
waabid: I strongly prefer casting the deps object as the deps necessary in this class.

The reason for this is that as the deps underneath this component changes, you will have to modify this object as well, even if this specific component does not consume that dependency, which minimizes the utility of having deps in the first place. Instead, it's better to only mention the dependencies that this component consumes and then cast that object; this is not something you can miss either as if you consume a dependency and it is not defined in that object (you will miss the type error; no error at compile time), your test will at least still fail at run time.

In this case, the deps would only include scanActionCreator because that is the only consumed dependency in that class.

Let me know if that makes sense.

src/tests/unit/tests/electron/views/automated-checks/__snapshots__/automated-checks-view.test.tsx.snap
smoralesd: from typemoq... it's reproducible... if not, the snapshot test will fail on each run... I'm not aware how do typemoq achieve that though.

src/tests/unit/tests/electron/platform/android/rule-information-provider.test.ts
smoralesd: I did not got any error from `'Color Contrast Ratio'` being null (I scan 3 or 4 real android apps) that's why that's not being changed here. I'd rather not change it as this scenario will quickly dissapear when we do filter any color contrast issue with confidence value None

src/electron/platform/android/rule-information-provider.ts
smoralesd: updated

src/tests/unit/tests/electron/views/automated-checks/automated-checks-view.test.tsx
smoralesd: As talked offline: casting at the props level technically speaking also cast the deps level. I did remove the deps we don't need for this unit tests though! Thanks for the feedback.

PR #1439, opened 2019-10-11
refactor: renamed and moved view model to separate files for cards

waabid:
#### Description of changes

This moves interfaces/types that are really only for the cards view to a separate place for them specifically.

#### Pull request checklist

### DIFF
--- a/src/common/types/store-data/card-view-model.ts
+++ b/src/common/types/store-data/card-view-model.ts
@@ -1,3 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
 import { GuidanceLink } from '../../../scanner/rule-to-links-mappings';
 import { InstanceResultStatus, UnifiedResult } from './unified-data-interface';
 export type CardRuleResultStatus = InstanceResultStatus | 'inapplicable';
--- a/src/common/types/store-data/card-view-model.ts
+++ b/src/common/types/store-data/card-view-model.ts
@@ -1,0 +1,15 @@
+import { GuidanceLink } from '../../../scanner/rule-to-links-mappings';
+import { InstanceResultStatus, UnifiedResult } from './unified-data-interface';
+export type CardRuleResultStatus = InstanceResultStatus | 'inapplicable';
+export interface CardRuleResult {
+    id: string;
+    nodes: CardResult[];
+    description: string;
+    url: string;
+    guidance: GuidanceLink[];
+}
+export type CardRuleResultsByStatus = {
+    [key in CardRuleResultStatus]: CardRuleResult[];
+};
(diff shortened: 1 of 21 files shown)


src/common/types/store-data/card-view-model.ts
```diff
@@ -0,0 +1,15 @@
+import { GuidanceLink } from '../../../scanner/rule-to-links-mappings';
```
RobGallo: No copyright header.

src/common/types/store-data/card-view-model.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { GuidanceLink } from '../../../scanner/rule-to-links-mappings';
+import { InstanceResultStatus, UnifiedResult } from './unified-data-interface';
```
smoralesd: nitpick: some new lines in between this type/interface definitions would improve readability.

src/tests/unit/tests/common/components/cards/sample-view-model-data.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { UnifiedResult, UnifiedRuleResult, UnifiedStatusResults } from '../../../../../../common/types/store-data/unified-data-interface';
+import { CardRuleResult, CardRuleResultsByStatus } from '../../../../../../common/types/store-data/card-view-model';
```
smoralesd: nitpick: this is a good moment to change this imports to non-relative paths

src/common/types/store-data/card-view-model.ts
waabid: Yep! Noticed it earlier; it should be there now

src/common/types/store-data/card-view-model.ts
```diff
@@ -0,0 +1,15 @@
...
+    nodes: CardResult[];
+    description: string;
+    url: string;
+    guidance: GuidanceLink[];
+}
+export type CardRuleResultsByStatus = {
+    [key in CardRuleResultStatus]: CardRuleResult[];
+};
+export interface CardResult extends UnifiedResult {}
+export const AllRuleResultStatuses: CardRuleResultStatus[] = ['pass', 'fail', 'unknown', 'inapplicable'];
```
RobGallo: Is there no way to iterate the type? It will be tough to remember to update this value if other values are added to the enum.

waabid: Unfortunately, I don't believe there is. I don't suspect we'll be adding to this that frequently though.

PR #1425, opened 2019-10-10
refactor: toast maintains its own 'visible' state

karanbirsingh:
#### Description of changes

This PR shows an idea for how the toast component could maintain its own state. The unit tests need to be updated to work properly with the react reference.

The team is looking at alternatives for the toast in general, so leaving this up as a draft

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [ ] Added relevant unit test for your changes. (`yarn test`)
- [ ] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [ ] Ran precheckin (`yarn precheckin`)
- [ ] (UI changes only) Added screenshots/GIFs to description above
- [ ] (UI changes only) Verified usability with NVDA/JAWS

PR #1440, opened 2019-10-11
refactor: toast maintains its own 'visible' state

AhmedAbdoOrtiga:
#### Description of changes

This PR shows an idea for how the toast component could maintain its own state. The unit tests need to be updated to work properly with the react reference.

The team is looking at alternatives for the toast in general, so leaving this up as a draft

**This PR include the changes were made in [PR 1425](https://github.com/microsoft/accessibility-insights-web/pull/1425)**

#### Pull request checklist


src/tests/unit/tests/common/components/cards/card-kebab-menu-button.test.tsx
```diff
@@ -245,13 +251,17 @@ describe('CardKebabMenuButtonTest', () => {
 
         expect(rendered.state().showNeedsSettingsContent).toBe(true);
 
-        issueFilingActionMessageCreatorMock.verifyAll();
+        verifyMocks([issueFilingActionMessageCreatorMock, issueFilingServiceProviderMock]);
     });
 
-    function getMenuItemWithKey(rendered: ShallowWrapper, itemKey: string): IContextualMenuItem {
+    function getMenuItemWithKey(rendered: any, itemKey: string): IContextualMenuItem {
```
dbjorge: Instead of `any`, `ReactWrapper | ShallowWrapper`?

src/tests/unit/tests/common/components/copy-issue-details-button.test.tsx
```diff
@@ -28,15 +37,35 @@ describe('CopyIssueDetailsButtonTest', () => {
...
-        expect(result.getElement()).toMatchSnapshot();
-        onClickMock.verifyAll();
+        // tslint:disable-next-line: await-promise
+        await button.simulate('click');
+
+        const toast = result.find(Toast);
+
+        expect(toast.state().toastVisible).toBe(true);
+        expect(toast.state().content).toBe('Failure details copied.');
+        expect(result.debug()).toMatchSnapshot();
```
dbjorge: I think maybe it would be better to snapshot the `toast.props()` rather than `result.debug()` - that would avoid this unit test's snapshot depending on the implementation details of the render methods of `Toast` and `DefaultButton` (they have their own unit tests for that :) )

AhmedAbdoOrtiga: - if I use the props to snapshot the toast, the snapshot result isn't that informative for example:
`const toast = result.find(Toast);
const toastResult = Enzyme.mount(<Toast {...toast.props()} />);
expect(toastResult.debug()).toMatchSnapshot();`
the result is 
`exports[CopyIssueDetailsButtonTest render after click shows toast 2] = "<Toast deps={{...}} timeoutLength={6000} />";
- We use debug instead of getElement because getElement  doesn't work while it's expected to do, we do it in other files like start-over-drop-down.test I think
`

PR #1441, opened 2019-10-11
Contextual menu fit and finish

haonliu:
#### Description of changes
- Update the background color of the contextual menu to match the mock.
- Update the background color of the menu button on hover to match the mock.
- Update the background color of the menu button on pressed to match the mock.
- Give the contextual menu a 4px border radius.
- Update the box shadow of the menu to match the mock.

Figma - hovered and pressed:
image

Screenshot of the change:
hover:
image

pressed:

image

#### Pull request checklist

### DIFF
--- a/src/common/components/cards/card-kebab-menu-button.scss
+++ b/src/common/components/cards/card-kebab-menu-button.scss
@@ -2,9 +2,31 @@
 // Licensed under the MIT License.
 @import '../../styles/colors.scss';
 
+.kebab-menu-callout {
+    box-shadow: 0px 6.4px 14.4px $box-shadow-132, 0px 1.2px 3.6px $box-shadow-108 !important;
+    border-radius: 4px;
+    border-width: 0px !important;
+    div.ms-Callout-main {
+        border-radius: 4px;
+    }
+}
+
 .kebab-menu {
+    background: $acrylic-light;
     i {
         color: $primary-text;
+    }
+    li {
+        background-color: $acrylic-light;
+    }
+
+    button.ms-ContextualMenu-link {
+        &:hover {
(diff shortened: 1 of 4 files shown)


src/common/styles/colors.scss
```diff
@@ -62,6 +62,9 @@
     --neutral-outcome: var(--neutral-60);
     --incomplete-color: var(--neutral-60);
     // Additional colors
+    --acrylic-light: #f6f6f6;
```
dbjorge: Should the new colors have high contrast specializations?

src/common/components/cards/card-kebab-menu-button.scss
```diff
@@ -2,10 +2,32 @@
 // Licensed under the MIT License.
 @import '../../styles/colors.scss';
 
+.kebab-menu-callout {
+    box-shadow: 0px 6.4px 14.4px $box-shadow-132, 0px 1.2px 3.6px $box-shadow-108 !important;
```
dbjorge: Can we use a `elementType.className` style instead to avoid the `!important` (friendlier to user stylesheets)?

src/common/components/cards/card-kebab-menu-button.scss
```diff
@@ -2,10 +2,32 @@
...
+    border-radius: 4px;
+    border-width: 0px !important;
+    div.ms-Callout-main {
```
waabid: **Important note:** this will not work once it goes through webpack. The classes that are not controlled by us need to be utilized with only the global modifier. You can look at failure-instance-panel.scss as an example.

waabid: This also applies to the css below.

PR #1443, opened 2019-10-11
Add maximize/restore button for title bar

jalkire:
#### Description of changes
The restore icon for the electron title bar is not part of our office fabric icon set, so this PR adds it as an SVG and uses said icon in a new MaximizeRestoreButton, which displays either a maximize or restore icon. Once #1429 goes in, this should be hooked up.

Restore icon: 
image

image

#### Pull request checklist


src/electron/views/automated-checks/components/maximize-restore-button.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export interface MaximizeRestoreButtonProps {
+    isMaximized: boolean;
+    onClick: () => void;
+}
+
+export const MaximizeRestoreButton = NamedFC<MaximizeRestoreButtonProps>('TitleBar', (props: MaximizeRestoreButtonProps) => {
+    const content = props.isMaximized ? null : <RestoreIcon />;
```
dbjorge: `ActionButton` also has an `onRenderIcon` prop you could assign `RestoreIcon` to; would that be better? I'm not whether it makes a difference but I'm suspicious about why it'd be provided if there was no difference

src/electron/views/automated-checks/components/maximize-restore-button.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const icon = props.isMaximized ? { iconName: 'Stop' } : null;
```
dbjorge: It looks like we refer to the icons in camelCase (eg, `'stop'`) in most other places - should probably be consistent (not sure if there's any linux-specific bugs lurking here)

src/electron/views/automated-checks/components/maximize-restore-button.tsx
jalkire: It looks like the only difference in outcome is that `onRenderIcon` [passes in the button's props and renders before the button's text](https://github.com/OfficeDev/office-ui-fabric-react/blob/785a8733bedc868baa1a4ee17c0b551168cc710d/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx#L283), so there would be no difference in this situation. If there were also text associated with the button, it would impact the location of the icon (before the text if `onRenderIcon` was used, after if it was passed as a child). Given that this is supposed to behave like an icon, I think `onRenderIcon` makes more sense, even if there's no difference in this context

src/electron/views/automated-checks/components/maximize-restore-button.tsx
jalkire: We seem to [usually](https://github.com/microsoft/accessibility-insights-web/blob/ccdbcf4a141364dba315e6c56495ff2211f925e9/src/popup/components/header-contextual-menu.tsx#L37) [refer](https://github.com/microsoft/accessibility-insights-web/blob/ccdbcf4a141364dba315e6c56495ff2211f925e9/src/popup/launch-pad-row-configuration-factory.ts#L30) [to](https://github.com/microsoft/accessibility-insights-web/blob/ccdbcf4a141364dba315e6c56495ff2211f925e9/src/DetailsView/components/start-over-dropdown.tsx#L44) single word iconNames capitalized, so I have a difference instinct on this one...I don't particularly care one way or the other, but this seems closer to consistent in my opinion

jalkire:
@msft-github-bot delay merging this pull request for 1 hour

src/electron/views/automated-checks/components/maximize-restore-button.tsx
jalkire: Updated

src/electron/views/automated-checks/components/maximize-restore-button.tsx
```diff
@@ -0,0 +1,26 @@
+import { ActionButton } from 'office-ui-fabric-react/lib/Button';
+import * as React from 'react';
+
+import { RestoreIcon } from 'common/icons/restore-icon';
+import { NamedFC } from 'common/react/named-fc';
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
devaradhanm: move the copyright up

src/electron/views/automated-checks/components/maximize-restore-button.tsx
```diff
@@ -0,0 +1,26 @@
...
+    isMaximized: boolean;
+    onClick: () => void;
+}
+
+export const MaximizeRestoreButton = NamedFC<MaximizeRestoreButtonProps>('TitleBar', (props: MaximizeRestoreButtonProps) => {
+    const renderRestore = () => <RestoreIcon />;
+    const iconProps = { iconName: 'Stop' };
+    const onClick = () => props.onClick();
+
+    const button = props.isMaximized ? (
```
devaradhanm: we may have to conside full screen as well.

PR #1446, opened 2019-10-11
feat(android-cards-view): wire up device disconnected popup

smoralesd:
#### Description of changes

Wiring up the "Device disconnected" popup. When the user reaches the "Automated checks" view, we should the "scanning" spinner, if the scan failes, we'll be showing this popup that let's the user rescan the same device or navigate back to input a different port number.

image

**Notes for the reviews**
- This is only a wiring up PR, the dialog was already there
- The only exception to the previous point is: I made the device name a bold text to make it easier to catch
- There is a flickering on the window when navigating back to the port entry view. I had left that out of the scope of this PR as it touches a different piece of code and technically speaking, that bug was already there.

#### Pull request checklist

### DIFF
--- a/src/tests/unit/tests/electron/views/automated-checks/automated-checks-view.test.tsx
+++ b/src/tests/unit/tests/electron/views/automated-checks/automated-checks-view.test.tsx
@@ -1,11 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { ScanActionCreator } from 'electron/flux/action-creator/scan-action-creator';
+import { WindowStateActionCreator } from 'electron/flux/action-creator/window-state-action-creator';
 import { ScanStatus } from 'electron/flux/types/scan-status';
 import { AutomatedChecksView, AutomatedChecksViewProps } from 'electron/views/automated-checks/automated-checks-view';
+import { DeviceDisconnectedPopup } from 'electron/views/device-disconnected-popup/device-disconnected-popup';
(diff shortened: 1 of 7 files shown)


src/tests/unit/tests/electron/views/automated-checks/automated-checks-view.test.tsx
```diff
@@ -30,30 +33,99 @@ describe('AutomatedChecksView', () => {
...
+                deviceStoreData: {
+                    port,
+                },
+            } as AutomatedChecksViewProps;
+
+            const wrapped = shallow(<AutomatedChecksView {...props} />);
+
+            wrapped.find(DeviceDisconnectedPopup).prop('onRescanDevice')();
+
+            scanActionCreatorMock.verify(creator => creator.scan(port), Times.exactly(2));
```
dbjorge: This is confusing to me; why are there 2 scan messages rather than 1? If this is the expected behavior, can you leave a comment here explaining it?

dbjorge: Oh, I see, there's a scan on first mount too. Maybe a more isolated test would be to clear the mock expectations between lines 102 and 104 and then just have this test verify the one call unique to the event handler?

smoralesd: use the reset option, added a comment to explain why we do reset the mock! 
Thanks for the feedback.

PR #1448, opened 2019-10-11
contextual menu - high contrast mode

haonliu:
#### Description of changes
Have run the UX flow with Fer and confirmed this is what we want. 
This PR only includes the contextual menu control.

normal:
image
hover:
image
expand:
image

select a menu item:
image

When color contrast is off, nothing changes:
image

#### Pull request checklist

### DIFF
--- a/src/common/components/cards/card-kebab-menu-button.scss
+++ b/src/common/components/cards/card-kebab-menu-button.scss
@@ -2,17 +2,18 @@
 // Licensed under the MIT License.
 @import '../../styles/colors.scss';
 
-.kebab-menu-callout {
-    box-shadow: 0px 6.4px 14.4px $box-shadow-132, 0px 1.2px 3.6px $box-shadow-108 !important;
+div.kebab-menu-callout {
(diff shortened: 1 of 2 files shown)


dbjorge:
The focus state for the menu items in high-contrast has them using a light gray background instead of staying black-background like most of the other buttons in focus states do:

screenshot of card action menu item in focused state

screenshot of card action menu button in focused state

screenshot of left nav item in focused state

@ferBonnin @cheade Is this correct, or should menu items stay white-on-black in focused state? (it's similar for other menus, not a specific issue with this card action menu...)

ferBonnin:
> @ferBonnin @cheade Is this correct, or should menu items stay white-on-black in focused state? (it's similar for other menus, not a specific issue with this card action menu...)

I see this is consistent with what other buttons do in High contrast mode :)

cheade:
This behavior isn't necessary in high contrast mode as the point of the mode is to remove the subtlety. I don't believe the grey causes a contrast issue? 

If it's not a lot of work, I'm all for removing the grey and backing the background black. If it's time-consuming and not causing an issue, I think it's subtle enough that it shouldn't cause any trouble or visual disturbance to stay.

PR #1451, opened 2019-10-14
added text field validation via Enter key

JGibson2019:
#### Description of changes

users can now hit the "Enter" key and validate the port number. This PR also updated device connect port entry test and snapshots

#### Pull request checklist

- [N/A] Addresses an existing issue: Fixes #0000

### DIFF
--- a/src/tests/unit/tests/electron/views/device-connect-view/components/device-connect-port-entry.test.tsx
+++ b/src/tests/unit/tests/electron/views/device-connect-view/components/device-connect-port-entry.test.tsx
@@ -9,15 +9,18 @@ import { shallow } from 'enzyme';
 import { Button } from 'office-ui-fabric-react/lib/Button';
 import * as React from 'react';
 import { EventStubFactory } from 'tests/unit/common/event-stub-factory';
-import { IMock, Mock, MockBehavior, Times } from 'typemoq';
+import { IMock, It, Mock, MockBehavior, Times } from 'typemoq';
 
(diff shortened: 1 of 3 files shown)


Shobhit1:
I know this might be a more usability question; but is it fair to consider a lost `focus` event here as well with the `Enter` press to start validation; like pressing `Tab`.

src/tests/unit/tests/electron/views/device-connect-view/components/device-connect-port-entry.test.tsx
```diff
@@ -102,5 +104,25 @@ describe('DeviceConnectPortEntryTest', () => {
...
+                },
+                viewState: {
+                    deviceConnectState: DeviceConnectState.Default,
+                },
+            } as DeviceConnectPortEntryProps;
+            const rendered = shallow(<DeviceConnectPortEntry {...props} />);
+            rendered.setState({ port: testPortNumber });
+            const textField = rendered.find(MaskedTextField);
+
+            textField.simulate('keydown', { keyCode: KeyCodeConstants.ENTER }, enterEventStub);
```
smoralesd: do you actually need to pass `enterEventStub`? I though we figure is not needed

JGibson2019: You're correct; we don't need it and I've removed it in my most recent commit

PR #1456, opened 2019-10-14
adding title bar support for mac

devaradhanm:
#### Description of changes

- As part of this PR, we handle mac os separately. If mac, we use window frame with titlebar style set to hidden. This adds 3 icons to top left corner of window. So, we need to add some padding to the title bar to cover this icons. Also, we remove the title bar if the window is in full screen mode.

- For windows & mac, we show the minimize, close & maximize icons on the right side.

- As part of this change, we have also made changes to only listen to maximize, unmaximize, enter-full-screen & leave-full-screen. We don't have to listen to minimize action, since we don't change state based on whether the window is minimized or not.

**Device connect view:**
image

**Results view in maximized mode:**
image

**Results view in full screen view**
image

#### Pull request checklist

### DIFF
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -32,11 +32,11 @@ const createWindow = () => {
 };
 
 const enableDevMode = (window: BrowserWindow) => {
-    // if (process.env.DEV_MODE === 'true') {
-    window.webContents.openDevTools({
-        mode: 'detach',
-    });
-    // }
+    if (process.env.DEV_MODE === 'true') {
+        window.webContents.openDevTools({
+            mode: 'detach',
+        });
+    }
 };
 
 app.on('ready', createWindow);
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -2,17 +2,22 @@
 // Licensed under the MIT License.
 import { app, BrowserWindow } from 'electron';
 import * as path from 'path';
+import { PlatformInfo, OSType } from 'electron/platform-info';
 
 let mainWindow: BrowserWindow;
-
+let platformInfo = new PlatformInfo(process);
 const createWindow = () => {
+    const os = platformInfo.getOs();
     mainWindow = new BrowserWindow({
         show: false,
         webPreferences: { nodeIntegration: true },
-        frame: false,
+        titleBarStyle: 'hidden',
         width: 600,
         height: 391,
+        frame: os == OSType.Mac ? true : false,
     });
+    mainWindow.setMenuBarVisibility(false);
+    mainWindow.setSheetOffset(22);
 
     mainWindow
         .loadFile(path.resolve(__dirname, '../electron/views/index.html'))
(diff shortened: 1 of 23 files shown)


src/electron/main/main.ts
```diff
@@ -27,11 +32,11 @@ const createWindow = () => {
 };
 
 const enableDevMode = (window: BrowserWindow) => {
-    if (process.env.DEV_MODE === 'true') {
-        window.webContents.openDevTools({
-            mode: 'detach',
-        });
-    }
+    // if (process.env.DEV_MODE === 'true') {
```
Shobhit1: is this intentional?

src/electron/views/device-connect-view/components/window-title.tsx
```diff
@@ -1,32 +1,48 @@
...
+    }
+
+    const windowtitleClassNames = [windowTitle, props.className].filter(c => c != null);
+
+    if (props.deps.platformInfo.isMac()) {
+        windowtitleClassNames.push(macWindowTitle);
+    }
     return (
-        <header className={[windowTitle, props.className].filter(c => c != null).join(' ')}>
+        <header className={windowtitleClassNames.join(' ')}>
```
smoralesd: so... this line: 
```typescript
 const windowtitleClassNames = [windowTitle, props.className].filter(c => c != null);
```
and this line 
```typescript
windowtitleClassNames.join(' ')
```
is it worth it to use office fabric `css` function? (which does the same)

src/tests/electron/window-frame-listener.test.ts
```diff
@@ -20,75 +20,67 @@ describe(WindowFrameListener, () => {
...
-            eventsToListen.forEach(eventName => {
-                browserWindowMock
-                    .setup(b => b.on(eventName as any, It.isAny()))
-                    .callback((event, cb) => {
-                        eventCallbacks.push(cb);
-                    })
-                    .verifiable(Times.once());
-            });
+        beforeEach(() => {
+            setupVerifiableWindowEventCallback('maximize', cb => (maximizeCallback = cb));
```
smoralesd: should each test set mocks only for what each test need? 
Having all mocks set up for all cases make the code a little hard to follow as in, I have no idea which mock is really needed for each unit test.

Shobhit1:
I feel like this PR can be broken into at least 2 more small PRs accomplishing the subtasks that you listed in the PR description. Is that too much work if we want to separate out this PR into smaller PRs? Please check.

src/tests/electron/window-frame-listener.test.ts
devaradhanm: i made the browserWindowMock to be strict to control the access to browser window. So, I may need to do the mock on all the types anyway. Or i need to have 1 test with strict mock on initialize & then loose mock for others. Do you prefer that?

devaradhanm:
@Shobhit1 , I agree. I could split this in to two prs. I open to make this change. I could have split platform-info & window listener changes in to 1 pr & then remaining in another pr. I ended up converting my spike to productizable code.

PR #1461, opened 2019-10-14
refactor: extract full interface from user configuration action creator

smoralesd:
#### Description of changes

`UserConfigMessageCreator` is one of the several indirect dependencies of the `CardsView` component so we need to make the message creator API to match that of the action creator. 
In this PR:
- Extract interface from the `UserConfigurationActionCreator`
- Make both message & action creator API to match the new interface
- Update all usages to match new API
- Use the interface for dependency (instead of depending on the concrete classes)

#### Pull request checklist

   - not affected
   - not affected

### DIFF
--- a/src/background/global-action-creators/user-configuration-action-creator-impl.ts
+++ b/src/background/global-action-creators/user-configuration-action-creator-impl.ts
@@ -1,19 +1,23 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { UserConfigurationActionCreator } from 'background/global-action-creators/types/user-configuration-action-creator';
+
 import {
     SaveIssueFilingSettingsPayload,
     SetHighContrastModePayload,
     SetIssueFilingServicePayload,
     SetIssueFilingServicePropertyPayload,
+    SetTelemetryStatePayload,
 } from '../actions/action-payloads';
 import { UserConfigurationActions } from '../actions/user-configuration-actions';
 
-export class UserConfigurationActionCreator {
+export class UserConfigurationActionCreatorImpl implements UserConfigurationActionCreator {
     constructor(private readonly userConfigActions: UserConfigurationActions) {}
 
     public getUserConfigurationState = () => this.userConfigActions.getCurrentState.invoke(null);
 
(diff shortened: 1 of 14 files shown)


src/background/global-action-creators/user-configuration-action-creator-impl.ts
```diff
@@ -1,19 +1,23 @@
...
     SaveIssueFilingSettingsPayload,
     SetHighContrastModePayload,
     SetIssueFilingServicePayload,
     SetIssueFilingServicePropertyPayload,
+    SetTelemetryStatePayload,
 } from '../actions/action-payloads';
 import { UserConfigurationActions } from '../actions/user-configuration-actions';
 
-export class UserConfigurationActionCreator {
+export class UserConfigurationActionCreatorImpl implements UserConfigurationActionCreator {
```
devaradhanm: may be rename 1 class to UserConfigurationWebActionCreator & the other to UserConfigurationElectronActionCreator?
(this may not just invoke action directly in future, if we go with multiple window approach)

smoralesd: both `UserConfigurationActionCreator` & `UserConfigMessageCreator` are in full use on the browser extesion so using the word `Electron` to name one of them doesn't apply.
The main purpose of this refactor is to enable use to follow a design principle we stablish while ago when we started to build the browser extension: flux classes with "doubles" on both sides (the presentation layer and the background context) should be interchangeable. This is true for the stores in general as both the `BaseStoreImpl` and the `StoreProxy` implements `BaseStore` but the principle doesn't really hold for the action & message creators... this PR is just one step in that direction. Why doing this now? We are in a scenario where we want this principle to be follow so we can enable ourselves to re-use browser extension components in the electron app. Just to clarify this point a little bit more: the message creators don't make full sense in the electron app as there are no several context of excecution on the presentation layer (there is only one renderer process) and we keep the stores (the "real", holder-of-the-truth stores) on the one renderer process we use. The only thing that would make sense from the message creator is the public API but, as pointed out above, since action & message creator API are not the same, they are not really interchangeable... hence, this PR

PR #1473, opened 2019-10-15
feat(android-cards-view): dispatcher adapter to reuse action message creators

smoralesd:
#### Description of changes

In order to consume the `CardsView` on the electron app, we need to pass a lot of dependencies. Some of this dependencies are the **"message creators"**. The message creators are made so the "dispatch" a message to the background (this in on the browser extension context) but we don't have communication on the electron app (an no need for it just yet).

This PR approach to solve the reusability problem by creating a different implementation of the browser extension's `ActionMessageDispatcher` which will not send the message through a communication channel but instead, will call the interpreter directly.

In this PR:
- extract interface from current `ActionMessageDispatcher`
- replace all references (as dependencies) to `ActionMessageDispatcher` to now mention the new interface `Dispatcher`
- add new `ActionMessageDispatcherAdapter` to be used in electron
- update the electron initializer to use the new dispatcher adapter

PS: specials thanks to @dbjorge for the idea on the dispatcher adapter

#### Pull request checklist

   - not affected
   - not affected


src/electron/adapters/action-message-dispatcher-adapter.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { Interpreter } from 'background/interpreter';
+import { TelemetryData } from 'common/extension-telemetry-events';
+import { Message } from 'common/message';
+import { Dispatcher } from 'common/message-creators/types/dispatcher';
+
+export class ActionMessageDispatcherAdapter implements Dispatcher {
```
devaradhanm: it may be better to convey where it is dispatching to instead of just naming it adapter, now that we have an interface.
having ActionMessageDispatcherAdapter  & ActionMessageDispatcher as class names are confusing.

smoralesd: @devaradhanm any suggestions? I did not spend much time thinking of the name, I was expecting some feedback about it to be honest

dbjorge: Really, both types of dispatcher are dispatching action messages, they're just doing it differently. Maybe the interface can be `ActionMessageDispatcher` and the implementations can be `RemoteActionMessageDispatcher` and `DirectActionMessageDispatcher`?

smoralesd: I though about using `Direct`... I'll go with you're suggestion

PR #1475, opened 2019-10-15
feat(android-cards-view): update styles for automated checks view

smoralesd:
#### Description of changes

Mostly, add margins/paddings values on the automated checks view main content and the command bar.

image

#### Pull request checklist

   - not affected

### DIFF
--- a/src/electron/views/automated-checks/automated-checks-view.scss
+++ b/src/electron/views/automated-checks/automated-checks-view.scss
@@ -2,6 +2,9 @@
 // Licensed under the MIT License.
 .automated-checks-view {
     main {
-        margin: 16px 24px;
+        margin-top: 16px;
+        margin-bottom: 16px;
+        margin-left: 24px;
+        margin-right: 24px;
     }
 }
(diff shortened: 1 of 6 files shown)


src/electron/views/automated-checks/components/command-bar.scss
```diff
@@ -12,4 +12,12 @@
     font-size: 14px;
     font-weight: normal;
     border-bottom: 1px solid $neutral-alpha-8;
+    padding: 0px 16px;
+
+    #rescan-button {
```
dbjorge: Is it necessary to use an id rather than a class for this? If there's a specificity issue you might be able to work around it by setting the classname you need as an `iconProps` property instead

src/electron/views/automated-checks/automated-checks-view.scss
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+.automated-checks-view {
+    main {
+        margin: 16px 24px;
```
dbjorge: Consider using the named versions (`margin-top`, etc) for setting multiples like this; harder to make mistakes/miss stuff in review if that's our default practice

smoralesd: updated

src/electron/views/automated-checks/components/command-bar.scss
smoralesd: Good call. I updated and put the `className` on the `iconProps` as you suggested

PR #1478, opened 2019-10-15
feat: convert unified result for target page

waabid:
#### Description of changes

This PR introduces the use of a feature flag and converting the new data format to the old data format for the target page UI to consume.

This converter is aware of specific properties on the data model and I'm not fully sure how I feel about this and would love to see/hear opinions; the data this converter will receive is always sure to be web since this can only exist in the web, so that was my rationale for doing this (and limiting it here). However, I also don't like accessing non-strongly typed values like this: should we look into having strongly-typed platform specific identifiers/descriptors/resolution?

#### Pull request checklist


src/injected/element-based-view-model-creator.ts
```diff
@@ -0,0 +1,74 @@
...
+export class ElementBasedViewModelCreator {
+    constructor(
+        private getDecoratedAxeNode: GetDecoratedAxeNodeCallback,
+        private getHighlightedResultInstanceIds: GetHighlightedResultInstanceIdsCallback,
+    ) {}
+
+    public getElementBasedViewModel: GetElementBasedViewModelCallback = (
+        rules: UnifiedRule[],
+        results: UnifiedResult[],
+        cardSelectionData: CardSelectionStoreData,
```
RobGallo: Did you mean for this to be CardSelectionViewData? That is the type with 'highlightedResultUids'.

src/injected/element-based-view-model-creator.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export type GetElementBasedViewModelCallback = (
+    rules: UnifiedRule[],
+    results: UnifiedResult[],
+    cardSelectionData: CardSelectionStoreData,
+) => DictionaryStringTo<AssessmentVisualizationInstance>;
+
+export type GetHighlightedResultInstanceIdsCallback = (
+    cardSelectionData: CardSelectionStoreData,
```
RobGallo: Did you mean for this to be CardSelectionViewData? That is the type with 'highlightedResultUids'.

src/injected/element-based-view-model-creator.ts
```diff
@@ -0,0 +1,74 @@
...
+        private getDecoratedAxeNode: GetDecoratedAxeNodeCallback,
+        private getHighlightedResultInstanceIds: GetHighlightedResultInstanceIdsCallback,
+    ) {}
+
+    public getElementBasedViewModel: GetElementBasedViewModelCallback = (
+        rules: UnifiedRule[],
+        results: UnifiedResult[],
+        cardSelectionData: CardSelectionStoreData,
+    ) => {
+        const stringToResult: DictionaryStringTo<AssessmentVisualizationInstance> = {};
```
RobGallo: I would love it if you renamed this variable to resultsDictionary or something like that. As is, the variable sounds like a function, which in JavaScript can be quite confusing.

src/injected/element-based-view-model-creator.ts
waabid: The "ElementBasedViewModelCreator" receives the callback that you created as a part of the constructor and then uses it when necessary. All this is, is an interface/type for your callback. Hope that makes sense.

src/injected/element-based-view-model-creator.ts
waabid: This classe uses a dependency, effectively your function, to get that from the store data. This class is basically where we take store data (rules, results, and selection store data) and does all the processing (or at least coordinates it). I did not want to get the highlighted/processed data in selector map helper, which utilizes this, since it felt more relevant here. Hope that makes sense.

PR #1483, opened 2019-10-15
feat(android-cards-view): no rule resources support

smoralesd:
#### Description of changes

Update `RuleResource` component to handle no rule url and no rule guidance links scenarios (basically rendering null in those cases)

#### Pull request checklist

### DIFF
--- a/src/common/components/cards/rule-resources.tsx
+++ b/src/common/components/cards/rule-resources.tsx
@@ -4,9 +4,10 @@ import { GuidanceLinks } from 'common/components/guidance-links';
 import { GuidanceTags, GuidanceTagsDeps } from 'common/components/guidance-tags';
 import { NewTabLink } from 'common/components/new-tab-link';
 import { NamedFC } from 'common/react/named-fc';
+import { UnifiedRule } from 'common/types/store-data/unified-data-interface';
+import { isEmpty } from 'lodash';
 import * as React from 'react';
 
-import { UnifiedRule } from '../../../common/types/store-data/unified-data-interface';
 import { moreResourcesTitle, ruleDetailsId, ruleMoreResources } from './rule-resources.scss';
 
 export type RuleResourcesDeps = GuidanceTagsDeps;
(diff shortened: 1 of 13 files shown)


src/tests/unit/tests/common/components/cards/rule-resources.test.tsx
```diff
@@ -2,21 +2,52 @@
...
 import { shallow } from 'enzyme';
+import { cloneDeep } from 'lodash';
 import * as React from 'react';
 
 import { exampleUnifiedRuleResult } from './sample-view-model-data';
 
 describe('RuleResources', () => {
-    it('renders', () => {
-        const rule = exampleUnifiedRuleResult;
+    describe('renders', () => {
```
dbjorge: Test case for the null case?

smoralesd: I'll add it

src/common/components/cards/rule-resources.tsx
```diff
@@ -17,6 +18,10 @@ export type RuleResourcesProps = {
 };
 
 export const RuleResources = NamedFC<RuleResourcesProps>('RuleResources', ({ deps, rule }) => {
+    if (rule.url == null || isEmpty(rule.guidance)) {
```
devaradhanm: may be use isNil?

dbjorge: Is this really intended to be `||` rather than `&&`? I'd have expected that we want to show the resources in the event that we have either of them

smoralesd: mmm we need to cover `[]`

src/tests/unit/tests/common/components/cards/rule-resources.test.tsx
```diff
@@ -2,21 +2,81 @@
...
+            const wrapper = shallow(<RuleResources {...props} />);
 
-        expect(wrapper.getElement()).toMatchSnapshot();
+            expect(wrapper.getElement()).toMatchSnapshot();
+        });
+
+        it('no url, only guidance links', () => {
+            const rule = cloneDeep(exampleUnifiedRuleResult);
+            rule.url = null;
+
```
devaradhanm: why not make this a data driven test since u are duplicating the test method code anyways?

the testCase can have the rule value & guidance value.

src/common/components/cards/rule-resources.tsx
smoralesd: Make sense

src/tests/unit/tests/common/components/cards/rule-resources.test.tsx
smoralesd: Updated

PR #1484, opened 2019-10-15
updating scroll bar styling

devaradhanm:
#### Description of changes

(Ignore the blue line inside the screenshot. I added a div with border to test the scroll bar.)

image

When you have a child that with min width & cannot be broken, you will see a horizontal scrollbar
image

#### Pull request checklist

### DIFF
--- a/src/electron/views/device-connect-view/device-connect-view.scss
+++ b/src/electron/views/device-connect-view/device-connect-view.scss
@@ -5,7 +5,7 @@
 body {
     margin: 0px;
     height: 100%;
-    widows: 100%;
+    width: 100%;
     overflow: hidden;
 }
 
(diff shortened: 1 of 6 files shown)


src/electron/views/device-connect-view/device-connect-view.scss
```diff
@@ -4,13 +4,32 @@
 
 body {
     margin: 0px;
+    height: 100%;
+    widows: 100%;
```
dbjorge: "widows"?

devaradhanm:
You only see horizontal scrollbar when u add a component that has a min width & cannot be spread vertically. the scrollbar u see here is because i added a div with min width. so it cannot be broken. i will add a screenshot for cards ui without the horizontal scrollbar

PR #1489, opened 2019-10-16
feat(android-cards-view): add how-to-fix CardRow for android 

devaradhanm:
#### Description of changes

Sample screenshot with dummy data
image

#### Pull request checklist

### DIFF
--- a/src/common/components/cards/fix-instruction-color-box.scss
+++ b/src/common/components/cards/fix-instruction-color-box.scss
@@ -1,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../../common/styles/colors.scss';
+
+.how-to-fix-card-row {
+    span.fix-instruction-color-box {
+        width: 14px;
+        height: 14px;
+        display: inline-block;
+        vertical-align: -5%;
+        margin: 0px 2px;
+        border: 1px solid $always-black;
+    }
+
+    .screen-reader-only {
+        position: absolute;
+        left: -10000px;
+        top: auto;
+        width: 1px;
+        height: 1px;
+        overflow: hidden;
+    }
+}
--- a/src/common/components/cards/how-to-fix-android-card-row.tsx
+++ b/src/common/components/cards/how-to-fix-android-card-row.tsx
@@ -32,6 +32,7 @@ type HowToFixSplit = {
 function getHowToFixContent(props: HowToFixAndroidCardRowProps): (JSX.Element | string)[] {
     const propertyData = props.propertyData;
     let howToFixSplit: HowToFixSplit[] = [{ str: propertyData.howToFix }];
+    const result: (JSX.Element | string)[] = [];
 
     if (!isEmpty(propertyData.formatAsCode)) {
         propertyData.formatAsCode.forEach(item => {
@@ -39,9 +40,22 @@ function getHowToFixContent(props: HowToFixAndroidCardRowProps): (JSX.Element | 
         });
     }
 
-    return howToFixSplit.map((item, index) => {
-        return isNil(item.str) ? <Term key={`strong-how-to-fix-${props.index}-${index}`}>{item.match}</Term> : item.str;
+    howToFixSplit.forEach((item, index) => {
+        const key = `strong-how-to-fix-${props.index}-${index}`;
+        const content = isNil(item.str) ? item.match : item.str;
+
+        if (content[0] === ' ' && index !== 0) {
+            result.push(<span key={key + '-before-space'}>&nbsp;</span>);
+        }
+
+        result.push(isNil(item.str) ? <Term key={key}>{content}</Term> : <span key={key}>{content}</span>);
+
+        if (content[content.length - 1] === ' ') {
+            result.push(<span key={key + '-after-space'}>&nbsp;</span>);
+        }
     });
+
+    return result;
 }
 
 function getHowToFixSplitsForPattern(pattern: string, previousHowToFixSplit: HowToFixSplit[]): HowToFixSplit[] {
--- a/src/common/components/cards/how-to-fix-android-card-row.tsx
+++ b/src/common/components/cards/how-to-fix-android-card-row.tsx
@@ -56,7 +56,13 @@ function getHowToFixContent(props: HowToFixAndroidCardRowProps): (JSX.Element | 
             result.push(<span key={key + '-before-space'}>&nbsp;</span>);
         }
 
-        result.push(isEmpty(item.str) ? <Term key={key}>{content}</Term> : <span key={key}>{content}</span>);
+        result.push(
+            isEmpty(item.str) ? (
+                <Term key={key}>{content}</Term>
+            ) : (
+                <span key={key}>{props.deps.fixInstructionProcessor.process(content)}</span>
+            ),
+        );
 
         if (content[content.length - 1] === ' ') {
             result.push(<span key={key + '-after-space'}>&nbsp;</span>);
--- a/src/common/components/cards/how-to-fix-android-card-row.tsx
+++ b/src/common/components/cards/how-to-fix-android-card-row.tsx
@@ -17,7 +17,7 @@ export const HowToFixAndroidCardRow = NamedFC<HowToFixAndroidCardRowProps>('HowT
     return (
         <SimpleCardRow
             label="How to fix"
-            content={<>{getHowToFixContent(props)}</>}
+            content={<span>{getHowToFixContent(props)}</span>}
             rowKey={`how-to-fix-row-${props.index}`}
             contentClassName="how-to-fix-card-row"
         />
--- a/src/common/components/cards/how-to-fix-android-card-row.tsx
+++ b/src/common/components/cards/how-to-fix-android-card-row.tsx
@@ -3,14 +3,14 @@
 import * as React from 'react';
 
 import { Term } from 'assessments/markup';
-import { UnifiedFormatAsCodeResolution } from 'common/types/store-data/unified-data-interface';
+import { FormatAsCodeResolution } from 'common/types/store-data/unified-data-interface';
 import { isEmpty, isNil } from 'lodash';
 import { CardRowProps } from '../../../common/configs/unified-result-property-configurations';
 import { NamedFC } from '../../../common/react/named-fc';
 import { SimpleCardRow } from './simple-card-row';
 
 export interface HowToFixAndroidCardRowProps extends CardRowProps {
-    propertyData: UnifiedFormatAsCodeResolution;
+    propertyData: FormatAsCodeResolution;
 }
 
 export const HowToFixAndroidCardRow = NamedFC<HowToFixAndroidCardRowProps>('HowToFixAndroidCardRow', props => {
@@ -31,7 +31,7 @@ type HowToFixSplit = {
 
 function getHowToFixContent(props: HowToFixAndroidCardRowProps): (JSX.Element | string)[] {
     const propertyData = props.propertyData;
-    let howToFixSplit: HowToFixSplit[] = [{ str: propertyData.howToFixSummary }];
+    let howToFixSplit: HowToFixSplit[] = [{ str: propertyData.howToFix }];
 
     if (!isEmpty(propertyData.formatAsCode)) {
         propertyData.formatAsCode.forEach(item => {
--- a/src/common/components/cards/how-to-fix-android-card-row.tsx
+++ b/src/common/components/cards/how-to-fix-android-card-row.tsx
@@ -1,0 +1,85 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as React from 'react';
+
+import { isEmpty, isNil } from 'lodash';
+import { CardRowProps } from '../../../common/configs/unified-result-property-configurations';
+import { NamedFC } from '../../../common/react/named-fc';
+import { SimpleCardRow } from './simple-card-row';
+import { Term } from 'assessments/markup';
+
+export interface HowToFixAndroidCardRowPropertyData {
+    howToFixSummary: string;
+    formatAsCode?: string[];
+}
+export interface HowToFixAndroidCardRowProps extends CardRowProps {
+    propertyData: HowToFixAndroidCardRowPropertyData;
+}
+
+export const HowToFixAndroidCardRow = NamedFC<HowToFixAndroidCardRowProps>('HowToFixAndroidCardRow', props => {
+    return (
+        <SimpleCardRow
+            label="How to fix"
+            content={<>{getHowToFixContent(props)}</>}
+            rowKey={`how-to-fix-row-${props.index}`}
+            contentClassName="how-to-fix-card-row"
+        />
+    );
+});
+
+type HowToFixSplit = {
+    str?: string;
+    match?: string;
+};
+
+function getHowToFixContent(props: HowToFixAndroidCardRowProps): (JSX.Element | string)[] {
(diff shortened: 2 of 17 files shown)


src/common/components/cards/how-to-fix-android-card-row.tsx
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as React from 'react';
```
smoralesd: should this file be under `electron` folder? maybe even `electron\views\automated-checks\components\cards`

src/tests/unit/tests/common/components/cards/how-to-fix-android-card-row.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+                howToFixSummary: "This isn't a simple text. It has some words that will be marked as bold.",
+            },
+        };
+
+        const testSubject = shallow(<HowToFixAndroidCardRow {...props} />);
+
+        expect(testSubject.getElement()).toMatchSnapshot();
+    });
+
+    it('renders without matches', () => {
```
smoralesd: should this be `"without formatAsCode property"`??

src/common/components/cards/how-to-fix-android-card-row.tsx
devaradhanm: this will be used by the configuration property, just like how we do for web https://github.com/microsoft/accessibility-insights-web/blob/cda39b427c80b79f26fc4a40c8b165617b3c21fd/src/common/configs/unified-result-property-configurations.tsx#L43

src/electron/platform/android/rule-information-provider.ts
```diff
@@ -92,8 +92,8 @@ export class RuleInformationProvider {
         );
     };
 
-    private buildUnifiedResolution(unformattedText: string, codeStrings: string[] = null): UnifiedResolution {
-        return { howToFixSummary: unformattedText, formatAsCode: codeStrings };
+    private buildUnifiedResolution(unformattedText: string, codeStrings: string[] = null): UnifiedFormatAsCodeResolution {
```
DaveTryon: howToFixFormat is not necessarily limited to formatting as code. It could include italics, links, or any other number of things. The type name should not be bound to "FormatAsCode". UnifiedFormattedResolution, perhaps?

Also, the rename should extend to include RuleInformation's GetUnifiedResolutionDelegate type and getUnifiedResolution method

src/electron/platform/android/rule-information-provider.ts
```diff
@@ -92,8 +92,8 @@ export class RuleInformationProvider {
...
+        return { howToFixSummary: unformattedText, howToFixFormat: { howToFix: unformattedText, formatAsCode: codeStrings } };
```
DaveTryon: Why are we duplicating the unformattedText in the object? Lisa and I intentionally changed this recently to avoid exactly this behavior, and I'd be very sad to see it return. My guess is that there's downstream code that consumes the howToFixFormat property. I'll argue that the downstream code should consume the UnifiedFormattedResolution (or whatever we end up calling it)--that way it has all of the information that it needs in a single object, with no duplication.

src/common/components/cards/how-to-fix-android-card-row.tsx
smoralesd: I saw the same while working on rendering `text`, `contentDescription` and `className` so I think this is fine until we properly decide what to do with it

src/electron/platform/android/rule-information-provider.ts
devaradhanm: buildUnifiedFormattedResolution

smoralesd:
2 comments from the screenshot:
- `TouchSizeWcag`'s `how to fix` text seems to be running together (there is not enough space between the words, specially the bolded ones)
- `ColorContrast`'s `how to fix` seems to be missing the colored squares right before the color hex value:
image

I think the first comment should be part of this PR.
I'm not sure about the second one, I guess is a nice to have at this point

src/common/components/cards/how-to-fix-android-card-row.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    howToFixSplit.forEach((item, index) => {
+        const key = `strong-how-to-fix-${props.index}-${index}`;
+        const content = isNil(item.str) ? item.match : item.str;
+
+        if (content[0] === ' ' && index !== 0) {
+            result.push(<span key={key + '-before-space'}>&nbsp;</span>);
+        }
+
+        result.push(isNil(item.str) ? <Term key={key}>{content}</Term> : <span key={key}>{content}</span>);
```
smoralesd: 01 - isEmptry vs isNill result on a string
 are you sure you don't want to use `isEmpty` for checking this string values?

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -73,6 +73,15 @@ export type UnifiedResolution = {
     howToFixSummary: string;
 } & InstancePropertyBag;
 
+export type FormattedResolution = {
+    howToFix: string;
+    formatAsCode?: string[];
+};
+
+export type UnifiedFormattedResolution = {
+    howToFixFormat?: FormattedResolution;
```
smoralesd: should this be named `howToFixFormatted` or `formattedHowToFix`??? `Format` implies a different meaning I guess

src/electron/platform/android/fetch-scan-results.ts
```diff
@@ -9,7 +10,7 @@ export type HttpGet = typeof axios.get;
 
 export const createFetchScanResults = (httpGet: HttpGet): FetchScanResultsType => {
     return async (port: number) => {
-        const response = await httpGet(`http://localhost:${port}/axe/result`);
-        return new ScanResults(response.data);
+        //  const response = await httpGet(`http://localhost:${port}/axe/result`);
```
smoralesd: comented code here.

src/tests/unit/tests/common/components/cards/how-to-fix-android-card-row.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+            howToFix: 'This is a sample string ',
+        },
+        {
+            label: 'when match is on a word',
+            formatAsCode: ['sample'],
+            howToFix: 'This is a sample string ',
+        },
+    ];
+    renderWithSpaceTestCases.forEach(testCase => {
+        it(`renders with whitespace - ${testCase.label}`, () => {
```
smoralesd: is it only because you only want `testCase.label` to show up on the test description console output that you are not using `it.each` here?

src/common/components/cards/how-to-fix-android-card-row.tsx
```diff
@@ -0,0 +1,96 @@
...
+import * as React from 'react';
+
+import { Term } from 'assessments/markup';
+import { FormattedResolution } from 'common/types/store-data/unified-data-interface';
+import { isEmpty, isNil } from 'lodash';
+import { CardRowProps } from '../../../common/configs/unified-result-property-configurations';
+import { NamedFC } from '../../../common/react/named-fc';
+import { SimpleCardRow } from './simple-card-row';
+
+export interface HowToFixAndroidCardRowProps extends CardRowProps {
```
smoralesd: **not for this PR but**: by inheriting from `CardRowProps`, `HowToFixAndroidCardRowProps` contains as a deps a:
```typescript
    fixInstructionProcessor: FixInstructionProcessor;
```
which means, we should be able to add the colored square for color contrast failures.
Again, I don't think we need to do that as part of this PR

src/tests/unit/tests/common/components/cards/how-to-fix-android-card-row.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { Mock } from 'typemoq';
+
+import { HowToFixAndroidCardRow, HowToFixAndroidCardRowProps } from 'common/components/cards/how-to-fix-android-card-row';
+import { FixInstructionProcessor } from '../../../../../../injected/fix-instruction-processor';
+
+describe(HowToFixAndroidCardRow, () => {
+    it('renders with matches', () => {
+        const fixInstructionProcessorMock = Mock.ofType(FixInstructionProcessor);
+        const props: HowToFixAndroidCardRowProps = {
+            deps: { fixInstructionProcessor: fixInstructionProcessorMock.object },
```
smoralesd: `fixInstructionProcessor` is not really been used on `HowToFixAndroidCardRow`... can't it be null?

src/tests/unit/tests/common/components/cards/how-to-fix-android-card-row.test.tsx
devaradhanm: we do have this usage in many places. for eg: content-page.test.tsx

src/tests/unit/tests/common/components/cards/how-to-fix-android-card-row.test.tsx
devaradhanm: i will be sending another pr to do the color processing

src/common/components/cards/how-to-fix-android-card-row.tsx
devaradhanm: i will be sending another pr for thi, once this is merged

devaradhanm: i have added support for color processing now.

src/tests/unit/tests/common/components/cards/how-to-fix-android-card-row.test.tsx
devaradhanm: added support for color processing now

src/common/components/cards/fix-instruction-color-box.scss
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../../common/styles/colors.scss';
```
devaradhanm: we have duplicated this code everywhere. I have created this new file for styling color box. We need to import this file in all other usages. but, using this new file in other places is not part of this pr.

src/electron/platform/android/rule-information-provider.ts
```diff
@@ -79,21 +79,21 @@ export class RuleInformationProvider {
         return result;
     }
 
-    private getTouchSizeUnifiedResolution = (ruleResultsData: RuleResultsData): UnifiedResolution => {
+    private getTouchSizeUnifiedResolution = (ruleResultsData: RuleResultsData): UnifiedFormattableResolution => {
```
DaveTryon: Should this be getTouchSizeUnifiedFormattedResolution instead of getTouchSizeUnifiedResolution?

DaveTryon:
>     private getColorContrastUnifiedResolution = (ruleResultsData: RuleResultsData): UnifiedResolution => {

This should also return a UnfiedFormattableResolution. Rename function to getColorContrasUnifiedFormattableResolution ?

---
Refers to: src/electron/platform/android/rule-information-provider.ts:59 in 27969d4. [](commit_id = 27969d4e541db70ff37746e70df208bff69c0748, deletion_comment = False)

src/electron/views/device-connect-view/device-connect-view.scss
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 @import './components/telemetry-permission-dialog.scss';
 @import '../../../reports/components/outcome.scss';
+@import '../../../common/components/cards/fix-instruction-color-box.scss';
```
DaveTryon: Can we remove the `../../..` ?
```suggestion
@import 'common/components/cards/fix-instruction-color-box.scss';
```

src/electron/platform/android/rule-information-provider.ts
DaveTryon: As we discussed offline, we should have a design conversation to reconcile the differences between issue filing (which operates on the entire `UnifiedResolution` object, and cards, which operates on the `howToFixFormat` object. That's outside the scope of this PR, though, so we'll live with temporary duplication of data

src/electron/platform/android/rule-information-provider.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { UnifiedResolution } from 'common/types/store-data/unified-data-interface';
+import { UnifiedFormattableResolution, UnifiedResolution } from 'common/types/store-data/unified-data-interface';
```
DaveTryon: (Can't add a note on line 59, so adding it here)
All references to `UnifiedResolution` should be replaced with `UnifiedFormattedResolution`. Line 59 is still returning a `UnifiedResolution`. In addition, `getColorContrastUnifiedResolution` might be more appropriate as `getColorContrastUnifiedFormattedResolution`

src/common/components/cards/fix-instruction-color-box.scss
DaveTryon: I think the `../../..` can be omitted:
```suggestion
@import 'common/styles/colors.scss';
```

src/electron/platform/android/rule-information.ts
```diff
@@ -12,12 +12,12 @@ export class RuleInformation {
     constructor(
         readonly ruleId: string,
         readonly ruleDescription: string,
-        readonly getUnifiedResolutionDelegate: GetUnifiedResolutionDelegate,
+        readonly getUnifiedFormattableResolutionDelegate: GetUnifiedResolutionDelegate,
```
DaveTryon: Should `GetUnifiedResolutionDelegate` be renamed to `GetUnifiedFormattableResolutionDelegate` ?

src/tests/unit/tests/electron/platform/android/rule-information-provider.test.ts
```diff
@@ -49,7 +49,7 @@ describe('RuleInformationProvider', () => {
 
     function validateUnifiedResolution(ruleId: string, ruleResult: RuleResultsData): UnifiedResolution {
```
DaveTryon:
```suggestion
    function validateUnifiedResolution(ruleId: string, ruleResult: RuleResultsData): UnifiedFormattableResolution {
```

src/tests/unit/tests/electron/platform/android/scan-results-helpers.ts
```diff
@@ -93,11 +93,11 @@ export function buildRuleInformation(ruleId: string, includeResults: boolean = t
     return {
         ruleId: ruleId,
         ruleDescription: 'This describes ' + ruleId,
-        getUnifiedResolutionDelegate: r => {
+        getUnifiedFormattableResolutionDelegate: r => {
             expect('getUnifiedResolution').toBe('This code should never execute');
```
DaveTryon:
```suggestion
            expect('getUnifiedFormattableResolution').toBe('This code should never execute');
```

DaveTryon:
We should have a chat about how to handle the color sample squares. I'd suggest that we think about adding background and foreground color information to the howToFixFormat object, as opposed to trying to somehow parse it out of the unformatted text. Seems out of scope for this PR either way.

src/common/components/cards/fix-instruction-color-box.scss
smoralesd: that works for typescript file... we haven't done anything like this for scss... don't get me wrong, it may work, but also, it may not

src/electron/views/device-connect-view/device-connect-view.scss
smoralesd: same as before... works for typescript. not sure about scss

PR #1497, opened 2019-10-17
fix: prevent injected dialog buttons wrapping from new Toast layout

dbjorge:
#### Description of changes

Fixes a regression introduced by our recent toast aria-live changes where its container div causes an undesired layout change in the injected target page dialog.

Before:

screenshot of dialog before change showing buttons wrapping

After:

screenshot of dialog after change with normal button layout

#### Pull request checklist

### DIFF
--- a/src/common/components/toast.scss
+++ b/src/common/components/toast.scss
@@ -5,7 +5,11 @@
 $toast-width: 316px;
 $toast-padding: 20px;
 
-.toast {
+.toast-container {
+    display: inline-block;
+}
+
+.toast-content {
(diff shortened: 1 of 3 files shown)


src/common/components/toast.scss
```diff
@@ -5,7 +5,11 @@
 $toast-width: 316px;
 $toast-padding: 20px;
 
-.toast {
+.toast-container {
+    display: inline-block;
```
devaradhanm: or you could use display: flex on the container.

PR #1502, opened 2019-10-17
fix: make "copy failure details" button in target page dialog work again

### LINKED ISSUE #1501: Copy failure details fail on the target page
## Describe the bug

Activating "Copy failure details" button on the target page's failure details dialog shows the "Failed to copy failure details. Please try again." toast.

**To Reproduce**

1. Go to [accessible university before](http://www.washington.edu/accesscomputing/AU/before.html)
2. Run automated checks
3. Activate any "!" icon
4. On the failure details dialog, activate "Copy failure details" button
5. Observe: the "Failed to copy failure details. Please try again." toast shows up. No details are copied to the system clipboard

## CodePen repro example

Not needed

## Expected behavior

Copy failure details should work properly

## Screenshots

01 - copy failure details on target page fails

- OS: Windows 10 Enterprise 10.0.18362 Build 18362
- Extension Version: insider 2.10.1 (on edge dev)
- Browser Version Microsoft Edge version 79.0.294.1
- Target Page http://www.washington.edu/accesscomputing/AU/before.html

Yes

No

## Additional context

**Not** present on production

dbjorge:
#### Description of changes

* Adds `console.error` logging with the full error message on copy failure, for diagnosability
* Wires up `NavigatorUtils` in the target page to the dialog renderer (it was missing before)

screenshot of copying working in target page

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1501 

### DIFF
--- a/src/common/components/cards/card-kebab-menu-button.tsx
+++ b/src/common/components/cards/card-kebab-menu-button.tsx
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { MoreActionsMenuIcon } from 'common/icons/more-actions-menu-icon';
-import { createDefaultLogger } from 'common/logging/default-logger';
-import { Logger } from 'common/logging/logger';
 import { ActionButton } from 'office-ui-fabric-react/lib/Button';
 import { DirectionalHint, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
 import * as React from 'react';
--- a/src/common/components/cards/card-kebab-menu-button.tsx
+++ b/src/common/components/cards/card-kebab-menu-button.tsx
@@ -142,6 +142,7 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
         try {
             await this.props.deps.navigatorUtils.copyToClipboard(text);
         } catch (error) {
+            console.error(`Failed to copy failure details: ${error}`);
             this.toastRef.current.show('Failed to copy failure details. Please try again.');
             return;
         }
--- a/src/common/components/cards/card-kebab-menu-button.tsx
+++ b/src/common/components/cards/card-kebab-menu-button.tsx
@@ -1,10 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { MoreActionsMenuIcon } from 'common/icons/more-actions-menu-icon';
+import { createDefaultLogger } from 'common/logging/default-logger';
+import { Logger } from 'common/logging/logger';
 import { ActionButton } from 'office-ui-fabric-react/lib/Button';
 import { DirectionalHint, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
 import * as React from 'react';
-
-import { MoreActionsMenuIcon } from 'common/icons/more-actions-menu-icon';
 import { IssueDetailsTextGenerator } from '../../../background/issue-details-text-generator';
 import { DetailsViewActionMessageCreator } from '../../../DetailsView/actions/details-view-action-message-creator';
 import { IssueFilingDialog } from '../../../DetailsView/components/issue-filing-dialog';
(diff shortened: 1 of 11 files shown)


src/common/components/cards/card-kebab-menu-button.tsx
```diff
@@ -142,6 +142,7 @@ export class CardKebabMenuButton extends React.Component<CardKebabMenuButtonProp
         try {
             await this.props.deps.navigatorUtils.copyToClipboard(text);
         } catch (error) {
+            console.error(`Failed to copy failure details: ${error}`);
```
smoralesd: should use `Logger` from `src\common\logging\logger.ts`

src/common/components/copy-issue-details-button.tsx
```diff
@@ -40,6 +40,7 @@ export class CopyIssueDetailsButton extends React.Component<CopyIssueDetailsButt
         try {
             await this.props.deps.navigatorUtils.copyToClipboard(this.getIssueDetailsText(this.props.issueDetailsData));
         } catch (error) {
+            console.error(`Failed to copy failure details: ${error}`);
```
smoralesd: should use `Logger` from `src\common\logging\logger.ts`

src/tests/unit/tests/injected/visualization/drawer-provider.test.ts
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { NavigatorUtils } from 'common/navigator-utils';
 import { IMock, Mock } from 'typemoq';
 import { BrowserAdapter } from '../../../../../common/browser-adapters/browser-adapter';
```
smoralesd: probably a good time to update this import paths to non-relative ones

src/injected/dialog-renderer.tsx
```diff
@@ -82,6 +84,7 @@ export class DialogRenderer {
                 fixInstructionProcessor,
                 issueDetailsTextGenerator,
                 windowUtils: this.windowUtils,
+                navigatorUtils: this.navigatorUtils,
                 targetPageActionMessageCreator: mainWindowContext.getTargetPageActionMessageCreator(),
                 issueFilingActionMessageCreator: mainWindowContext.getIssueFilingActionMessageCreator(),
                 browserAdapter: this.browserAdapter,
```
haonliu: I think with this change we can remove the casting on Line 93 now. I'm concerned it would cause more issue in the future. But not necessarily have to be in this PR since you did not touch that.

dbjorge: Good point, addressed

src/injected/dialog-renderer.tsx
```diff
@@ -82,6 +84,7 @@ export class DialogRenderer {
                 fixInstructionProcessor,
                 issueDetailsTextGenerator,
                 windowUtils: this.windowUtils,
+                navigatorUtils: this.navigatorUtils,
```
peterdur: Would the compiler have caught this if this were `const deps: LayeredDetailsDialogDeps = { ... }` instead of `const deps = { ... } as LayeredDetailsDialogDeps`? If so, should we prefer that way?

haonliu: I agree. we should try to avoid casting.

dbjorge: Already resolved

src/tests/unit/tests/injected/visualization/drawer-provider.test.ts
dbjorge: I'd rather keep those to separate PRs

src/common/navigator-utils.ts
```diff
@@ -31,7 +31,12 @@ export class NavigatorUtils {
         return userAgent.substring(versionOffset + versionPrefix.length + 1).split(' ')[0];
     }
 
-    public copyToClipboard(data: string): Promise<void> {
-        return this.navigatorInfo.clipboard.writeText(data);
+    public async copyToClipboard(data: string): Promise<void> {
+        try {
```
smoralesd: is the try-catch needed?
can it be something like:
```typescript
public async copyToClipboard(data: string): Promise<void> {
   return this.navigatorInfo.clipboard.writeText(data).catch(error => {
      this.logger.error('Error during copyToClipboard: ', error);
      throw error;
   });
}
```

dbjorge: I've changed this because I don't think it's any worse of a solution, but I also don't think it's better than the original code using try-catch; I think this is a personal style preference for which we do not currently have a consistent team style to follow, so I feel a little bit frustrated to be spending time addressing it :(

PR #1511, opened 2019-10-18
Extract dimensions from bounding rectangle

DaveTryon:
#### Description of changes

We discovered that the Axe service has changed the format of the data in the TouchSizeWcag rule result. The new format looks wrong, so we're extracting the dimensions directly from the bounding rectangle. This should work in both old and new versions of the service.

#### Pull request checklist

### DIFF
--- a/src/electron/platform/android/rule-information-provider.ts
+++ b/src/electron/platform/android/rule-information-provider.ts
@@ -81,8 +81,9 @@ export class RuleInformationProvider {
 
     private getTouchSizeUnifiedFormattableResolution = (ruleResultsData: RuleResultsData): UnifiedFormattableResolution => {
         const dpi = ruleResultsData.props['Screen Dots Per Inch'] as number;
-        const physicalWidth = ruleResultsData.props['width'] as number;
-        const physicalHeight = ruleResultsData.props['height'] as number;
+        const boundingRect = ruleResultsData.props['boundsInScreen'];
+        const physicalWidth = (boundingRect['right'] as number) - (boundingRect['left'] as number);
+        const physicalHeight = (boundingRect['bottom'] as number) - (boundingRect['top'] as number);
         const logicalWidth = physicalWidth / dpi;
         const logicalHeight = physicalHeight / dpi;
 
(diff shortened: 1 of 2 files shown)


src/electron/platform/android/rule-information-provider.ts
```diff
@@ -80,8 +80,9 @@ export class RuleInformationProvider {
 
     private getTouchSizeUnifiedFormattableResolution = (ruleResultsData: RuleResultsData): UnifiedFormattableResolution => {
         const dpi = ruleResultsData.props['Screen Dots Per Inch'] as number;
-        const physicalWidth = ruleResultsData.props['width'] as number;
-        const physicalHeight = ruleResultsData.props['height'] as number;
+        const boundingRect = ruleResultsData.props['boundsInScreen'];
+        const physicalWidth = (boundingRect['right'] as number) - (boundingRect['left'] as number);
```
Shobhit1: Do we want to make this much more type strict by not casting the `boundingRect['right']` as number and making sure that `physicalWidth` is declared as `const physicalWidth: number`? Seems like that would be a better check.

Shobhit1: Because `string` to `number` conversions in `javascript` (which TS will eventually compile down to) are so uncertain; I think we should use TS to strictly here to avoid those potential runtime problems.

src/tests/unit/tests/electron/platform/android/rule-information-provider.test.ts
```diff
@@ -16,10 +16,20 @@ describe('RuleInformationProvider', () => {
 
     function buildTouchSizeWcagRuleResultObject(status: string, dpi: number, height: number, width: number): RuleResultsData {
         const props = {};
-        // This is based on the output of the Android service
+        // This is based on the output of the Android service.
         props['Screen Dots Per Inch'] = dpi;
-        props['height'] = height;
-        props['width'] = width;
+
+        const left = 123 as number;
```
Shobhit1: Pertaining to my comment on the actual test; does `'123' as number` compile? Just curious.

src/electron/platform/android/rule-information-provider.ts
DaveTryon: Changed as suggested

src/tests/unit/tests/electron/platform/android/rule-information-provider.test.ts
DaveTryon: No. '123' as number generates an error.

PR #1519, opened 2019-10-18
fix: ensure scanning spinner is announce by SC

smoralesd:
#### Description of changes

Currently, the "scanning..." spinner under fast pass' automated checks is not being announced by the screen reader. 

Using a `<div role="alert">` to wrap the spinner component solve this issue.

Additionally, I update the import paths on the changes file (different commit, same PR)

#### Pull request checklist

   - check with both NVDA and JAWS, the spinner is properly announce on both


src/DetailsView/components/issues-table.tsx
```diff
@@ -182,7 +182,11 @@ export class IssuesTable extends React.Component<IssuesTableProps> {
     }
 
     private renderSpinner(label: string): JSX.Element {
-        return <Spinner className="details-view-spinner" size={SpinnerSize.large} label={label} />;
+        return (
+            <div role="alert">
+                <Spinner className="details-view-spinner" size={SpinnerSize.large} label={label} />
```
karanbirsingh: This change improves what we have in `master`. I wonder if we should explore these options too as a follow-up:
- Using the [status](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_status_role) or [progressbar](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role) role if we feel either represents our spinner more accurately
- There is a `ScanningSpinner` in the electron folder; is something about that one electron specific? If not we may be able to reuse it

src/DetailsView/components/issues-table.tsx
```diff
@@ -182,7 +182,11 @@ export class IssuesTable extends React.Component<IssuesTableProps> {
     }
 
     private renderSpinner(label: string): JSX.Element {
-        return <Spinner className="details-view-spinner" size={SpinnerSize.large} label={label} />;
+        return (
+            <div role="alert">
```
karanbirsingh: It says [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) that:

> To maximize compatibility, some people recommend adding a redundant aria-live="assertive" when using this role. However, adding both aria-live and role="alert" causes double speaking issues in VoiceOver on iOS.

I'm not sure which is better, but `ScanningSpinner` has the redundant property, so we may want to be consistent

src/DetailsView/components/issues-table.tsx
smoralesd: I don't think we have a good case for progressbar basically because we don't have progression information from axe (it's "just a callback" that you have to wait until it finishes, no updates in the middle of it).
As per the electron specific spinner: for some reason using `aria-live="assertive"` on the browser extension does not work as expected so I decided to apply the role directly in the case of `issues-table.tsx`

src/DetailsView/components/issues-table.tsx
smoralesd: Yeap, I tried using '`aria-live`' here same as on the electron case, but for some reason, NVDA and JAWS didn't produce the desire effect (I haven't test VoiceOver though). This reason is related to my answer on your other comment.

src/DetailsView/components/issues-table.tsx
smoralesd: I'm updating `ScanningSpinner` so we can pass more properties (`aria-live` and `label`) so we can actually use the same component on both cases. Thanks for the feedback.

smoralesd: Updated

src/DetailsView/components/issues-table.tsx
smoralesd: Updated

src/DetailsView/components/issues-table.tsx
smoralesd: So, we decided to leave the usage of status or progressbar role outside of this PR scope. It's a very good point that may help us improve even more the spinner experience for AT users but it seems that we need a little investigation before moving forward with this idea.

PR #1523, opened 2019-10-21
fix: add failures detected AT announcement on CardsView

smoralesd:
#### Description of changes

Our current failure details list (using Office UI Fabric' DetailsList) includes an AT alert. When the scan is finished and there are failures, AT announce: "X failures detected". Our current `CardsView` does not have this alert.

In this PR:
- Add the alert on the `CardsView`, i.e. on the details view.
- Make sure the automated checks report does not have the alert (as the report is an static document)

#### Pull request checklist

   - there is no issue filed against GitHub
   - not affected
   - NVDA/JAWS under Windows 10, no VoiceOver

### DIFF
--- a/src/common/components/cards/result-section-title.tsx
+++ b/src/common/components/cards/result-section-title.tsx
@@ -11,13 +11,21 @@ export type ResultSectionTitleProps = {
     title: string;
     badgeCount: number;
     outcomeType: InstanceOutcomeType;
+    shouldAlertFailuresCount?: boolean;
 };
 
 export const ResultSectionTitle = NamedFC<ResultSectionTitleProps>('ResultSectionTitle', props => {
+    const failureTerm = props.badgeCount > 1 ? 'failures were' : 'failure was';
+    const alertingFailuresCount = (
+        <span role="alert">
+            {props.badgeCount} {failureTerm} detected.
+        </span>
+    );
+
     return (
         <span className={resultSectionTitle}>
             <span className="screen-reader-only">
-                {props.title} {props.badgeCount}
+                {props.title} {props.shouldAlertFailuresCount ? alertingFailuresCount : props.badgeCount}
             </span>
             <span className={title} aria-hidden="true">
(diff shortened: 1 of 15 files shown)


src/common/components/cards/result-section-title.tsx
```diff
@@ -11,13 +11,21 @@ export type ResultSectionTitleProps = {
     title: string;
     badgeCount: number;
     outcomeType: InstanceOutcomeType;
+    shouldAlertFailuresCount?: boolean;
 };
 
 export const ResultSectionTitle = NamedFC<ResultSectionTitleProps>('ResultSectionTitle', props => {
+    const failureTerm = props.badgeCount > 1 ? 'failures were' : 'failure was';
```
DaveTryon: In a case with 0 failures, will this announce "0 failures was..." ?

src/tests/unit/tests/reports/components/report-sections/__snapshots__/report-body.test.tsx.snap
```diff
@@ -86,6 +86,7 @@ exports[`ReportBody renders 1`] = `
         "violations": Array [],
       }
     }
+    shouldAlertFailuresCount={false}
```
DaveTryon: Observation: Every instance in the snapshot sets shouldAlertFailuresCount to false. Should there be any test cases where it's set to true?

smoralesd: should not as `ReportBody` is only used on the automated check report, which by definition, should not have any alert.

src/common/components/cards/result-section-title.tsx
smoralesd: Good catch, I was fighting with some weird behavior and totally forgot about 0. Will update presently

smoralesd: updated.

PR #1532, opened 2019-10-22
feat(card-selection): add accessibility announcements for cards selection 

Shobhit1:
#### Description of changes

After discussion with Rob, Fer & Wahaj, we decided to add some explicit text to be announced on card focus and kebab menu focus.

#### NVDA output:

For card:
`html card  row  not selected` 

For Kebab Menu:

`More Actions html html-has-lang
menu button  collapsed  subMenu`

#### Pull request checklist

### DIFF
--- a/src/common/components/cards/instance-details.tsx
+++ b/src/common/components/cards/instance-details.tsx
@@ -67,31 +67,33 @@ export const InstanceDetails = NamedFC<InstanceDetailsProps>('InstanceDetails', 
 
     const kebabMenuAriaLabel = `More Actions ${result.identifiers.identifier} ${result.ruleId}`;
     return (
-        <div
-            className={instanceDetailsCardStyling}
-            tabIndex={0}
-            onClick={cardClickHandler}
-            onKeyDown={cardKeyPressHandler}
-            aria-selected={result.isSelected}
-            aria-label={`${result.identifiers.identifier} card`}
-            role="row"
-        >
-            <table className={reportInstanceTable}>
-                <tbody>
-                    {renderCardRowsForPropertyBag(result.identifiers)}
-                    {renderCardRowsForPropertyBag(result.descriptors)}
-                    {renderCardRowsForPropertyBag(result.resolution)}
-                </tbody>
-            </table>
-            <InstanceDetailsFooter
-                deps={deps}
-                result={result}
-                highlightState={highlightState}
-                userConfigurationStoreData={userConfigurationStoreData}
-                rule={rule}
-                targetAppInfo={targetAppInfo}
-                kebabMenuAriaLabel={kebabMenuAriaLabel}
-            />
+        <div role="table">
+            <div
+                className={instanceDetailsCardStyling}
+                tabIndex={0}
+                onClick={cardClickHandler}
+                onKeyDown={cardKeyPressHandler}
+                aria-selected={result.isSelected}
+                aria-label={`${result.identifiers.identifier} card`}
+                role="row"
+            >
+                <table className={reportInstanceTable}>
+                    <tbody>
+                        {renderCardRowsForPropertyBag(result.identifiers)}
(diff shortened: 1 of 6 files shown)


src/tests/unit/tests/common/components/cards/card-kebab-menu-button.test.tsx
```diff
@@ -144,6 +144,22 @@ describe('CardKebabMenuButtonTest', () => {
...
+        ariaLabel => {
+            const newProps = {
+                ...defaultProps,
+                kebabMenuAriaLabel: ariaLabel,
+            };
+            const rendered = shallow(
+                <CardKebabMenuButton {...newProps} deps={{ ...defaultDeps, cardInteractionSupport: allCardInteractionsSupported }} />,
+            );
+
+            expect(rendered.debug()).toMatchSnapshot();
```
smoralesd: why `.debug()` instead of `.getElement()`?

src/tests/unit/tests/common/components/cards/card-kebab-menu-button.test.tsx
```diff
@@ -144,6 +144,22 @@ describe('CardKebabMenuButtonTest', () => {
...
+            expect(rendered.find(ActionButton).prop('menuProps')).toMatchSnapshot();
```
smoralesd: I think in general, when having more than one `.toMatchSnapshot();` in the same test, we add a comment so it's a little easier to differentiate the snapsnots on the .snap file

smoralesd: if you look at the .snap file on this PR it shows:
`CardKebabMenuButtonTest renders per snapshot with allCardInteractionsSupported and undefined aria label passed as prop` witth 2 different suffixes: `1` and `2`... not sure which one is which, having the comment helps readability

src/common/components/cards/instance-details.tsx
```diff
@@ -65,8 +65,17 @@ export const InstanceDetails = NamedFC<InstanceDetailsProps>('InstanceDetails',
...
     return (
-        <div className={instanceDetailsCardStyling} tabIndex={0} onClick={cardClickHandler} onKeyDown={cardKeyPressHandler}>
+        <div
+            className={instanceDetailsCardStyling}
+            tabIndex={0}
+            onClick={cardClickHandler}
+            onKeyDown={cardKeyPressHandler}
+            aria-selected={result.isSelected}
+            aria-label={`${result.identifiers.identifier} card`}
+            role="row"
```
smoralesd: Adding this comment base on @DaveTryon comment on teams:
Why is `role="row"' necessary here?

Shobhit1: Because `aria-selected` only works with certain `roles` and `row` seems to be the most appropriate for the card here.
https://www.w3.org/WAI/PF/aria/states_and_properties#aria-selected

src/tests/unit/tests/common/components/cards/card-kebab-menu-button.test.tsx
Shobhit1: Sure. Will add a comment.

src/tests/unit/tests/common/components/cards/card-kebab-menu-button.test.tsx
Shobhit1: `debug` gave me more information than `getElement` in this case; so used that.

src/common/components/cards/instance-details.tsx
peterdur: Using [role="row"](https://www.w3.org/WAI/PF/aria/roles#row) requires being contained in a rowgroup, grid, or treegrid... the current state results in `A11YSelfValidator.validate();` automated checks failures when rows are visible. 

We probably want an E2E test that does a check on the cards UX with some rules expanded.

PR #1533, opened 2019-10-22
test: add e2e for automated checks

smoralesd:
#### Description of changes

Add end to end test for the new automated checks view on AI-android.

**Notes**
- first 5 commits are actually a small refactor to enable code reuse on the new test

#### Pull request checklist


src/tests/electron/tests/automated-checks.view.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import { Application } from 'spectron';
+import { createApplication } from 'tests/electron/common/create-application';
+import { dismissTelemetryOptInDialog } from 'tests/electron/common/dismiss-telemetry-opt-in-dialog';
+import { AutomatedChecksViewSelectors } from 'tests/electron/common/element-identifiers/automated-checks-view-selectors';
+import { CommonSelectors } from 'tests/electron/common/element-identifiers/common-selectors';
+import { testResourceServerConfig } from 'tests/electron/setup/test-resource-server-config';
+import { DEFAULT_ELECTRON_TEST_TIMEOUT_MS } from 'tests/electron/setup/timeouts';
+import * as WebDriverIO from 'webdriverio';
+
+// spectron wraps calls to electron APIs as promises. Unfortunately, only electron typings are used,
```
dbjorge: How many of these cases are there? I think it would be nicer for us to add typings files overriding the default electron typings rather than disabling the warning; practically speaking, the "no-floating-promises" warnings are also going to be disabled by this issue, which have historically been a really common source of test flakiness

src/tests/electron/tests/automated-checks.view.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    it('show automated checks results', async () => {
+        const ruleGroups = await client.$$(`div${collapsibleRuleDetailsGroupSelector}`);
+
+        expect(ruleGroups).toHaveLength(4);
+
+        await assertExpandedRuleGroup(1, 'ImageViewName', 1);
+        await assertExpandedRuleGroup(2, 'ActiveViewName', 2);
+        await assertExpandedRuleGroup(3, 'TouchSizeWcag', 1);
+        await assertExpandedRuleGroup(4, 'ColorContrast', 1);
+    });
```
dbjorge: The web E2E tests apply timeout parameters to the test cases; I'd expect the electron cases to have similar startup time requirements, so maybe we should be doing that in the electron e2es also?

src/tests/electron/tests/automated-checks.view.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+// spectron wraps calls to electron APIs as promises. Unfortunately, only electron typings are used,
+// so tslint thinks some of the methods do not return promises.
+// tslint:disable: await-promise
+
+describe('AutomatedChecksView', () => {
+    let app: Application;
+    let client: WebDriverIO.Client<void>;
+
+    const collapsibleRuleDetailsGroupSelector = '.collapsible-rule-details-group';
```
dbjorge: Should this (and the other selectors/selector generation templates in this file) be in `automated-checks-view-selectors`?

src/tests/electron/tests/automated-checks.view.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    async function assertExpandedRuleGroup(position: number, expectedTitle: string, expectedFailures: number): Promise<void> {
+        const title = await client.$(`div${collapsibleRuleDetailsGroupSelector}:nth-of-type(${position}) .rule-details-id`).getText();
+        expect(title).toEqual(expectedTitle);
+
+        const failures = await client.$$(`div${collapsibleRuleDetailsGroupSelector}:nth-of-type(${position}) li`);
+        expect(failures).toHaveLength(expectedFailures);
+    }
+
+    async function ensureAppIsInDeviceConnectionDialog(): Promise<void> {
+        await client.waitForVisible(CommonSelectors.rootContainer, DEFAULT_ELECTRON_TEST_TIMEOUT_MS);
```
dbjorge: In the web e2e tests we use separate timeouts for element waits vs entire test timeouts, since it's awkward to debug cases where the test times out before individual element awaits do. Should we be doing that here, rather than re-using the test-level timeout for element-level interactions?

src/tests/electron/tests/automated-checks.view.test.ts
```diff
@@ -0,0 +1,94 @@
...
+        client = app.client;
+    });
+
+    beforeEach(async () => {
+        await dismissTelemetryOptInDialog(app);
+
+        await ensureAppIsInDeviceConnectionDialog();
+
+        // inputs the port, validate and start scanning
+        await client.click(CommonSelectors.portNumber);
```
dbjorge: Would it be better to add an `AutomatedChecksViewPage` type controller now like we have for the web pages?

src/tests/electron/tests/automated-checks.view.test.ts
smoralesd: That's an issue I think. 
On one side, my understanding is that we are "forced" to use certaing version of `webdriverio` based on the version of `spectron` we use. I'm not sure why this is or if I'm right about it.
On another side, I tried to add a dev dependency to `@types/webdriverio` but types seems incorrect to me (no Promises for example). I guess we could investigate this a little bit more but that should be a different piece of work, a different PR

src/tests/electron/tests/automated-checks.view.test.ts
smoralesd: This is a good point! I'll look at what browser extension e2e have for timeouts and I'll try to follow the same/similar pattern

src/tests/electron/tests/automated-checks.view.test.ts
smoralesd: Good catch. I got totally absorb by a different problem with the selectors. I was actually looking forward to comments around those. I'll move this selectors to the apropiate place.

src/tests/electron/tests/automated-checks.view.test.ts
smoralesd: Same as before: I'll look at what browser extension e2e have for timeouts and I'll try to follow the same/similar pattern

src/tests/electron/tests/automated-checks.view.test.ts
smoralesd: This is also a great point. I'm still concerned about how "usable" is to use spectron/webdriveio. I don't feel it the tests are easy to write considering I spend some good time figuring out how to get the elements I needed to click or verify. If you look at the selectors, they are no simple selectors (basically, div + class follow by another class or `li`), I didn't find a way to select and element and then select childrens from said element, which from my point of view would be ideal. The main reason for this is: webdriveio's Client does not return an actual html element (which have the `querySelector`/`querySelectorAll` function) but returns a webdriveio element wrapper.

So, to summaries, I agree that we should follow the "page controller" pattern we already have on the browser extension but I'm still worried about what spectron/webdriveio allows to do for writing 2e2 tests

smoralesd:
I'm going to close this PR for now. I'm creating smaller PR's for some of the "required" changes for this new test to be added: #1540 and #1541 
I'll create a new PR with just the new changes for adding the 2e2 test, considering the comments on this PR.

PR #1547, opened 2019-10-23
feat: update highlight state in cards

waabid:
#### Description of changes

This updates the view model for cards to include data about the highlight state of a card and additionally shows it in the UI. Since it is in the CardResult, this means we do not have it as a property in the footer.

image

#### Pull request checklist


src/tests/unit/common/rule-based-view-model-provider.test.ts
```diff
@@ -34,24 +34,24 @@ describe('RuleBasedViewModelProvider', () => {
     });
 
     const testStubCombinations = [
-        { isExpanded: true, isSelected: true },
-        { isExpanded: false, isSelected: false },
-        { isExpanded: true, isSelected: false },
+        { isExpanded: true, isSelected: true, isHighlighted: true },
```
smoralesd: this make me think about our model:
- is it possible to have something selected but not highlighted? 
- is it possible to have something highlighted but not selected?

If it's possible, we're missing test cases here. If its **not** possible, then the follow up question is, do we need both flags?

waabid: It is possible; good call. Will add a stub for this.

PR #1570, opened 2019-10-28
chore: css modules will always use the hash value for generating class names

waabid:
#### Description of changes

We've had a few issues recently where the css loads fine locally but then we run into issues in prod/insider. This is occurring due to how css-modules work. A bit of context:

The way to use css-modules is to import classNames in our code (ts/tsx files) straight from our css files. This is beneficial for us in the sense that our code/components will be directly linked to our styles rather than softly coupling our code by explicitly referencing class names in two separate places (both ts and css files).

Additionally, css-modules has a feature to uniquely generate classNames on compile time, ensuring that if a className, such as "header" is used in separate components, their css will not class because at compile time they may be generated as "headerX" and "headerY" (where X and Y are unique hashes; in our implementation, these are based off filenames). Previously, we did not do this for local builds and our e2e tests. As a result, if there was a component with some class name that was relying on css-modules but that class name was also referenced in our non-bundled css file (i.e. details-view.css), there would be a style collision **locally** but not in production/insider (since the class name here would be hashed). As such, you'd get this odd behavior where css would look one way locally but another way in insider/production, increasing the chances for bugs.

This PR ensures we use the hashed css names locally as well. The benefit of using the non-hashed names was that we would not into issues if we wanted to query for selectors in our e2e tests and makes css otherwise a bit easier to understand when working locally. The former can be dealt with by using data-* attributes (in this case, I went with data-automation-id) and the latter while a nice to have is not as important as ensuring we see these css issues locally.

#### Pull request checklist

### DIFF
--- a/src/tests/end-to-end/common/element-identifiers/details-view-selectors.ts
+++ b/src/tests/end-to-end/common/element-identifiers/details-view-selectors.ts
@@ -1,10 +1,10 @@
+import { overviewHeadingAutomationId } from 'DetailsView/components/overview-content/overview-heading';
+
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 export const detailsViewSelectors = {
     previewFeaturesPanel: '.preview-features-panel',
-    previewFeaturesPanelToggleList: '.preview-feature-toggle-list',
 
-    testNavArea: 'nav',
     testNavLink: (testName: string): string => `nav [name=${testName}] a`,
 
     mainContent: '[role=main]',
@@ -19,5 +19,5 @@ export const detailsViewSelectors = {
 
 export const overviewSelectors = {
     overview: '.overview',
-    overviewHeading: '.overview-heading',
+    overviewHeading: `[data-automation-id=${overviewHeadingAutomationId}]`,
 };
--- a/src/tests/end-to-end/tests/details-view/__snapshots__/preview-features.test.ts.snap
+++ b/src/tests/end-to-end/tests/details-view/__snapshots__/preview-features.test.ts.snap
@@ -71,7 +71,7 @@ exports[`Preview Features Panel Normal mode should match content in snapshot 1`]
             class="ms-Panel-content content-000"
           >
             <div
-              class="no-preview-feature-message"
+              class="no-preview-feature-message116bs"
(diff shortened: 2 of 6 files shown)


webpack.config.js
```diff
@@ -7,7 +7,7 @@ const TerserWebpackPlugin = require('terser-webpack-plugin');
 const CaseSensitivePathsPlugin = require('case-sensitive-paths-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 
-const getCSSModulesLoadersConfig = isDevMode => {
+const getCSSModulesLoadersConfig = () => {
```
dbjorge: It might be more clear to inline this into `commonConfig` alongside the `.tsx?` rule and remove the separate `devModules`/`prodModules` intermediates, now that it isn't devMode-dependent.

src/tests/end-to-end/tests/details-view/__snapshots__/preview-features.test.ts.snap
```diff
@@ -71,7 +71,7 @@ exports[`Preview Features Panel Normal mode should match content in snapshot 1`]
             class="ms-Panel-content content-000"
           >
             <div
-              class="no-preview-feature-message"
+              class="no-preview-feature-message116bs"
```
dbjorge: Consider normalizing the hash suffix in snapshots so e2e tests don't get broken if whitespace changes in a referenced CSS file; you could do this by plugging into the same helper we use to normalize the numbers in office fabric classnames in `src/tests/end-to-end/common/element-snapshot-formatter.ts` (it might be easier to distinguish the hashes if you change the `localIdentName` format to include some consistent separator for the hash, eg `'[local]---[hash:base64:5]'`)

src/tests/end-to-end/common/element-identifiers/details-view-selectors.ts
```diff
@@ -1,10 +1,10 @@
+import { overviewHeadingAutomationId } from 'DetailsView/components/overview-content/overview-heading';
```
dbjorge: Nice

dbjorge:
Per discussion with @waabid , picking this up to continue in #1811, closing this PR in favor of the newer one

PR #1583, opened 2019-10-30
feat: tab controller only fires tab update actions when URL has changed

waabid:
#### Description of changes

We don't really care for all tab changes; only ones that affect the URL (although I can see us caring for title at some point as well). As a result, we shouldn't fire tab update actions until we see this.

The result is we do not throw data away after scroll events cause tab update events (for loading, for example).

#### Pull request checklist

### DIFF
--- a/src/background/tab-controller.ts
+++ b/src/background/tab-controller.ts
@@ -49,7 +49,7 @@ export class TabController {
         this.browserAdapter.addListenerToTabsOnRemoved(this.onTargetTabRemoved);
         this.browserAdapter.addListenerOnWindowsFocusChanged(this.onWindowFocusChanged);
         this.browserAdapter.addListenerToTabsOnActivated(this.onTabActivated);
-        this.browserAdapter.addListenerToTabsOnUpdated(this.handleTabUpdate);
+        this.browserAdapter.addListenerToTabsOnUpdated(this.handleTabUpdateOnUrlHasChanged);
 
         this.detailsViewController.setupDetailsViewTabRemovedHandler(this.onDetailsViewTabRemoved);
     }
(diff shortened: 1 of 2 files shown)


src/background/tab-controller.ts
```diff
@@ -107,6 +107,12 @@ export class TabController {
         }
     };
 
+    private handleTabUpdateOnUrlHasChanged = (tabId: number, changeInfo: chrome.tabs.TabChangeInfo): void => {
```
dbjorge: The naming here confused me a little on first reading; I think it might be more clear to make this new helper be named `handleTabUpdate` and rename the original thing to be named `onTabUrlUpdate` or something

src/tests/unit/tests/background/tab-controller.test.ts
```diff
@@ -653,11 +653,80 @@ describe('TabControllerTest', () => {
...
+                getTabCallback = cb;
+                onReject = reject;
+            })
+            .verifiable(Times.once());
+        logMock.setup(log => log(`changed tab with Id ${tabId} not found`)).verifiable(Times.once());
+
+        testSubject = createTabControllerWithoutFeatureFlag(tabInterpreterMap);
+        testSubject.initialize();
+        tabUpdatedCallback(tabId, { url: 'some url' });
+        getTabCallback(getTabCallbackInput as any);
```
dbjorge: Does Chrome actually have a case where it will invoke both the success *and* the reject callback in response to a single getTab call? Would it make more sense for "call getTabCallback" and "call onReject" cases to be split out into separate test cases?

smoralesd:
This PR should fix #1583

src/background/tab-controller.ts
waabid: I changed it up a bit; let me know what you think!

PR #1590, opened 2019-10-31
refactor: remove unused updateVisibility workflow

karanbirsingh:
#### Description of changes

In some of our visualizations, we used to check whether elements were still visible in the DOM and update our highlighting to match. We don't do this anymore. For example, microsoft.com has a carousel. Running the headings ad-hoc test will show which headings were visible at a single point in time. The heading box stays there even if the carousel moves and the heading disappears. This is true in production.

Some of the code that enabled the old workflow (dynamically add and remove visualizations based on element visibility) still remains in our codebase but isn't exercised. This PR removes all instances I could find of leftover instance-visibility concepts. Much of the client logic / usages used to be in  `instance-visibility-checker` which does not exist anymore: https://github.com/microsoft/accessibility-insights-web/blob/v2.10.0/src/injected/instance-visibility-checker.ts

The assessment-store in production is not aware of instance-visibility - assuming I'm not missing anything, removing the remaining plumbing (action creator, etc) should have no impact on behavior.

This PR includes many files in order to keep these deletions together semantically - that way in the future it might be easier to remember what happened.

This PR does not include any updates to feature flags. It looks like we have a feature flag for showing instance visibility. A different PR will remove this.

#### Pull request checklist

### DIFF
--- a/src/ad-hoc-visualizations/color/visualization.tsx
+++ b/src/ad-hoc-visualizations/color/visualization.tsx
@@ -48,6 +48,5 @@ export const ColorAdHocVisualization: VisualizationConfiguration = {
     getDrawer: provider => provider.createSingleTargetDrawer('insights-grey-scale-container'),
     getSwitchToTargetTabOnScan: () => false,
     getInstanceIdentiferGenerator: () => generateUID,
-    getUpdateVisibility: () => false,
     guidance,
 };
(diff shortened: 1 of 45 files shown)


karanbirsingh:
Missed some files - I'll add those in too so it's a single semantic change

PR #1604, opened 2019-11-01
feat: rendering the screenshot from the store in ScreenshotView

JGibson2019:
#### Description of changes
This PR represents the ability to render the screenshot image stored in data from a scan; if there is no screenshot image, the view will inform the users. I also cleaned up scan-results-helper, as the way I was creating a ScreenshotData object was incorrect. 

Image of the screenshot being rendered from the scan data:
image

#### Pull request checklist

- [N/A] Addresses an existing issue: #0000
- [X] Ran `yarn fastpass`
- [X] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). Check workflow guide at: `<rootDir>/docs/workflow.md`


src/electron/views/screenshot/screenshot-view.tsx
```diff
@@ -1,13 +1,31 @@
...
-            <h1 tabIndex={0}>This is the screenshot view</h1>
-        </div>
-    );
+import { Screenshot } from './screenshot';
+
+export type ScreenshotViewProps = {
+    screenshotData: ScreenshotData;
+};
+
+const screenshotAltText: string = 'axe android results screenshot';
```
lisli1: Maybe out of scope for this PR, but we should think about where to get the alt text for the screenshot and if this default alt text is sufficient.

JGibson2019:
Just found that I broke the automated-checks-view-test, so I'm closing until I fix that!

src/electron/views/screenshot/screenshot-view.tsx
JGibson2019: I agree; where should we get the alt-text from @ferBonnin and @dbjorge? Or is hard coding it ok for now?

ferBonnin: mmm I might add something like 
"Automated checks results highlighted in a screenshot of the app"

JGibson2019: Updated in the most recent commit

src/electron/views/automated-checks/automated-checks-view.tsx
```diff
@@ -68,7 +69,8 @@ export class AutomatedChecksView extends React.Component<AutomatedChecksViewProp
                         </main>
                     </div>
                     <div className={screenshotView}>
-                        <ScreenshotView />
+                        {this.renderScreenshotView()}
+                        {/* <ScreenshotView screenshotData={this.props.unifiedScanResultStoreData.screenshotData} /> */}
```
smoralesd: Should remove commented code.

src/electron/views/screenshot/screenshot-view.tsx
```diff
@@ -1,13 +1,31 @@
...
+import { Screenshot } from './screenshot';
+
+export type ScreenshotViewProps = {
+    screenshotData: ScreenshotData;
+};
+
+const screenshotAltText: string = 'axe-android results screenshot with highlighted components';
+
+export const ScreenshotView = NamedFC<ScreenshotViewProps>('ScreenshotView', (props: ScreenshotViewProps) => {
+    if (isUndefined(props.screenshotData) || isEmpty(props.screenshotData)) {
```
smoralesd: lodash's `isEmpty` already cover the `undefined` case:

image

src/electron/views/screenshot/screenshot-view.tsx
```diff
@@ -1,13 +1,31 @@
...
+        return (
+            <div>
```
smoralesd: is it absolutely necessary to have this sctructure? `<div>` > `<p>`?

src/electron/views/screenshot/screenshot-view.tsx
```diff
@@ -1,13 +1,31 @@
...
+                <p>Screenshot for scan is unavailable</p>
+            </div>
+        );
+    } else {
+        return (
+            <div>
+                <h1 tabIndex={0}>Page state screenshot</h1>
```
smoralesd: 2 questions here:
- is it **Page state screenshot** the right string here? (I didn't find figma mocks for this feature)
- is the `tabIndex` on purpose?

src/tests/unit/tests/electron/views/screenshot/screenshot-view.test.tsx
```diff
@@ -5,10 +5,22 @@ import { shallow } from 'enzyme';
 import * as React from 'react';
 
 describe('ScreenshotView', () => {
-    it('renders', () => {
-        const props: ScreenshotViewProps = {};
-        const wrapper = shallow(<ScreenshotView {...props} />);
+    describe('renders', () => {
+        it('when passed a value for screenshotData', () => {
+            const props: ScreenshotViewProps = {
+                screenshotData: { base64PngData: 'string' },
```
smoralesd: instead of `string` which is a type in typescript, would it be better to have something in the lines of `"test-base-64-png-data"`??

src/electron/views/automated-checks/automated-checks-view.tsx
```diff
@@ -120,4 +122,11 @@ export class AutomatedChecksView extends React.Component<AutomatedChecksViewProp
             />
         );
     }
+
+    private renderScreenshotView(): JSX.Element {
+        if (isUndefined(this.props.unifiedScanResultStoreData)) {
```
smoralesd: do we want to use `isEmpty` here??

src/electron/views/automated-checks/automated-checks-view.tsx
```diff
@@ -68,7 +69,8 @@ export class AutomatedChecksView extends React.Component<AutomatedChecksViewProp
                         </main>
                     </div>
                     <div className={screenshotView}>
```
smoralesd: The structure here is interesting! if  `this.props.unifiedScanResultStoreData` is undefined (acording to line 127 on the changed file) we'll get something like
```jsx
<div ...>
   {null}
</div>
```
Is that on purpose?

src/tests/unit/tests/electron/views/screenshot/screenshot-view.test.tsx
```diff
@@ -5,10 +5,22 @@ import { shallow } from 'enzyme';
...
+            };
+            const wrapper = shallow(<ScreenshotView {...props} />);
 
-        expect(wrapper.getElement()).toMatchSnapshot();
+            expect(wrapper.getElement()).toMatchSnapshot();
+        });
+
+        it('when passed an empty value for screenshotData', () => {
+            const props = {} as ScreenshotViewProps;
```
smoralesd: I think you should cover `props.screenshotData` being empty. It may also be worthy to check for `null` and `undefined` values, as it can be done using `it.each`

src/electron/views/screenshot/screenshot-view.tsx
smoralesd: If the alt text is not being use anywhere else, it should be fine here. We don't have a proper "resources" system right now.

src/electron/views/screenshot/screenshot-view.tsx
JGibson2019: 1. tabIndex is on purpose; the screen reader I tested with skipped the heading if it wasn't there on my last PR
2. I got this phrase from the figma

lisli1: There are some figma mocks here: https://www.figma.com/file/Bugs8uNCHdCzOKzhq06nNK2o/Mobile---User-flow?node-id=249%3A2058

smoralesd: - The tabIndex sounds odd, can we double check? (I can help with this from my box)
- Can we confirm? It sounds odd to me

src/electron/views/automated-checks/automated-checks-view.tsx
JGibson2019: yes; changing it

src/electron/views/screenshot/screenshot-view.tsx
JGibson2019: Changed

src/electron/views/screenshot/screenshot-view.tsx
JGibson2019: It's not; removed the <p> tags

src/tests/unit/tests/electron/views/screenshot/screenshot-view.test.tsx
JGibson2019: changing!

src/electron/views/automated-checks/automated-checks-view.tsx
JGibson2019: done

ferBonnin:
talked whit Sebastian on the "page state screenshot" I agree its on the Figma, but it does sound very odd. Lets do "Target App Screenshot"

src/electron/views/screenshot/screenshot-view.tsx
smoralesd: I checked with NVDA and I'm able to reach the header with arrow keys and the headings shortcut (I removed the `tabIndex` value locally to test)

src/electron/views/automated-checks/automated-checks-view.tsx
JGibson2019: I changed the renderScreenshotView method at some point Friday; is the current version (on the most recent commit) in line with the ideal structure?

JGibson2019:
> talked with Sebastián on the "page state screenshot" I agree its on the Figma, but it does sound very odd. Lets do "Target App Screenshot"

Updated in the latest commit

src/electron/views/automated-checks/automated-checks-view.tsx
JGibson2019: @smoralesd

src/tests/unit/tests/electron/views/screenshot/screenshot-view.test.tsx
```diff
@@ -1,19 +1,43 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { ScreenshotData } from 'common/types/store-data/unified-data-interface';
 import { ScreenshotView } from 'electron/views/screenshot/screenshot-view';
 import { ScreenshotViewModel } from 'electron/views/screenshot/screenshot-view-model';
 import { shallow } from 'enzyme';
 import * as React from 'react';
 
 describe('ScreenshotView', () => {
-    it('renders', () => {
```
dbjorge: I think we want to keep a test of the happy path where it renders the available screenshot layout

src/tests/unit/tests/electron/views/screenshot/screenshot-view.test.tsx
```diff
@@ -1,19 +1,43 @@
...
-        const viewModel: ScreenshotViewModel = {
-            screenshotData: null,
-            highlightBoxRectangles: [],
-            deviceName: null,
-        };
-        const wrapper = shallow(<ScreenshotView viewModel={viewModel} />);
+    describe('renders', () => {
+        it('when passed a null value for screenshotData', () => {
```
dbjorge: The view model provider already has the responsibility of normalizing "empty" screenshot data as null, so I'm not sure we need the other 2 cases here; if we do want them, it would probably be a little easier to read to use `.each`, like:

```ts
const emptyScreenshotDataCases: ScreenshotData[] = [null, undefined, {} as ScreenshotData];
it.each(emptyScreenshotDataCases)('when passed empty screenshotData %p', (screenshotData: ScreenshotData) => {
    // the same test implementation, but using the passed in screenshotData
});
```

src/electron/views/screenshot/screenshot-view.tsx
```diff
@@ -1,14 +1,26 @@
...
+const screenshotAltText: string = 'axe-android results screenshot with highlighted components';
+
+export const ScreenshotView = NamedFC<ScreenshotViewProps>('ScreenshotView', (props: ScreenshotViewProps) => {
+    if (isEmpty(props.viewModel.screenshotData)) {
+        return <div>Screenshot for scan is unavailable</div>;
+    }
     return (
         <div>
-            <h1 tabIndex={0}>This is the screenshot view</h1>
+            <h1 tabIndex={0}>Target Page Screenshot</h1>
```
smoralesd: We still have the `tabIndex` here... I think my original comment got lost somewhere

dbjorge: This is intentional because there has to be an element in tab order inside the independently scrollable region to accomodate https://dequeuniversity.com/rules/axe/3.3/scrollable-region-focusable - it's not sufficient for it to be arrow/header navigable since the rule is still applicable for non-screen-reader keyboard users

src/tests/unit/tests/electron/views/screenshot/screenshot-view.test.tsx
```diff
@@ -1,19 +1,34 @@
...
-        const viewModel: ScreenshotViewModel = {
-            screenshotData: null,
-            highlightBoxRectangles: [],
-            deviceName: null,
-        };
-        const wrapper = shallow(<ScreenshotView viewModel={viewModel} />);
+    describe('renders', () => {
+        it('when passed a value for screenshotData', () => {
+            const viewModel: ScreenshotViewModel = {
+                screenshotData: { base64PngData: 'testScreenshotPath' },
```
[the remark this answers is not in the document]
JGibson2019: Understood. I'll just change it to `testScreenshotData`

src/electron/views/screenshot/screenshot-view.tsx
smoralesd: leaving a comment sounds good.

JGibson2019: updating and adding the comment to link to the axe-core 3.3. rule on scrollabe regions!

PR #1612, opened 2019-11-05
feat(consistent-cards-UI): Add CardSelectionStore.toggleVisualHelper implementation

RobGallo:
#### Description of changes

Intended effects:

- When visual helper is toggled on
	- visual helper is enabled
	- nothing else changes
- When visual helper is toggled off
	- all selected cards are deselected
	- all deselected cards remain deselected
	- no rule expansion changes occur
	- visual helper is disabled
- When a card is selected and visual helper is disabled
	- the given card is selected
	- visual helper becomes enabled

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/background/stores/card-selection-store.ts
+++ b/src/background/stores/card-selection-store.ts
@@ -45,6 +45,12 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
         });
     };
 
+    private deselectAllCards = (): void => {
+        forOwn(this.state.rules, rule => {
+            this.deselectAllCardsInRule(rule);
+        });
+    };
+
     private toggleRuleExpandCollapse = (payload: RuleExpandCollapsePayload): void => {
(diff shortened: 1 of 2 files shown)


src/background/stores/card-selection-store.ts
```diff
@@ -74,11 +80,16 @@ export class CardSelectionStore extends BaseStoreImpl<CardSelectionStoreData> {
 
         rule.cards[payload.resultInstanceUid] = !rule.cards[payload.resultInstanceUid];
 
+        // whenever a card is selected, the visual helper is enabled
```
waabid: We strongly prefer to avoid using comments in our code and I'm not sure if it's necessary in this case.

I think we can create a variable that represents the "isSelected" property and instead the if check would read fairly well. This would also simplify some of your code above and make it more readable as you reach further into the objects more than once.

PR #1613, opened 2019-11-05
fix: adjust extension permissions

### LINKED ISSUE #1591: Canary hasn't updated since 10/24
## Describe the bug

Our canary releases are "in pending review" by the Chrome web store.
Canary is stuck since 10/24, push to insider took 1.5 days and Playground is also not in review.

Let's review Google's guidance and check if there is any permissions we can remove/edit to avoid having a compliance review for each release.

## Expected behavior

release should be available as fast as it was before (<1 hour)

## Additional context
This is blocking our ability to do fast development

smoralesd:
#### Description of changes

Using `"activeTab"` permissions instead of wide-matching patterns (`"https://*/*", "http://*/*", "file://*/*"`) in order to avoid getting flagged with **in pending review** status when updating the extension on google web store.

We have filed issue #1591 to track cannary not being properly update since 10/24 but this affect all of our extensions. This is related to chrome extension permissions and an effort to prevent malisious extension on the web store. Chrome developer website has a [transition guide](https://developer.chrome.com/extensions/runtime_host_permissions).

While changing the permissions seems to be working fine on the extension itself(we have a pending general-pass validation session from the team), the change has broke end to end test where we need to inject js/css to the target page (e.g, there is a test for the ad hoc panel where we enable each and every toggle that breaks with a permissions denied-kind of error).

This is due to how the `"activeTab"` permission works. Simply put, our extension gets permissions to inject js/css when the user activates our extension. There are basically 2 ways this happen:
1. the user clicks/press the extension icon on the extensions bar righ by the address bar
1. the user users our extension's shortcut to activate it.

We're using `Puppeteer` for our end-to-end tests and there is currently no API to efectivle use 1. as mentioned [here](https://github.com/GoogleChrome/puppeteer/issues/2486).

I tried option 2. by using `Puppeteer`'s [Keyboard](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-keyboard) api to no success. I also looked at a couple of npm packages that send keyboard strokes at the OS level (most notably: [kbm-robot](https://github.com/kylepaulsen/kbm-robot), [node-key-sender](https://github.com/garimpeiro-it/node-key-sender)) but they relay on `Java`, which seems a medium to mayor size/impact decission for our infrastructure. After that, I didn't want to invest more time with this approach as we need to unblock ourselves as fast as we can, finding a balance between time spend, robustness and correctness on the approach we choose.

The choosen approach was based on @dbjorge suggestion to follow this chrome ligthouse [PR](https://github.com/GoogleChrome/lighthouse/pull/4640) where they decided to modified the `manifest.json`, adding a permission to facilitated e2e testing.

#### Pull request checklist

- [x] Addresses an existing issue: fixes #1591
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/end-to-end/common/browser-factory.ts
+++ b/src/tests/end-to-end/common/browser-factory.ts
@@ -1,32 +1,54 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { generateUID } from 'common/uid-generator';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as Puppeteer from 'puppeteer';
 import * as util from 'util';
-import { generateUID } from '../../../common/uid-generator';
 import { Browser } from './browser';
 import { popupPageElementIdentifiers } from './element-identifiers/popup-page-element-identifiers';
 import { DEFAULT_BROWSER_LAUNCH_TIMEOUT_MS } from './timeouts';
 
 export const chromeLogsPath = path.join(__dirname, '../../../../test-results/e2e/chrome-logs/');
 
-export function browserLogPath(browserInstanceId: string): string {
-    return path.join(chromeLogsPath, browserInstanceId);
-}
+export const browserLogPath = (browserInstanceId: string): string => path.join(chromeLogsPath, browserInstanceId);
+
+const fileExists = util.promisify(fs.exists);
+const writeFile = util.promisify(fs.writeFile);
+const readFile = util.promisify(fs.readFile);
 
 export interface ExtensionOptions {
     suppressFirstTimeDialog: boolean;
+    addLocalhostToPermissions?: boolean;
 }
 
 export async function launchBrowser(extensionOptions: ExtensionOptions): Promise<Browser> {
     const browserInstanceId = generateUID();
-    const puppeteerBrowser = await launchNewBrowser(browserInstanceId);
-    const browser = new Browser(browserInstanceId, puppeteerBrowser);
+
+    // only unpacked extension paths are supported
+    const devExtensionPath = `${(global as any).rootDir}/drop/extension/dev/product`;
+    const manifestPath = getManifestPath(devExtensionPath);
+
+    let onClose: () => Promise<void>;
+
+    if (extensionOptions.addLocalhostToPermissions) {
+        const originalManifest = await readFile(manifestPath);
+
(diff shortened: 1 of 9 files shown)


src/tests/end-to-end/common/browser-factory.ts
```diff
@@ -55,10 +77,15 @@ async function verifyExtensionIsBuilt(extensionPath: string): Promise<void> {
     }
 }
 
-async function launchNewBrowser(browserInstanceId: string): Promise<Puppeteer.Browser> {
-    // only unpacked extension paths are supported
-    const extensionPath = `${(global as any).rootDir}/drop/dev/extension/`;
+function addLocalhostPermissionsToManifest(originalManifest: string): string {
+    const manifest = JSON.parse(originalManifest);
+
+    manifest['permissions'].push('http://localhost:9050/*');
```
dbjorge: Consider pulling this number out of `testResourceServerConfig.port`

src/tests/end-to-end/common/browser-factory.ts
```diff
@@ -1,32 +1,54 @@
...
-}
+export const browserLogPath = (browserInstanceId: string): string => path.join(chromeLogsPath, browserInstanceId);
+
+const fileExists = util.promisify(fs.exists);
+const writeFile = util.promisify(fs.writeFile);
+const readFile = util.promisify(fs.readFile);
 
 export interface ExtensionOptions {
     suppressFirstTimeDialog: boolean;
+    addLocalhostToPermissions?: boolean;
```
dbjorge: I think it would probably help prevent bugs for this to be non-optional; if a test wants to not have the extra permissions, it has to choose that explicitly

smoralesd: I see it the other way around: we don't give those permissions, unless the test explicitly set this flag to `true`

dbjorge: I don't think it should default to true, I'm saying I think it should maybe not have any default value (force every test to choose explicitly)

smoralesd: It defaults to `false` and having optional make test that don't need this permissions less verbose about it.

smoralesd:
This PR was more of an exploration and proof-of-concept. We have prove the permissions works and found at least one issue. I'll will abandon this PR and re-work it from latest master again. I'll be also taking care of the comments left on this PR.

PR #1615, opened 2019-11-05
refactor: introduce cards view model (which contains results by status in a property)

waabid:
#### Description of changes

Previously we had only card related data (in properties like: ruleResultsByStatus). We will now be adding additional data, such as whether the visual helper toggle is enabled or disabled and whether all rule groups are expanded; the former is the primary motivator to make this change.

Most of these changes are structuring the data to this new type.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/rule-based-view-model-provider.ts
```diff
@@ -9,20 +9,21 @@ import {
...
-export type GetUnifiedRuleResultsDelegate = (
+export type GetCardViewData = (
     rules: UnifiedRule[],
     results: UnifiedResult[],
     cardSelectionViewData: CardSelectionViewData,
-) => CardRuleResultsByStatus;
+) => CardsViewModel;
 
-export const getUnifiedRuleResults: GetUnifiedRuleResultsDelegate = (
+export const getUnifiedRuleResults: GetCardViewData = (
```
DaveTryon: Should this method be renamed to getCardViewData or something similar?

waabid: Updated

src/common/components/cards/result-section.tsx
```diff
@@ -4,8 +4,6 @@ import { css } from '@uifabric/utilities';
 import { NamedFC } from 'common/react/named-fc';
 import * as React from 'react';
 
-import { TargetAppData } from '../../../common/types/store-data/unified-data-interface';
```
DaveTryon: Edits in this file seems unrelated to the rest of the change. Opportunistic cleanup or is it actually needed?

waabid: The former; it was a bit confusing why these properties were declared twice, effectively speaking.

PR #1623, opened 2019-11-05
chore: add electron to dependabot ignore

Shobhit1:
#### Description of changes
Our `electron` dependency is based on the update frequency from internal electron build and so it doesn't make sense for dependabot to spam us on update from electron.
Adding `electron` to the ignored list.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -31,3 +31,8 @@ update_configs:
           # on false positives and breaking ai-web, so avoiding
           # dependabot auto-update for them like other deps.
           dependency_name: "axe-core"
+      - match:
+          # All electron builds are going to be manually updated
+          # since we use internal microsoft electron distributable
+          # doesn't make any sense to update this from dependabot
+          dependency_name: "electron"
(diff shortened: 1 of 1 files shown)


.dependabot/config.yml
```diff
@@ -31,3 +31,8 @@ update_configs:
           # on false positives and breaking ai-web, so avoiding
           # dependabot auto-update for them like other deps.
           dependency_name: "axe-core"
+      - match:
+          # All electron builds are going to be manually updated
+          # since we use internal microsoft electron distributable
```
dbjorge: I feel like this comment sounds a little ominous for an external user finding it ("why is Microsoft using an internal electron distributable? what are they injecting in there?").

Maybe slightly expand it to clarify "...since our release builds use a non-standard Electron build with distribution-restricted media codecs stripped out"?

Shobhit1: Updated the comment. Thanks for checking this!

PR #1624, opened 2019-11-05
fix: properly null check the view model in failed instances section

waabid:
#### Description of changes

The view model has the option be null when generated. This was missed in the re-factor.

#### Pull request checklist

- [n/a] Ran `yarn fastpass`


src/common/components/cards/failed-instances-section.tsx
```diff
@@ -20,7 +20,7 @@ export type FailedInstancesSectionProps = {
 export const FailedInstancesSection = NamedFC<FailedInstancesSectionProps>(
     'FailedInstancesSection',
     ({ cardsViewData, deps, userConfigurationStoreData, targetAppInfo, shouldAlertFailuresCount }) => {
-        if (cardsViewData.cards == null) {
+        if (cardsViewData == null) {
```
DaveTryon: Would we also want to catch the scenario where cardsViewData.cards is null?
if (cardsViewData == null || cardsViewData.cards == null)

PR #1596, opened 2019-11-01
feat(screenshot): create screenshot highlight box component

lisli1:
#### Description of changes

Created a highlight box component that will be used to visually highlight elements on the screenshot. The highlight boxes will render on top of the screenshot image.

**Important note:** The highlight box will position itself relative to its parent component due to the `position: absolute` property, but only when its parent component has `position: relative`. This is necessary to simplify rendering the highlight boxes on top of a screenshot image. The screenshot container that will render the screenshot image and all highlight boxes on top will need to include `position: relative` in its properties. This will be done in a future PR.

**Screenshot of a highlight box with hard-coded size and position on top of a screenshot image**
image

#### Pull request checklist

- [x] Addresses an existing issue: # 1628935
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). Check workflow guide at: `<rootDir>/docs/workflow.md`
- [x] (UI changes only) Added screenshots/GIFs to description above
- [ ] (UI changes only) Verified usability with NVDA/JAWS

PR #1629, opened 2019-11-06
feat(screenshot): create screenshot container

lisli1:
#### Description of changes

- Created a screenshot container with CSS `position: relative` to house the screenshot and highlight boxes. This is necessary to have the highlight boxes properly render on top of the screenshot, as mentioned previously in PR #1596.
- Updated relevant files and tests

#### Pull request checklist

- [x] Addresses an existing feature: # 1582212
- [x] Ran `yarn fastpass`


src/electron/views/screenshot/screenshot-container.scss
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+.screenshot-container {
+    position: relative !important;
```
dbjorge: Is the `!important` necessary?

src/electron/views/screenshot/highlight-box.tsx
```diff
@@ -8,7 +8,7 @@ import { CSSProperties } from 'react';
 
 export interface HighlightBoxDeps {}
 export interface HighlightBoxProps {
-    deps: HighlightBoxDeps;
+    deps?: HighlightBoxDeps;
```
dbjorge: We don't usually require that components be able to operate without their deps; if this component doesn't actually have any deps, it's best to just omit the deps prop and type entirely.

src/electron/views/screenshot/screenshot-container.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { BoundingRectangle } from 'electron/platform/android/scan-results';
+import { HighlightBox } from 'electron/views/screenshot/highlight-box';
+import { Screenshot } from 'electron/views/screenshot/screenshot';
+import { screenshotContainer } from 'electron/views/screenshot/screenshot-container.scss';
+import { isEmpty } from 'lodash';
+import * as React from 'react';
+
+export interface ScreenshotContainerProps {
+    screenshotData?: ScreenshotData;
+    highlightBoxRectangles?: BoundingRectangle[];
```
dbjorge: `highlightBoxRectangles` can be non-optional (it's not optional in the view model, so it's the view model provider's job to normalize it as an empty list)

src/electron/views/screenshot/screenshot-container.tsx
```diff
@@ -0,0 +1,44 @@
...
+    );
+});
+
+function renderHighlightBoxes(highlightBoxRectangles: BoundingRectangle[]): JSX.Element[] {
+    if (isEmpty(highlightBoxRectangles)) {
+        return;
+    }
+
+    const renderedBoxes: JSX.Element[] = [];
+    let keyId = 1;
```
dbjorge: You don't have to track the id yourself, `forEach` and `map` will pass a second index parameter to their callbacks that you can use:

```ts
const renderedBoxes = highlightBoxRectangles.map((rectangle, index) => {(
    <HighlightBox key={index} boundingRectangle={rectangle} />
)});
```

dbjorge: Using an index for a key id is probably acceptable for our purposes, but react is actually capable of being smarter about re-renders if we can use a key that remains stable when boxes are added/removed from the list. Probably the uid of the corresponding result would be ideal; I don't think changing the view model layout to pipe that association around is necessary for this PR, but it might be worth noting in a comment here that using the result uid would be better so we don't forget to make that update in the "interactive screenshot" feature when we'll need to start piping the UID in anyway.

src/electron/views/screenshot/screenshot-container.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    highlightBoxRectangles.forEach(instance => {
+        renderedBoxes.push(<HighlightBox key={keyId} boundingRectangle={instance} />);
```
dbjorge: Using an index for a key id is probably acceptable for our purposes, but react is actually capable of being smarter about re-renders if we can use a key that remains stable when boxes are added/removed from the list. Probably the uid of the corresponding result would be ideal; I don't think changing the view model layout to pipe that association around is necessary for this PR, but it might be worth noting in a comment here that using the result uid would be better so we don't forget to make that update in the "interactive screenshot" feature when we'll need to start piping the UID in anyway.

src/electron/views/screenshot/screenshot.tsx
```diff
@@ -5,9 +5,11 @@ import * as React from 'react';
 
 export type ScreenshotProps = {
     encodedImage: string;
-    altText: string;
+    altText?: string;
```
dbjorge: Seeing the optional altText param made me expect that there wouldn't be any alt text if it wasn't passed; I think if we're going to have Screenshot be the thing with responsibility for producing the alt text, it would maybe be more clear to just omit it from the props here

src/tests/unit/tests/electron/views/screenshot/__snapshots__/screenshot-view.test.tsx.snap
```diff
@@ -7,27 +7,49 @@ exports[`ScreenshotView renders when passed a value for screenshotData 1`] = `
...
 exports[`ScreenshotView renders when passed empty screenshotData {} 1`] = `
 <div>
-  Screenshot for scan is unavailable
+  <h1
+    tabIndex={0}
+  >
+    Target Page Screenshot
+  </h1>
+  <ScreenshotContainer
+    screenshotData={Object {}}
```
dbjorge: I think these empty cases in `screenshot-view.test.tsx` are probably obsolete and can be removed now that the "screenshot is unavailable" message responsibility has moved to the screenshot container

src/tests/unit/tests/electron/views/screenshot/screenshot-container.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+// Licensed under the MIT License.
+import { ScreenshotData } from 'common/types/store-data/unified-data-interface';
+import { BoundingRectangle } from 'electron/platform/android/scan-results';
+import { ScreenshotContainer, ScreenshotContainerProps } from 'electron/views/screenshot/screenshot-container';
+import { shallow } from 'enzyme';
+import * as React from 'react';
+
+describe('ScreenshotContainer', () => {
+    const basicScreenshotData: ScreenshotData = { base64PngData: 'test-base-64-png-data' };
+    const highlightBoxRectangles: BoundingRectangle[] = [
```
dbjorge: nit: this has a pretty general name but is only used in a single test case; I think it might be a little easier to follow that one test case if this were moved there.

src/electron/views/screenshot/screenshot-container.scss
lisli1: I thought it was necessary because the highlight boxes might not render on top of the screenshot correctly if it has some other value for the position property, and it's also used on the `insights-highlight-box` in `injected.scss`. Did you have any concerns regarding this that I should consider?

src/tests/unit/tests/electron/views/screenshot/screenshot-container.test.tsx
lisli1: Moved to the one test case

src/tests/unit/tests/electron/views/screenshot/__snapshots__/screenshot-view.test.tsx.snap
lisli1: Removed obsolete "empty screenshot" test cases

src/electron/views/screenshot/screenshot.tsx
lisli1: Removed altText param

src/electron/views/screenshot/screenshot-container.tsx
lisli1: Added comment for the future interactive screenshot feature

src/electron/views/screenshot/screenshot-container.tsx
lisli1: Thanks for the tip!

src/electron/views/screenshot/screenshot-container.tsx
lisli1: Made highlightBoxRectangles prop required and updated screenshot-container.test.tsx accordingly

src/electron/views/screenshot/highlight-box.tsx
lisli1: Removed HighlightBoxDeps

src/electron/views/screenshot/screenshot-container.scss
dbjorge: Using `!important` in CSS is usually an anti-pattern; it doesn't cooperate will with users that have special need for custom stylesheets, and it tends to be "leaky" (it forces other parts of CSS to use !important where they otherwise wouldn't need to). It is best to avoid using it unless there is some particularly compelling reason to need it; here, we shouldn't need it (there isn't anything that we'd normally expect to be overriding the class selector).

injected.scss is a bad example to follow because it is an exceptional case; it makes heavy use of `!important` because it has to contend with target pages whose styles we have no control over, which might themselves be using `!important` in ways that would otherwise override our injected styles. This component doesn't have that concern.

PR #1633, opened 2019-11-06
feat(consistent-cards-ui): Abstract out ExportResults properties

DaveTryon:
#### Description of changes

This connects the "Export results" component into the automated checks when cardsUI is enabled. To do this, the code to determine the properties for the "Export results" is moved into `report-export-component-props-factory.ts`, then plumbed through as needed. The `CommandBars` class is split into 2 classes, `AutomatedChecksCommandBar` and `AssessmentCommandBar`, although they may end up merged again once cardsUI is no longer behind a feature flag.

One of the benefits of this approach is that the code to build the CommandBar and the code to build the properties are now loosely coupled, so that the tests for each are isolated.

This doesn't touch the "Export results" component in the Issues table, since that is working fine for now and will eventually go away.

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/DetailsView/components/details-view-command-bar.tsx
```diff
@@ -60,41 +71,24 @@ export class DetailsViewCommandBar extends React.Component<DetailsViewCommandBar
...
                 <ReportExportComponent
-                    deps={deps}
-                    reportGenerator={reportGenerator}
-                    pageTitle={tabStoreData.title}
-                    exportResultsType={'Assessment'}
-                    scanDate={deps.getCurrentDate()}
-                    htmlGenerator={htmlGenerator}
-                    updatePersistedDescription={this.updatePersistedDescription}
-                    getExportDescription={this.getExportDescription}
+                    deps={this.props.reportExportComponentProps.deps}
```
waabid: Might be a good idea to use the spread operator instead:

```
<ReportExportComponent {...this.props.reportExportComponentProps} />
```

src/DetailsView/components/details-view-command-bar.tsx
```diff
@@ -10,23 +12,32 @@ import { FeatureFlagStoreData } from '../../common/types/store-data/feature-flag
...
     actionMessageCreator: DetailsViewActionMessageCreator;
     assessmentStoreData: AssessmentStoreData;
     assessmentsProvider: AssessmentsProvider;
-    renderExportAndStartOver: boolean;
+    reportExportComponentProps: ReportExportComponentProps;
+    renderStartOver: boolean;
     rightPanelConfiguration: DetailsRightPanelConfiguration;
+    visualizationScanResultData: VisualizationScanResultData;
+    cardsViewData: CardsViewModel;
+    reportExportComponentPropertyFactory?: ReportExportComponentPropertyConverter;
```
waabid: In what case does the `DetailsViewCommandBar `need both the factory and the ```reportExportComponentProps```?

As far as I can see it's not being used, so maybe just missed on clean up?

waabid: Ah, I see you're using the same interface for exposing the Assessment and FastPass command bars: I would instead create  a new interface for that purpose that includes this factory and not make it optional.

src/DetailsView/components/automated-checks-command-bar.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { NamedFC } from 'common/react/named-fc';
+import { ReportExportComponentProps } from 'DetailsView/components/report-export-component';
+import { getReportExportComponentPropsForAutomatedChecks } from 'DetailsView/components/report-export-component-props-factory';
+import { CommandBarProps, ReportExportComponentPropertyConverter } from './details-view-command-bar';
+import { DetailsViewCommandBar } from './details-view-command-bar';
+
+export const AutomatedChecksCommandBar = NamedFC<CommandBarProps>('AutomatedChecksCommandBar', props => {
+    const converter: ReportExportComponentPropertyConverter = props.reportExportComponentPropertyFactory;
+    let reportExportComponentProps: ReportExportComponentProps;
+    if (converter) {
```
waabid: I'm not fully following why this if-block is necessary in either command bars: when would the ```reportExportComponentPropertyFactory``` not exist?

src/DetailsView/components/details-view-switcher-nav.ts
```diff
@@ -41,17 +49,20 @@ export type GetDetailsSwitcherNavConfigurationProps = {
 
 const detailsViewSwitcherNavs: { [key in DetailsViewPivotType]: InternalDetailsViewSwitcherNavConfiguration } = {
     [DetailsViewPivotType.assessment]: {
-        CommandBar: CommandBarWithExportAndStartOver,
+        CommandBar: AssessmentCommandBar,
```
waabid: Hm, so I think we're effectively coupling the pivot type with the export button twice. Once here and another time inside of the specific command bars.

I think it'd be better to maybe move the methods to deps and then the command bars decided which one to use (eliminating the need to specify them here).

src/DetailsView/components/automated-checks-command-bar.tsx
DaveTryon: The override was added primarily for testing, so that I could test just the command bar in isolation from the factory method. In actual usage, the factory property is not being set. I'll try to move it into the deps

src/DetailsView/components/details-view-switcher-nav.ts
DaveTryon: The only difference now between AutomatedChecksCommandBar and AssessmentChecksCommandBar is the NamedFC. There is still an element of double coupling--let's discuss the value of the NamedFC distinction. If it's unimportant, then CommandBar can just be DetailsViewCommandBar and we're good to go

src/DetailsView/components/details-view-command-bar.tsx
DaveTryon: Done

src/DetailsView/components/details-view-command-bar.tsx
DaveTryon: Moved to deps. As discussed offline, the single interface is intentional

DaveTryon: As discussed, ran into problems with deps. Now picking up the property factories from the switcherNavConfiguration property

src/tests/unit/tests/DetailsView/components/details-view-command-bar.test.tsx
```diff
@@ -62,24 +69,41 @@ describe('DetailsViewCommandBar', () => {
...
+            getDateFromTimestamp: () => theDate,
         };
 
         return {
             deps,
             featureFlagStoreData,
             actionMessageCreator: actionMessageCreatorMock.object,
             tabStoreData,
-            renderExportAndStartOver,
+            reportExportComponentProps,
```
DaveTryon: This should go away

DaveTryon: Removed

src/DetailsView/details-view-body.tsx
```diff
@@ -72,7 +73,13 @@ export class DetailsViewBody extends React.Component<DetailsViewBodyProps> {
 
     private renderCommandBar(): JSX.Element {
         const { deps, switcherNavConfiguration } = this.props;
-        return <switcherNavConfiguration.CommandBar actionMessageCreator={deps.detailsViewActionMessageCreator} {...this.props} />;
+
+        const detailsViewCommandBarProps: DetailsViewCommandBarProps = {
```
waabid: I know this was there before but since you are touching these components, it might make sense to move the action message creator into deps inside the DetailsViewCommandBarProps.

src/tests/unit/tests/DetailsView/components/details-view-command-bar.test.tsx
```diff
@@ -98,31 +121,36 @@ describe('DetailsViewCommandBar', () => {
     });
 
     function testOnPivot(givenRenderExportAndStartOver: boolean): void {
-        renderExportAndStartOver = givenRenderExportAndStartOver;
-        const props = getProps();
+        const theHtml = 'this is the HTML';
+        let theDescription = null;
+        reportExportComponentProps = givenRenderExportAndStartOver
+            ? {
```
waabid: I would create a variable to represent the reportExportComponentProps object here; makes it easier to parse and understand that way.

src/tests/unit/tests/DetailsView/components/report-export-component-props-factory.test.ts
```diff
@@ -0,0 +1,165 @@
...
+                    tabStoreData,
+                    theDescription,
+                ),
+            )
+            .returns(() => theGeneratorOutput);
+    }
+
+    function setAutomatedChecksReportGenerator(): void {
+        reportGeneratorMock
+            .setup(x =>
```
waabid: I think we should have a name that identifies that the `x` here is a reportGenerator.

src/tests/unit/tests/DetailsView/components/details-view-command-bar.test.tsx
DaveTryon: Done

src/tests/unit/tests/DetailsView/components/report-export-component-props-factory.test.ts
DaveTryon: Done

src/DetailsView/details-view-body.tsx
DaveTryon: I agree with the idea here, but actionMessageCreator shows up in many, many, props, and this feels like a change thought we should consider holistically. I'll look into it as a separate PR, but it doesn't feel like it belongs in this one

PR #1634, opened 2019-11-07
feat: added wire-up for new actions in the cards UI

waabid:
#### Description of changes

A few notes: there seems to be a few missing tests for existing functions that I will not be adding for this PR in the action-creator (I did so for the message action creator). Similarly, there is missing telemetry for all this stuff and I will not be 'fixing' that in this PR either but it's a notable miss from previous cards UI work.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/extension-telemetry-events.ts
```diff
@@ -56,6 +56,7 @@ export const CONTENT_PAGE_OPENED: string = 'contentPageOpened';
 export const CONTENT_HYPERLINK_OPENED: string = 'contentHyperLinkOpened';
 export const CARD_SELECTION_TOGGLED: string = 'cardSelectionToggled';
 export const RULE_EXPANSION_TOGGLED: string = 'ruleExpansionToggled';
+export const CARDS_VISUAL_HELPER_TOGGLED: string = 'cardsVisualHelperToggled';
```
peterdur: Not used in this PR? (for future telemetry?)

waabid: Good catch: yes, not used here. I'll remove it and add it in a future PR.

Unfortunately it seems nothing is firing telemetry on the Cards UI due to some missing items in the message/action creator side of things, so I'll cover that in another PR.

waabid:
> Wahaj, the PR looks good (except for the telemetry value that isn't used). However,I'd like to see the missing unit test added. Also, will the missing telemetry be added as a task?

The telemetry value has been removed.

Regarding back-filling tests and adding telemetry: considering both are already in master and that not being the point of this PR, I'd like to move forward with this PR as is.

src/tests/unit/tests/background/actions/card-selection-action-creator.test.ts
```diff
@@ -38,6 +38,42 @@ describe('CardSelectionActionCreator', () => {
         );
     });
 
+    test('onToggleVisualHelper', () => {
+        const actionMock = createActionMock(null);
```
smoralesd: The name of the`const` seems very generic. I notice previously written tests use the name of the action in the mock name, e.g `toggleCardSelectionMock`. Should we do the same here to improve readability and consistency?

waabid: I'm not for it because I think it's not really a specific action's mock at the moment but consistency is a good reason so I'll update it 😄

PR #1641, opened 2019-11-07
feat(and-screenshot): make screenshot+highlights rescale dynamically to fit their pane

dbjorge:
#### Description of changes

Before this PR, the screenshot and highlights were rendered at actual size, using both horizontal and vertical scrollbars if that was bigger than the pane containing them, like this:

screenshot of screenshot not resizing with scrollbars

After this PR, the screenshot and highlights both resize dynamically to fit the pane containing them, like this:

screenshot of screenshot resizing to fit pane

Making the screenshot size itself dynamically was straightforward, just a bit of CSS (in particular, `object-fit: contain` to deal with maintaining the aspect ratio).

Making the highlight boxes respect the size was trickier. I considered using a CSS scale transform, but we didn't want the box outline width or the `!` sizes to change, just their positions, so that wasn't going to work. To achieve the desired effect, I switched the boxes from calculating positions as absolute px offsets relative to their container div to instead calculate relative % offsets relative to the container div. Since this required more a more interesting calculation involving knowledge of the viewport size, I decided to move the calculation out of the HighlightBox component and into the ViewModel, and to calculate it in the ViewModelProvider; updating the piping from ViewModelProvider -> HighlightBox accordingly is where most of the code change in the PR comes from.

#### Pull request checklist

- [x] Addresses an existing issue: 1640687 (partial)
- [x] Ran `yarn fastpass`


src/electron/views/screenshot/screenshot-container.tsx
```diff
@@ -2,16 +2,16 @@
...
 import { Screenshot } from 'electron/views/screenshot/screenshot';
 import { screenshotContainer } from 'electron/views/screenshot/screenshot-container.scss';
+import { HighlightBoxViewModel } from 'electron/views/screenshot/screenshot-view-model';
 import { isEmpty } from 'lodash';
 import * as React from 'react';
 
 export interface ScreenshotContainerProps {
     screenshotData?: ScreenshotData;
-    highlightBoxRectangles: BoundingRectangle[];
+    highlightBoxViewModels: HighlightBoxViewModel[];
```
smoralesd: does it make sense to make this new property optional the same as `screenshotData`? Alternative, does it make sense to make `screenshotData` mandatory to match the new one? I feel this 2 props are related in a way that you either have both or none, is this assumtion right?

dbjorge: I like making it mandatory. Moved the responsibility for laying out the unavailable message to the view to accommodate this.

src/electron/views/screenshot/screenshot-view.tsx
```diff
@@ -2,26 +2,35 @@
...
-                screenshotData={props.viewModel.screenshotData}
-                highlightBoxRectangles={props.viewModel.highlightBoxRectangles}
-            />
-            {renderDeviceNameCaption(props.viewModel.deviceName)}
+            {isUnavailable ? renderUnavailableMessage() : renderScreenshotContainer(props.viewModel)}
         </div>
     );
 });
 
-function renderDeviceNameCaption(deviceName?: string): JSX.Element {
```
smoralesd: I'm guessing this (removing the device name) was updated on the figma mock. Just checking this wasn't remove inadvertently.

dbjorge: Correct, apparently it confused several users in user testing into thinking that the thing labelled "Android Emulator" was going to be an actual interactive emulator, so the caption was removed to reduce confusion

PR #1645, opened 2019-11-08
feat(and-screenshot): hook up initial electron card selection support

dbjorge:
#### Description of changes

This PR hooks up the common/web collapsible control and the associated action/messages to electron so we can have basic card selection behavior working. The screenshot highlights already used the card selection data appropriately and needed no further changes to support the right highlighting.

screenshot of electron app with a card selected, filtering highlighting on the screenshot

(note: that screenshot was from these changes as-rebased onto some in-progress fit and finish work, which this PR does not include)

I updated the card footer to allow it to avoid rendering its action menu if in a highlighting-only support state, since supporting the copy failure details action would have required much more work than I wanted in one PR.

I removed the now-obsolete electron specific collapsible card component as part of this.

The common collapsible card component uses collapsed-by-default instead of expanded-by-default, so I also updated the e2e tests for the expand/collapse behavior accordingly. collapsed-by-default is the desired behavior per current mocks.

#### Pull request checklist

- [x] Addresses an existing issue: early start on 1582215
- [x] Ran `yarn fastpass`


src/electron/views/renderer-initializer.ts
```diff
@@ -164,16 +169,19 @@ getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch).then((persistedDat
 
     const fixInstructionProcessor = new FixInstructionProcessor();
 
+    const cardSelectionMessageCreator = new CardSelectionMessageCreator(dispatcher);
```
smoralesd: would it be better to put the action creator close the the action message creators as we have the user config action/message creators now? They are somehow related... maybe have them close togheter makes it easier to read/understand

PR #1654, opened 2019-11-08
feat(and-screenshot): fit and finish for automated checks view/screenshot view

dbjorge:
#### Description of changes

Assorted minor fit and finish changes to bring the screenshot view more in line with the mock for the feature.

Also included a few updates to the rest of the automated checks view to bring it in line with the mock, in cases where the existing inconsistencies clashed with the screenshot view updates (most notably, some header fonts/background color). The only one of these that affects non-electron is that the font/margin change on the "Failed instances" header will also effect web/report; I verified that the figma mocks for those scenarios are consistent in wanting this change.

Out of scope:
* highlight box styling
* complete mock adherence for components outside screenshot view

Before:

screenshot before changes in this PR

After:

screenshot with changes from this PR

Mock (reference):

screenshot from figma mockup of desired end state

#### Pull request checklist

- [x] Addresses an existing issue: 1640687
- [x] Ran `yarn fastpass`


src/electron/views/automated-checks/components/header-section.tsx
```diff
@@ -2,16 +2,16 @@
 // Licensed under the MIT License.
 import { NamedFC } from 'common/react/named-fc';
 import * as React from 'react';
-import { h1, text } from './header-section.scss';
+import * as styles from './header-section.scss';
```
waabid: Any reason we can't import the styles like before from this instead of importing *?

dbjorge: I've been trying out using `styles` as a consistent identifier for css classes after seeing it as a common pattern in a lot of the videos/blogs/etc I watched during day of learning; I think it makes it a little more readable during use for them to have a common prefix since they often have names that are otherwise easy to confuse with the classes/variables in use for the implementation of the class

waabid: Gotcha! I see what you mean; maybe bring this up in engineering discussion as well?

PR #1657, opened 2019-11-09
feat: add license to installers for linux, windows and macOS

AhmedAbdoOrtiga:
#### Description of changes
- Add license to android installers for:
  - **Windows:**
image

  - **Linux:** 
image

  - **macOS**
image

#### Pull request checklist

- [ ] Ran `yarn fastpass`


electron-builder.yml
```diff
@@ -1,17 +1,24 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
-appId: com.aimobile.app
+appId: com.aiandroid.app
 directories:
     app: drop/electron/electron
+    buildResources: src/electron/resources
```
Shobhit1: Curious if this cannot read the License file directly? the one that is already in our repo

karanbirsingh: @Shobhit1 text is different - the one in our repo refers to AI for Web

AhmedAbdoOrtiga: yeah, plus as mentioned here https://www.electron.build/configuration/dmg
you have to add the license in the buildResources folder and it should be picked up, and to do that we have to define the buildResources folder path

.prettierignore
```diff
@@ -18,10 +18,13 @@ AppXManifest.xml
...
+src/electron/resources/license_en.txt
+src/electron/resources/license.txt
 docs/NOTICE.html
 drop/
 extension/
 LICENSE
 new-appx.ps1
 test-results/
 yarn.lock
+LICENSE.txt
```
smoralesd: I see a `LICENSE.txt` under the `docs` folder but not directly on the root folder. Is this addition to `.prettierignore` meant to ignore a file we didn't add on this PR?

src/electron/resources/license.txt
```diff
@@ -0,0 +1,12 @@
+Accessibility Insights for Android
```
smoralesd: do we need both, `license.txt` and `license_en.txt`?

AhmedAbdoOrtiga: Yep, for mac it will use license_en.txt, for linux it uses license.txt, for windows either will work

.prettierignore
AhmedAbdoOrtiga: I should have deleted it, but I forgot

PR #1658, opened 2019-11-09
feat(consistent-cards-ui): added buttons for collapse/expand all and visual helper toggle

waabid:
#### Description of changes

I made the button container here something passed by deps: this allows us to make use a null pattern here (specifically use for the report scenario).

image

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/components/cards/cards-visualization-modifier-buttons.tsx
```diff
@@ -0,0 +1,53 @@
...
+    cardSelectionMessageCreator: CardSelectionMessageCreator;
+};
+
+export type CardsVisualizationModifierButtonsProps = {
+    deps: CardsVisualizationModifierButtonsDeps;
+    visualHelperEnabled: boolean;
+    allCardsCollapsed: boolean;
+};
+
+export const CardsVisualizationModifierButtons = NamedFC<CardsVisualizationModifierButtonsProps>(
```
smoralesd: not thrilled about the name of this component... is this something close to a command bar? The `VisualizationModifier` part of the name kind of doesn't match the expand/collapse button... thoughts?

src/common/components/cards/cards-visualization-modifier-buttons.tsx
```diff
@@ -0,0 +1,53 @@
...
+    'CardsVisualizationModifierButtons',
+    props => {
+        const { deps, visualHelperEnabled, allCardsCollapsed } = props;
+
+        let onClick: (_: SupportedMouseEvent) => void = deps.cardSelectionMessageCreator.collapseAllRules;
```
smoralesd: - is it necesary to wrap `cardSelectionMessageCreator.collapseAllRules` here? is it because of the event type doesn't match?
- the name here, `onClick` is not to informative, as we have to `onClick` props down below (in the `ActionButton` and the `Toggle`). I do see that the on click function changes base on the state. Still feels to generic for a name.

src/common/components/null-component.tsx
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { NamedFC } from 'common/react/named-fc';
+
+export const NullComponent = NamedFC<{}>('NullComponent', () => {
```
smoralesd: I don't think we need the `{}` on the props type template. Something like this should work just fine:
```typescript
export const NullComponent = NamedFC('NullComponent', () => null);
```

src/tests/unit/tests/common/components/cards/cards-visualization-modifiers-buttons.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { IMock, Mock } from 'typemoq';
+
+describe('CardsVisualizationModifierButtons', () => {
+    let cardSelectionMessageCreatorMock: IMock<CardSelectionMessageCreator>;
+
+    beforeEach(() => {
+        cardSelectionMessageCreatorMock = Mock.ofType<CardSelectionMessageCreator>();
+    });
+
+    test('with all cards collapsed', () => {
```
smoralesd: do we test the `onClick` handle for the `ActionButton` and the `Toggle` elsewhere?

src/common/components/cards/cards-visualization-modifier-buttons.tsx
waabid: Discussed offline: will leave as is.

src/common/components/cards/cards-visualization-modifier-buttons.tsx
waabid: Not necessary; updated.

src/tests/unit/tests/common/components/cards/cards-visualization-modifiers-buttons.test.tsx
waabid: Usually we just test whether something is being passed in ... I figured we could/should do the test here though, so I added them to verify the correct functions are being invoked.

PR #1666, opened 2019-11-12
fix: match cards title to mocks

waabid:
#### Description of changes

As per a comment in the previous PR, the text for the rule should not go under the outcome chip. This modifies the component to make that happen. Gif below of how it looks:

cardstitleresizewithminwidth

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/report-sections/minimal-rule-header.tsx
+++ b/src/reports/components/report-sections/minimal-rule-header.tsx
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 import { NamedFC } from 'common/react/named-fc';
 import * as React from 'react';
+import { outcomeChipContainer } from 'reports/components/report-sections/minimal-rule-header.scss';
 
 import { InstanceOutcomeType } from '../instance-outcome-type';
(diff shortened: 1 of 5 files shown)


src/reports/components/report-sections/minimal-rule-header.tsx
```diff
@@ -36,8 +37,9 @@ export const MinimalRuleHeader = NamedFC<MinimalRuleHeaderProps>('MinimalRuleHea
 
     return (
         <span className="rule-detail">
+            <span className={outcomeChipContainer}>{renderCountBadge()}</span>
             <span>
-                {renderCountBadge()} {renderRuleName()}: {renderDescription()}
```
DaveTryon: Should the renderCountBadge function also be deleted? This seems to be the only place where it's called

smoralesd: hey @DaveTryon, we're still using `renderCountBadge` we just move it to a different place inside the same component.

PR #1676, opened 2019-11-13
feat(consistent-cards-ui): Add Start over button to automated checks

DaveTryon:
#### Description of changes

Implement the __Start over__ button for FastPass with cards UI (works for both automated checks and tabstops). Assessment view still uses the StartOverDropdown; FastPass view uses an ActionButton. When clicked, the ActionButton uses a method on the 

Earlier code in this feature added the `StartOverComponentPropertyFactory` and `ReportExportComponentPropertyFactory` classes, which returned property objects, then relied upon other code to build the object. This PR changes that approach so the factories (now named `StartOverComponentFactory` and `ReportExportComponentFactory`) each return a JSX.Element object that can be consumed directly. This brings the property and creation code into a single location for each type. It also reduces overlap between tests

#### Pull request checklist

- [x] Addresses an existing issue: 1638935
- [x] Ran `yarn fastpass`
- [pending] (UI changes only) Added screenshots/GIFs to description above
- [pending] (UI changes only) Verified usability with NVDA/JAWS


src/DetailsView/components/report-export-component-factory.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const reportGenerator = deps.reportGenerator;
+
+    const reportExportComponentProps: ReportExportComponentProps = {
+        deps: deps,
+        scanDate: scanDate,
+        reportGenerator: reportGenerator,
+        pageTitle: tabStoreData.title,
+        exportResultsType: 'AutomatedChecks',
+        htmlGenerator: description =>
+            reportGenerator.generateFastPassAutomateChecksReport(
```
DaveTryon: The method being called here should arguably be renamed to `generateFastPassAutomatedChecksReport`, but that will be a separate PR

src/DetailsView/components/start-over-component-factory.tsx
```diff
@@ -0,0 +1,57 @@
...
+        return null;
+    }
+
+    const selectedTest = props.visualizationStoreData.selectedFastPassDetailsView;
+    const actionMessageCreator = props.actionMessageCreator;
+
+    return (
+        <ActionButton
+            iconProps={{ iconName: 'Refresh' }}
+            onClick={(event: SupportedMouseEvent) => onClickFastPass(event, actionMessageCreator, selectedTest)}
```
waabid: You can actually do something like this here:

```(event: SupportedMouseEvent) => actionMessageCreator.rescanVisualization(event, selectedTest)```

File: src/tests/unit/tests/DetailsView/components/start-over-component-factory.test.tsx [hunk 3]
```diff
@@ -100,12 +103,22 @@ describe('StartOverComponentPropsFactory', () => {
...
         const props = getProps(false);
         const rendered = getStartOverComponentForFastPass(props);
 
         expect(rendered).toMatchSnapshot();
     });
+
+    test('getStartOverComponentPropsForFastPass, CardsUI is true, scanResults is not null, scanning is true, component matches snapshot', () => {
+        setScanResult();
+        setCardsUiFlag(true);
+        scanning = 'true';
```
waabid: Just a heads up we don't usually set scanning to "true"; the string is usually a string identifier for a test.

waabid:
Just a heads up we don't usually set scanning to "true"; the string is usually a string identifier for a test.

So it alternates between being null (nothing scanning) and scanning (where it carries the value of what's being scanned).

PR #1679, opened 2019-11-13
feat: introduce permissions ensurer

smoralesd:
#### Description of changes

As part of the **Reduce permissions for Accessibility Insights for Web** feature, we need a way to handle permissions (check & request basically).

In this PR:
- update browser/chrome adapter to include `chrome.permissions` wrappers
- introduce the new permissions ensurer 

**Notes** The class is not yet being used but it has been tested in the [active-tab-permissions](https://github.com/smoralesd/accessibility-insights-web/tree/active-tab-permission) branch

#### Pull request checklist

- [x] Addresses an existing issue: part of WI # 1645760
- [x] Ran `yarn fastpass`


src/background/permissions/permissions-ensurer.ts
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
+
+export class PermissionsEnsurer {
```
waabid: Not sure how I feel about the name: why can't we go with PermissionGetter?

Ensuring we have permissions feels similar to getting permissions to me?

dbjorge: For me, `PermissionGetter` doesn't make me expect that it would be requesting permissions so much as returning which permissions we already have. I like `PermissionEnsurer` or `PermissionRequester`

dbjorge: or `BrowserPermissionsRequester` maybe

smoralesd: I wasn't sure about the name either. Considering what this class does:
- check permissions (by injecting code)
- request permissions if needed
- interpreting error codes from chrome API
All that make me thing that is way more than a getter. Also, consider we may end up using this right before injecting our code into the target page. 
I can go with requester as technically speaking, we can't ensure we have permissions, in the sense of, the user may as well denied permissions to our extension.
I'm open to renaming here.

smoralesd:
I just realize that I created this PR from the wrong branch. 
The conversion around `PermissionsEnsurer` class is still valid so I'll copy it over to the new PR

PR #1682, opened 2019-11-13
feat: introduce permissions ensurer

smoralesd:
#### Description of changes

As part of the **Reduce permissions for Accessibility Insights for Web** feature, we need a way to handle permissions (check & request basically).

In this PR:
- update browser/chrome adapter to include `chrome.permissions` wrappers
- introduce the new permissions ensurer 

**Notes** The class is not yet being used but it has been tested in the [active-tab-permissions](https://github.com/smoralesd/accessibility-insights-web/tree/active-tab-permission) branch

#### Pull request checklist

- [x] Addresses an existing issue: part of WI # 1645760
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/permissions/permissions-ensurer.ts
+++ b/src/background/permissions/permissions-ensurer.ts
@@ -3,6 +3,9 @@
 import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
 
 export class PermissionsEnsurer {
+    public static noPermissionsMessageEnding = 'Extension manifest must request permission to access this host.';
+    public static noUserGestureMessage = 'This function must be called during a user gesture';
+
     constructor(private readonly browserAdapter: BrowserAdapter) {}
 
     public async ensureInjectPermissions(tabId: number): Promise<boolean> {
@@ -31,7 +34,7 @@ export class PermissionsEnsurer {
 
             return true;
         } catch (error) {
-            if (error.message.endsWith('Extension manifest must request permission to access this host.')) {
+            if (error.message.endsWith(PermissionsEnsurer.noPermissionsMessageEnding)) {
                 return false;
             }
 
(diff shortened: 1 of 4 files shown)


src/background/permissions/permissions-ensurer.ts
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
+
+export class PermissionsEnsurer {
```
smoralesd: from @waabid Not sure how I feel about the name: why can't we go with PermissionGetter?

Ensuring we have permissions feels similar to getting permissions to me from a consumer perspective and ensurer seems like an uncommon term for this kind of situation?

src/background/permissions/permissions-ensurer.ts
```diff
@@ -0,0 +1,67 @@
...
+    public static noPermissionsMessageEnding = 'Extension manifest must request permission to access this host.';
```
dbjorge: Have you verified whether these messages get translated for users with non-english chrome language settings?

src/background/permissions/permissions-ensurer.ts
```diff
@@ -0,0 +1,67 @@
...
+    private requestPermissions(tabId: number): Promise<boolean> {
+        return new Promise((resolve, reject) => {
+            this.browserAdapter.getTab(tabId, async tab => {
+                const tabUrl = new URL(tab.url);
+                const permissionsToRequest = {
+                    origins: [`${tabUrl.origin}/`],
+                };
+
+                try {
+                    await this.browserAdapter.requestPermissions(permissionsToRequest).then(resolve);
```
dbjorge: Would there be any value to having this check `UrlValidator.isSupportedUrl` first?

src/background/permissions/permissions-ensurer.ts
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT License.
+import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
+
+export class PermissionsEnsurer {
+    public static noPermissionsMessageEnding = 'Extension manifest must request permission to access this host.';
+    public static noUserGestureMessage = 'This function must be called during a user gesture';
+
+    constructor(private readonly browserAdapter: BrowserAdapter) {}
+
+    public async ensureInjectPermissions(tabId: number): Promise<boolean> {
```
dbjorge: It feels a little off for the caller to have to understand 2 different error paths (resolve(false) vs reject(error)). Is there a use case we're anticipating where we need to respond differently to those 2 cases?

dbjorge: (if not, maybe `Promise<void>` would be more appropriate)

src/background/permissions/permissions-ensurer.ts
dbjorge: (from other thread): I'm not a fan of `PermissionGetter`, I don't think it's clear enough that it can be taking a user-visible action rather than just returning a bit of state. My favorite is `BrowserPermissionsRequester`, but I think `Ensurer` is also an okay noun (I think "ensure" is a reasonably common term for this sort of operation)

waabid: I get what you mean: I like requester!

src/background/permissions/permissions-ensurer.ts
smoralesd: maybe not... I think we are doing that before we get to the point we need to inject our scripts on the target page... I could be wrong though

Issue #1219, opened 2019-09-06
WCAG 2.1 1.4.12 Text spacing in the Switcher 


ferBonnin:
## Describe the bug

The switcher to change from fastpass to assessment has overlapping text

**To Reproduce**
1. Go to  text legibility> text spacing
2. run the following function (Same as the bookmarklet)
`(function () { document.querySelector("body").classList.add("text-spacing"); var style = document.createElement("style"), styleContent = document.createTextNode(".text-spacing * { line-height: 1.5 !important; letter-spacing: 0.12em !important; word-spacing: 0.16em !important; } .text-spacing p{ margin-bottom: 2em !important; } "); style.appendChild(styleContent); var caput = document.getElementsByTagName("head"); caput[0].appendChild(style); })();
`
3. check the switcher

## Expected behavior

all text is visible

## Screenshots

![Switcher with overlapping text](https://user-images.githubusercontent.com/9062104/64457949-ecf3e200-d0a8-11e9-9ed5-e999d033393a.png)

## Extension (please complete the following information)

- OS: [e.g. Windows/Mac] Windows
- Version [e.g. 1.140.1] 2.9.1
- Browser Version [e.g Chrome 71.0] 75
- Target Page [e.g www.bing.com] any

## Are you willing to submit a PR?

No
## Did you search for similar existing issues?

Yes

AhmedAbdoOrtiga:
I can't seem to repro, the only difference is my Chrome version is "Version 76.0.3809.132 (Official Build) (64-bit)", can you please advise?

AhmedAbdoOrtiga:
Worked with @peterdur, we were able to repro it by running the code in the DevTool console

peterdur:
Looks to be related to .ms-Dropdown-title having width: 139px;

varshannagarajan:
Hey, I was wondering if I could work on this issue.

ferBonnin:
Hi @varshannagarajan for sure! thanks for the help :)

varshannagarajan:
Hi, @ferBonnin I was trying out a few different changes and was wondering if the drop down box should remain the same size or if it was OK if it changed according to the length of the text?

Static:
![StaticDropdown](https://user-images.githubusercontent.com/48139461/67823211-8e3f5900-fa98-11e9-836a-c686affa117d.gif)

Dynamic:
[...]

PR #1693, opened 2019-11-14
fix: stopped switcher from moving outside the header

varshannagarajan:
#### Description of changes
Changed the header bar to be a flex box. This was done to make sure all components of the header bar were centered at all times. Fixes the switcher position problem as mentioned in issue #1219 . I wrapped the logo, header text and switcher div keeping the gear option outside.

No bookmarklet:
image

With bookmarklet without fix:
image

With bookmarklet and fix:
image

#### Pull request checklist

- [x] Addresses an existing issue: #1219 
- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/Styles/detailsview.scss
+++ b/src/DetailsView/Styles/detailsview.scss
@@ -234,10 +234,6 @@ div.insights-file-issue-details-dialog-container {
 #details-container {
     display: flex;
     flex-direction: column;
-
-    .details-view-body {
-        margin-top: $detailsViewHeaderBarHeight;
-    }
     .ms-Nav-compositeLink a {
         border-left: $pivotItemLeftBorderWidth $pivotItemBorderStyle $neutral-0;
         border-top: 0px;
@@ -308,8 +304,18 @@ div.insights-file-issue-details-dialog-container {
     }
     .header-bar {
         padding-left: 12px;
-        position: fixed;
+        position: relative;
         z-index: 1000;
+        display: flex;
+        justify-content: space-between;
+        min-width: 550px;
+        .header-text {
+            padding-top: unset;
+        }
+        .header-text-switcher {
+            display: flex;
+            align-items: center;
+        }
         .gear-options-button-component {
             float: right;
             margin-right: 16px;
--- a/src/DetailsView/components/header.tsx
+++ b/src/DetailsView/components/header.tsx
@@ -24,9 +24,11 @@ export class Header extends React.Component<HeaderProps> {
     public render(): JSX.Element {
         return (
             <header className="header-bar">
-                <HeaderIcon deps={this.props.deps} />
-                <div className="ms-font-m header-text">{title}</div>
-                {this.renderSwitcher()}
+                <div className="header-text-switcher">
+                    <HeaderIcon deps={this.props.deps} />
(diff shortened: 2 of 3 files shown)


varshannagarajan:
I don't believe I modified anything that would cause the exiting with error code 1.

peterdur:
@varshannagarajan Thanks for following up on your earlier PR with this fix! Looks like re-running the tests addressed the test issue you mentioned above. We want to take a little extra time to validate this fix across multiple scenarios; between that, the holiday week, and ongoing work on #1591, it may be a little while longer before we review and merge.

src/DetailsView/Styles/detailsview.scss
```diff
@@ -308,8 +304,18 @@ div.insights-file-issue-details-dialog-container {
     }
     .header-bar {
         padding-left: 12px;
-        position: fixed;
+        position: relative;
         z-index: 1000;
+        display: flex;
```
dbjorge: It would be good to remove the redundant `float:` attributes from the direct children, since the flex display will now be laying them out appropriately

src/DetailsView/Styles/detailsview.scss
```diff
@@ -308,8 +304,18 @@ div.insights-file-issue-details-dialog-container {
     }
     .header-bar {
         padding-left: 12px;
-        position: fixed;
+        position: relative;
         z-index: 1000;
```
dbjorge: The high z-index is probably unnecessary after this change

src/DetailsView/Styles/detailsview.scss
```diff
@@ -308,8 +304,18 @@ div.insights-file-issue-details-dialog-container {
...
+        display: flex;
+        justify-content: space-between;
+        min-width: 550px;
```
dbjorge: I'm not sure the min-width is necessary; even with it, the "Accessibility Insights for Web" text gets split over 2 lines when resizing the window down to its minimum width at 100% zoom, and using the min-width smaller than the 500px min-width of the window means that the settings button slides out of visibility at the minimum window width.

Without min-width, the text still splits unpleasantly, but the gear stays visible, which I think is more important. Maybe the best option would be to omit min-width and also set up the title text to use `text-overflow: ellipsis;`?

dbjorge: current production, before bookmarklet
screenshot in captioned state

with current iteration of change, before bookmarklet:
screenshot in captioned state

with current iteration of change, after bookmarklet:
screenshot in captioned state

this iteration + removing min-width, before bookmarklet:
screenshot in captioned state

this iteration + removing min-width, after bookmarklet:
screenshot in captioned state

dbjorge: On further investigation, this issue actually wasn't introduced by this PR; it was introduced by a separate change that's happened since the last production release.

peterdur:
@varshannagarajan, thanks again for sending us this PR! During our validation pass this week, we discovered some related issues around header bar behavior at small sizes. @dbjorge on our team, having reviewed your proposed changes earlier, was able to extend what you did to cover those issues as well in #1803. Since that PR includes the changes here, this PR is now obsolete. We'll make sure to include you in our contributor thanks in the next release!

PR #1697, opened 2019-11-15
Refactor: Move details view action message creator to deps

DaveTryon:
#### Description of changes
One of the PR comments of #1676 was that the `DetailsViewActionMessageCreator` object should be passed in as part of the deps, instead of as a property. Rather than making that change as a one-off, it seemed to make sense to approach that holistically. This PR updates _every_ class that uses `DetailsViewActionMessageCreator` (or that has a dependency that uses it) as follows:

1. Add a `detailsViewActionMessageCreator` member to the deps type (if no deps type existed, create a new one and add it to the props interface)
2. Remove the `DetailsViewActionMessageCreator` from the props interface
3. Update all places where the `DetailsViewActionMessageCreator` gets consumed
4. Update all places where the `DetailsViewActionMessageCreator` gets passed in (this is usually just passing along the deps object)
5. Standardize on `detailsViewActionMessageCreator` as the name in the code for `DetailsViewActionMessageCreator` objects
6. In cases where the `detailsViewActionMessageCreator` object initializes an object, use the streamlined syntax (as opposed to specifying it once as a property name and again as an obejct)

In addition, this also updates the unit tests as follows:
1. Standardize on the names of `detailsViewActionMessageCreatorMock` for mocks of `DetailViewActionMessageCreator` objects

Validated by running through various scenarios, found no problems.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/assessment-view.tsx
+++ b/src/DetailsView/components/assessment-view.tsx
@@ -127,14 +127,7 @@ export class AssessmentView extends React.Component<AssessmentViewProps> {
     }
 
     private renderTargetChangeDialog(): JSX.Element {
-        return (
-            <TargetChangeDialog
-                deps={this.props.deps}
-                prevTab={this.props.prevTarget}
-                newTab={this.props.currentTarget}
-                actionMessageCreator={this.props.deps.detailsViewActionMessageCreator}
-            />
-        );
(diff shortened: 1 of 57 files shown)


RobGallo:
This PR is too large for me to review effectively (in less than an hour anyway). However, I have looked at the first 10 files and they look good.

PR #1712, opened 2019-11-19
chore: change line width to 80 for src/popup

Shobhit1:
#### Description of changes
Change line width to 80 for `src/popup` folder.

#### Screenshot

image

#### Pull request checklist

- [x] Addresses an existing issue: #1713
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/popup/__snapshots__/incompatible-browser-renderer.test.tsx.snap
+++ b/src/tests/unit/tests/popup/__snapshots__/incompatible-browser-renderer.test.tsx.snap
@@ -20,13 +20,15 @@ exports[`IncompatibleBrowserRenderer renders 1`] = `
         >
           We don’t currently support your browser.
           <br />
-          Please check 
+          Please check
+           
           <NewTabLink
             href="https://accessibilityinsights.io/"
           >
             accessibilityinsights.io
           </NewTabLink>
-           for download options.
+           
+          for download options.
         </div>
       </div>
     </div>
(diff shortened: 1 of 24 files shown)


src/tests/unit/tests/popup/__snapshots__/incompatible-browser-renderer.test.tsx.snap
```diff
@@ -20,13 +20,15 @@ exports[`IncompatibleBrowserRenderer renders 1`] = `
         >
           We don’t currently support your browser.
           <br />
-          Please check 
+          Please check
+
```
smoralesd: this one looks odd... looks like we added new lines on the content itself... can you check (and maybe post a screenshot) of how this actually looks? I not is from the incompatible browser (i.e. old edge) so I'm not sure how critical this is

Shobhit1: Yeah. Need a windows machine for this. I was not able to test this on Mac. 😸

smoralesd: I think if you look at the actual file (not the test) you'll find `{' '}` being added as part of the reformat. If you remove them, the snapshot should not need to be change. Notice too that this is the only snapshot being change while the general expectation is no snapshot should be change (there maybe some edge cases where updating snapshot may be actually needed but I can't really imagine one scenario where this is true, also, I don't think this is one of those cases)

PR #1714, opened 2019-11-19
chore: move over macOS builds to windows-latest & ubuntu till the issue with macOS is fixed

Shobhit1:
#### Description of changes

Move over our builds to `windows` hosted agents from `macOS` till `macOS` connection issue is fixed. We already have a `sev-2` `IcM` with the respective team and will work with @JGibson2019 to see that through. But this will get us going on our daily work!

commit 08dada688bc5814cd9e070dc6aac5983404e6f2b is where i am updating snapshot for the node js version change.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -4,7 +4,7 @@ jobs:
     - job: 'unit_tests_and_lints'
 
       pool:
-          vmImage: 'macOS-10.14'
+          vmImage: 'macOS-latest'
 
       steps:
           - template: pipeline/install-node-prerequisites.yaml
@@ -66,7 +66,7 @@ jobs:
 
     - job: 'e2e_mac_web_electron'
       pool:
-          vmImage: macOS-10.14
+          vmImage: macOS-latest
       steps:
           - template: pipeline/install-node-prerequisites.yaml
           - template: pipeline/e2e-test-from-agent.yaml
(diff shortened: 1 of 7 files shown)


smoralesd:
I was thinking about the same... I remember one of the reason (maybe not the chief reason though) we use mac os was the case-sensitive nature of how mac os interprets/uses paths... I think linux works similar while windows ise case-insensitve... I'm wondering if it's more suitable for our purposes to use ubuntu for this change instead of mac.
Good approach!

Shobhit1:
> I was thinking about the same... I remember one of the reason (maybe not the chief reason though) we use mac os was the case-sensitive nature of how mac os interprets/uses paths... I think linux works similar while windows is case-insensitve... I'm wondering if it's more suitable for our purposes to use ubuntu for this change instead of mac.
> Good approach!

You are right and we should move some builds to ubuntu. I think a good approach here will be to move e2e to `windows` as they need less changes for them to work on windows and are more stable on that platform (mac is the most stable in that case).
But i will move test and build stuff on ubuntu and check whats the status for our builds after that.

Shobhit1:
Closing this PR as the issue with `macOS` is fixed now!

PR #1717, opened 2019-11-20
fix: replace all-origin permissions with activeTab permission

### LINKED ISSUE #1591: Canary hasn't updated since 10/24

dbjorge:
#### Description of changes

This purpose of this PR is to resolve #1591 by replacing our `manifest.json` permissions for `https://*/*", "http://*/*", "file://*/*` with the `activeTab` permission.

To achieve this, it updates the way we track creation and update events for tabs such that we now:
* treat opening the popup view (which gives our extension permissions for the currently-active tab for as long as it stays on the same origin) as "opening" that tab
* treat tab updates/navigation events that change the origin of a tab as having "closed" that tab

Example gif of the new interaction:

animation of changing the origin of a tab and having the details view treat it as being closed

This is a draft PR for reference as we test out this approach in Playground. It's not quite complete yet. In particular, the following elements aren't done:

* Thinking about whether we need to update any of the "tab closed" page text accordingly
* Renaming the "newTabCreated" tab action and associated machinery to better represent what it's now doing, probably "tabPermissionsGranted"
* Updating the chrome command handlers for the global keyboard shortcuts to also result in the same tabPermissionsGranted action

This PR only slightly changes the `target-page-controller` implementation, but its tests were a real challenge to follow originally, so the PR completely rewrites `target-page-controller.test.ts` (don't bother comparing them side-by-side, treat it as a new test file for the class).

#### Pull request checklist

- [x] Addresses an existing issue: #1591
- [x] Ran `yarn fastpass`


src/tests/unit/common/simulated-browser-adapter.ts
```diff
@@ -0,0 +1,80 @@
...
+    // These simulate real "update"/"activate" events (they update the windows/tabs state and send the notifications)
+    updateTab: (tabId: number, changeInfo: chrome.tabs.TabChangeInfo) => void;
+    activateTab: (tab: chrome.tabs.Tab) => void;
+};
+
+export function createSimulatedBrowserAdapter(tabs: chrome.tabs.Tab[], windows: chrome.windows.Window[]): SimulatedBrowserAdapter {
+    const mock: Partial<SimulatedBrowserAdapter> = Mock.ofType<BrowserAdapter>();
+    mock.tabs = [...tabs];
+    mock.windows = [...windows];
+    mock.setup(m => m.addListenerOnConnect(It.isAny())).callback(c => (mock.notifyOnConnect = c));
```
smoralesd: should we use `It.is(isFunction)` here instead (`isFunction` can be imported from `lodash`)

src/tests/unit/tests/background/target-page-controller.test.ts
```diff
@@ -2,748 +2,330 @@
...
+        mockDetailsViewController = setupMockDetailsViewController();
+        mockTabInterpreters = {};
+        mockTabInterpreters[EXISTING_ACTIVE_TAB_ID] = Mock.ofType<Interpreter>();
+        mockTabInterpreters[EXISTING_INACTIVE_TAB_ID] = Mock.ofType<Interpreter>();
+        mockTabInterpreters[NEW_TAB_ID] = Mock.ofType<Interpreter>();
+        mockTabContextFactory = setupMockTabContextFactory(mockTabInterpreters);
 
         testSubject = new TargetPageController(
-            tabInterpreterMap,
+            {},
```
smoralesd: should this be `mockTabInterpreters` ???

dbjorge: No (the intention is to track those separately to be able to test this class's ability to do its own tracking/mutation of the tabToContextMap), but I didn't notice the first time around that the mutations to the tabToContextMap are externally observable contracts of this class's behavior, so you're right that we should be passing a value for this and testing those mutations. Updated tests.

src/tests/end-to-end/tests/details-view/headings.test.ts
```diff
@@ -14,8 +14,9 @@ describe('Details View -> Assessment -> Headings', () => {
     let headingsPage: DetailsViewPage;
 
     beforeAll(async () => {
-        browser = await launchBrowser({ suppressFirstTimeDialog: true });
+        browser = await launchBrowser({ suppressFirstTimeDialog: true, addLocalhostToPermissions: true });
         targetPage = await browser.newTargetPage();
+        await browser.newPopupPage(targetPage); // Required for the details view to register as having permissions/being open
```
smoralesd: Did I forget to delete this after changing how we launch the browser (i.e. after adding the `addLocalhostToPermissions` flag)

src/tests/end-to-end/tests/details-view/overview.test.ts
```diff
@@ -15,6 +15,7 @@ describe('Details View -> Overview Page', () => {
     beforeAll(async () => {
         browser = await launchBrowser({ suppressFirstTimeDialog: true });
         targetPage = await browser.newTargetPage();
+        await browser.newPopupPage(targetPage); // Required for the details view to register as having permissions/being open
```
smoralesd: same here: Did I forget to delete this after changing how we launch the browser (i.e. after adding the `addLocalhostToPermissions` flag)

src/tests/end-to-end/tests/details-view/headings.test.ts
dbjorge: No, I added these afterwards; these are required *in addition to* the `addLocalhostToPermissions` flag (the flag ensures that Chrome gives us permission to inject, this line ensures that our extension understands that it has those permissions)

src/tests/end-to-end/tests/details-view/overview.test.ts
dbjorge: same as other one

PR #1730, opened 2019-11-21
test: update e2es to use the new insightsUserConfiguration controller for setting high-contrast mode

dbjorge:
#### Description of changes

This updates all the existing E2E tests to use the new insightsUserConfiguration controller in the background page to set high-contrast mode for accessibility tests, rather than using the details page UI to do it. This:

* Cuts 2-3min off e2e runtime
* Removes a bunch of operations that would otherwise be opportunities for flakiness
* Cooperates better with upcoming permissions changes that would otherwise prevent the use of the details view UI without popping the popup UI first

Mechanically, this allows for a bunch of simplification of the existing tests, since they no longer need to spawn separate browser instances/pages in a lot of cases; there was a pattern of having 2 top level describe blocks for "normal mode" vs "high contrast" mode that I've eliminated and replaced with updating a11y tests to just do `it.each` over the 2 HC states.

As part of this, I updated the guidance content tests to work more similar to every other test case, where we do a11y scans in both HC/non-HC modes but only snapshot the components in one of the 2 modes. This is where most of the "30k lines deleted" comes from.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -92,7 +98,16 @@ export class Page {
     }
 
     public async evaluate(fn: Puppeteer.EvaluateFn, ...args: any[]): Promise<any> {
-        return await this.screenshotOnError(async () => await this.underlyingPage.evaluate(fn, ...args));
+        const timeout = DEFAULT_PAGE_ELEMENT_WAIT_TIMEOUT_MS;
+        return await this.screenshotOnError(
```
smoralesd: is this `return await` redundant here? will `return this.screenshotOnError(` be sufficient?

src/tests/end-to-end/common/browser.ts
```diff
@@ -89,6 +89,11 @@ export class Browser {
         return page;
     }
 
+    public async gotoContentPage(existingPage: ContentPage, newContentPath: string): Promise<void> {
```
smoralesd: Not a high priority comment here: this seems like a static function... I'm wondering if that's desirable or not... or maybe we can move this to the `ContentPage` class (if it exist anyway)

dbjorge: It's not quite static because forming the URL depends on the background page to identify the extension URL origin

src/tests/end-to-end/common/page-controllers/page.ts
dbjorge: The return is required here because the function propogates the returned value of `fn`, similar to the original Puppeteer `evaluate` method (note the function return type is `Promise<any>`, not `Promise<void>`)

dbjorge: Sorry, I misinterpreted the original comment; I see now you were asking about removing the `await`, not the `return`.

I think https://github.com/standard/standard/issues/1442 has a very good overview of why I think it is better to include the await as a general rule

smoralesd: After readying the v8 article I agree!

PR #1731, opened 2019-11-21
Initial version of the Reporter. Not yet functional

markreay:
#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/reports/library/report-html-generator-initializer.ts
+++ b/src/reports/library/report-html-generator-initializer.ts
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { AutomatedChecksReportSectionFactory } from 'reports/components/report-sections/automated-checks-report-section-factory';
+import { getDefaultAddListenerForCollapsibleSection } from 'reports/components/report-sections/collapsible-script-provider';
+import { ReactStaticRenderer } from 'reports/react-static-renderer';
+import { ReportHtmlGenerator } from 'reports/report-html-generator';
+import { getPropertyConfiguration } from '../../common/configs/unified-result-property-configurations';
+import { DateProvider } from '../../common/date-provider';
+import { EnvironmentInfoProvider } from '../../common/environment-info-provider';
+import { initializeFabricIcons } from '../../common/fabric-icons';
+import { GetGuidanceTagsFromGuidanceLinks } from '../../common/get-guidance-tags-from-guidance-links';
+import { FixInstructionProcessor } from '../../injected/fix-instruction-processor';
+
+const browserVersion = "PLACEHOLDER_FOR_BROWSER_VERSION";
+const browserSpec = "PLACEHOLDER_FOR_BROWSER_SPEC";
+const axeVersion = "PLACEHOLDER_FOR_AXE_VERSION";
+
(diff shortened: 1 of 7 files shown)


src/tests/unit/tests/reports/library/report-html-generator-initializer.test.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { reportHtmlGeneratorInitializer } from "reports/library/report-html-generator-initializer";
+import { ReportHtmlGenerator } from "reports/report-html-generator";
+
+describe('ReportHtmlGeneratorInitializer', () => {
+
+    it('returns a valid generator', () => {
+        const initalizer = reportHtmlGeneratorInitializer();
```
RobGallo: Spelling: initalizer

src/reports/library/report-html-generator-initializer.ts
```diff
@@ -0,0 +1,38 @@
...
+import { EnvironmentInfoProvider } from '../../common/environment-info-provider';
+import { initializeFabricIcons } from '../../common/fabric-icons';
+import { GetGuidanceTagsFromGuidanceLinks } from '../../common/get-guidance-tags-from-guidance-links';
+import { FixInstructionProcessor } from '../../injected/fix-instruction-processor';
+
+const browserVersion = "PLACEHOLDER_FOR_BROWSER_VERSION";
+const browserSpec = "PLACEHOLDER_FOR_BROWSER_SPEC";
+const axeVersion = "PLACEHOLDER_FOR_AXE_VERSION";
+
+export const reportHtmlGeneratorInitializer = () => {
```
RobGallo: I think I would have grasped the purpose of this more quickly if it were named something like htmlReportGeneratorFactory or better still: createHtmlReportGenerator.
Also, the return type of the function isn't declared. I noticed there is a unit test to check the return type, but couldn't it be checked at compiletime by TypeScript?

src/tests/unit/tests/reports/library/report-html-generator-initializer.test.ts
```diff
@@ -0,0 +1,14 @@
...
+
+        expect(initalizer).toBeInstanceOf(ReportHtmlGenerator);
```
RobGallo: Is there an advantage to using toBeInstanceOf vs. notToBeNull?

src/tests/unit/tests/reports/library/report-html-generator-initializer.test.ts
markreay: Yes. Correct. Thanks.

src/reports/library/report-html-generator-initializer.ts
markreay: You are probably right. It started as a refactor from the DetailsViewInitializer so that's where I got the name. Now that we are here, it needs a better name. 

Thanks for the feedback.

src/tests/unit/tests/reports/library/report-html-generator-initializer.test.ts
markreay: It validates that it returns an object of the correct type. Just a little more rigorous.

PR #1734, opened 2019-11-22
refactor: automated checks report does not uses ScanResults

smoralesd:
#### Description of changes

Current `ReportHtmlGenerator.generateHtml` expect a `scanResult: ScanResults` and a `cardsViewData: CardsViewModel` parameters.
Both `scanResult` and `cardsViewData` have analogous information (rule results, passes, fails, etc) so technically speaking we don't need both.

In this PR I'm removing the old `scanResult` in favor on using the `cardsViewData` only.

Checked the report works properly with both the old details list and the new cards UI.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/report-sections/full-rule-header.tsx
+++ b/src/reports/components/report-sections/full-rule-header.tsx
@@ -19,12 +19,12 @@ export type FullRuleHeaderDeps = {
 
 export type FullRuleHeaderProps = {
     deps: FullRuleHeaderDeps;
-    cardResult: CardRuleResult;
+    cardRuleResult: CardRuleResult;
     outcomeType: InstanceOutcomeType;
 };
 
 export const FullRuleHeader = NamedFC<FullRuleHeaderProps>('FullRuleHeader', props => {
-    const { outcomeType, deps, cardResult } = props;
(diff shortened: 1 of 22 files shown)


src/reports/components/report-sections/full-rule-header.tsx
```diff
@@ -19,15 +19,15 @@ export type FullRuleHeaderDeps = {
 
 export type FullRuleHeaderProps = {
     deps: FullRuleHeaderDeps;
-    rule: RuleResult;
+    cardResult: CardRuleResult;
```
Shobhit1: Question: Why link this name with a particular UI implementation (i.e. Card in this case). Like why not pass a more generic name like `results` and then pass appropriate data when we initialize these variables.

smoralesd: The main reason is `CardRuleResults` is part of the `CardsViewModel`. Being a view model, it's couple to the view components that use it. i.e. this is not a domain model, hence I think it's fine the name of the prop reflect this. It also reflects the type name. I did change the name of the property to `cardRuleResult` though (here and on other files.

PR #1741, opened 2019-11-25
feat(report package) Implemented report npm package

markreay:
#### Description of changes

Creates an npm package to emit the Accessibility Insights automated checks report for a given axe-core scan result.

To build the npm package run `yarn build:package:report` and the package is in the directory `package/drop`.

#### Pull request checklist

- [x] Ran `yarn fastpass`


Gruntfile.js
```diff
@@ -310,6 +340,17 @@ module.exports = function(grunt) {
         console.log(`${targetName} extension is in ${dropExtensionPath}`);
     });
 
+    grunt.registerTask('package', function() {
```
smoralesd: `'package'` sounds too generic, should we include `'report'` in the name of this task (an related code)?

src/reports/package/accessibilityInsightsReport.d.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+declare namespace AccessibilityInsightsReport {
```
smoralesd: I think this is the first `declare namespace` usage on our repo... I'm wondering about the side-effects this have. I seem to remember we avoid using `namespace`s for some reason but I forgot. I'm mostly checking if this is absolutely necessary to introduce at this time.

src/reports/package/accessibilityInsightsReport.d.ts
```diff
@@ -0,0 +1,24 @@
...
+    export type Report = {
+        asHTML(): string;
+    };
+
+    export type ReportOptions = {
```
smoralesd: I'm not so sure 'Options` is a proper noun for this type. It's more data you want to add to the report, certainly it's not the results from axe but contextual data to make sense of the axe result themselves. I'm sorry I don't have a better suggestion for the name.

src/reports/package/reporter-factory.ts
```diff
@@ -0,0 +1,70 @@
...
+import { initializeFabricIcons } from '../../common/fabric-icons';
+import { GetGuidanceTagsFromGuidanceLinks } from '../../common/get-guidance-tags-from-guidance-links';
+import { FixInstructionProcessor } from '../../injected/fix-instruction-processor';
+import AccessibilityInsightsReport from './accessibilityInsightsReport';
+
+const axeResultsReportGenerator = (results: axe.AxeResults, options: AccessibilityInsightsReport.ReportOptions) => {
+    const { browserVersion, browserSpec, pageTitle: targetPageTitle } = options;
+    const axeVersion = results.testEngine.version;
+
+    initializeFabricIcons();
```
smoralesd: Is this the right place to call `initializeFabricIcons`? It seems to me that this call should only be needed when you actually generate the html report (basically `src/reports/package/axe-results-report.ts`). Since this call have side effects, maybe it's a good idea to move where said side effects are actually needed, right?

src/reports/package/root/package.json
```diff
@@ -0,0 +1,29 @@
+{
+    "name": "accessibility-insights-report",
+    "version": "0.0.1",
+    "description": "Accessibility Insights Report",
+    "license": "MIT",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/Microsoft/accessibility-insights-web"
+    },
+    "dependencies": {
```
smoralesd: I curious about this list of dependency, some of them strike me as not needed for the report, e.g `applicationinsights-js` (are we sending telemetry from the report or when building the report?), `idb-keyval` (are we accessing indexed db -from the browser)... I guess this maybe indirect dependencies for the report, if that's so, it sounds like some refactoring list for the near feature to clean/split up the report dependencies so we can reduce this list of dependencies.

src/tests/unit/tests/reports/package/reporter-factory.test.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
smoralesd: I'm wondering if we should have a `integration` folder for this test...

Gruntfile.js
markreay: Or maybe `reporter-package`? Or `report-package`?

src/reports/package/accessibilityInsightsReport.d.ts
markreay: This seems like how libraries typically share their types. The interesting thing about this file is that it is copied verbatim into the exported package.

src/reports/package/accessibilityInsightsReport.d.ts
markreay: That's fair. I confess I didn't put much energy into this. These are just the 4 data items that we need to build the report.

src/reports/package/reporter-factory.ts
markreay: Yeah. That's another head scratcher for me. You get a warning when you call it more than once to generate the report.
 I hate this global state kind of thing.

src/reports/package/root/package.json
markreay: We could trim these for sure.

src/tests/unit/tests/reports/package/reporter-factory.test.ts
markreay: That makes sense. I didn't want to add that burden to this initial PR but I would like to have several scan files with expected results.

Gruntfile.js
markreay: Probably so. How about reporter? Or maybe reporter-package? Or report-package?

smoralesd: `report-package` seems to cover purpose of the new package and let the door open to any other package we may want to introduce in the future.

markreay: I changed it to `package-report`. What do you think?

src/reports/package/reporter-factory.ts
markreay: I disabled the warnings so that is no longer an issue.

src/tests/unit/tests/reports/package/reporter-factory.test.ts
markreay: Discussed with @smoralesd and agreed that at a minimum these tests should be in a file named integration for now. Deferring on creating a new category of tests for now given that these tests are still very quick to run.

src/reports/package/accessibilityInsightsReport.d.ts
markreay: Any ideas for a better name here?

src/reports/package/accessibilityInsightsReport.d.ts
smoralesd: That sounds reasonable

markreay: I'm inclined to continue with this unless someone has a better idea about how to represent the external interface of this library.

src/reports/package/root/package.json
markreay: I pruned the list of dependencies to just the following:
```
        "axe-core": "3.3.2",
        "classnames": "^2.2.6",
        "lodash": "^4.17.15",
        "moment": "^2.24.0",
        "office-ui-fabric-react": "6.210.5",
        "react": "^16.12.0",
        "react-dom": "^16.12.0",
        "uuid": "^3.3.3"
```

peterdur:
> To build the npm package run `yarn build:package` and the package is in the directory `package/drop`.

Should this be `yarn build:package:report` ?

markreay:
> > To build the npm package run `yarn build:package` and the package is in the directory `package/drop`.
> 
> Should this be `yarn build:package:report` ?

It is actually. It was changed to exactly that.

src/reports/package/accessibilityInsightsReport.d.ts
smoralesd: This made me remember our `EnvironmentInfo` type.
There is also a `ScanDetailsReportModel` interface (the type is used on a `scanDetails` property somewhere else).
None of those types have exactly the properties you have here but they are kind of similar (from the usage point of view anyway).
Maybe those names/types are a good starting point?

src/reports/package/accessibilityInsightsReport.d.ts
smoralesd: I think I'm fine with this and we have some time before we actually publish this npm package to re-review this.

src/reports/package/accessibilityInsightsReport.d.ts
markreay: One point we may wish to consider is what would be a good name from the point of view of the caller. From the caller, something simple and straightforward would be preferred.

src/reports/package/accessibilityInsightsReport.d.ts
markreay: We don't have a lot of time before we need to publish. I was hoping to publish tomorrow. However, we will be going out as oh-dot initially and we will only have the ai-service as our initial consumer.

src/reports/package/accessibilityInsightsReport.d.ts
markreay: One possibility might be to consider all five data elements a report "definition" so it could be called a ReportDefinition, or ReportRequest.

peterdur: `ReportOptions` sounds like it would select different behavior during report generation, which isn't right. Two of the fields describe the browser in which the scan for results occurred, and two fields describe the page that was scanned. Maybe something like `ScanContext` or `ResultsEnvironment`?

smoralesd: I like `ScanContext`!!!

src/reports/package/accessibilityInsightsReport.d.ts
```diff
@@ -1,19 +1,27 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+
+import axe from 'axe-core';
```
pownkel: Could this be `import * as axe from 'axe-core'` instead? I'm still seeing some build issues from the services side.

markreay: let me see about that

markreay: Changed as requested.

PR #1768, opened 2019-11-27
feat(report-package) Customized report footer for service.

markreay:
#### Description of changes

Report footer has been modified with the following text:

> This automated checks result was generated using the Mark's Accessibility Service that helps find some of the most common accessibility issues. The scan was performed using axe-core 3.3.2 and MarkBrowser Version 0.0.0.0.0.0.1. For a complete WCAG 2.1 compliance assessment please visit http://aka.ms/AccessibilityInsights.

##### Screenshot
image

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/reports/package/reporter-factory.ts
```diff
@@ -40,8 +41,11 @@ const axeResultsReportGenerator = (parameters: AccessibilityInsightsReport.AxeRe
     const reactStaticRenderer = new ReactStaticRenderer();
     const fixInstructionProcessor = new FixInstructionProcessor();
 
+    const sectionFactory = { ...AutomatedChecksReportSectionFactory,
+        FooterSection: PackageReportFooter};
```
smoralesd: I like the idea of overriding section at the root composition level but this leads to duplication on the new footer component. Basically, `ReportFooter` (under `src\reports\components\report-sections\report-footer.tsx`) and the new `PackageReportFooter` are almost exactly the same. I put some more comments on `PackageReportFooter` directly

src/reports/package/package-report-footer.tsx
```diff
@@ -0,0 +1,25 @@
...
+import * as React from 'react';
+
+import { NamedFC } from 'common/react/named-fc';
+import { serviceTitle, toolName } from 'content/strings/application';
+import { SectionProps } from 'reports/components/report-sections/report-section-factory';
+import { ToolLink } from 'reports/components/report-sections/tool-link';
+
+export type ReportFooterProps = Pick<SectionProps, 'environmentInfo'>;
+
+export const PackageReportFooter = NamedFC<ReportFooterProps>('ReportFooter', ({ environmentInfo }) => {
```
smoralesd: This has a lot of duplication from `ReportFooter` (under `src\reports\components\report-sections\report-footer.tsx`). I'm wondering if it's better to change the props in a way that we can reuse the component but pass different props to customize, let's say, the tool name.

src/reports/package/package-report-footer.tsx
```diff
@@ -0,0 +1,25 @@
...
+    const { browserSpec, axeCoreVersion } = environmentInfo;
+    return (
+        <div className="report-footer-container">
+            <div className="report-footer" role="contentinfo">
+                This automated checks result was generated using the {serviceTitle}{' '}
```
smoralesd: so `serviceTitle` resolves to `Accessibility Insights Service` (by looking at `src/content/strings/application.ts`) but the package can be run outside the service, right? If that's true then we are not fully solving this issue (of having AI-Web name in the report when it was not generated that way)

devaradhanm: can we make the title as an option that will be passed to this package?

markreay: > Left some comments on specific files. I also notice on the screenshot we have **axe-core 3.3.2 and BROWSER_SPEC**... I'm guessing we are not passing browser spec values properly. I'm also wondering if we have missing unit tests around that.

That is the version of axe-core that created the test file and the magic string I passed into my example code that I generated the screenshot with.

src/reports/package/package-report-footer.tsx
markreay: I will look into that. I dealt with the ToolLink but there is other duplication that we can address.

src/reports/package/reporter-factory.ts
markreay: Refactored to remove duplication. Some textual duplication remains but there should be virtually no code duplication.

src/reports/package/package-report-footer.tsx
markreay: Added serviceName that can be passed in by the caller.

src/reports/components/report-sections/report-footer.tsx
```diff
@@ -2,28 +2,16 @@
 // Licensed under the MIT License.
 import * as React from 'react';
 
-import { NewTabLink } from 'common/components/new-tab-link';
 import { NamedFC } from 'common/react/named-fc';
-import { toolName } from 'content/strings/application';
 import { SectionProps } from './report-section-factory';
 
-export type ReportFooterProps = Pick<SectionProps, 'environmentInfo'>;
+export type ReportFooterProps = Pick<SectionProps, 'environmentInfo'> & Pick<SectionProps, 'footerText'>;
```
smoralesd: I think the syntax of `Pick` let you do this:
```typescript
Pick<SectionProps, 'environmentInfo' | 'footerText'>
```
I think this is less verbose

src/reports/package/package-report-footer.tsx
markreay: Refactored the duplication away.

src/reports/components/report-sections/tool-link.tsx
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as React from 'react';
+
+import { NewTabLink } from 'common/components/new-tab-link';
+import { NamedFC } from 'common/react/named-fc';
+import { toolName } from 'content/strings/application';
+
+export const ToolLink = NamedFC('ToolLink', () => (
+    <NewTabLink
```
devaradhanm: will this be shown for a11y insights service report as well? will the text be :

Get more information and download Accessibility Insights Service?
(we can't download the service)

devaradhanm: never mind. tool name will always be accessibility insights web.

src/tests/unit/tests/reports/package/__snapshots__/integration.test.ts.snap
```diff
@@ -1,5 +1,5 @@
...
...95 5.00381C4.65948 5.35528 4.65948 5.92513 5.01095 6.2766L6.72443 7.99007L4.9837 9.7308C4.63222 10.0823 4.63222 10.6521 4.9837 11.0036C5.33517 11.3551 5.90502 11.3551 6.25649 11.0036L7.99722 9.26287L9.73816 11.0038C10.0896 11.3553 10.6595 11.3553 11.011 11.0038C11.3624 10.6523 11.3624 10.0825 11.011 9.73101L9.27001 7.99007L10.9837 6.27639Z\\" fill=\\"white\\"></path></svg></span></span><span class=\\"count\\" aria-hidden=\\"true\\"> 1</span><span class=\\"screen-reader-only\\">1 Failed</span></span></span></span><span><span class=\\"rule-details-id\\">page-has-heading-one</span>: <span class=\\"rule-details-description\\">Ensure that the page, or at least one of its frames contains a level-one heading</span></span></span></button></div><div id=\\"content-container-page-has-heading-one\\" class=\\"collapsible-content\\" aria-hidden=\\"true\\"><div class=\\"ruleMoreResources\\"><div class=\\"moreResourcesTitle\\">Resources for this rule</div><span class=\\"ruleDetailsId\\"><a target=\\"_blank\\" href=\\"https://dequeuniversity.com/rules/axe/3.3/page-has-heading-one?application=webdriverjs\\" class=\\"ms-Link insights-link root-49\\">More information about page-has-heading-one</a></sp
```
devaradhanm: is there a way to pretty print the html so that it is easier to find out what changed?

src/reports/components/report-sections/report-footer-text.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export type ReportFooterTextProps = EnvironmentInfo;
+
+export type ReportFooterText = React.FC<EnvironmentInfo>;
+
+export const DefaultFooterText: ReportFooterText = NamedFC<ReportFooterTextProps>(
+    'DefaultFooterText',
+    ({ axeCoreVersion, browserSpec, extensionVersion }) => (
+        <>
+            This automated checks result was generated using {`${toolName} ${extensionVersion} (Axe ${axeCoreVersion})`}, a tool that helps
```
devaradhanm: for service report this will say, that the report was generated using "Accessibility Insights for Web", since toolName here is not configurable

src/reports/components/report-sections/report-footer.tsx
markreay: Yes. Will change that.

src/tests/unit/tests/reports/package/__snapshots__/integration.test.ts.snap
markreay: I was wondering the same thing.

src/reports/components/report-sections/report-section-factory.tsx
```diff
@@ -29,6 +30,7 @@ export type SectionProps = {
     userConfigurationStoreData: UserConfigurationStoreData;
     targetAppInfo: TargetAppData;
     shouldAlertFailuresCount?: boolean;
+    footerText: ReportFooterText;
```
smoralesd: This feels odd... why not go through `ReportSectionFactory.FooterSection` directly? you can wrap both the "default" behavior (as in, the current AI-Web behavior) and the package behavior in a common component and change the value of `FooterSection` when the instance of `ReportSectionFactory` is created, which is a different place in each case.

src/reports/components/report-sections/report-footer-text.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { EnvironmentInfo } from 'common/environment-info-provider';
+import { NamedFC } from 'common/react/named-fc';
+import { toolName } from 'content/strings/application';
+import { ToolLink } from 'reports/components/report-sections/tool-link';
+
+export type ReportFooterTextProps = EnvironmentInfo;
+
+export type ReportFooterText = React.FC<EnvironmentInfo>;
+
+export const DefaultFooterText: ReportFooterText = NamedFC<ReportFooterTextProps>(
```
devaradhanm: where is DefaultFooterText used? I dont see this usage in this pr

src/reports/components/report-sections/report-footer.tsx
```diff
@@ -2,28 +2,16 @@
...
 
-export const ReportFooter = NamedFC<ReportFooterProps>('ReportFooter', ({ environmentInfo }) => {
-    const { extensionVersion, browserSpec, axeCoreVersion } = environmentInfo;
+export const ReportFooter = NamedFC<ReportFooterProps>('ReportFooter', ({ environmentInfo, footerText: Text }) => {
```
devaradhanm: footerText property is not used

src/reports/components/report-sections/report-footer-text.tsx
markreay: src/DetailsView/details-view-initializer.ts: 231

src/reports/components/report-sections/report-section-factory.tsx
markreay: Yeah. Maybe.

src/reports/components/report-sections/report-footer.tsx
markreay: It's used, it's aliased to Text - It has to be uppercase to render.

src/reports/components/report-sections/report-section-factory.tsx
markreay: I was trying to avoid shimming it like that.

markreay: Ok, I have an idea that might be more open-closed. This is taking forever, but I'm having fun.

src/reports/components/report-sections/report-footer.tsx
devaradhanm: nice. i thought it was the type of footerText.

src/reports/components/report-sections/report-section-factory.tsx
markreay: Ok, I did the shim or decorator approach and I think it's an improvement. @smoralesd -- thoughts?

PR #1769, opened 2019-11-27
feat(permissions): update target page close content

smoralesd:
#### Description of changes

As part of the browser permissions changes, we are re-using the `TargetPageClosedView` component to handle the case where the user changes the target page url. The message on that component only mention the case where the user closes the target page tab and there is nothing about url changes and what to do.

This is a proposal for changing the content/instructions we show the user when target page is close or it's url has change.

image

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/no-content-available.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { targetPageClosed } from './target-page-closed-view.scss';
+
+export const NoContentAvailable = NamedFC('NoContentAvailable', () => {
+    return (
+        <main className={targetPageClosed}>
+            <h1>No content available</h1>
+            <ul>
+                <li>If the target page was closed, you can close this tab or reuse it for something else.</li>
+                <li>
+                    If the URL was change on the target page, you can reactivate this tab by pressing{' '}
```
peterdur:
```suggestion
                    If the URL was changed on the target page, you can reactivate this tab by pressing{' '}
```
I think "change" should be "changed" here.

peterdur:
```suggestion
                    If the URL was change on the target page, you can reactivate this tab by pressing the{' '}
```
I think we should have a "the" before the product name (...pressing **the** [AI-Web] extension icon...)...)

src/DetailsView/components/no-content-available.tsx
```diff
@@ -0,0 +1,23 @@
...
+                    <Markup.Term>{productName}</Markup.Term> extension icon on the browser toolbar or by using the extension shortcuts.
```
peterdur:
```suggestion
                    <Markup.Term>{productName}</Markup.Term> extension icon on the browser toolbar or by using the extension keyboard shortcuts.
```
May be clearer and more consistent with language elsewhere if we call them **keyboard** shortcuts?

src/DetailsView/components/no-content-available.tsx
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as Markup from 'assessments/markup';
+import { NamedFC } from 'common/react/named-fc';
+import { productName } from 'content/strings/application';
+import * as React from 'react';
+
+import { targetPageClosed } from './target-page-closed-view.scss';
```
peterdur: Should we update the CSS class name to match the component name?
```suggestion
import { noContentAvailable } from './no-content-available-view.scss';
```

src/DetailsView/components/no-content-available.tsx
smoralesd: updated

smoralesd: updated

src/DetailsView/components/no-content-available.tsx
smoralesd: updated

src/DetailsView/components/no-content-available.tsx
smoralesd: Good catch! Updated.

PR #1777, opened 2019-12-03
refactor: update all lodash imports

smoralesd:
#### Description of changes

Update all usages of `import * as _ from 'lodash'` to explicitly import what we are actually using from lodash. This should impact the bundle size in a positive way as the former style ends up including all of `lodash` code, while the latter helps webpack with the tree shaking. I do not have numbers for this though.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/assessment-report-model-builder.ts
+++ b/src/reports/assessment-report-model-builder.ts
@@ -6,7 +6,8 @@ import { Assessment } from 'assessments/types/iassessment';
 import { ManualTestStatus } from 'common/types/manual-test-status';
 import { AssessmentData, AssessmentStoreData, TestStepInstance } from 'common/types/store-data/assessment-result-data';
 import { TabStoreData } from 'common/types/store-data/tab-store-data';
-import * as _ from 'lodash';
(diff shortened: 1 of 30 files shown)


src/reports/assessment-report-model-builder.ts
```diff
@@ -6,7 +6,8 @@ import { Assessment } from 'assessments/types/iassessment';
 import { ManualTestStatus } from 'common/types/manual-test-status';
 import { AssessmentData, AssessmentStoreData, TestStepInstance } from 'common/types/store-data/assessment-result-data';
 import { TabStoreData } from 'common/types/store-data/tab-store-data';
-import * as _ from 'lodash';
+import { keys } from 'lodash';
+import { find } from 'office-ui-fabric-react';
```
lisli1: I'm curious why the change here uses `find` from `office-ui-fabric-react` instead of `lodash`?

smoralesd: Good catch! Thank for finding this.

PR #1783, opened 2019-12-03
refactor: move fix-instructions-panel to common folder from injected and move to css modules

Shobhit1:
#### Description of changes
This PR cleans out some of the stuff from `FixInstructionsPanel`. This achieves 2 things:

1. move component `FixInstructionsPanel` to common, which is where it belongs since its being used by `fastpass`, and `detailsDialog` both. `Injected` didn't seem like a good place to put this component.
2. move the css to css module to avoid having `injected.css` in our electron bundle due to classes being referenced from `injected.css`.

Next PR/Step: 
Try to find exact classes or components that are referencing `injected.scss` in electron and move them to css modules.

#### Pull request checklist

- [x] Ran `yarn fastpass`

#### Screenshots

Screen Shot 2019-12-03 at 2 43 36 PM

### DIFF
--- a/src/common/components/fix-instruction-panel.scss
+++ b/src/common/components/fix-instruction-panel.scss
@@ -1,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../styles/colors.scss';
+
+.insights-fix-instruction-list {
+    li {
+        span.fix-instruction-color-box {
+            width: 14px;
+            height: 14px;
+            display: inline-block;
+            vertical-align: -5%;
+            margin: 0px 2px;
+            border: 1px solid $always-black;
+        }
+    }
+
+    .screen-reader-only {
+        position: absolute;
+        left: -10000px;
+        top: auto;
+        width: 1px;
+        height: 1px;
+        overflow: hidden;
+    }
+}
--- a/src/common/components/fix-instruction-panel.tsx
+++ b/src/common/components/fix-instruction-panel.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { FixInstructionProcessor } from 'common/components/fix-instruction-processor';
 import { CheckType } from 'injected/components/details-dialog';
-import { FixInstructionProcessor } from 'injected/fix-instruction-processor';
 import * as React from 'react';
 import * as styles from './fix-instruction-panel.scss';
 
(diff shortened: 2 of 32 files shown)


src/common/components/fix-instruction-panel.scss
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../styles/colors.scss';
+
+.insights-fix-instruction-list {
```
dbjorge: With this in place, can the corresponding styles be removed from the higher-level scss files where it's repeatedly defined? (`how-to-fix-card-row.scss`, `detailsview.scss`, `injected.scss`, `instance-details.scss`)

src/common/components/fix-instruction-panel.tsx
```diff
@@ -47,8 +48,8 @@ export const FixInstructionPanel = NamedFC<FixInstructionPanelProps>('FixInstruc
 
     return (
         <div>
-            {props.renderTitleElement(title, 'insights-fix-instruction-title')}
-            <ul className="insights-fix-instruction-list">{renderInstructions(props.checkType)}</ul>
+            {props.renderTitleElement(title, null)}
```
dbjorge: It looks like this is the only spot where the function in question is invoked; should the second parameter be removed?

src/tests/unit/tests/common/components/cards/how-to-fix-card-row.test.tsx
```diff
@@ -5,7 +5,7 @@ import { shallow } from 'enzyme';
 import * as React from 'react';
 import { Mock } from 'typemoq';
 
-import { FixInstructionProcessor } from '../../../../../../injected/fix-instruction-processor';
+import { FixInstructionProcessor } from '../../../../../../common/components/fix-instruction-processor';
```
dbjorge: baseUrl-style?

src/common/components/fix-instruction-panel.tsx
Shobhit1: The function definition doesn't have this as an optional param and everywhere this is used, its being passed a classname, so didn't touch the definition.

src/common/components/fix-instruction-panel.scss
Shobhit1: Yes. Will Do.

Shobhit1:
> inline

Updated.

src/common/components/fix-instruction-panel.tsx
dbjorge: I'm saying that I believe this callsite is the only non-test callsite where a classname was ever passed to any concrete implementation, so I think it would make sense to remove the classname param from the implementations now that it's obsolete

Shobhit1: Done.

PR #1786, opened 2019-12-03
refactor: Simplified report footer text logic

markreay:
#### Description of changes

Refactored report footer text to simplify the report footer to use a FooterText element instead of an property that can be overridden.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/reports/components/report-sections/report-footer.tsx
```diff
@@ -2,35 +2,16 @@
...
-    ),
-);
-
-export type ReportFooterProps = {
-    footerText?: ReportFooterText;
-} & Pick<SectionProps, 'environmentInfo'>;
-
-export const ReportFooter = NamedFC<ReportFooterProps>('ReportFooter', ({ environmentInfo, footerText }) => {
-    const Text = footerText || DefaultFooterText;
+export const ReportFooter = NamedFC<ReportFooterProps>('ReportFooter', ({ environmentInfo, children }) => {
```
smoralesd: looks like we don't need `environmentInfo` prop anymore, right?

src/reports/package/footer-text-for-service.tsx
```diff
@@ -2,19 +2,18 @@
...
+    const footerText = ({ environmentInfo: { axeCoreVersion, browserSpec } }) =>
         <>
             This automated checks result was generated using the {serviceName}{' '}
             that helps find some of the most common accessibility issues. The scan was
             performed using axe-core {axeCoreVersion} and {browserSpec}. For a complete
             WCAG 2.1 compliance assessment please visit{' '}<ToolLink />.
         </>;
 
-    return ((props: ReportFooterProps) => reportFooter({ ...props, footerText })) as ReactFCWithDisplayName<ReportFooterProps>;
+    return footerText as typeof FooterText;
```
smoralesd: I think this type should be couple to the `ReportSectionFactory` (which can be considered abstract in this context) instead of the `FooterText` (which can be considered an implementation in this context).

src/reports/components/report-sections/report-body.tsx
```diff
@@ -38,7 +39,9 @@ export const ReportBody = NamedFC<ReportBodyProps>('ReportBody', props => {
                     <NotApplicableChecksSection {...sectionProps} />
                 </ResultsContainer>
             </ContentContainer>
-            <FooterSection {...sectionProps} />
+            <FooterSection {...sectionProps}>
```
smoralesd: Since we're not using `environmentInfo` on the `FooterSection` anymore, I'm wondering if we can avoid passing `sectionProps` here.

src/reports/components/report-sections/report-footer.tsx
markreay: Yes

src/reports/components/report-sections/report-body.tsx
markreay: Yes

PR #1794, opened 2019-12-04
feat(electron-update): call autoupdate after mainwindow is loaded

Shobhit1:
#### Description of changes

This PR fixes a small issue where `autoUpdate` was not getting called even once when the window was loaded and this would have created a situation where the autoupdate is just called after the user has spent an hour on the app, which is not what we want.

After this change, the check for update will be called once when the main window is loaded and that function also sets up an 1 hour recurring check

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -13,7 +13,6 @@ const electronAutoUpdateCheck = new AutoUpdaterClient(autoUpdater);
 let recurringUpdateCheck;
 
 const createWindow = () => {
-    recurringUpdateCheck = setupRecurringUpdateCheck();
     const os = platformInfo.getOs();
     mainWindow = new BrowserWindow({
         show: false,
@@ -41,6 +40,10 @@ const createWindow = () => {
         mainWindow.show();
         enableDevMode(mainWindow);
     });
+
+    checkForUpdates()
+        .then(() => console.log('checked for updates'))
+        .catch(console.log);
 };
 
 const enableDevMode = (window: BrowserWindow) => {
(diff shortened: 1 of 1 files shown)


smoralesd:
I saw another feat tag recently: (electron-update) is this the same feature using a different tag? Consistency on tagging PR's help whoever had to prepare release notes and the test plan.

Shobhit1:
> I saw another feat tag recently: (electron-update) is this the same feature using a different tag? Consistency on tagging PR's help whoever had to prepare release notes and the test plan.

Updated

src/electron/main/main.ts
```diff
@@ -41,6 +40,14 @@ const createWindow = () => {
...
         enableDevMode(mainWindow);
     });
+
+    electronAutoUpdateCheck
+        .check()
+        .then(() => {
+            console.log('checked for updates');
+            setupRecurringUpdateCheck();
+        })
+        .catch(console.log);
```
smoralesd: should this use `console.error`?

src/electron/main/main.ts
```diff
@@ -41,6 +40,14 @@ const createWindow = () => {
         mainWindow.show();
         enableDevMode(mainWindow);
     });
+
+    electronAutoUpdateCheck
+        .check()
+        .then(() => {
+            console.log('checked for updates');
```
smoralesd: is this `console.log` needed?

Shobhit1: This wont be logged anywhere on the console, so doesn't hurt and its easier to check while developing! So, yes I am inclined to keep this

src/electron/main/main.ts
Shobhit1: Thats how it was for the `loadUrl` catch, so following same patterns!

PR #1797, opened 2019-12-04
chore: better memory management with electron app

Shobhit1:
#### Description of changes
Just adding some more event handlers on window closing and trying to make sure that app quits completely. Also dereferencing `mainWindow` for better window management.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -41,11 +41,8 @@ const createWindow = () => {
         enableDevMode(mainWindow);
     });
 
-    // Emitted when the window is closed.
     mainWindow.on('closed', () => {
-        // Dereference the window object, usually you would store windows
-        // in an array if your app supports multi windows, this is the time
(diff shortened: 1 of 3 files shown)


karanbirsingh:
some context that helped me from Peter / Shobhit

https://electronjs.org/docs/api/browser-window#event-closed

> Emitted when the window is closed. After you have received this event you should remove the reference to the window and avoid using it any more.

https://electronjs.org/docs/api/app#event-window-all-closed

> Emitted when all windows have been closed.
> If you do not subscribe to this event and all windows are closed, the default behavior is to quit the app; however, if you subscribe, you control whether the app quits or not. If the user pressed Cmd + Q, or the developer called app.quit(), Electron will first try to close all the windows and then emit the will-quit event, and in this case the window-all-closed event would not be emitted.

PR #1803, opened 2019-12-05
fix: avoid header bar components overflowing out of the header bar

dbjorge:
#### Description of changes

This PR fixes an issue caught during v2.11.0 release validation introduced in #1606 where if you resize the details view down to the minimum window size allowed by Chrome (500px) at default zoom levels, the settings gear icon overflows out of the header bar:

screenshot with missing settings icon

@varshannagarajan 's outstanding PR #1693 fixes a variant of this issue for the switcher dialog at high zoom/text spacing levels. This PR takes the same idea as #1693 but expands it to the entire header bar (it therefore obsoletes #1693)
* It runs the *entire* header bar into a flexbox
* It enables text-overflow behavior on the extension title, so it will shrink before functional elements do
* It adjusts styling slightly such that everything fits at default zoom level + Chrome's default minimum window size (500px)

Since this essentially amounted to a rewrite of the relevant styling, I took it as an opportunity to pull it out of the big shared scss files and into modules. This involved a little gymnastics for some of the office fabric style overrides that previously relied on being within a `#details-container` selector to achieve the necessary specificity.

To enable re-use of the new modularized styles in guidance pages, this also extracts out the common parts of the header to a new common Header component for the guidance content page to reuse. This happened to be the first point where the guidance content pages required css module styles, so I also added the necessary reference from `insights.html` to `bundle/insights.css`.

The most questionable bit of CSS is the use of `margin-left: auto` on the settings icon styling to get it to functionally float right inside the flexbox; I thought using this was a little more pleasant to reason about than the alternative of using nested flexboxes in the same direction, but was a bit on the fence about which technique to use.

Screenshots:

**details view, 500px 100% zoom**
image

**details view, 500px 100% zoom, text-spacing bookmarklet**
image

**details view, 1280px, 100% zoom**
image

**details view, 1280px 400% zoom**
image

**fastpass report, 500px 100% zoom**
image

**assessment report, 500px 100% zoom**
image

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/interactive-header.tsx
```diff
@@ -2,33 +2,30 @@
...
+export interface InteractiveHeaderProps {
+    deps: InteractiveHeaderDeps;
     dropdownClickHandler: DropdownClickHandler;
     featureFlagStoreData: FeatureFlagStoreData;
     tabClosed: boolean;
     selectedPivot: DetailsViewPivotType;
 }
 
-export class Header extends React.Component<HeaderProps> {
+export class InteractiveHeader extends React.Component<InteractiveHeaderProps> {
```
smoralesd: is this a good opportunity to convert this component to use `NamedFC`???

src/DetailsView/components/switcher.scss
```diff
@@ -0,0 +1,47 @@
...
+        :global(.ms-Dropdown-caretDownWrapper) {
+            top: 4px;
+        }
+        :global(.ms-Dropdown-caretDown) {
+            color: $neutral-0;
+        }
+    }
+
+    &:focus {
+        .ms-Dropdown-title {
```
smoralesd: does `.ms-Dropdown-title` here needs to be wrapped into `global` just like in line 13?

dbjorge: Yes, good catch

src/DetailsView/components/interactive-header.tsx
dbjorge: Good idea, much cleaner this way

src/common/components/header.scss
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../common/styles/colors.scss';
+@import '../../common/styles/common.scss';
+@import '../../common/styles/fonts.scss';
+
+@mixin ellipsedText {
```
waabid: Any chance this could be in common instead?

waabid: As in, this looks fairly generic and I think it could go in common but I'm not sure if we have a "common" standard for this already (that is not equivalent).

src/reports/components/report-sections/header-section.scss
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 @import '../../../common/styles/colors.scss';
+@import '../../../common/styles/common.scss';
 @import '../../../common/styles/fonts.scss';
 
 @mixin ellipsedText {
```
waabid: If moved to common, we can remove this.

PR #1806, opened 2019-12-05
refactor: store proxy takes in tabId via constructor

waabid:
#### Description of changes

As per my previous PR, I feel like this PR will reduce future issues arising when creating new stores and using them in details/pop-up/etc views.

#### Pull request checklist

- [x] Ran `yarn fastpass`


waabid:
> This is great! I like we are updating the `constructor` to more clearly represent the preconditions for the `StoreProxy` to work, i.e, you do need the `tabId`.
> 
> I notice something while going through the changes though; now, every time we instanciate `StoreProxy` the same last 2 values appear on each line we do so: `browserAdapter, tab.id`. Does it make sense to extract those parameter (maybe in a binded-function, maybe in a store proxy provider) and in that way we enforce the tab id even more?

I thought about that as well, but figured that could be done in a future re-factor PR.

PR #1811, opened 2019-12-06
chore: make CSS module class names consistent between dev and prod

dbjorge:
#### Description of changes

This is a continuation of @waabid 's #1570. I've addressed the feedback that was there and fixed the e2e tests.

The most important bit of change is in `webpack.config.js`. Most of the other changes are either:
* Fixing pre-existing cases where we re-used class names across different modules. In particular, automated-checks-view and device-connect-view and the assorted cards report components did a lot of this between each other.
* Replacing e2e tests' use of classnames with `data-automation-id` attributes (a few web e2es and almost all electron e2es)

There is also a change in `app-controller.ts` that forces the electron E2Es to test against the minimum supported window size; this fixed an issue that came up during debugging where some test failures were happening in a way that reproed on build agents but not on dev machines.

Out of scope: updating all web CSS selectors used by E2Es to use `data-automation-id` instead (this would be a good idea to do to pre-empt tests breaking when moving more components to CSS modules, but leaving it for a separate PR)

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/components/cards/rules-with-instances.tsx
```diff
@@ -15,6 +14,8 @@ import { CollapsibleComponentCardsProps } from './collapsible-component-cards';
 import { RuleContent, RuleContentDeps } from './rule-content';
 import * as styles from './rules-with-instances.scss';
 
+export const ruleGroupAutomationId = 'cards-rule-group';
```
waabid: Any reason we can't move this to the place where the automation id is used (as in the collapsible cards component)?

PR #1822, opened 2019-12-09
refactor: promisify create tab new window

smoralesd:
#### Description of changes

In this case, promisifying `createTabInNewWindow`.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/details-view-controller.ts
+++ b/src/background/details-view-controller.ts
@@ -34,6 +34,21 @@ export class DetailsViewController {
         );
     }
 
+    public showDetailsViewP(targetTabId: number): Promise<void> {
+        const detailsViewTabId = this.tabIdToDetailsViewMap[targetTabId];
+
+        if (detailsViewTabId != null) {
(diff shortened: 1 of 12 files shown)


src/background/details-view-controller.ts
```diff
@@ -18,20 +18,19 @@ export class DetailsViewController {
...
 
         if (detailsViewTabId != null) {
             this.browserAdapter.switchToTab(detailsViewTabId);
             return;
         }
 
-        this.browserAdapter.createTabInNewWindow(
-            this.getDetailsUrl(targetTabId),
-            (tab: chrome.tabs.Tab) => {
+        return this.browserAdapter
```
dbjorge: If you use async/await in this sort of scenario, you'll get more useful error stacks in the event that something goes wrong (the stack will include this function with await, but will omit it without)

smoralesd: @dbjorge updated

PR #1847, opened 2019-12-12
fix: electron e2e issue with resize by putting a timeout after resize

Shobhit1:
#### Description of changes
Fix the electron e2e issue with button click by making window wait till its fully loaded (resize ends with a load event)

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/tests/electron/common/view-controllers/app-controller.ts
+++ b/src/tests/electron/common/view-controllers/app-controller.ts
@@ -28,7 +28,7 @@ export class AppController {
     }
 
     public async openDeviceConnectionDialog(): Promise<DeviceConnectionDialogController> {
-        await this.setToMinimumSupportedWindowSize();
+        this.resizeWindow();
 
         await dismissTelemetryOptInDialog(this.app);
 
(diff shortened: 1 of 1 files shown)


src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -45,11 +45,19 @@ export class AppController {
...
 
-        this.setToMinimumSupportedWindowSize();
+        this.resizeWindow();
 
         return automatedChecksView;
     }
 
+    private resizeWindow(): void {
+        let resizeId;
+        this.app.browserWindow.addListener('resize', () => {
```
dbjorge: I think there's maybe a misunderstanding here; I don't think this ever actually invokes this.setToMinimumSupportedWindowSize, it's just adding a listener such that it *would* be invoked if some other thing resized the window. I tried running the electron tests locally with this change to confirm my understanding and found that the electron window does not appear to get resized down to minimum with this code in place.

src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -45,11 +45,19 @@ export class AppController {
         const automatedChecksView = new AutomatedChecksViewController(this.client);
         await automatedChecksView.waitForViewVisible();
 
-        this.setToMinimumSupportedWindowSize();
```
dbjorge: I think the original name was more clear; consider keeping this name for the outer function and instead renaming (or inlining) the inner function

dbjorge:
Taking over in #1862 since Shobhit is out and this is causing noise in PRs/CI

PR #1851, opened 2019-12-12
feat(permissions): add ScanIncompleteWarning data into result stores

dbjorge:
#### Description of changes

This PR:

* Adds `scanIncompleteWarnings` data to the assessment and unified result stores
* Adds a new `ScanIncompleteWarningDetector` to handle populating the new data
* Updates the previously-done analyzer piping to pipe around the new `ScanIncompleteWarningDetector` instead of an `IframeDetector`
* Initializes the new detector in `main-window-initializer`

Out of scope (separate PRs):
* Wiring up the data from the permissions store @lisli1 sent out for PR recently to the new detector
* Consuming the new data in the new UI component @waabid has out for draft PR

#### Pull request checklist

- [x] Addresses an existing issue: #1655462
- [x] Ran `yarn fastpass`
- [0/a] (UI changes only) Verified usability with NVDA/JAWS

### DIFF
--- a/src/common/types/scan-incomplete-warnings.ts
+++ b/src/common/types/scan-incomplete-warnings.ts
@@ -1,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// This is a string-enum, but we only one type right now
+export type ScanIncompleteWarningId = 'missing-required-cross-origin-permissions';
--- a/src/common/types/scan-incomplete-warnings.ts
+++ b/src/common/types/scan-incomplete-warnings.ts
@@ -1,5 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-// This is a string-enum, but we only one type right now
+// This is a string-enum, but we only have one type right now
 export type ScanIncompleteWarningId = 'missing-required-cross-origin-permissions';
--- a/src/common/types/store-data/assessment-result-data.ts
+++ b/src/common/types/store-data/assessment-result-data.ts
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { ScanIncompleteWarningId } from 'common/types/scan-incomplete-warnings';
 import { DictionaryStringTo } from '../../../types/common-types';
 import { Tab } from '../../itab';
 import { ManualTestStatus, ManualTestStatusData } from '../manual-test-status';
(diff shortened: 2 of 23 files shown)


src/common/types/scan-incomplete-warnings.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// This is a string-enum, but we only one type right now
```
smoralesd: missing word: **have** => we only **have** one...

src/common/types/store-data/assessment-result-data.ts
```diff
@@ -28,6 +29,7 @@ export interface AssessmentData {
     generatedAssessmentInstancesMap?: InstanceIdToInstanceDataMap;
     manualTestStepResultMap?: RequirementIdToResultMap;
     testStepStatus: ManualTestStatusData;
+    scanIncompleteWarnings?: ScanIncompleteWarningId[];
```
waabid: Is there any particular reason for this being optional (here and other places)?

src/injected/scan-incomplete-warning-detector.ts
```diff
@@ -0,0 +1,20 @@
...
+    hasCrossOriginPermissions(): boolean;
+}
+
+export class ScanIncompleteWarningDetector {
+    constructor(private iframeDetector: IframeDetector, private crossOriginPermissionDetector: CrossOriginPermissionDetector) {}
+
+    public detectScanIncompleteWarnings = () => {
+        const warnings: ScanIncompleteWarningId[] = [];
+        if (this.iframeDetector.hasIframes() && !this.crossOriginPermissionDetector.hasCrossOriginPermissions()) {
+            warnings.push('missing-required-cross-origin-permissions');
```
waabid: Can we store this as a variable instead and reference it here and in the type declaration?

waabid:
LGTM, just a couple minor comments

src/common/types/store-data/assessment-result-data.ts
dbjorge: It's intentional for it to be optional in the store data sections because it has to be backwards compatible with users that upgrade with old store data in place.

src/injected/scan-incomplete-warning-detector.ts
dbjorge: It would be technically possible to do this in `scan-incomplete-warnings`:

```ts
export const missingRequiredCrossOriginPermissions = 'missing-required-cross-origin-permissions' as const;

// This is a string-enum, but we only have one type right now
export type ScanIncompleteWarningId = typeof missingRequiredCrossOriginPermissions;
```

...but I find this less readable than just using the string directly and relying on the string-enum type to enforce the use of specific string values, personally. @smoralesd / @lisli1 , care to provide a tiebreaker opinion? :)

smoralesd: I also prefer the use of the string directly, it reads better... the only downside I can see (I have actually seen it before) shows up if we decide to change the values of the string enum (e.g. in the case of a typo). I think in that case, VSCode rename won't work. I still prefer it this way though.

PR #1854, opened 2019-12-13
test: add all-origin cross-origin iframe scanning e2e test

smoralesd:
#### Description of changes

Adding e2e test to check we do get results from inside cross-origin iframes we the extension does have all-origins permissions.

Sadly, I have found no way to grant permissions to our extension like a user should by using puppeteer, so instead, we are following the same pattern we had before: altering the manifest -adding the permissions- just before launching the browser.

#### Pull request checklist

- [x] Addresses an existing issue: completes WI # 1655465
- [x] Ran `yarn fastpass`


src/tests/end-to-end/common/browser-factory.ts
```diff
@@ -29,17 +31,7 @@ export async function launchBrowser(extensionOptions: ExtensionOptions): Promise
...
-    if (extensionOptions.addLocalhostToPermissions) {
-        const originalManifest = await readFile(manifestPath);
-
-        const permissiveManifest = addLocalhostPermissionsToManifest(originalManifest.toString());
-
-        await writeFile(manifestPath, permissiveManifest);
-
-        onClose = async () => await writeFile(manifestPath, originalManifest.toString());
-    }
+    const onClose = await alterManifestWithPermissions(extensionOptions, manifestPath);
```
dbjorge: This line feels a little surprising to read (I assumed it was a bug at first based on the function name); maybe more clear would be to move it into a class where there can be separate `setTemporaryManifestPermissions` and `restoreOriginalManifestPermissions` functions?

src/tests/end-to-end/common/browser-factory.ts
```diff
@@ -17,9 +17,11 @@ const fileExists = util.promisify(fs.exists);
 const writeFile = util.promisify(fs.writeFile);
 const readFile = util.promisify(fs.readFile);
 
+export type ExtraPermissions = 'localhost' | 'all-origins';
```
dbjorge: It may not be clear to an unfamiliar reader why they'd ever want to use `localhost`, given that the only `permissions.request` calls we'll have in the product will be for all-origins; maybe naming something like `fake-activeTab` or something would be more clear, possibly with a comment explaining and/or linking the puppeteer issue?

src/tests/end-to-end/tests/details-view/cross-origin-iframe.test.ts
```diff
@@ -16,42 +16,73 @@ const testResourceServerConfig: ResourceServerConfig = {
...
+            }
+        });
+
+        it('does not get results from inside cross-origin iframes', async () => {
+            const automatedChecks = await formatPageElementForSnapshot(
+                fastPassAutomatedChecks,
+                fastPassAutomatedChecksSelectors.ruleDetailsGroups,
+            );
+
+            expect(automatedChecks).toMatchSnapshot();
```
dbjorge: It's already other tests' responsibility to test the E2E html generated to represent the fastpass cards view, right? Maybe this test should stick to just testing that the expected counts are present, rather than conflating the formatting testing vs the origin permissions testing?

dbjorge: (eg, see `it('supports expanding and collapsing rule groups', ...)` in `/src/tests/electron/tests/automated-checks-view.test.ts`)

src/tests/end-to-end/common/browser-factory.ts
smoralesd: updated

src/tests/end-to-end/common/browser-factory.ts
smoralesd: updated.

src/tests/end-to-end/tests/details-view/cross-origin-iframe.test.ts
smoralesd: @dbjorge is this closer to what you had in mind?

src/tests/end-to-end/tests/details-view/cross-origin-iframe.test.ts
```diff
@@ -43,16 +43,20 @@ describe('scanning', () => {
...
+                'frame-title': 2,
+                'html-has-lang': 1,
+            };
+
+            await assertFailureCounts(ruleDetails, expectedCounts);
         });
     });
 
-    describe('with all-origins permissions', () => {
+    describe.only('with all-origins permissions', () => {
```
dbjorge: stray `.only`?

dbjorge: (there must be a lint rule for this...)

smoralesd: ups

PR #1855, opened 2019-12-13
chore: add basic logging for autoUpdater

karanbirsingh:
#### Description of changes

This PR includes basic autoUpdater logging for debugging convenience. After this PR, logs from the autoUpdater process can retroactively be found in paths [here](https://github.com/megahertz/electron-log#electron-log). Without this PR, getting these logs requires restarting the application from the command line instead of the start menu.

Out of scope of this PR / to do later:
- decide whether to pipe existing renderer process logs to disk as well
- decide whether/how to combine autoUpdater logs & general logging
- react to unhandled errors from node
- decide whether to include electron/chromium logs

[electron-log](https://github.com/megahertz/electron-log#electron-log) and [winston](https://github.com/winstonjs/winston) are two popular logging libraries - I chose electron-log for now but it should be simple to switch later if needed.

The logger is set in `AutoUpdaterClient` instead of `main.ts` and I've omitted unit tests for whether the logger is non-null. If this would provide value, I can add it.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/auto-update/auto-updater-client.ts
```diff
@@ -1,8 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import log from 'electron-log';
 
 export class AutoUpdaterClient {
-    constructor(private readonly autoUpdater) {}
+    constructor(private readonly autoUpdater) {
```
smoralesd: is there any strong reason to not have a type for `autoUpdater`? looks like `AppUpdater` is the right type to use

src/electron/auto-update/auto-updater-client.ts
```diff
@@ -1,8 +1,12 @@
...
+        log.transports.file.level = 'info';
```
smoralesd: In general, we try to avoid having global dependencies, instead we favor passing any dependency through the constructor. In the end, the composition root is the responsible to get the right instance for each dependency.

src/electron/auto-update/auto-updater-client.ts
```diff
@@ -1,8 +1,12 @@
...
+        autoUpdater.logger = log;
```
smoralesd: Is there any way to test this is called? or at least that `logger` is indeed `log` ???
I'm surprise no test file has been changed as part of the changes on this constructor.

src/electron/auto-update/auto-updater-client.ts
karanbirsingh: I didn't notice the missing type, I agree with you - changed

src/electron/auto-update/auto-updater-client.ts
karanbirsingh: I alternated between these options for a while & have pushed a new in-between iteration given your comment
```
// in main.ts:
import log from 'electron-log';
autoUpdater.logger = log;
const electronAutoUpdateCheck = new AutoUpdaterClient(autoUpdater);
```
or
```
// in auto-updater-client.ts:
import log from 'electron-log';
autoUpdater.logger = log; // in constructor
```

karanbirsingh: spoke with @dbjorge / @waabid and settled on the first option above for now. We will look into `main.ts` refactoring separately, including how much of the update behavior to move into `AutoUpdaterClient` & how much of `main.ts` should be tested explicitly.

PR #1864, opened 2019-12-13
chore: Enable CSP script-src for electron product

karanbirsingh:
#### Description of changes

The electron security page [recommends](https://electronjs.org/docs/tutorial/security#csp-meta-tag) a [CSP policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) to restrict the resources that can be loaded into our `BrowserWindow`. Adding this policy revealed that we share the same electron configuration for dev and prod - both use `eval-source-map` for webpack's [devtool](https://webpack.js.org/configuration/devtool/#devtool). `script-src` doesn't allow script execution methods like `eval`. I've updated devtool to `source-map` (separate file) for now so that we can add CSP. I was still able to debug the app in vscode locally.

Other options include:
- using a [nonce for webpack](https://webpack.js.org/guides/csp/) to allow its scripts
- conditioning the CSP tag & devtool setting on dev/prod environment
- a less stringent directive than `script-src: self`

#### Pull request checklist

- [x] Ran `yarn fastpass`


webpack.config.js
```diff
@@ -96,7 +96,7 @@ const electronConfig = {
     entry: electronEntryFiles,
     name: 'electron',
     mode: 'development',
-    devtool: 'eval-source-map',
```
JGibson2019: Why does `script-src` not allow `eval` as a prefix?

PR #1865, opened 2019-12-14
feat(permissions): create BrowserPermissionsTracker

lisli1:
#### Description of changes

Created `BrowserPermissionsTracker` to check for all URL and all file permissions on the extension when it initializes and whenever permissions are added or removed.

#### Pull request checklist

- [x] Addresses an existing issue: # 1655063
- [x] Ran `yarn fastpass`


src/background/permissions-state-tracker.ts
```diff
@@ -0,0 +1,35 @@
...
+    public initialize(): void {
+        this.browserAdapter.addListenerOnPermissionsAdded(this.notifyChange);
+        this.browserAdapter.addListenerOnPermissionsRemoved(this.notifyChange);
+        this.notifyChange();
+    }
+
+    public notifyChange(): void {
+        const allUrlAndFilePermissions: string = '*://*/*';
+
+        // tslint:disable-next-line: no-floating-promises
```
dbjorge: The tslint error here is legitmate; it's warning you that there's no guarantee about what will happen if an error is thrown from within the promise. Probably better than suppressing the warning would be to add a `.catch` (or convert to an `async` function with a try/catch) that uses `logger.error` to report any exceptions that occur. Probably we should also consider an exception to be the same as a "false" result for the purposes of sending the state update message.

src/background/permissions-state-tracker.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { Interpreter } from 'background/interpreter';
+import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
+import { Message } from 'common/message';
+import { Messages } from 'common/messages';
+
+export class PermissionsStateTracker {
```
dbjorge: Thinking about this name more, maybe `BrowserPermissionsTracker` would be a more descriptive name?

src/tests/unit/tests/background/permissions-state-tracker.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    beforeEach(() => {
+        browserAdapterMock = Mock.ofType<BrowserAdapter>();
+        interpreterMock = Mock.ofType<Interpreter>();
+
+        testSubject = new PermissionsStateTracker(browserAdapterMock.object, interpreterMock.object);
+    });
+
+    it.each([true, false])('notifyChange sends the expected interpreter message', async permissionsState => {
+        browserAdapterMock
+            .setup(adapter => adapter.containsPermissions(It.isAny()))
```
dbjorge: Probably this should setup should require that it gets the expected origin pattern, rather than accepting any input

src/tests/unit/tests/background/permissions-state-tracker.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    let interpreterMock: IMock<Interpreter>;
+
+    beforeEach(() => {
+        browserAdapterMock = Mock.ofType<BrowserAdapter>();
+        interpreterMock = Mock.ofType<Interpreter>();
+
+        testSubject = new PermissionsStateTracker(browserAdapterMock.object, interpreterMock.object);
+    });
+
+    it.each([true, false])('notifyChange sends the expected interpreter message', async permissionsState => {
```
dbjorge: With `it.each`, it's helpful to put placeholders in the test names for the test data (otherwise it's hard to distinguish the 2 cases in test result output). Like this, I'd expect both test cases to show up in the console as `PermissionsStateTracker notifyChange sends the expected interpreter message`; better would maybe be `notifyChange sends the expected interpreter message when browser indicates permissions are %p`

dbjorge: I think this test case is verifying implementation details; the user of the class doesn't really need to know about the private `notifyChange` function, they care that notifications are sent in response to browser events and `initialize()`.

To test the browser events, the usual pattern is to set up a mock browser adapter to record the onAddListener/onRemoveListener calls and save any listeners they received, then test the results of calling those saved listeners. You can either set up the mocks for this manually here, or you could extend the `SimulatedBrowserAdapter` we have for this purpose with logic similar to what it already does for many of the other `addListener` browserAdapter methods.

src/background/permissions-state-tracker.ts
lisli1: Thanks for the suggestions! I like it much more than what I originally had.

src/background/browser-permissions-tracker.ts
```diff
@@ -0,0 +1,44 @@
...
+        private readonly logger: Logger,
+    ) {}
+
+    public async initialize(): Promise<void> {
+        this.browserAdapter.addListenerOnPermissionsAdded(this.notifyChange);
+        this.browserAdapter.addListenerOnPermissionsRemoved(this.notifyChange);
+        await this.notifyChange();
+    }
+
+    private notifyChange = async (): Promise<void> => {
```
waabid: notifyChange is probably not a great name for this. Something like "updatePermissionState" or something might be more appropriate.

src/background/browser-permissions-tracker.ts
```diff
@@ -0,0 +1,44 @@
...
+                origins: [allUrlAndFilePermissions],
+            });
+        } catch (error) {
+            payload = false;
+            this.logger.log('Error occurred while checking browser permissions');
+        } finally {
+            const message: Message = {
+                messageType: Messages.PermissionsState.SetPermissionsState,
+                payload: payload,
+                tabId: null,
```
waabid: Does tabId need to be set as null? Can it be left out entirely?

src/tests/unit/tests/background/browser-permissions-tracker.test.ts
```diff
@@ -0,0 +1,151 @@
...
+        it('registers the expected listeners', async () => {
+            await testSubject.initialize();
+
+            browserAdapterMock.verify(m => m.addListenerOnPermissionsAdded(It.is(isFunction)), Times.once());
+            browserAdapterMock.verify(m => m.addListenerOnPermissionsRemoved(It.is(isFunction)), Times.once());
+        });
+    });
+
+    describe('onPermissionsAdded', () => {
+        it.each([true, false])(
```
waabid: The tests look good! However, I do notice a lot of overlap in these scenarios that you can likely refactor.

An example would be a "getExpectedMessage" function with the expected browserPermissions value as a parameter. Something similar might be done with the browser adapter setup.

src/background/browser-permissions-tracker.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { Interpreter } from 'background/interpreter';
+import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
+import { Logger } from 'common/logging/logger';
+import { Message } from 'common/message';
+import { Messages } from 'common/messages';
+
+export const allUrlAndFilePermissions: string = '*://*/*';
```
waabid: Instead of storing this, you could just store the permissions object that we always ask for.

src/background/browser-permissions-tracker.ts
lisli1: You're right; I took out the tabId.

PR #1900, opened 2019-12-20
fix: NVDA Details Dialog Label Bug

JGibson2019:
#### Description of changes
With the previous pull request, the labels were only being read by Narrator and JAWS. The changes in this pull request ensure that all three screen readers (JAWS,  Narrator, and NVDA) are reading out the labels in the pictured dialog:

image

Whether or not "Region" is announced is AT specific. For example, users hear "Rule name aria-hidden-focus link" for Narrator and "Rule name region aria-hidden-focus-link" for NVDA. 

#### Pull request checklist

- [X] Addresses an existing issue: Bug #1644157
- [X] Ran `yarn fastpass`
- X ] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). Check workflow guide at: `<rootDir>/docs/workflow.md`


src/injected/components/details-dialog.tsx
```diff
@@ -186,9 +186,11 @@ export class DetailsDialog extends React.Component<DetailsDialogProps, DetailsDi
         const ruleNameID = 'rule-name';
 
         return (
-            <div className="insights-dialog-rule-name" aria-labelledby={ruleNameID}>
-                {this.renderSectionTitle('Rule name', ruleNameID)}
-                <NewTabLink href={fixUrl(rule.helpUrl)}>{rule.ruleId}</NewTabLink>
+            <div className="insights-dialog-rule-name">
```
smoralesd: did you try changing the original `<div>` to be a `<section>` instead of adding a new `<section>` tag as direct child of the origin `<div>`... mostly curious about the exploration process than anything.

JGibson2019: I did not try changing the original `<div>` to be a `<section>` tag; if I remember correctly, the example that we looked at to come up with this fix had the section tag nested within a div, and we were trying to keep the same structure. That said, I can try making the parent div a section tag and add the aria-labelledBy attribute back? Or are you ok with keeping it like this?

smoralesd: I think less "extra" tags in the general html structure is better, if we can go with changing the original `<div>` to `<section>`, that would be better.

JGibson2019: Note: updated to follow @smoralesd 's suggestion; converted `<div>` to a `<section>` tag and have both the className and aria-labelledBy as attributes for the `<section>` tag

PR #1901, opened 2019-12-20
refactor: promisify tabs query

smoralesd:
#### Description of changes

In this case, promisifying `tabsQuery`.

#### Pull request checklist
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/browser-message-broadcaster-factory.ts
+++ b/src/background/browser-message-broadcaster-factory.ts
@@ -14,9 +14,7 @@ export class BrowserMessageBroadcasterFactory {
         // condition described by #1816
         await this.sendMessageToFrames(message);
 
-        const allTabs: chrome.tabs.Tab[] = await new Promise(resolve => {
-            this.browserAdapter.tabsQuery({}, resolve);
-        });
+        const allTabs = await this.browserAdapter.tabsQueryP({});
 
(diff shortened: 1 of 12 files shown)


src/background/browser-message-broadcaster-factory.ts
```diff
@@ -14,9 +14,7 @@ export class BrowserMessageBroadcasterFactory {
         // condition described by #1816
         await this.sendMessageToFrames(message);
 
-        const allTabs: chrome.tabs.Tab[] = await new Promise(resolve => {
-            this.browserAdapter.tabsQuery({}, resolve);
-        });
+        const allTabs = await this.browserAdapter.tabsQuery({});
```
lisli1: Should we have a type for `allTabs`?

smoralesd: That was necessary before because by using the promise to wrap a call to the callback-based `tabsQuery` we get a result type of `unknown` (even though the callback of that function does have proper typing), hence, we need to be explicit with the types. If we don't, we get type errors like it's shown in this screenshot
image.
By changing the signature of `tabsQuery` to explicitly return a `Promise<Tabs.Tab[]>` the need to be explicit goes away, as the function is already being explicit with the return type and typescript is able to assign the proper type to `allTabs`.

PR #1911, opened 2019-12-24
refactor: rename chrome command handler

smoralesd:
#### Description of changes

We have a class responsible to trigger actions when the user uses a keyboard shortcut. The name of the class includes the word `Chrome`, which we should not use as the extension works on Edge too.

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/background/background-init.ts
```diff
@@ -121,7 +121,7 @@ async function initialize(): Promise<void> {
         logger,
     );
 
-    const chromeCommandHandler = new ChromeCommandHandler(
+    const chromeCommandHandler = new KeyboardShortcutHandler(
```
lisli1: Should we also change the variable name `chromeCommandHandler` to `keyboardShortcutHandler`?

smoralesd: Good catch!
Updated.

PR #1916, opened 2019-12-26
feat(permissions): iframe warning notification content update

smoralesd:
#### Description of changes

Currently, the issue notification (when the user uses keyboard shortcuts) shows up only when we detect no issues on the page.

We want to update this logic so:
1. We add text to mention we detected iframes (only if we detected)
2. Show a notification for both cases, issues found or no issues found.

This 2 points make up for 4 scenarios:
1. yes issues, yes iframes => issues found AND iframe warning
1. no issues, yes iframes => congrats message but with iframe warning
1. yes issues, no iframes => issues found
1. no issues, no iframes => congrats message

This PR update the issues notification message logic to account for those cases. All the other notifications (color, tab stops, headings and landmarks) are left untouched.

I took screenshots from windows notification center to show the text from the 4 scenarios we want to cover and how the fit on the notification popup:
notifications on windows

I also checked on Ubuntu but it's hard to get a proper screenshot. I'm going to wait until I can check on MacOS but besides that, this code changes are ready for review.

#### Pull request checklist

- [x] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`


src/ad-hoc-visualizations/issues/visualization.tsx
```diff
@@ -56,10 +56,26 @@ export const IssuesAdHocVisualization: VisualizationConfiguration = {
         }),
     getIdentifier: () => AdHocTestkeys.Issues,
     visualizationInstanceProcessor: () => VisualizationInstanceProcessor.nullProcessor,
-    getNotificationMessage: selectorMap =>
-        isEmpty(selectorMap)
-            ? 'Congratulations!\n\nAutomated checks found no issues on this page.'
-            : null,
+    getNotificationMessage: (selectorMap, key, warnings) => {
+        const iframeCoda =
+            '\nThere are iframes in the target page. Use FastPass or Assessment to provide additional permissions.';
```
waabid: This assumes that all warnings will be iframe warnings: are we sure we want to make that assumption (we are not doing so as far as I am aware elsewhere)?

src/ad-hoc-visualizations/issues/visualization.tsx
```diff
@@ -56,10 +56,26 @@ export const IssuesAdHocVisualization: VisualizationConfiguration = {
...
+
+        if (isEmpty(selectorMap)) {
+            if (isEmpty(warnings)) {
+                return 'Congratulations!\n\nAutomated checks found no issues on this page.';
+            }
+
+            return `No automated checks issues found.${iframeCoda}`;
+        }
+
+        const noIssuesFound = 'Automated checks found issues.';
```
waabid: This variable name should probably be "issues found", no?

src/ad-hoc-visualizations/issues/visualization.tsx
```diff
@@ -56,10 +56,26 @@ export const IssuesAdHocVisualization: VisualizationConfiguration = {
         }),
     getIdentifier: () => AdHocTestkeys.Issues,
     visualizationInstanceProcessor: () => VisualizationInstanceProcessor.nullProcessor,
-    getNotificationMessage: selectorMap =>
-        isEmpty(selectorMap)
-            ? 'Congratulations!\n\nAutomated checks found no issues on this page.'
-            : null,
+    getNotificationMessage: (selectorMap, key, warnings) => {
```
waabid: Not a necessity but I'd strongly like to move this out of here and have it as a floating function; I don't want to give the impression that we should be writing significant code (which I'd say this is) in the configuration files, if that makes sense.

src/ad-hoc-visualizations/issues/visualization.tsx
smoralesd: shameful display :( I was moving code around, changing the "direction" of the conditionals... I guess this was named on a previous iteration. Thanks for catching this.

src/ad-hoc-visualizations/issues/visualization.tsx
smoralesd: I was on the fence about this... right now this is the only place where we have a "complex" function to create the notification message, but I also knew we're going to need to update this as soon as we want to remove the old fast pass automated check view.

src/ad-hoc-visualizations/issues/visualization.tsx
smoralesd: Make sense. Updated to explicitly check for the proper id. I also refactor the code to reduce the amount of nested if's.

smoralesd:
@waabid Great review. I addressed all the comments and the code looks way better than the original proposal.

PR #1919, opened 2019-12-27
refactor: promisify switch to tab

smoralesd:
#### Description of changes

In this case, promisifying `switchToTab`.

`switchToTab` actually have a little logic inside (so not strictly an adapter function): 1) set the tab to active; 2) set the window to focused. Simple enough but doesn't really match the Adapter Pattern. As a future work (so out of scope for this PR) we may want to extract that bit of logic to, says, `TabSwitcher` class (or plainly a function).

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/background/target-tab-controller.test.ts
+++ b/src/tests/unit/tests/background/target-tab-controller.test.ts
@@ -57,7 +57,7 @@ describe('TargetTabControllerTest', () => {
             browserAdapterMock.verifyAll();
         });
 
-        it('propagates error from swith to tab', async () => {
+        it('propagates error from switch to tab', async () => {
             getSwitchToTargetTabCallbackMock
                 .setup(cm => cm(step))
                 .returns(() => true)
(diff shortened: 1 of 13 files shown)


src/tests/unit/tests/background/target-tab-controller.test.ts
```diff
@@ -25,46 +24,53 @@ describe('TargetTabControllerTest', () => {
...
+            await testSubject.showTargetTab(tabId, test, step);
+            browserAdapterMock.verifyAll();
+        });
 
-    it('showTargetTab: test does switch to target tab', () => {
-        getSwitchToTargetTabCallbackMock
-            .setup(cm => cm(step))
-            .returns(() => true)
-            .verifiable(Times.once());
+        it('propagates error from swith to tab', async () => {
```
lisli1: Looks like a typo: 

```suggestion
        it('propagates error from switch to tab', async () => {
```

PR #1975, opened 2020-01-13
feat(office-fabric): udpate tooltip host content

smoralesd:
#### Description of changes

After updating to office fabric 7, the **Assessment in progress** dialog got a layout-changing behavior when the user hover over the second target page link (this user story has more details 1663263).

This happen because we're mixing inline and block elements inside the dialog content. In order to fix this, I forced the `<Link>` component to render as an anchor tag (which is an inline type of element).

#### Pull request checklist
- [x] Addresses an existing issue: completes WI # 1663263
- [x] Ran `yarn fastpass`


dbjorge:
Approved assuming this passes E2Es, but it feels unlikely to me that this wouldn't break any E2E snapshots

PR #1981, opened 2020-01-14
fix: element snapshot formatter does not handle more than one class name value

smoralesd:
#### Description of changes

While reviewing #1979, @dbjorge notice we are not properly handling a class attribute with more than one class value inside of it, like in this real case: `ms-Panel-headerText header-text--0nPhV headerText-000`.

The expected result from the example above is for the formater/normalizer to change `header-text--0nPhV` into `header-text--{{CSS_MODULE_HASH}}`. The position of the class name value should not matter; the same goes for the amount of class names the class attribute has.

Because we're using regex to do the replacement, I added unit tests for those functions (that way it was easier to found which of the functions was broken) and then updated the logic of the broken one to properly handle more than one class name value.

#### Pull request checklist
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/end-to-end/tests/popup/__snapshots__/launchpad.test.ts.snap
+++ b/src/tests/end-to-end/tests/popup/__snapshots__/launchpad.test.ts.snap
@@ -103,7 +103,7 @@ exports[`Popup -> Launch Pad content should match snapshot 1`] = `
               >
                 <div
                   aria-hidden="true"
-                  class="ms-Grid-col ms-sm3 popup-start-dialog-icon-circle"
+                  class="ms-Grid-col ms-sm000 popup-start-dialog-icon-circle"
                 >
                   <i
                     aria-hidden="true"
@@ -151,7 +151,7 @@ exports[`Popup -> Launch Pad content should match snapshot 1`] = `
               >
                 <div
                   aria-hidden="true"
-                  class="ms-Grid-col ms-sm3 popup-start-dialog-icon-circle"
+                  class="ms-Grid-col ms-sm000 popup-start-dialog-icon-circle"
                 >
                   <i
                     aria-hidden="true"
@@ -199,7 +199,7 @@ exports[`Popup -> Launch Pad content should match snapshot 1`] = `
               >
                 <div
                   aria-hidden="true"
-                  class="ms-Grid-col ms-sm3 popup-start-dialog-icon-circle"
+                  class="ms-Grid-col ms-sm000 popup-start-dialog-icon-circle"
                 >
                   <i
                     aria-hidden="true"
--- a/src/tests/unit/common/__snapshots__/element-snapshot-formatter.test.ts.snap
+++ b/src/tests/unit/common/__snapshots__/element-snapshot-formatter.test.ts.snap
@@ -4,7 +4,7 @@ exports[`element-snapshot-formatter format html for snapshot normalizing ms-Pane
 <DocumentFragment>
   <div>
     <div
-      class="ms-Panel-headerText header-text--000nPhV headerText-000"
+      class="ms-Panel-headerText header-text--{{CSS_MODULE_HASH}} headerText-198"
     >
       Hello world!
     </div>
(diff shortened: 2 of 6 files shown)


src/tests/unit/common/__snapshots__/element-snapshot-formatter.test.ts.snap
```diff
@@ -0,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`element-snapshot-formatter format html for snapshot normalizing ms-Panel-headerText header-text--0nPhV headerText-198 1`] = `
+<DocumentFragment>
+  <div>
+    <div
+      class="ms-Panel-headerText header-text--{{CSS_MODULE_HASH}} headerText-198"
```
dbjorge: Shouldn't this be `headerText-000`?

src/tests/unit/common/element-snapshot-formatter.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    describe('normalize office fabric generated class names', () => {
+        describe.each`
+            actualClassName              | expectedClassName
+            ${'thumb-123'}               | ${'thumb-000'}
+            ${'ms-Panel-234'}            | ${'ms-Panel-000'}
+            ${'thumb-321 ms-Panel-432'}  | ${'thumb-000 ms-Panel-000'}
+            ${'no-numbers'}              | ${'no-numbers'}
+            ${'no-numbers thumb-543'}    | ${'no-numbers thumb-000'}
+            ${'Panel765-word'}           | ${'Panel000-word'}
+            ${'Panel765-word thumb-123'} | ${'Panel000-word thumb-000'}
```
dbjorge: I think this is missing a test case that would fail for `thumb-123 name-not-ending-in-digit`

src/tests/unit/common/__snapshots__/element-snapshot-formatter.test.ts.snap
dbjorge: (I imagine this from the module normalization applying first and then the `{` and `}` characters confusing the OF normalization regex)

src/tests/unit/common/element-snapshot-formatter.test.ts
smoralesd: Added

src/tests/unit/common/__snapshots__/element-snapshot-formatter.test.ts.snap
smoralesd: This is fixed now.

src/tests/end-to-end/tests/popup/__snapshots__/launchpad.test.ts.snap
```diff
@@ -103,7 +103,7 @@ exports[`Popup -> Launch Pad content should match snapshot 1`] = `
               >
                 <div
                   aria-hidden="true"
-                  class="ms-Grid-col ms-sm3 popup-start-dialog-icon-circle"
+                  class="ms-Grid-col ms-sm000 popup-start-dialog-icon-circle"
```
smoralesd: This is a side effect of the changes I made on the office fabric class name normalizer. It's unfortunate as this classes are related to the office fabric grid and are manually added.
Getting the regexp to skip this classes would add a lot of more complexity than I would like to add. Also, we should move away from office fabric grid and favor CSS grid and flex box instead.
Base on those arguments, I think it's fine to live with this small side effect for now.

src/tests/unit/tests/reports/package/__snapshots__/integration.test.ts.snap
```diff
@@ -117,7 +117,7 @@ exports[`report package integration with issues 1`] = `
             color-interpolation-filters="sRGB"
             filterUnits="userSpaceOnUse"
             height="29.5269"
-            id="filter0_d"
+            id="filter000_d"
```
smoralesd: Similar to my other comment about ms-grid related classes, this is a side effect from the office fabric class name normalizer update. Unfortunate but I think is not too bad.

smoralesd:
closing this PR in favor of newer PR #1987

PR #1984, opened 2020-01-15
feat(release-channels): publish intermediate electron release channel bundles

karanbirsingh:
#### Description of changes

In order to create release pipelines for insider/production, each pipeline needs input artifacts to pack and sign. This PR publishes an initial set of files from the canary pipeline so that the release work can be unblocked.

Not in this PR:
- instead of rebuilding just to publish, reuse the canary bundle that was tested in e2e tests
- several gruntfile refactors
- e2e tests of all flavors
- webpack config changes per flavor

sample run in internal release build

#### Pull request checklist

- [ ] Ran `yarn fastpass`


targets.config.js
```diff
@@ -116,4 +116,30 @@ module.exports = {
         bundleFolder: 'electronBundle',
         mustExistFile: 'main.bundle.js',
     },
+    electronInsider: {
```
jalkire: I know you mentioned that this PR isn't meant to include some refactoring work, but what would you thoughts be on updating the current `electron` target here to `electronCanary` so that it's consistent with the new `electronInsider` and `electronProduction` targets?

pipeline/electron/build-publish-electron-flavors.yaml
```diff
@@ -0,0 +1,24 @@
...
+      displayName: 'copy files to publish'
+      inputs:
+          SourceFolder: '$(System.DefaultWorkingDirectory)'
+          Contents: |
+              **/drop/**
+              electron-builder.yml
+              **/extension/icons/**
+              **/src/electron/resources/**
+              **/pipeline/scripts/**
+          TargetFolder: '$(System.DefaultWorkingDirectory)/builtproduct'
```
smoralesd: can this be a variable? so we re-use it between copy files target folder and publish path to publish?

targets.config.js
karanbirsingh: I am hoping we can come back and change the target names to either only be <channel> organized underneath product, or <product+channel>. My preference is the former, but in either case I tend towards leaving the current electron canary build alone until we settle on the others. This will avoid unintentional regressions in gruntfile.js and/or the current electron drop. It also feels like we might want an 'electronDev' target. @dbjorge would you prefer I update the electron target to electronCanary here?

pipeline/electron/build-publish-electron-flavors.yaml
karanbirsingh: discussed offline, will include in future PR

PR #2005, opened 2020-01-18
fix: fixing missing wcag links in generated report via report package

devaradhanm:
#### Description of changes
This PR fixes a bug in html report generation using the npm report package. The WCAG links were missing.

(ruleToLinkConfiguration in resultDecorator object was undefined when using it in report package)

Added react-helmet as a dependency for report package, since this change made the output js file to take dependency on react-helmet.
 
image

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/scanner/exposed-apis.ts
+++ b/src/scanner/exposed-apis.ts
@@ -36,6 +36,7 @@ export let scan = (
         documentUtils,
         messageDecorator,
         (ruleId, axeHelpUrl) => helpUrlGetter.getHelpUrl(ruleId, axeHelpUrl),
+        ruleToLinkConfiguration,
     );
     const launcher = new Launcher(axe, scanParameterGenerator, document, options);
     const axeResponseHandler = new AxeResponseHandler(
@@ -43,8 +44,6 @@ export let scan = (
         errorCallback,
         resultDecorator,
     );
-
-    resultDecorator.setRuleToLinksConfiguration(ruleToLinkConfiguration);
     launcher.runScan(axeResponseHandler);
 };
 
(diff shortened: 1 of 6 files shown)


src/scanner/exposed-apis.ts
```diff
@@ -36,15 +36,14 @@ export let scan = (
         documentUtils,
         messageDecorator,
         (ruleId, axeHelpUrl) => helpUrlGetter.getHelpUrl(ruleId, axeHelpUrl),
+        ruleToLinkConfiguration,
     );
     const launcher = new Launcher(axe, scanParameterGenerator, document, options);
     const axeResponseHandler = new AxeResponseHandler(
         successCallback,
         errorCallback,
```
JGibson2019: With the `setRuleToLinksConfiguration()` call being deleted below, what other purpose does `resultDecorator` serve?

devaradhanm: having a separate method setRuleToLinksConfiguration to set a dependency that is mandatory for resultDecorator to work is error prone. Instead with this pr, i am passing the ruleToLinkConfiguration in constructor. Also, this object will never change anyway.

devaradhanm: see integration.test.ts.snap. this will tell you what i am fixing.

JGibson2019: Oh I see what you're fixing, but I was wondering what else `resultDecorator` did in this file and if this instance of it could be deleted

smoralesd:
@devaradhanm could you fill out the **Pull request checklist**??

PR #2026, opened 2020-01-22
feat(unified-channels): Add pipeline definition for the new unsigned build

karanbirsingh:
#### Description of changes

This PR builds on a subset of #2006 - 

> This PR implements the new signed/unsigned build machinery we'll be using to produce signed builds across all channel+platform combinations we want to support. In particular, the expectation for setting up the pipelines will be:
> - `build-unsigned-release-packages.yaml` will trigger automatically with each passing build of our internal security/compliance check build 
> - it will need a non-secret build variable set for the shared app insights instrumentation key

This PR includes some changes found after validation of #2006. Many of them are small syntax changes described in [commit logs](https://github.com/microsoft/accessibility-insights-web/commits/channels) starting at abeee22. Some changes are mentioned explicitly below:
- Linux e2e tests run differently than non-Linux e2e tests. This PR modifies `electron-e2e-test-interactive.yaml` to manage this. We could have composed the YAML files to accomplish this as well, but this would require separating `build-unsigned-release-packages.yaml` per platform
- YAML files are moved into new `unified` folder
- injected electron-mirror args into packing step

Remaining work / out of scope:
- everything described as remaining work in #2006 
- signing packages, this will happen in an upcoming PR
- improving the e2e tests in this pipeline, which currently include an extra dev build step (1 min)
- removing redundancy between `electron-e2e-test-interactive.yaml` and `electron-e2e-test-linux.yaml` - both include build steps that are unnecessary for us, we will improve this in an upcoming PR

Impact:
- Adding the new YAML files should have no impact to the current release. Once it is in, we will create new ADO pipelines to begin consuming these YAML files.
- `electron-e2e-test-interactive.yaml` is used in other YAML pipelines, but should have no impact as all of them are called from mac/windows platforms.

#### Pull request checklist

- [ ] Ran `yarn fastpass`


pipeline/unified/build-unsigned-release-packages.yaml
```diff
@@ -0,0 +1,45 @@
...
+steps:
+    - template: ../install-node-prerequisites.yaml
+
+    - template: ../electron/download-electron-mirror.yaml
+
+    - template: ../electron/electron-e2e-test-interactive.yaml
+
+    - template: ../electron/electron-e2e-publish-results.yaml
+
+    - script: yarn build:electron:all --unified-version=$(Build.BuildNumber) --app-insights-instrumentation-key=$(AppInsightsInstrumentationKey)
```
jalkire: Is `Build.Buildnumber` what we want for version? I thought we set up `name` above for this purpose

karanbirsingh: I think they are identical - [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/run-number?view=azure-devops&tabs=yaml) it says `name` is used to specify run number and that you can reference it via `Build.BuildNumber`. It looks like we can do [either](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/run-number?view=azure-devops&tabs=yaml#how-can-you-reference-the-run-number-variable-within-a-script), so let me know if you think we should prefer one.

PR #2033, opened 2020-01-23
fix: remove operation.name from telemetry

karanbirsingh:
#### Description of changes

Application Insights logs [operation name](https://docs.microsoft.com/en-us/azure/azure-monitor/app/data-model-context#operation-name) which can include user paths. We don't need this information and would rather not have user paths in our telemetry, so this PR overwrites it to an empty string. This follows the example [here](https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#telemetrycontext).

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/background/telemetry/app-insights-telemetry-client.ts
+++ b/src/background/telemetry/app-insights-telemetry-client.ts
@@ -78,6 +78,7 @@ export class AppInsightsTelemetryClient implements TelemetryClient {
     }
 
     private initializeInternal(): void {
+        this.appInsights.context.operation.name = '';
         this.appInsights.context.addTelemetryInitializer((envelope: ExtendedEnvelop) => {
             const baseData = envelope.data.baseData;
             baseData.properties = {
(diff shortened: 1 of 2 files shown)


src/background/telemetry/app-insights-telemetry-client.ts
```diff
@@ -78,6 +78,7 @@ export class AppInsightsTelemetryClient implements TelemetryClient {
     }
 
     private initializeInternal(): void {
+        this.appInsights.context.operation.name = '';
```
smoralesd: I'm guessing this affects both AI-Web and AI-Android right?
I'm curious... one of the links you posted mention that **"Operation Name... [is] usually the URL of the HTTP request."** (I'm paraphrasing here). Some time ago, we did have the url as part of our telemetry and we did work to remove it (for GDPR compliance). Now I'm a little worried about this finding. I don't recall we talk about this one while removing the url and I'm wondering if we do actually see any url from **operation.name** property on telemetry.

karanbirsingh: @smoralesd based on what Fer showed me for web, it seems those URLs are relative to the chrome extension, e.g. something like 'background/background.html'. It's possible the electron URLs are this way because of how we're bundling canary; removing the field seems fair just in case

PR #2044, opened 2020-01-24
fix: avoid invalid div/span nesting in left nav items

### LINKED ISSUE #1801: Cards UI and parsing
## Describe the bug

Cards UI fails parsing requirement

**To Reproduce**

1. Run automated checks, fast pass in a website with issues
2. open inspect and run the following command in the console: 
```javascript
(function(){function c(a,b){var c=document.createElement("textarea");c.name=a;c.value=b;d.appendChild(c)}var e=function(a){for(var b="",a=a.firstChild;a;){switch(a.nodeType){case Node.ELEMENT_NODE:b+=a.outerHTML;break;case Node.TEXT_NODE:b+=a.nodeValue;break;case Node.CDATA_SECTION_NODE:b+="<![CDATA["+a.nodeValue+"]]\>";break;case Node.COMMENT_NODE:b+="<\!--"+a.nodeValue+"--\>";break;case Node.DOCUMENT_TYPE_NODE:b+="<!DOCTYPE "+a.name+">\n"}a=a.nextSibling}return b}(document),d=document.createElement("form");d.method="POST";d.action="https://validator.w3.org/nu/";d.enctype="multipart/form-data";d.target="_blank";d.acceptCharset="utf-8";c("showsource","yes");c("content",e);document.body.appendChild(d);d.submit()})();
```
3. "then run the [check for WCAG 2.1 parsing compliance bookmarklet](https://validator.w3.org/nu/about.html)
4. See errors
 * Error: Start tag seen without seeing a doctype first. Expected <!DOCTYPE html>. 
* Error: Element div not allowed as child of element span in this context. (Suppressing further errors from this subtree.) 
* Error: Element div not allowed as child of element button in this context. (Suppressing further errors from this subtree.) 

## Expected behavior

No errors should appear

- OS Name & Version:  Windows 10
- AI-Web Version & Environment:  Insider 2.11.0
- Browser Version:  Google Chrome 78
- Target Page:  any

## Additional context

Found during validation

dbjorge:
#### Description of changes

This fixes the last subitem of #1801, where our left nav items were failing the WCAG 4.1.1 parsing test due to us nesting `<div>`-based content for our left nav items inside of span/button elements. The outer `<span>` is controlled by office fabric in this case, so this updates all the inner elements to use `<span>` rather than `<div>`. This was a noop in most cases since almost all of the nav item content rendered inline/with overridden display styles anyway.

The sole exception was the overview button's layout where it has the label on one line then the percentage complete on the next line, which relied on the implicit block display; the style update in `detailsview.scss` is there to prevent this impact.

There is no visual or AT change as part of this update. After this update, the only remaining nu validator WCAG violations are false positives (see #1801 for more info)

#### Pull request checklist

- [x] Addresses an existing issue: fixes #1801
- [x] Ran `yarn fastpass`
- [n/a, no change] (UI changes only) Added screenshots/GIFs to description above

### DIFF
--- a/src/DetailsView/components/left-nav/left-nav-icon.tsx
+++ b/src/DetailsView/components/left-nav/left-nav-icon.tsx
@@ -12,11 +12,7 @@ export type LeftNavIconProps = {
 export const LeftNavStatusIcon = NamedFC<LeftNavIconProps>('LeftNavStatusIcon', props => {
     const { item } = props;
 
-    return (
-        <span>
-            <StatusIcon status={item.status} className={'dark-gray'} level="test" />
-        </span>
-    );
+    return <StatusIcon status={item.status} className={'dark-gray'} level="test" />;
 });
 
 export const LeftNavIndexIcon = NamedFC<LeftNavIconProps>('LeftNavIndexIcon', props => {
--- a/src/DetailsView/components/left-nav/left-nav-icon.tsx
+++ b/src/DetailsView/components/left-nav/left-nav-icon.tsx
@@ -13,14 +13,14 @@ export const LeftNavStatusIcon = NamedFC<LeftNavIconProps>('LeftNavStatusIcon', 
     const { item } = props;
 
     return (
-        <div>
+        <span>
             <StatusIcon status={item.status} className={'dark-gray'} level="test" />
-        </div>
+        </span>
     );
 });
 
 export const LeftNavIndexIcon = NamedFC<LeftNavIconProps>('LeftNavIndexIcon', props => {
     const { item } = props;
 
-    return <div className={'index-circle'}>{item.index}</div>;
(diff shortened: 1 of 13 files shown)


src/DetailsView/components/left-nav/left-nav-icon.tsx
```diff
@@ -13,14 +13,14 @@ export const LeftNavStatusIcon = NamedFC<LeftNavIconProps>('LeftNavStatusIcon',
     const { item } = props;
 
     return (
-        <div>
+        <span>
```
smoralesd: I'm wondering if this `<span>` is actually needed here

dbjorge: Looks like not! updated

PR #2063, opened 2020-01-29
refactor: popup header menu buttons

smoralesd:
#### Description of changes

While working on #2053, I tried to convert the related components to use CSS Modules but that turned out to be harder than it should. I think the main reason this happened is because of the way we decided to implement the buttons on the header (**gear** and **hamburger**):

1. **gear**: it's actually using `<Link>` and `<ContextualMenu>`
1. **hamburger**: it's using `<IconButton>` and `<ContextualMenu>`

This decision have some side effects:
1. Using two different component types makes it harder to style the components so they look the same.
1. Accessibility is not correct, as both elements are being announced as _plain_ buttons. There is no indication that activating either of the buttons will show the contextual menu.
1. Manually using `<ContextualMenu>` force us to have state-full components, which in general, tend to be more complex than stateless components.

This side effects seem to go away if we choose to use `<IconButton>` **and** to pass `menuProps` property when rendering the buttons. Passing `menuProps` effectively convert the elements into **menu** buttons, which means an AT, NVDA for instance, will announce the element something like this:
```
manage settings
menu button collapsed submenu
```
Note that **manage settings** is our own, custom 'aria-label' value, while the role **menu** and the state **collapsed** comes directly from using `<IconButton>` + `menuProps` property, there is nothing extra we need to do to achieve this.
Also, the new components will be stateless as the `<IconButton>` will take that responsibility now.

**NOTE**
I know this looks like a big PR. I did try to break it on smaller chunks but it was too hard to achieve at both the typescript and css level without doing improper things. I guess it could be easier to review commit by commit instead.

01 - collapsed default state
02 - collapsed hover state
03 - expanded no item hover selected
04 - expanded item hover
05 - collapsed default state
06 - expanded default state
07 - expanded item hover
08 - collapsed default state HC
09 - collapsed hover state HC
10 - expanded no item hover selected HC
11 - expanded item hover HC
12 - collapsed default state HC
13 - expanded default state HC
14 - expanded item hover HC

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/common/components/gear-menu-button.tsx
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the MIT License.
+import { NamedFC } from 'common/react/named-fc';
+import { IconButton, IContextualMenuItem } from 'office-ui-fabric-react';
+import * as React from 'react';
+import { DropdownClickHandler } from '../dropdown-click-handler';
+import { FeatureFlags } from '../feature-flags';
+import { FeatureFlagStoreData } from '../types/store-data/feature-flag-store-data';
+import * as styles from './gear-menu-button.scss';
+
+export interface GearOptionsButtonComponentProps {
```
karanbirsingh: We can rename this to reflect the new component name

src/common/components/header.tsx
```diff
@@ -7,15 +7,22 @@ import { NamedFC } from 'common/react/named-fc';
 import { title } from 'content/strings/application';
 import * as styles from './header.scss';
 
-export type HeaderProps = { deps: HeaderDeps };
 export type HeaderDeps = HeaderIconDeps;
 
+export type HeaderProps = {
+    deps: HeaderDeps;
+    nearItems?: JSX.Element;
```
karanbirsingh: The names near and far didn't make sense to me at first, but I'm not sure if names like 'leftItems' would be better or if they're too specific

smoralesd: So `nearItems` and `farItems` follow the naming choice of Office UI Fabric [CommandBar](https://developer.microsoft.com/en-us/fabric#/controls/web/commandbar). I think they avoid using `leftItems` and `rightItems` to account for RTL languages where using left and right to name items and mention how they are positioned in the general UI layout is not correct. 
I could change the name but I agree with you: left and right doesn't make more/better sense to me.

src/common/components/gear-menu-button.tsx
smoralesd: Good catch

src/common/components/header.tsx
peterdur: `leftItems`/`rightItems` would assume a specific reading direction, which isn't true for all languages (though for now that's not an issue for AI-Web). I do note that the Office Fabric [CommandBar](https://developer.microsoft.com/en-us/fabric#/controls/web/commandbar) uses `items` and `farItems`; would it make sense to be consistent with this? Or is there other precedent for `nearItems`?

smoralesd: I swear I saw `nearItems`... I'll update to use `items` and `farItems`.

smoralesd: Updated

src/DetailsView/components/interactive-header.tsx
```diff
@@ -25,13 +24,14 @@ export const InteractiveHeader = NamedFC<InteractiveHeaderProps>('InteractiveHea
...
 
-    return (
-        <Header deps={props.deps}>
-            <Switcher deps={props.deps} pivotKey={props.selectedPivot} />
-            <GearOptionsButtonComponent
-                dropdownClickHandler={props.dropdownClickHandler}
-                featureFlags={props.featureFlagStoreData}
-            />
-        </Header>
+    const getNearItems = () => <Switcher deps={props.deps} pivotKey={props.selectedPivot} />;
```
peterdur: Update to match rename below?
```suggestion
    const getItems = () => <Switcher deps={props.deps} pivotKey={props.selectedPivot} />;
```

smoralesd: updated

smoralesd:
> I wonder if using MockBehavior.Strict in the button tests would allow us to remove the verify(Times.never()) calls before each click - I'm OK either way

Is this on the `HamburgerMenuButton` & `GearMenuButton` tests?

If I understand correctly your comment, you are talking about creating the mocks like:
```typescript
Mock.ofType<LaunchPanelHeaderClickHandler>()
```
which makes the mock to effectively use `MockBehavior.Loose`. This allows to directly validate if a function on the mock has been called.
On the other hand, if you use `MockBehavior.Strict` that force you into a 2 step mock usage:
- first: you need to setup (in which case, I strongly feel I'll use `verifiable(Times.once())` any way)
- second: you need to call `verifyAll` on the mock.

I know we have instances of both cases. I think in this case I do prefer to use loose behave mocks though.

I'll completed this PR and check with you offline in case we want to revisit and update.

PR #2081, opened 2020-01-31
refactor: generic toggle to use css module

smoralesd:
#### Description of changes

Convert `<GenericToggle>` to use css module.

This will help to have the correct stylings on both **AI-Web** and **AI-Android** (the latter will come up as part as an upcoming feature to bring the settings panel to AI-Android).

Additionally, fixed the description text color which was missed from a previous CSS-Module refactor on the `<SettingsPanel>` 

**Wrong color (on prod 2.14.1)**
01 - incorrect color from prod 2 14 1

**Fixed color**
02 - fixed color on this PR

#### Pull request checklist
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/settings-panel/settings-panel.scss
+++ b/src/DetailsView/components/settings-panel/settings-panel.scss
@@ -3,10 +3,6 @@
 @import '../../../common/styles/colors.scss';
 
 .settings-panel {
-    :global(.toggle-description) {
-        color: $secondary-text;
-    }
-
     h3 {
         font-size: 17px;
         margin-block-end: 8px;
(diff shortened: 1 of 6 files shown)


src/DetailsView/components/settings-panel/settings-panel.scss
```diff
@@ -3,7 +3,7 @@
 @import '../../../common/styles/colors.scss';
 
 .settings-panel {
-    .generic-toggle-component .toggle-description {
+    :global(.toggle-description) {
```
dbjorge: Dumb question: is it actually desirable/necessary for the description text color to be secondary-text in the settings panel but the default primary text color in the preview features panel (the only other place where generic toggle appears to be used)? I'd have expected us to want those to look consistent, in which case this probably just belongs in `generic-toggle.scss`

smoralesd: Good point. Honestly, I was following on what we had before (as far as I know) but it does make sense and I check with Fer too.

PR #2092, opened 2020-02-04
refactor: is-mac-os class on AI-Android

smoralesd:
#### Description of changes

Currently, AI-Android uses different css styles for the window title bar base on which OS the app is running on (mac os vs others). The different styles are being applied at the `<WindowTitle>` level, not at a higher level (`<body>` element for instance) like in the case of `'high-contrast-theme'`.

Working on adding the **settings panel** to AI-Android, I noticed there will be a need to do differential behavior for the settings panel component based on OS as well, so it makes sense to move and rename `.mac-window-title` from it's current placement to the `<body>` element, so any component that needs different styles based on OS can use the same class. 

Doing this matches what we do with high contrast theme, which is something we need to bring to AI-Android as part of the settings panel feature anyway. This PR will also help with that part.

**Notes** there is no actual UI/UX behavior changes, only how we apply class names and styles

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/electron/views/common/body-class-modifier/body-class-modifier.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export type BodyClassModifierDeps = {
+    platformInfo: PlatformInfo;
+};
+
+export type BodyClassModifierProps = {
+    deps: BodyClassModifierDeps;
+};
+
+export class BodyClassModifier extends React.Component<BodyClassModifierProps> {
```
smoralesd: I'm open to a better name for this component

src/electron/views/common/window-title/window-title.tsx
```diff
@@ -43,9 +37,11 @@ export const WindowTitle = NamedFC<WindowTitleProps>('WindowTitle', (props: Wind
 });
 
 function getIconsContainer(props: WindowTitleProps): JSX.Element {
-    if (!props.deps.platformInfo.isMac() && !isEmpty(props.actionableIcons)) {
-        return <div className={styles.actionableIconsContainer}>{props.actionableIcons}</div>;
+    const { platformInfo } = props.deps;
+
+    if (platformInfo.isMac() || isEmpty(props.actionableIcons)) {
```
dbjorge: I feel a little uneasy about this change; I worry that it's adding new complexity/global state without removing the original complexity, in that the component in question still uses the existing mechanism for passing this flag globally (`platformInfo.isMac()`), but now additionally relies on the second new mechanism for passing the same information globally (`:global(.is-mac-os)`).

smoralesd: As mention offline and also on the description, moving `is-mac-os` class to the `<body>` element follows the same pattern we have for `high-contrast-theme`. It moves the logic of adding the class to a unique point in the code and then, it let's each component, through it css module, to adjust styles accordingly if there is ever a need to style differently per OS.

Doing this now will also help in bringing high contrast theme to AI-android in the near future.

I don't think this 2 concerns (applying the class and styling differently base on the present of the tag) need to be address on the same place in the code, _au contraire_, the class is indeed a global concern (the whole app is running on mac os, or is not) while the styling is a local concern (not all the components will need to use different styles base on OS). This matches almost 100% with the logic we follow when implementing high contrast theme.

PR #2113, opened 2020-02-06
feat(unified-channels): enable icons on linux for unified electron client

karanbirsingh:
#### Description of changes

This PR adds icons to the Linux AppImage for the unified electron client:
- we use an icns file, aligning with [guidance](https://www.electron.build/icons.html#linux) from electron-builder. This allows tools like appimaged or AppImageLauncher to product thumbnail icons in Ubuntu from the installer
- we set the icon explicitly via electron so that the running-app-icon appears regardless of whether the user has integrated the installer via appimaged/AppImageLauncher

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/electron/main/main.ts
```diff
@@ -11,14 +13,21 @@ import { mainWindowConfig } from './main-window-config';
...
+
 log.transports.file.level = 'info';
 autoUpdater.logger = log;
 
 let recurringUpdateCheck;
 const electronAutoUpdateCheck = new AutoUpdaterClient(autoUpdater);
 
+const os = platformInfo.getOs();
+
+const iconBaseName = path.join(__dirname, '..', config.getOption('electronIconBaseName'));
```
dbjorge: I wonder if we should extract this bit out to a separate file; it's pretty small but also pretty independent/testable

smoralesd:
is it `feat(release-channels)` or `feat(unified-channels)`??? or are they different features?

karanbirsingh:
> is it feat(release-channels) or feat(unified-channels)??? or are they different features?

Looks like I was using the wrong name - I've updated it

PR #2120, opened 2020-02-07
feat(android-settings): bring settings panel to AI-Android

smoralesd:
#### Description of changes

This PR hook up all necessary pieces to bring the settings panel (from AI-Web) into AI-Android.

There was only one `NullObject` needed to pull this off.

Styles needed to be adjusted for the `<Panel>` to not be render on top of the window title bar. The styles also account for the height difference when running AI-Android on MacOS.

After this PR is completed, bringing **high contrast mode** should be straight forward, although, we'll need to check if all related styles are being applied correctly and that nothing is left behind (i.e. elements that are not affected by high contrast, giving the app and mixed-style look)

01 - settings panel

**Notes**
In order to keep this PR small, e2e tests are consider out of scope of the PR, but not from the ~not-yet-active~ feature.

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/background/details-view-controller.ts
```diff
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { DetailsViewControllerType } from 'background/details-view-controller-type';
 import { BrowserAdapter } from '../common/browser-adapters/browser-adapter';
 import { DictionaryStringTo } from '../types/common-types';
 
-export class DetailsViewController {
+export class DetailsViewController implements DetailsViewControllerType {
```
dbjorge: I felt a little confused by this naming. Maybe better would be `ExtensionDetailsViewController implements DetailsViewController`? Or is `DetailsViewController` used in too many places to want to do the refactoring in the same PR?

src/electron/settings/android-settings-provider.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { createSettingsProvider } from 'DetailsView/components/settings-panel/settings/settings-provider';
+import { TelemetrySettings } from 'DetailsView/components/settings-panel/settings/telemetry/telemetry-settings';
+
+export const AndroidSettingsProvider = createSettingsProvider([TelemetrySettings]);
```
dbjorge: `UnifiedSettingsProvider`

src/electron/views/renderer-initializer.ts
```diff
@@ -85,6 +96,10 @@ const windowStateActions = new WindowStateActions();
 const scanActions = new ScanActions();
 const unifiedScanResultActions = new UnifiedScanResultActions();
 const cardSelectionActions = new CardSelectionActions();
+const detailsViewActions = new DetailsViewActions();
+const previewFeaturesActions = new PreviewFeaturesActions(); // not really used but needed by DetailsViewStore
```
dbjorge: It feels like there are a lot of dependencies being pulled in just because of the use of the DetailsViewStore. I wonder if it would make more sense to do a PR prior to this one that separates out a `SettingsPanelStore` or something?

dbjorge: (I almost wrote the same comment earlier in this review up at `automated-checks-view`)

src/background/details-view-controller.ts
smoralesd: updated

src/electron/settings/android-settings-provider.ts
smoralesd: Updated

src/electron/views/renderer-initializer.ts
smoralesd: @dbjorge 
I kind of agree but I have made several PR's before this in preparation to bring the settings panel as cleanly as possible to AI-Android. I rather complete this PR and take on the refactor later. Thoughts?

dbjorge: Separate PR seems reasonable to me!

PR #2148, opened 2020-02-12
draft mac update

karanbirsingh:
#### Description of changes

This PR contains changes to fix updating on macOS for the unified client. Because this branch was associated with several builds during testing, I will recreate the PR with a new branch when this changeset looks good


pipeline/scripts/remove-pkg.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+const child_process = require('child_process');
```
dbjorge: It would maybe be nice to leave a comment in this file describing what its purpose is and why it's necessary

pipeline/scripts/update-latest-yml.js
```diff
@@ -6,6 +6,7 @@ const YAML = require('js-yaml');
 const path = require('path');
 
 const parentDir = process.argv[2];
+const platform = process.argv.length > 3 ? process.argv[3] : process.platform;
 
 const getLatestYAMLPath = parentDir => {
     const platformModifier = process.platform == 'darwin' ? '-mac' : process.platform == 'linux' ? '-linux' : '';
```
dbjorge: Should we be replacing the other usages of `process.platform` in this file with the new parameter?

pipeline/scripts/zip-mac-folder.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+const child_process = require('child_process');
```
dbjorge: Here too, would be nice to have a comment describing why this is necessary/referencing appropriate github issue

pipeline/unified/channel/sign-release-package.yaml
```diff
@@ -34,8 +35,12 @@ jobs:
                 signConfigType: inlineSignParams
                 inlineOperation: ${{ parameters.inlineSignParams }}
 
-          - script: yarn update:electron-checksum signing-in-progress/${{ parameters.signedArtifactName }}
-            displayName: update electron checksum after signing
+          - script: node ./pipeline/scripts/remove-pkg.js signing-in-progress/${{ parameters.signedArtifactName }}
+            displayName: remove .pkg binary from zip file
+            condition: eq(variables['Agent.OS'], 'Darwin')
+
+          - script: yarn update:latest-yml signing-in-progress/${{ parameters.signedArtifactName }} ${{ parameters.platform }}
```
dbjorge: Should we get rid of the `update:latest-yml` package.json entry and just call `node ./pipeline/scripts/etc.js` directly like on line 38? The more direct path feels a little nicer to me if this is the only place it's used; less unnecessary stuff in package.json and one less layer to have to follow to understand the pipeline

PR #2149, opened 2020-02-12
feat(android-settings): refactor how we assign class to the body

smoralesd:
#### Description of changes

Currently, we have 2 different approach and 2 slightly different needs to assign a `class` to the body of an html page. On one side, we have the high contrast theme on AI-Web. On the other side, we have the `is-mac-os` on AI-Android. In both cases, we use `<Helmet>` for doing this.

As part of the android settings feature, we need to bring high contrast mode to AI-Android and it would be nice if both, AI-Web and AI-Android, use the same mechanism.

In this PR, I'm updating `BodyClassAssigner` to use the newly introduce `ClassAssigner` type. This will allow for:
- Separate concerns, as now the `BodyClassAssigner` does not know about the logic to apply each specific class (`is-mac-os` and `high-contrast-theme`) nor does it care
- It makes `BodyClassAssigner` open/close by accepting an array of `ClassAssigners` and calling them one by one to apply classes according to their own logic.
- It allows to instanciate the `BodyClassAssigner` with different `ClassAssigners` depending on the needs:
   - AI-Web will only use `HighContrastClassAssigner` (TBD, this will come in a future PR)
   - AI-Android will use both `MacOsClassAssigner` and `HighContrastClassAssigner`

#### Pull request checklist
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/views/common/body-class-modifier/body-class-modifier.tsx
+++ b/src/electron/views/common/body-class-modifier/body-class-modifier.tsx
@@ -1,12 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { PlatformInfo } from 'electron/window-management/platform-info';
+import { ClassAssigner } from 'electron/views/common/body-class-modifier/class-assigner';
+import { uniq } from 'lodash';
 import { css } from 'office-ui-fabric-react';
 import * as React from 'react';
 import Helmet from 'react-helmet';
 
 export type BodyClassModifierDeps = {
-    platformInfo: PlatformInfo;
+    classAssigners: ClassAssigner[];
 };
 
 export type BodyClassModifierProps = {
@@ -15,15 +16,11 @@ export type BodyClassModifierProps = {
 
 export class BodyClassModifier extends React.Component<BodyClassModifierProps> {
     public render(): JSX.Element {
-        const classNames = [];
-
-        if (this.props.deps.platformInfo.isMac()) {
-            classNames.push('is-mac-os');
-        }
+        const classNames = this.props.deps.classAssigners.map(assigner => assigner.assign());
 
         return (
             <Helmet>
-                <body className={css(...classNames)} />
+                <body className={css(...uniq(classNames))} />
             </Helmet>
         );
     }
--- a/src/electron/views/renderer-initializer.ts
+++ b/src/electron/views/renderer-initializer.ts
@@ -51,10 +51,8 @@ import { createScanResultsFetcher } from 'electron/platform/android/fetch-scan-r
 import { ScanController } from 'electron/platform/android/scan-controller';
 import { createDefaultBuilder } from 'electron/platform/android/unified-result-builder';
 import { UnifiedSettingsProvider } from 'electron/settings/unified-settings-provider';
-import {
-    RootContainerDeps,
-    RootContainerState,
-} from 'electron/views/root-container/components/root-container';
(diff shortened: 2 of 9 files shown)


src/electron/views/common/body-class-modifier/body-class-modifier.tsx
```diff
@@ -1,12 +1,13 @@
...
-import { PlatformInfo } from 'electron/window-management/platform-info';
+import { ClassAssigner } from 'electron/views/common/body-class-modifier/class-assigner';
+import { uniq } from 'lodash';
 import { css } from 'office-ui-fabric-react';
 import * as React from 'react';
 import Helmet from 'react-helmet';
 
 export type BodyClassModifierDeps = {
-    platformInfo: PlatformInfo;
+    classAssigners: ClassAssigner[];
```
dbjorge: I worry that encapsulating the information that determines className behind a function call like this may lead to issues with React not propogating updates to this component like we'd expect; it doesn't matter here when it's just `platformInfo` since it doesn't change over time, but it might when we add high contrast mode and need React to understand to rerender this component when the high contrast mode settings change.

I wonder if it would make more sense to re-use the existing `Theme` component that the extension uses for the high contrast body classes, give it an `extraClassNames` prop, and pass it `calculatePlatformClassNames(platformInfo)` from the unified initializer similar to what you're doing to pass the classAssigners in this PR.

smoralesd: So... In a different branch ([this one](https://github.com/smoralesd/accessibility-insights-web/tree/high-contrast-theme-class-assigner)), I'm introducing `ThemeClassAssigner`. I converted `BodyClassAssigner` so it uses `withStoreSubscription`. This ensure the updates you're worried about happen when the user enable/disable high contrast mode (it's working properly on the branch I mention). I'm not 100% convinced on the design though. The fact that `BodyClassAssigner` "listen" to store changes to trigger the class assigners but doesn't explicitly pass the state to them (nor it uses the state directly) seems odd to me. I'm open to suggestions.

dbjorge: I agree that it doesn't feel right for the `BodyClassModifier` to appear to listen to state changes but not directly use the state; I think it would be best to make the coupling more obvious. Maybe instead of `assign()` and `update()`, just do `assign(userConfig: UserConfigurationStoreData)`, and make `ThemeClassAssigner` stateless?

src/electron/views/renderer-initializer.ts
```diff
@@ -314,6 +317,7 @@ getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch).then(
             ...cardsViewDeps,
             storeActionMessageCreator: new NullStoreActionMessageCreator(),
             settingsProvider: UnifiedSettingsProvider,
+            classAssigners,
```
dbjorge: I felt confused by the name `classAssigners` here; I wasn't able to understand based on the name what the responsibility of the thing being passed here was (it's much more specific than "assigning classes"). Maybe `bodyClassAssigner`, at least?

("assigner" feels like maybe not the right verb, since they're more determining which classes should be assigned than performing the assignment, but I'm struggling to come up with a better verb. "determiner"? doesn't sound great)

smoralesd: `bodyClassAssigner` doesn't sound right either... the assigner only get the class to be applied... it doesn't need to now to which element is applied. That's the `BodyClassModifier` responsibility...

PR #2168, opened 2020-02-18
refactor: ad hoc tools panel to use CSS Grid

smoralesd:
#### Description of changes

Currently, the ad hoc tools panel is using `ms-Grid` styles to layout the toggles. This put a lot of knowledge about layout on the component. Also, using `ms-Grid` is not that great, specially compared to using CSS Grid.

In this PR: refactor the component to use CSS Grid instead.

**NOTE** there are no visible UI changes.

#### Pull request checklist
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/popup/components/ad-hoc-tools-panel.tsx
+++ b/src/popup/components/ad-hoc-tools-panel.tsx
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { Icon } from 'office-ui-fabric-react';
-import { Link } from 'office-ui-fabric-react';
+import { NamedFC } from 'common/react/named-fc';
+import { css, Icon, Link } from 'office-ui-fabric-react';
 import * as React from 'react';
-
+import * as styles from './ad-hoc-tools-panel.scss';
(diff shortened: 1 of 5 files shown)


src/popup/components/ad-hoc-tools-panel.tsx
```diff
@@ -1,75 +1,64 @@
...
 
 export interface AdHocToolsPanelProps {
     backLinkHandler: () => void;
     diagnosticViewToggleFactory: DiagnosticViewToggleFactory;
 }
 
-export class AdHocToolsPanel extends React.Component<AdHocToolsPanelProps, {}> {
-    private sliceSize: number = 3;
+export const AdHocToolsPanel = NamedFC<AdHocToolsPanelProps>('AdHocToolsPanel', props => {
+    const getTogglesWithDividers = () => {
```
dbjorge: A nice further improvement would be to move these dividers to be CSS borders rather than elements in the layout, but I think it's reasonable to defer that for a separate PR

smoralesd: Agree... I'll have a second look as soon as I have some slack time.

PR #2172, opened 2020-02-19
feat(docs): Changed productName to use deps.applicationTitle

markreay:
#### Description of changes

Another refactor to support sharing UI components into the docs repo.

There is another reference to a hard-coded constant in the Markup classes. It uses `productName` which is identical to `applicationTitle`. Changed to use the `applicationTitle` found in `deps` to abstract.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/views/content/markup.tsx
+++ b/src/views/content/markup.tsx
@@ -44,7 +44,7 @@ export type Markup = {
 };
 
 export type MarkupDeps = {
-    textContent: Pick<TextContent, 'applicationTitle'>,
+    textContent: Pick<TextContent, 'applicationTitle'>;
     contentActionMessageCreator: ContentActionMessageCreator;
 };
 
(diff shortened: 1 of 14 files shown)


src/views/content/markup.tsx
```diff
@@ -43,21 +42,23 @@ export type Markup = {
     Include: React.FC<{ content: ContentPageComponent }>;
 };
 
-export type MarkupDeps = { contentActionMessageCreator: ContentActionMessageCreator };
+export type MarkupDeps = {
+    applicationTitle: string;
```
smoralesd: I'm not sure how I feel about this. I think we have use **deps** to put functions or instances of classes we need to pass around for components to work but not string. I think we have done this more or less on purpose so having a string on the deps feels odd. I'm not 100% sure this is not already happening; also, I don't have a better approach right now, as passing `applicationTitle` as a "normal" prop probably means a lot of piping. Thoughts?

markreay: @smoralesd, For what it's worth, it is already happening but that doesn't mean it should be revisited. 

It could be a function returning a string? Or it could be a constants object with a string. 

What do you suggest?

markreay: After discussion with Sebastian we will try to go with an object named `textContent`

markreay: Changed as discussed.

PR #2182, opened 2020-02-20
feat(android-settings): enable multiple BodyClassManipulators to co-exist

dbjorge:
#### Description of changes

This PR:
* Extracts the platformInfo-based responsiblities out of `BodyClassManipulator` into `PlatformBodyClassManipulator`
* Moves `BodyClassManipulator` to common in preparation for `Theme` to start using it
* Updates `BodyClassManipulator` to manage body class state itself (using new `DocumentManipulator` functionality) rather than using Helmet, to enable multiple manipulators to co-exist peacefully (as opposed to Helmet's behavior of only allowing one Helmet to work at a time on the `<body>`)

A subsequent PR will update `<Theme>` to use `BodyClassManipulator` in place of `Helmet`; that will enable `Theme` and `PlatformBodyClassManipulator` to co-exist peacefully in `RootContainerRenderer`

#### Pull request checklist

- [x] Addresses an existing issue: part of 1677805
- [x] Ran `yarn fastpass`


src/common/components/body-class-modifier.tsx
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { DocumentManipulator } from 'common/document-manipulator';
+import * as React from 'react';
+
+export type BodyClassModifierProps = {
+    classNames: string[];
+    documentManipulator: DocumentManipulator;
```
smoralesd: should this be a dep instead of a prop?

src/common/components/body-class-modifier.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+// We user our own BodyClassModifier rather than using Helmet's similar functionality because we
+// need multiple BodyClassModifiers to be able to co-exist without overwriting each other or the
+// body classes office fabric maintains.
+export class BodyClassModifier extends React.Component<BodyClassModifierProps> {
+    public render(): JSX.Element {
+        return null;
```
smoralesd: _"funny"_ comment: we went from a component that listen to the store for updates and render, calling the dependencies but not passing any of the new state, which was awkward, to a component that does pass params to it's dependencies on updates but that renders null every time :P Don't get me wrong, this looks way better than my original attempt, but it's still a tad odd that the component renders null no matter what.

src/common/components/body-class-modifier.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    public componentDidUpdate(prevProps: BodyClassModifierProps): void {
+        this.updateClassNames(prevProps.classNames, this.props.classNames);
+    }
+
+    public componentWillUnmount(): void {
+        this.updateClassNames(this.props.classNames, []);
+    }
+
+    private updateClassNames(oldClassNames: string[], newClassNames): void {
```
smoralesd: missing type for `newClassNames`

src/common/components/body-class-modifier.tsx
```diff
@@ -0,0 +1,44 @@
...
+        const { documentManipulator } = this.props;
+
+        const addedClassNames = newClassNames.filter(n => !oldClassNames.includes(n));
```
smoralesd: I know the `n` parameter can't be more local than this... but I still prefer to name this params more verbosely, should `className` be a better name here?

src/common/components/body-class-modifier.tsx
```diff
@@ -0,0 +1,44 @@
...
+        const removedClassNames = oldClassNames.filter(n => !newClassNames.includes(n));
+
+        // document.body may include classes not managed by this BodyClassModifier; leave them as-is
+        let allBodyClassNames = documentManipulator.bodyClassNames;
+        allBodyClassNames = allBodyClassNames.filter(n => !removedClassNames.includes(n));
+        allBodyClassNames.push(...addedClassNames);
+
+        documentManipulator.bodyClassNames = allBodyClassNames;
```
smoralesd: This sintax here has always been a little alien to me. basically because `bodyClassNames` is not a normal prop, it's basically a setter here, and the setter does have logic, so this line is not just an assign statement, something else is happening behind the scene (in this case is the call to `css` before actually assigning the class to the document's body).

Would it be easier to read and understand this is not just an assign statement if we have a proper function (instead of a setter)?

src/common/document-manipulator.ts
```diff
@@ -1,8 +1,17 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { css } from 'office-ui-fabric-react';
+
 export class DocumentManipulator {
     constructor(private document: Document) {}
 
+    public get bodyClassNames(): string[] {
+        return this.document.body.className.split(' ');
+    }
```
smoralesd: missing new line between the methods here

src/electron/views/root-container/components/platform-body-class-modifier.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export type PlatformBodyClassModifierProps = {
+    deps: PlatformBodyClassModifierDeps;
+};
+
+export const PlatformBodyClassModifier = NamedFC<PlatformBodyClassModifierProps>(
+    'PlatformBodyClassModifier',
+    props => {
+        if (props.deps.platformInfo.isMac()) {
```
smoralesd: does it make sense to reverse this conditional, make the clause guard the one that return null? Doing so makes the "then" block a one-liner which is always easier to read.

src/tests/unit/tests/common/components/body-class-modifier.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    });
+
+    it('should add classNames to documentManipulator.bodyClassNames when mounted', () => {
+        shallow(<BodyClassModifier classNames={['test-class']} documentManipulator={documentManipulator} />).render();
+        expect(documentManipulator.bodyClassNames).toEqual(['test-class']);
+    });
+
+    it('should update classNames in documentManipulator.bodyClassNames when props are updated', () => {
+        const testSubject = shallow(<BodyClassModifier classNames={['from-original-props']} documentManipulator={documentManipulator} />);
+        testSubject.render();
```
smoralesd: question: why do we need to explicitly call `render`? Isn't `shallow` "rendering" the component anyway?

src/tests/unit/tests/electron/views/root-container/components/platform-body-class-modifier.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+        expect(testSubject.getElement()).toBeNull();
+    });
+
+    it('Renders per snapshot for mac', () => {
+        mockPlatformInfo.setup(m => m.isMac()).returns(() => true);
+
+        const testSubject = shallow(<PlatformBodyClassModifier deps={mockDeps} />);
+
+        expect(testSubject.debug()).toMatchSnapshot();
```
smoralesd: the snapshot for this test seems interesting to me... I'm guessing is because the use of `debug()` on this line. Does using `getElement()` get us `documentManipulator={...}` (or similar)?? Is that the reason you used `debug()` instead of `getElement()`?

src/common/components/body-class-modifier.tsx
dbjorge: I agree that it feels a little odd; it's oddness we already had anyway (just indirectly via Helmet, before), but it's definitely a little strange-feeling for a react component to be side-effect-only

src/common/components/body-class-modifier.tsx
dbjorge: I'm glad to hear this comment, I debated exactly this question as I was writing it; will switch to the setter method

src/tests/unit/tests/common/components/body-class-modifier.test.tsx
dbjorge: Good catch, this is leftover from when I was misunderstanding the lifecycle model and had implemented `componentDidUpdate` without `componentDidMount`. Removed.

src/common/components/body-class-modifier.tsx
dbjorge: I don't feel strongly about it either way, switched

Issue #2043, opened 2020-01-24
Text changes in Electron app


ferBonnin:
## Describe the bug

Two text changes needed in Electron app:
- in port number screen, should be Android not android (uppercase A)
- in Congratulations message, when no automated checks errors are found. Need to add further clarification that more testing is required; suggestion:
"No failed automated checks were found.  Continue investigating your app's accessibility compliance through manual testing"

## Expected behavior
New text changes are added 

## Additional context
Uppercase A is required per branding guidelines.
Adding more clarification per user's feedback.

ferBonnin:
Marking as ready for triage so we can review text suggestions before changing them

ferBonnin:
per conversation. For web, similar text (note that the differences are bolded)
No failed automated checks were found. Continue investigating your **website's** accessibility compliance through manual testing **using Tab stops and Assessment**"

karanbirsingh:
This text is still not done:

`No failed automated checks were found. Continue investigating your website's accessibility compliance through manual testing using Tab stops and Assessment` because it requires different electron & web behavior. Getting the other static parts in now

ferBonnin:
Per conversation with Peter, let's timebox getting the two different texts for Web and Unified to a day.
Android: "No failed automated checks were found. Continue investigating your app's accessibility compliance through manual testing"

Web: "No failed automated checks were found. Continue investigating your website's accessibility compliance through manual testing using Tab stops and Assessment"

 If there is something too complex to do this, talk with @ferBonnin for a generic text for both

smoralesd:
@ferBonnin this message also shows up on the automated checks report. Should we use the "web" version of this message on the report too?

ferBonnin:
@smoralesd yes, please

DaveTryon:
@ferBonnin, this has been merged into master and should soon be present in canary releases of both AIWeb and the unified client. Please regress at your convenience.

ferBonnin:
validated in unified and web canary. looks good.

PR #2188, opened 2020-02-20
fix: Add platform-specific congrats messages

DaveTryon:
#### Description of changes

As called out in #2043, we want to update the messaging after automated tests find no issues. This fix updates the congratulations message in Web, Android, and reports. It adds a deps value that gets passed into the class that generates the message, so that the default value can be overridden. I went with the override approach because the report package doesn't currently pass in the necessary props. There's probably a future PR in the reports package to enable this, but since there was a stated goal to time-box this fix, I used default + override.

Part of the fix was updating the word-break property to break properly on word boundaries.

#### Pull request checklist

- [x] Addresses an existing issue: #2043 
- [x] Ran `yarn fastpass`

Congrats message in web automated results:
image

Congrats message in generated reports:
image

Congrats message in unified (clipped to display just the message):
image


src/electron/views/renderer-initializer.ts
```diff
@@ -292,6 +292,8 @@ getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch).then(
             unifiedResultToIssueFilingDataConverter: null, // we don't support issue filing right now
             windowUtils: null,
             setFocusVisibility,
+            customCongratsMessage:
+                "No failed automated checks were found. Continue investigating your app's accessibility compliance through manual testing",
```
smoralesd: is there a missing `.` at the end of the string?

src/reports/components/report-sections/no-failed-instances-congrats.tsx
```diff
@@ -5,16 +5,30 @@ import * as React from 'react';
...
+export type NoFailedInstancesCongratsProps = {
+    deps: NoFailedInstancesCongratsDeps;
+};
+
+export const NoFailedInstancesCongrats = NamedFC<NoFailedInstancesCongratsProps>(
+    'NoFailedInstancesCongrats',
+    props => {
+        const message = props.deps.customCongratsMessage
+            ? props.deps.customCongratsMessage
+            : "No failed automated checks were found. Continue investigating your website's accessibility compliance through manual testing using Tab stops and Assessment in Accessibility Insights for Web";
```
smoralesd: would it be better to pass this text (the "coda" of the congrats message for AI-Web) as a deps the same as we do for AI-Android?
Doing so makes all the cases follow the same pattern, which is easier to understand.

DaveTryon: The reports always pass in a null value, so it's not really feasible to use a code there. Given the messages, I'm not sure how much would be the "base" and how much would be the "coda". My inclination is to stick with this for now, then (after we update the reports to pass in a value), follow a pattern where we have no default and the string is completely specified by the caller. Happy to chat more about this if you want.

src/electron/views/renderer-initializer.ts
DaveTryon: Yes. I'l add it to both strings.

src/reports/components/report-sections/no-failed-instances-congrats.tsx
```diff
@@ -5,16 +5,30 @@ import * as React from 'react';
...
+};
+
+export type NoFailedInstancesCongratsProps = {
+    deps: NoFailedInstancesCongratsDeps;
+};
+
+export const NoFailedInstancesCongrats = NamedFC<NoFailedInstancesCongratsProps>(
+    'NoFailedInstancesCongrats',
+    props => {
+        const message = props.deps.customCongratsMessage
```
karanbirsingh: you might be able to use ?? here instead of ternary

DaveTryon: Will do this when I fix the merge conflict

PR #2199, opened 2020-02-21
feat(android-settings): "immutable" state on base store

smoralesd:
#### Description of changes

Currently, our `Store`s use mutable state object, i.e. every time an action causes a state to change, the `Store` is only updating values inside the state, the object that holds the state is fundamentally the same.

This doesn't seem to have any interesting effect on AI-Web basically because of the messaging infrastructure we use to communicate between the background (where stores live) and the "front-end" side, where we use `StoreProxy`, where we have this line of code to update the state:
```ts
this.state = message.payload;
```
Note that the `message` is being marshalled by chrome messaging API, meaning, it has the same values that the state we are sending from the background but it's not the same object (in the C world, basically they are 2 different references or pointers that happen to hold the same values).

Now, this happen to reveal a bug on AI-Android, around how we handle default/high contrast themes. On `ThemeInner` component, we have this couple of lines inside `componentDidUpdate`:
```ts
const enableHighContrastCurr = this.isHighContrastEnabled(this.props);
const enableHighContrastPrev = this.isHighContrastEnabled(prevProps);
```
then, base on the current and previous state of the high contrast flag we decide if we need to call `loadTheme` or not (`loadTheme` is the one that allows to update office fabric components styles).
This is fine on AI-Web because the `Store` <-> `StoreProxy` structure we have makes the previous and current state a different object every time we get an update from the background; this is **not** fine on AI-Android because we don't use `StoreProxy`, we don't have a background, basically, our "source of truth" is in the same page and we only use `Store` instances; this plus the fact our `Store` uses a mutable state make the comparison we use on `ThemeInner` to always be `false`  (basically previous and current state is the same all the time) which prevent us from calling `loadTheme`.

In order to solve this, we could make sure our `Store` uses an immutable `state` like redux does, i.e, every time we update the state we create a new object with all the same values plus the updated ones (that's kind of easy to do with the spread operator: `const newState = {...this.state}` ), but doing so in `BaseStoreImpl` or `Store` seems risky at this point.

Considering the last point, the approach I chose to tackle this is add a sort of decorator or processor that we use as part of the `getState` logic inside `BaseStoreImpl`, going from this code:
```ts
public getState(): TState {
    return this.state;
}
```
to this code:
```ts
public getState(): TState {
    return this.getStateProcessor(this.state);
}
````
By default, we set `getStateProcessor` to be the identity function (from `lodash`) and make the constructor param to be optional, this way, the store instances we use on AI-Web are left unchanged (both, on behavior and on how we instanciate them); but in the case of AI-Android we use lodash's `cloneDeep` as the `getStateProcessor`. This way, every time there is an update on the `Store` and a component ask for the `state` we get a new copy of the state, effectively making the previous state and current state different objects (at least in the Component level)

#### Pull request checklist
- [x] Addresses an existing issue: part of WI # 1677805
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/stores/base-store-impl.ts
+++ b/src/background/stores/base-store-impl.ts
@@ -1,14 +1,19 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { identity } from 'lodash';
 import { BaseStore } from '../../common/base-store';
 import { Store } from '../../common/flux/store';
 import { StoreNames } from '../../common/stores/store-names';
 
+export type StateProcessor<State> = (state: State) => State;
+
 export abstract class BaseStoreImpl<TState> extends Store implements BaseStore<TState> {
-    private storeName: StoreNames;
     protected state: TState;
 
-    constructor(storeName: StoreNames) {
+    constructor(
+        private readonly storeName: StoreNames,
+        private readonly getStateProcessor: StateProcessor<TState> = identity,
+    ) {
         super();
         this.storeName = storeName;
         this.onGetCurrentState = this.onGetCurrentState.bind(this);
@@ -28,7 +33,7 @@ export abstract class BaseStoreImpl<TState> extends Store implements BaseStore<T
     }
 
     public getState(): TState {
-        return this.state;
+        return this.getStateProcessor(this.state);
     }
 
     protected onGetCurrentState(): void {
--- a/src/background/stores/global/user-configuration-store.ts
+++ b/src/background/stores/global/user-configuration-store.ts
@@ -1,10 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { IndexedDBAPI } from 'common/indexedDB/indexedDB';
+import { StoreNames } from 'common/stores/store-names';
+import { UserConfigurationStoreData } from 'common/types/store-data/user-configuration-store';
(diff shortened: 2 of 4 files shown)


dbjorge:
It worries me a little to think of the cloning behavior being different between web and unified; it feels like it would be easy for it to result in one or the other being coded to rely on the behavior from one or the other.

Maybe cloning every time would be a safer way to go here? The performance implications don't feel very comfortable, but given that in web our store updates all already involve a round trip of inter-process communication, it's hard for me to imagine an extra `cloneDeep` making an appreciable difference; I'm expecting the IPC to be at least an order of magnitude or two slower than that anyway, and it hasn't proven to be a big problem in web.

dbjorge:
Would love to hear @peterdur and @waabid chime in :)

smoralesd:
pinging @markreay for his opinion too

markreay:
I like the approach and I also agree with Dan's comments generally. 

I think the key here is effective naming and also to make sure that the electron is not the exceptional case. 

I would prefer if we were explicit about the `stateProcessor` in all cases and ensure that we explicitly declare either `identity` or `cloneDeep` in every case. We have no real idea what the future holds and it's hard to know for sure which of these is the "default" case. (If we really felt there should be a default, I would recommend cloneDeep as the "safer" choice.)

smoralesd:
Abandoning this PR in favor of a simpler, more scoped approach that will only change the `UserConfigurationStore`

PR #2202, opened 2020-02-21
fix: Revise Aria label for device port number

DaveTryon:
#### Description of changes

Bug 1676735 calls out that the port number dialog lacks context when accessed via NVDA or JAWS. The suggested change from that bug is to update the Aria label from "Port label" to "Android Device port number e.g. 12345"

Tested with both NVDA and JAWS

#### Pull request checklist

- [x] Addresses an existing issue: 1676735
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/views/device-connect-view/components/device-connect-port-entry.tsx
+++ b/src/electron/views/device-connect-view/components/device-connect-port-entry.tsx
@@ -44,7 +44,7 @@ export class DeviceConnectPortEntry extends React.Component<
                 <h3>Android device port number</h3>
                 <MaskedTextField
                     data-automation-id={deviceConnectPortNumberFieldAutomationId}
-                    ariaLabel="Android Device port number e.g. 12345"
+                    label="Android device port number"
                     onChange={this.onPortTextChanged}
                     placeholder="Ex: 12345"
                     className={styles.portNumberField}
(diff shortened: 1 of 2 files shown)


DaveTryon:
@smoralesd, looking at your comments in bug 1676767, it seems like you may have had a different fix in mind for this?

src/electron/views/device-connect-view/components/device-connect-port-entry.tsx
```diff
@@ -44,7 +44,7 @@ export class DeviceConnectPortEntry extends React.Component<
                 <h3>Android device port number</h3>
                 <MaskedTextField
                     data-automation-id={deviceConnectPortNumberFieldAutomationId}
-                    ariaLabel="Port number"
+                    ariaLabel="Android Device port number e.g. 12345"
```
smoralesd: should the aria label match the place holder a little bit more? 
The aria label is using `e.g.` while the place holder uses `Ex:`

Also, this is kind of related to what we were talking offline: if we convert the `<h3>` to a label and use the `for` attribute to associate the label with this input, I think the AT will use that label to announce the input, thus, if we keep this aria label, there is a chance of duplicated text being announced (I'm not 100% sure about that but I think the possibility is real)

That being said, I think this issue (the aria label) and the "let's not use `<h3>` issue should be tackle togheter.

Thoughts?

DaveTryon: Makes sense to me. I'll see what I can come up with.

DaveTryon: I dug into the label question a little more, but there's a problem. We're using an office fabric MaskedTextField, which contains 2 divs inside it. HTML doesn't allow a label to apply to something inside a div, so there's no way to bind the current h3 to the edit unless we completely replace the MaskedTextField (which I don't think we want to do at this point). As it turns out, though, MaskedTextField has a label property that does exactly what we want it to do in both NVDA and JAWS.

DaveTryon:
I spoke too soon. Closing this for now, will reopen once I've worked out some details

PR #2204, opened 2020-02-21
feat(android-settings): respect system high contrast settings

dbjorge:
#### Description of changes

This PR implements a `NativeHighContrastModeListener` that hooks into [Electron's nativeTheme APIs](https://www.electronjs.org/docs/api/native-theme) to respond to system high contrast mode notifications.

This PR implements what our feature crew has discussed as "MVP" functionality:
* On app startup, the user config store's high contrast settings are overridden by the system settings
* When the system settings change, the user config store's settings are overridden
* The user can toggle high contrast mode in the settings panel too, but any settings there are ephemeral and will be overridden at next app startup or when the system settings change.

A later user story tracks improving this to remember the app setting separately from the system setting, but that is out of scope here.

Since `nativeTheme` is only accessible from an Electron app's main process, but we need the information in our renderer process, most of this PR is spent on the communication layer between the two processes. Per guidance from the Electron/security folks, we're using `ipcMain`/`ipcRenderer` to achieve this (notably, avoiding the `remote` module), adapting the existing `Interpreter`/`Dispatcher`/`Message` infrastructure:
* The `NativeHighContrastModeListener` uses the existing `UserConfigMessageCreator` to send `highContrastEnabled` messages
* The main process maintains an instance of a new `IpcMessageDispatcher` class which routes messages from the message creator to `mainWindow.webContext.send`
* The renderer process maintains an instance of a new `IpcMessageReceiver` class which routes messages received from the `IpcMessageDispatcher` via Electron's `ipcRenderer` to the same `Interpreter` used by messages created within the renderer process.
* To avoid the main process sending messages before the renderer process is ready to receive them, the renderer process now sends an "initialized" message at the end of its initializer. Since the code to do this using the native `ipcRenderer.send('message')` + `ipcMain.on('message')` was already as minimal as I could imagine the registration being in the main/renderer initializers, I didn't wrap them in any additional abstractions, though I could imagine wanting to do so in a later PR if we ever need to send more complex messages in the other direction.

animation showing enabling windows high contrast mode affecting the device connect view

#### Pull request checklist

- [x] Addresses an existing issue: 1678719
- [x] Ran `yarn fastpass`


src/electron/ipc/ipc-message-dispatcher.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public dispatchType(messageType: string): void {
+        throw new Error('Method not implemented (not yet required for any main process events).');
+    }
+    public sendTelemetry(eventName: string, eventData: TelemetryData): void {
+        throw new Error('Method not implemented (not yet required for any main process events).');
+    }
+
+    private messageBacklog: Message[] = [];
```
smoralesd: It's odd that we have instance properties in the middle of the class. Should they be on the top?

src/electron/ipc/ipc-message-dispatcher.ts
```diff
@@ -0,0 +1,52 @@
...
+    private registeredSinks: IpcMessageSink[] = [];
+
+    public registerMessageSink = (sink: IpcMessageSink): void => {
+        this.registeredSinks.push(sink);
+
+        if (this.registeredSinks.length === 1) {
```
smoralesd: this `=== 1` here feels a little odd. I guess I kind of agree with your first gut reaction on having a queue-like behavior here: it would be better not to have it; that does raise the small concern of how do we ensure order of execution between instanciating this class and sending message. Then again, we own all the initializer so we can ensure that order.

src/electron/ipc/ipc-message-receiver.ts
```diff
@@ -0,0 +1,37 @@
...
+        private readonly underlyingIpcRenderer: IpcRenderer,
+        private readonly logger: Logger,
+    ) {}
+
+    public initialize(): void {
+        this.underlyingIpcRenderer.on(IPC_MESSAGE_CHANNEL_NAME, this.onMessageReceived);
+    }
+
+    private onMessageReceived = (event: IpcRendererEvent, ...args: any[]) => {
+        if (event.senderId !== 0) {
```
smoralesd: I'm curious: when does this happen?
This log here remind me of the log we added on AI-Web when we receive a message for a tab we are not tracking, that log became a spammer right away.

src/electron/main/native-high-contrast-mode-listener.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { UserConfigMessageCreator } from 'common/message-creators/user-config-message-creator';
+import { NativeTheme } from 'electron';
+
+export class NativeHighContrastModeListener {
+    private lastHighContrastUpdate: boolean;
```
smoralesd: the combination of name + type seems weird to me. From the name (specially the **update**, I'd expected this to hold a timestamp (maybe); from the type, I'd expect this to hold a flag to signal if we already process the update (like sending a message so the whole UI change theme)

src/electron/ipc/ipc-message-dispatcher.ts
dbjorge: Going to remove the queue behavior and just control the instantiation order

src/electron/main/native-high-contrast-mode-listener.ts
dbjorge: Updated to `highContrastStateAtLastNotify`

src/electron/ipc/ipc-message-receiver.ts
dbjorge: Per electron docs, senderId is:

> The webContents.id that sent the message, you can call `event.sender.sendTo(event.senderId, ...)` to reply to the message, see `ipcRenderer.sendTo` for more information. This only applies to messages sent from a different renderer. Messages sent directly from the main process set `event.senderId` to 0.

That is, we don't currently expect it in any circumstance, since we don't have other processes sending messages besides the main process

dbjorge: (updated the error message to make this more clear)

Issue #1813, opened 2019-12-06
Cards UI: selecting failure doesnt move target page


ferBonnin:
## Describe the bug

cannot select a failure in the details view and have the target page move to the location.

**To Reproduce**

1. Go to https://www.washington.edu/accesscomputing/AU/before.html
2. run FastPass > automated checks with Insider build
3. Scroll to the bottom in the target page
4. in the details view, select any of the color contrast failures

## Expected behavior

Target page should scroll up and make the failure visible (in production, this works)

## Context (please complete the following information)

- OS Name & Version:  Windows 10
- AI-Web Version & Environment:  Insider 2.11.0
- Browser Version:  Chrome 78

## Additional context
During a demo received customer feedback that this behaviour would be very helpful.

PR #2206, opened 2020-02-21
fix: card selection/toggle causes the target page to scroll it into view 

waabid:
#### Description of changes

This feature was never built into the new cards view. As a result, this modifies the focus change handler to handle both the card selection case and the old visualization store (still used for assessments).

#### Pull request checklist

- [x] Addresses an existing issue: #1813 
- [x] Ran `yarn fastpass`


src/injected/focus-change-handler.ts
```diff
@@ -34,4 +36,29 @@ export class FocusChangeHandler {
         this.scrollingController.processRequest(scrollingMessage);
         this.targetPageActionMessageCreator.scrollRequested();
     };
+
+    private getTarget(storeData: TargetPageStoreData): string[] {
+        return (
+            storeData.visualizationStoreData.focusedTarget || this.getCardResultTarget(storeData)
```
smoralesd: would it be worth it to have a comment here to explain why we need to check `visualizationStoreData` and not only the `cardSelectionStoreData`?

src/injected/focus-change-handler.ts
```diff
@@ -34,4 +36,29 @@ export class FocusChangeHandler {
...
+        ) {
+            return null;
+        }
+
+        const focusedResult = storeData.unifiedScanResultStoreData.results.find(
+            result => result.uid === storeData.cardSelectionStoreData.focusedResultUid,
+        );
+
+        if (focusedResult == null) {
+            throw 'focused result was not found';
```
smoralesd: I think we avoid throwing `string`, it make for weird behavior. Throwing `new Error('focused result was not found')` should be fine.

waabid: Done!

PR #2216, opened 2020-02-24
fix: true-up electron-storage-adapter

smoralesd:
#### Description of changes

Currently, our `ElectronStorageAdapter`, which implements the `StorageAdapter` interface and serves as a "double" of our `ChromeAdapter` (around accessing local storage) wraps any object we want to store around a single key: `installationData`. 

We are only using the `ElectronStorageAdapter` to store `installationData` so the issue mention above cause the storage to write something like this:
```js
{
installationData: {
  installationData: {
      month: 1,
      year: 2020,
      installationId: <guid>
    }
  }
}
```
Notice the duplicated `installationData` key on the object.

This causes our `InstallDataGenerator` to always generate a new `installationId` (basically, because is checking for an existing `installationId` at the "wrong" level of the object.

This PR fixes this issue by properly and directly using the keys of the object pass to `setUserData` to make calls to the underlying indexedDB API. Get and remove API's have been update accordingly too.

**installationData from master (bugged)**
01 - bug storage adapter

**installationData from this branch (fixed)**
02 - fixed storage adapter

**Notes**
- Notoriously, there was no need to update `renderer-initializer` which instanciate `ElectronStorageAdapter`
- The fix is applied seemingly by the `InstallDataGenerator`
- I test this fix under a "update scenario". Basically, build and run AI-Android from master, check the `installationData` using dev tools (it's wrong) and then build and run AI-Android from this branch and check again. There is no error and telemetry is being send properly, keeping the same `installationId` in between app restarts.

#### Pull request checklist
- [x] Addresses an existing issue: yes, but there is not an issue on the repo
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/adapters/electron-storage-adapter.ts
+++ b/src/electron/adapters/electron-storage-adapter.ts
@@ -7,11 +7,12 @@ export class ElectronStorageAdapter implements StorageAdapter {
     constructor(private readonly indexedDBInstance: IndexedDBAPI) {}
 
     public async setUserData(items: Object): Promise<void> {
-        const setItemPromises = Object.keys(items).map(itemKey =>
-            this.indexedDBInstance.setItem(itemKey, items[itemKey]),
-        );
+        const keys = Object.keys(items);
 
-        await Promise.all(setItemPromises);
+        for (let index = 0; index < keys.length; index++) {
+            const key = keys[index];
+            await this.indexedDBInstance.setItem(key, items[key]);
+        }
     }
 
     public async getUserData(keys: string[]): Promise<{ [key: string]: any }> {
--- a/src/electron/adapters/electron-storage-adapter.ts
+++ b/src/electron/adapters/electron-storage-adapter.ts
@@ -1,30 +1,35 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { IndexedDBDataKeys } from 'background/IndexedDBDataKeys';
 import { StorageAdapter } from 'common/browser-adapters/storage-adapter';
 import { IndexedDBAPI } from 'common/indexedDB/indexedDB';
-import { pick } from 'lodash';
 
 export class ElectronStorageAdapter implements StorageAdapter {
     constructor(private readonly indexedDBInstance: IndexedDBAPI) {}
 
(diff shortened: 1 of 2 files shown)


src/electron/adapters/electron-storage-adapter.ts
```diff
@@ -1,30 +1,35 @@
...
 import { IndexedDBAPI } from 'common/indexedDB/indexedDB';
-import { pick } from 'lodash';
 
 export class ElectronStorageAdapter implements StorageAdapter {
     constructor(private readonly indexedDBInstance: IndexedDBAPI) {}
 
     public async setUserData(items: Object): Promise<void> {
-        await this.indexedDBInstance.setItem(IndexedDBDataKeys.installation, items);
+        const setItemPromises = Object.keys(items).map(itemKey =>
+            this.indexedDBInstance.setItem(itemKey, items[itemKey]),
```
dbjorge: I wasn't able to find a guarantee in the [docs](https://github.com/jakearchibald/idb-keyval#readme) for the underlying API we're using that it's okay to interleave `set` calls like this; maybe safer to await each setItem in turn 
rather than using `Promise.all`

src/tests/unit/tests/electron/adapters/electron-storage-adapter.test.ts
```diff
@@ -22,112 +15,78 @@ describe('ElectronStorageAdapter', () => {
...
 
-        it('gets data, two keys', async () => {
-            const keys = Object.keys(testData);
+            await expect(testSubject.setUserData(items)).rejects.toMatch(reason);
+        });
+    });
 
-            indexedDBInstanceMock.setup(db => db.getItem(IndexedDBDataKeys.installation)).returns(() => Promise.resolve(testData));
+    describe('getUserData', () => {
+        it.only.each`
```
dbjorge: Stray `only`

smoralesd: good catch. removed.

src/electron/adapters/electron-storage-adapter.ts
smoralesd: Good call. Updated.

PR #2232, opened 2020-02-25
fix: use 7:1 contrast ratio for primary button background in high-contrast mode

dbjorge:
#### Description of changes

Previously, in high contrast mode, we were using our default "primary theme color" for primary button colors, which in high contrast mode is `#0078d4` and looks like this (with a 4:1 contrast ratio against the near-black text/background):

screenshot of button with old darker/lower-contrast background color

With this change, we now use our high contrast palette's "selected items/checkboxes/toggles" color (`#38A9FF`) instead, which has a 7.141:1 ratio against the background/text:

screenshot of button with new lighter/higher-contrast background color

This PR removes a style in web's `common.scss` that already overrode the default office fabric value, but did so in a way that did not meet contrast requirements, such that web's dialogs use the same high-contrast palette; this gets an early start on one of the requirements in feature 1559424

#### Pull request checklist

- [x] Addresses an existing issue: part of 1559424
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/styles/common.scss
+++ b/src/common/styles/common.scss
@@ -22,14 +22,6 @@ button::after {
     position: absolute;
 }
 
-.ms-Button.ms-Button--primary:not(.is-disabled) {
-    background: $communication-primary;
-
-    &:hover {
-        background: $communication-shade-10;
-    }
-}
-
 .ms-fontColor-neutralPrimary,
 .ms-fontColor-neutralPrimary--hover:hover {
     color: $neutral-100 !important;
--- a/src/common/styles/default-theme-palette.ts
+++ b/src/common/styles/default-theme-palette.ts
@@ -26,4 +26,7 @@ export const DefaultThemePalette: IPartialTheme = {
         black: '#1d1d1d',
         white: '#ffffff',
     },
+    semanticColors: {
+        primaryButtonBackground: '#106ebe',
+    },
 };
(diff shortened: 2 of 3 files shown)


src/common/styles/common.scss
```diff
@@ -22,14 +22,6 @@ button::after {
     position: absolute;
 }
 
-.ms-Button.ms-Button--primary:not(.is-disabled) {
-    background: $communication-primary;
-
-    &:hover {
-        background: $communication-shade-10;
```
dbjorge: Dropping this hover style is intentional; it results in using a reasonable looking white hover color that Fer was happy with

src/common/styles/default-theme-palette.ts
```diff
@@ -26,4 +26,7 @@ export const DefaultThemePalette: IPartialTheme = {
         black: '#1d1d1d',
         white: '#ffffff',
     },
+    semanticColors: {
+        primaryButtonBackground: '#106ebe',
```
smoralesd: I'm confused... should this color be '#0078d4' ??

dbjorge: No, `#106ebe` is intentiona; per the latest figma for both the android app and the component figma for web, it's supposed to be communication-primary (like it was originally in web via the old `common.scss` stylee) in both places

PR #2240, opened 2020-02-27
fix: fit-and-finish + refactor for cards modifier buttons subcomponents

dbjorge:
#### Description of changes

This PR is part 1 of work to move the "Visual helper" toggle out of the main cards view and into the screenshot view in unified; it extracts the "Visual Helper" toggle and the "Expand/Collapse All" buttons into separate components, so the next PR can use them independently.

I've labeled this "fix" and not "refactor" because as part of extracting the styles, I fixed up some minor inconsistencies the "Visual helper" toggle had with its figma:

* Padding between expand/collapse and visual helper updated from 25px to 27px
* Font weight for the "Visual helper" label made normal instead of fabric's default bold (by creating a new `text-style-body-m` mixin to match the `body-m` text style in our typography figma)
* Distance between the visual helper label and toggle updated from 12px -> 8px

Both the unified and web figmas agree on the latter two, so I put it unilaterally in the visual helper toggle component. The two products will disagree in the next PR on the top/bottom margins required (this is why I made `VisualHelperToggle` accept a className prop)

Canary on left, dev with these changes on right:

screenshot showing graphics delta introduced by this PR

#### Pull request checklist

- [x] Addresses an existing issue: part of 1676714
- [x] Ran `yarn fastpass`


src/common/components/cards/cards-visualization-modifier-buttons.tsx
```diff
@@ -1,57 +1,38 @@
...
-    deps: CardsVisualizationModifierButtonsDeps;
-    visualHelperEnabled: boolean;
-    allCardsCollapsed: boolean;
-};
+export type CardsVisualizationModifierButtonsProps = ExpandCollapseAllButtonProps &
+    VisualHelperToggleProps;
 
-export const CardsVisualizationModifierButtons = NamedFC<CardsVisualizationModifierButtonsProps>(
-    'CardsVisualizationModifierButtons',
+export const ExpandCollapseOnlyModifierButtons = NamedFC<CardsVisualizationModifierButtonsProps>(
```
smoralesd: Should the file name be updated as well?

src/common/styles/fonts.scss
```diff
@@ -66,3 +66,10 @@ $semiBoldFontFamily: make-font-family($semiBoldFonts) !default;
     font-size: 17px;
     line-height: 24px;
 }
+
+@mixin text-style-body-m {
```
smoralesd: The 'body' in the name seems odd. I notice we already have `text-style-title-s` which makes me thing its for a title. Should this new `mixin` be called `text-style-label-m` instead?

dbjorge: The `body-m` naming here matches the names of the text styles in our typography figma

src/common/components/cards/cards-visualization-modifier-buttons.tsx
dbjorge: I left the filename as-is (matching the shared props type) because this file intentionally exports 2 different variants of `CardsVisualizaitonModifierButton`; maybe exporting a type by that name that both impls adhere to would be more clear?

dbjorge: Added a commit doing this

Issue #1739, opened 2019-11-23
export: provide hosted export targets


bnb:
## Is your feature request related to a problem? Please describe.
Nope.

## Describe the desired outcome
Currently, if I export I just get a bare HTML file. If I want to share this with peers, I need to publish it myself.

While I personally know of and have access to a dozen ways to publish a single HTML file rapidly, this isn't true of everyone who may use Accessibility Insights.

I'd love to see something like "Export to..." built-in to this project - more specifically, it would be cool to see "Export to..." options for Glitch, CodePen, GitHub Pages (if authed and a repo is selected?), and possibly even enable the community to contribute additional targets.

## Describe alternatives you've considered
The status quo is one alternative - it doesn't add any complexity, but does make exporting and sharing results harder.

An additional option would be an interactive CLI published to npm that requires you have npm and Node.js installed. Running something like `npx insights-publish ./path/to/insights/export.html` could be another option. This seems like potentially a significantly larger amount of work would be needed, though if you wanted to build built-in plugin support you could add something like `npx insights-publish --to=glitch  ./path/to/insights/export.html` where the module resolves `--to=glitch` as an additional module with a prefix like `insights-plugin-glitch` that it will install and that can hook into an extension API and publish as the author of `insights-plugin-glitch` defines.

## Additional context
None.

adarshrema:
Hello @bnb, thank you for using Accessibility Insights for Web and this feature request. Looks like this would support a workflow specific to your team/ project. We welcome any contributions from your team to make this happen.

We are interested in gauging how attractive this feature might be for the larger community. However, given our current priorities and information we have about this feature request so far, it is not likely that we will be able to pick this up anytime soon.

juchampa:
Hi @bnb ,

We are happy to inform you this issue has been addressed through a public contribution to our open source repository [here](https://github.com/microsoft/accessibility-insights-web/pull/2252), by @ sean-beard. This feature is now under a public flag, so feel free to use this feature and contribute to it if you'd like.

bnb:
Thought I had responded to this initially, but apparently not.

> Hello @bnb, thank you for using Accessibility Insights for Web and this feature request. Looks like this would support a workflow specific to your team/ project. We welcome any contributions from your team to make this happen.

[...]

PR #2252, opened 2020-02-28
feat(report): support ability to export to CodePen

sean-beard:
#### Description of changes
These changes add the option to export reports to CodePen in addition to an HTML file. There is a [relevant open issue](https://github.com/microsoft/accessibility-insights-web/issues/1739) detailing a need for this type of functionality. 

I wasn't quite sure what the design should look like for this so I've went ahead and used a split `PrimaryButton` from `office-ui-fabric-react` in the `ExportDialog ` component but I'm definitely open to alternative approaches!

a11y-report-export

I was also considering splitting out the HTML and CSS before sending the payload to CodePen but I figured I'd get some feedback on the changes within this PR before moving forward with more in-depth functionality.

#### Pull request checklist

- [x] Addresses an existing issue: #1739
- [x] Ran `yarn fastpass`


smoralesd:
Hey @sean-beard I noticed you pushed some new changes. Should I go ahead and review them?

sean-beard:
Hey @smoralesd, sure, feel free to review! 

Just FYI - I'm still looking into our options for removing the `document` global dependency. I have some working code now but I'm not too satisfied with it, yet.

src/DetailsView/components/export-dialog.tsx
```diff
@@ -73,9 +83,36 @@ export const ExportDialog = NamedFC<ExportDialogProps>('ExportDialog', props =>
...
+                <PrimaryButton
+                    text="Export"
+                    split
+                    splitButtonAriaLabel="Export HTML to any of these format options"
+                    aria-roledescription="split button"
+                    menuProps={{
+                        items: ReportExportServiceProviderImpl.all().map(service => ({
+                            key: service.key,
+                            text: service.displayName,
+                            onClick: (e: React.MouseEvent<HTMLButtonElement>) => {
```
smoralesd: This `onClick` function does not seem to be covered by unit test.

src/DetailsView/components/export-dialog.tsx
```diff
@@ -73,9 +83,36 @@ export const ExportDialog = NamedFC<ExportDialogProps>('ExportDialog', props =>
...
+                    }}
+                    download={props.fileName}
+                    href={fileURL}
+                />
+                {ExportForm && (
+                    <ExportForm
+                        fileName={props.fileName}
+                        description={props.description}
+                        html={props.html}
+                        onSubmit={() => {
```
smoralesd: The `onSubmit` function here does not seem to be covered by unit test

src/DetailsView/components/export-dialog.tsx
```diff
@@ -27,18 +29,24 @@ export interface ExportDialogDeps {
 }
 
 export const ExportDialog = NamedFC<ExportDialogProps>('ExportDialog', props => {
+    const [format, setFormat] = React.useState<ExportFormat | null>(null);
```
smoralesd: @waabid @dbjorge This is the first time we'll be introducing a react hook.
I'm not against this but I do have concerns, specially the fact that this is a global dependency and how to properly test the state changes (I'm guessing the testing part is a solved problem at this point but I may be wrong).
Any thoughts?

src/report-export/services/codepen-report-export-service.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import {
+    ExportFormat,
+    ExportFormProps,
+    ReportExportService,
+} from 'report-export/types/report-export-service';
+
+const CodePenReportExportServiceKey: ExportFormat = 'codepen';
+
+const exportForm = NamedFC<ExportFormProps>(
```
smoralesd: The `exportForm` seems interesting enough that we should have unit tests for it.
The use of react hooks calls my attention here: currently on the repo, we don't use hooks so we don't have any example on how to unit tests this. It would be a very good contribution to write the first unit test around react hooks! If that seems to convoluted, there is always the possibility to go back to a more "traditional" react component so we can follow other examples in this repo on how to test it.

src/DetailsView/components/export-dialog.tsx
```diff
@@ -48,6 +56,8 @@ export const ExportDialog = NamedFC<ExportDialogProps>('ExportDialog', props =>
     };
 
     const fileURL = props.deps.fileURLProvider.provideURL([props.html], 'text/html');
+    const exportService = ReportExportServiceProviderImpl.forKey(format);
```
smoralesd: `ReportExportServicePRoviderImpl` is being use as a global dependency. Among other unwanted side effects, this forces the unit test to not be properly isolated (by depending on a concrete class/instance/implementation that's not the unit being tested).
One way to solve this is to follow the "`deps` pattern" we have across the repo:
Add a new property under `ExportDialogDeps` where we can pass an instance of the `ReportExportServiceProvider`. For the extension itself, this instance will be `ReportExportServiceProviderImpl` but for the unit test it can be a mock which does not need to change every time `ReportExportServicePRoviderImpl` changes, making the unit test properly isolated of other implementation details.

src/DetailsView/components/export-dialog.tsx
dbjorge: I don't have a particularly strong opinion on hooks vs class components as concepts, but I think it's better for maintainability to try to be consistent within a given codebase; I would lean towards using a class component instead for consistency's sake.

Issue #757, opened 2019-06-10
High Contrast mode: "Pass unmarked instances" button does not have enough contrast between 'active' and 'disabled' states


iamrafan:
## Describe the bug

It is tough to tell if the "Pass unmarked instances" button in assisted manual tests is active or disabled when in high contrast mode. The button is not having enough contrast across the states 

**To Reproduce**

1. Run assessment on any page
2. Enable high contrast mode from settings panel
3. Navigate to landmarks assessment -> 1. landmark roles (test step)
4. Click the "Pass unmarked instances" button
5. The button goes into disabled state
6. Observe that there is not much contrast between the active and disabled states

## Expected behavior

Users should be able to visually distinguish between active and disabled states 

## Screenshots

- Active state
![image](https://user-images.githubusercontent.com/22068434/59190573-eac5ca00-8b31-11e9-956b-325975add2a0.png)

- Disabled state
![image](https://user-images.githubusercontent.com/22068434/59190744-6162c780-8b32-11e9-97d4-b6fbcd8c47be.png)

## Extension

- Accessibility Insights for Web - prod
- Browser Version: Chrome 75.0.3770.80
- Target Page: Any page

## Did you search for similar existing issues?
Yes

ferBonnin:
@cheade Peter and I agree this is an issue, could you suggest other color?

cheade:
Hi guys!

![image](https://user-images.githubusercontent.com/40274682/59223675-5d11cb00-8b81-11e9-80b7-b087928ff4b0.png)

This purple should be in our high color contrast palette as a disabled color. 

Something else I've noticed in default windows high contrast apps, the actionable command buttons in high-contrast are outlined in white to further help define that it's an action, in lieu of the subtly that is removed in high contrast. (for active buttons)

karanbirsingh:
It seems to me the ActionButton OfficeFabric component does not support semantic styles yet for its disabled text color. I filed an issue here to validate this: https://github.com/OfficeDev/office-ui-fabric-react/issues/9418

If the button can use the `disabledText` semantic slot instead, then we'll just need to update OF once the fix is in. Otherwise we can look into workarounds on our end.

smoralesd:
@cheade I have a couple of questions:
- should we use #C285FF as the text color? or the background color?
- should the color be applied to all the buttons? That sounds like it'll need proper validation throughout the extension

cheade:
[...]

PR #2273, opened 2020-03-03
feat(compliance-contrast): use 3:1 contrast text for disabled command buttons in high contrast

dbjorge:
### Description of changes

This PR introduces a new `InsightsCommandButton` control to match the "Command button" in our [style figma](https://www.figma.com/file/Wj4Ggf6GGQBQkiDIaHfXRX2B/Accessibility-Insights%3A-Styles?node-id=1%3A27). I gave it a new `controls` folder because I expect it to be reasonable to use a similar structure for the other controls present in that Figma that we want to enforce specific insights styles for.

#757 only calls out the "Pass unmarked instances" button as failing contrast requirements in the disabled state; I audited all the other places we use `<ActionButton>`s for other cases with disable states and found that the same issue also applies to the FastPass start over button while a scan is in progress, so I fixed it while I was here.

For the initial PR, I kept its usage targeted to the specific use case we need to override the default fabric `ActionButton` styles for (the disabled text color, per #757); there are ~9 other places where we currently use an `ActionButton` that would be reasonable to replace with this new control, but I wanted to keep this PR small, so leaving them for a separate PR.

#### FastPass Start over button:
**Before**:
fastpass-start-over-before

**After**:
fastpass-start-over-after

#### Disabled "Pass all unmarked instances" button:
**Before**:
pass-all-unmarked-before

**After:
pass-all-unmarked-after

#### Pull request checklist

- [x] Addresses an existing issue: #757, 1686479
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/DetailsView/components/assessment-instance-table.test.tsx
+++ b/src/tests/unit/tests/DetailsView/components/assessment-instance-table.test.tsx
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { ActionButton } from 'office-ui-fabric-react';
 import { CheckboxVisibility, ConstrainMode, DetailsList, IColumn } from 'office-ui-fabric-react';
 import { Spinner, SpinnerSize } from 'office-ui-fabric-react';
 import * as React from 'react';
@@ -12,6 +11,7 @@ import {
     IGetMessageGenerator,
     IMessageGenerator,
 } from 'assessments/assessment-default-message-generator';
+import { InsightsCommandButton } from 'common/components/controls/insights-command-button';
 import { ManualTestStatus } from '../../../../../common/types/manual-test-status';
 import { AssessmentResultType, GeneratedAssessmentInstance } from '../../../../../common/types/store-data/assessment-result-data';
 import {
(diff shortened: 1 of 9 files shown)


smoralesd:
On the **Start over** button, there is a mismatch on the colors for the active state (when the user presses the button, without releasing the "click"): on the active state, the icon is white while the text is blue. I think we had the same on AI-Android but it should be fixed now. Do we care enough about that (maybe that can be address on a different PR)

src/tests/unit/tests/DetailsView/components/assessment-instance-table.test.tsx
```diff
@@ -108,9 +108,13 @@ describe('AssessmentInstanceTableTest', () => {
...
             <div>
-                <ActionButton iconProps={{ iconName: 'skypeCheck' }} onClick={testObject.getOnPassUnmarkedInstances()} disabled={true}>
+                <InsightsCommandButton
+                    iconProps={{ iconName: 'skypeCheck' }}
+                    onClick={testObject.getOnPassUnmarkedInstances()}
+                    disabled={true}
+                >
                     Pass unmarked instances
-                </ActionButton>
+                </InsightsCommandButton>
```
smoralesd: is it worth it to use this opportunity to convert this tests to use enzyme and snapshots?
The `TestableAssessmentInstanceTable` always gives me an itch.

dbjorge: I've done this, but the change to the tests is very substantial, so I'd rather put it in its own PR (I'll send that out after this one is merged)

dbjorge:
> On the **Start over** button, there is a mismatch on the colors for the active state (when the user presses the button, without releasing the "click"): on the active state, the icon is white while the text is blue. I think we had the same on AI-Android but it should be fixed now. Do we care enough about that (maybe that can be address on a different PR)

I think it would be good for us to address this, but per Fer, we want to keep the work for the immediate feature strictly scoped to required contrast fixes, so leaving it for a separate PR

PR #2292, opened 2020-03-04
feat(android-settings): make app HC setting persist across system HC toggling

dbjorge:
#### Description of changes

Before this PR, the interaction between the user and system high contrast states was that the native state overwrote the app setting on app startup and whenever the system state was toggled, and the user setting in our settings panel was strictly transient, forgotten with each new app launch.

With this PR, we now remember the last high contrast mode setting the user selected in the app across system high contrast mode switches. We still switch to high contrast mode if the system switches to that mode or if the app is started with the system in high contrast mode, but now if you switch the system out of high contrast mode (or start the app with the system not in high contrast mode), we will remember whatever setting you previously had set in the Settings panel and return to it.

In particular, if you toggle into high contrast mode explicitly in the settings panel, we will now "remember" that even if you toggle the system into and out of HC mode.

This behavior is based on the behavior of Team's theme selector.

Most of the code change here is in:

* Separating the wiring of the original "SetHighContrastMode" message (which previously we re-used for native contrast mode changes) to give native notifications a separate message
* Adding a new `lastSelectedHighContrast` property to `UserConfigurationStoreState` and updating all the tests that initialize test versions of that store state type

The interesting new logic to focus review attention on are in `src/background/stores/global/user-configuration-store.ts` and its associated new tests (@smoralesd / @JGibson2019 , this is unchanged from the version I mentioned in our teams chat earlier) 

Before (app state not persisted across system state being toggled):
animation showing app state not persisting

After (app state persists after toggling system state):
animation showing app state persisting

#### Pull request checklist

- [x] Addresses an existing issue: 1678723
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/stores/global/user-configuration-store.ts
+++ b/src/background/stores/global/user-configuration-store.ts
@@ -9,6 +9,7 @@ import {
     SetHighContrastModePayload,
     SetIssueFilingServicePayload,
     SetIssueFilingServicePropertyPayload,
+    SetNativeHighContrastModePayload,
 } from '../../actions/action-payloads';
 import { UserConfigurationActions } from '../../actions/user-configuration-actions';
 import { IndexedDBDataKeys } from '../../IndexedDBDataKeys';
@@ -19,6 +20,7 @@ export class UserConfigurationStore extends BaseStoreImpl<UserConfigurationStore
         isFirstTime: true,
         enableTelemetry: false,
         enableHighContrast: false,
+        lastSelectedHighContrast: false,
         bugService: 'none',
         bugServicePropertiesMap: {},
     };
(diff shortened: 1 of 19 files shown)


src/background/stores/global/user-configuration-store.ts
```diff
@@ -66,6 +71,14 @@ export class UserConfigurationStore extends BaseStoreImpl<UserConfigurationStore
 
     private onSetHighContrastMode = (payload: SetHighContrastModePayload): void => {
         this.state.enableHighContrast = payload.enableHighContrast;
+        this.state.lastSelectedHighContrast = payload.enableHighContrast;
+        this.saveAndEmitChanged();
+    };
+
+    private onSetNativeHighContrastMode = (payload: SetNativeHighContrastModePayload): void => {
+        this.state.enableHighContrast = payload.enableHighContrast
```
waabid: can be swapped for:

`this.state.enableHighContrast = payload.enableHighContrast || this.state.lastSelectedHighContrast`

If you're concerned about it not existing or something like that (shouldn't happen under TS but regardless), you can use !! in front.

dbjorge: I had it that way at first, but I kept having to look it over to convince myself it felt correct; I felt like this way was a little bit easier to read

PR #2297, opened 2020-03-04
feat(android-settings): add e2e tests for the settings panel

smoralesd:
#### Description of changes

Adding e2e test for the settings panel on AI-Android, including accessibility issues scan in default and high contrast theme.

The general strategy was to bring the same scenarios on the settings panel e2e test from AI-Web but translated to AI-Android.

**Note**
We have the idea to update the app controller to use the user config controller instead of hitting buttons in the UI for dismissing the first time telemetry dialog. This will lead to some changes to other e2e as well as adding a couple more e2e test. I left that for follow up PR to keep this one smaller.

#### Pull request checklist
- [x] Addresses an existing issue: part of WI # 1678715
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/electron/common/element-identifiers/automated-checks-view-selectors.ts
+++ b/src/tests/electron/common/element-identifiers/automated-checks-view-selectors.ts
@@ -5,6 +5,7 @@ import { instanceCardAutomationId } from 'common/components/cards/instance-detai
 import { ruleContentAutomationId } from 'common/components/cards/instance-details-group';
 import { ruleGroupAutomationId } from 'common/components/cards/rules-with-instances';
 import { automatedChecksViewAutomationId } from 'electron/views/automated-checks/automated-checks-view';
+import { commandButtonSettingsId } from 'electron/views/automated-checks/components/command-bar';
 import { highlightBoxAutomationId } from 'electron/views/screenshot/highlight-box';
 import { screenshotImageAutomationId } from 'electron/views/screenshot/screenshot';
 import { screenshotViewAutomationId } from 'electron/views/screenshot/screenshot-view';
@@ -24,7 +25,7 @@ export const AutomatedChecksViewSelectors = {
     nthRuleGroupInstances: (position: number) =>
         `${nthRuleGroup(position)} [data-automation-id="${instanceCardAutomationId}"]`,
 
-    settingsButton: 'button[aria-label="settings"]',
+    settingsButton: `[data-automation-id="${commandButtonSettingsId}"]`,
 };
 
 export const ScreenshotViewSelectors = {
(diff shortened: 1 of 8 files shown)


src/tests/electron/common/element-identifiers/automated-checks-view-selectors.ts
```diff
@@ -23,6 +23,8 @@ export const AutomatedChecksViewSelectors = {
         `${nthRuleGroup(position)} [data-automation-id="${cardsRuleIdAutomationId}"]`,
     nthRuleGroupInstances: (position: number) =>
         `${nthRuleGroup(position)} [data-automation-id="${instanceCardAutomationId}"]`,
+
+    settingsButton: 'button[aria-label="settings"]',
```
dbjorge: Consider instead creating a data-automation-id for it?

src/tests/electron/tests/unified-settings-panel.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        it('should default to "on" in the usual configuration our E2E tests use', async () => {
+            await automatedChecksView.expectToggleState(
+                settingsPanelSelectors.telemetryStateToggle,
+                true,
+            );
+        });
+    });
+
+    describe('High Contrast Mode toggle', () => {
+        it('should default to non-high-contrast mode', async () => {
```
dbjorge: Technically, it should default to the system's high contrast mode, but I'm not sure we have a good way of querying that from the test process...

dbjorge: If there isn't a good way to have the test do the right thing, it might be nice to at least use a specific error message to explain that case

src/tests/electron/common/element-identifiers/automated-checks-view-selectors.ts
smoralesd: done

src/tests/electron/tests/unified-settings-panel.test.ts
smoralesd: Skipping the test for now

src/tests/electron/tests/unified-settings-panel.test.ts
```diff
@@ -35,10 +35,14 @@ describe('AutomatedChecksView -> Settings Panel', () => {
     });
 
     describe('High Contrast Mode toggle', () => {
-        it('should default to non-high-contrast mode', async () => {
+        // skipping this test for now as we don't have a good, clean way to check if the system is using
+        // high contrast form the running test.
```
dbjorge: typo form -> from

smoralesd: updated

PR #2305, opened 2020-03-05
feat(android-settings): update how we suppress the first time dialog

smoralesd:
#### Description of changes

Currently, our AI-Android e2e tests dismiss the first time dialog (the telemetry opt-in) by programmatically interacting with it. This is not the way we do it on AI-Web, which uses a "window"-exposed controller.

In this PR:
- Update our app controller to expose a wait to interact with the user configuration controller on the browser window
- Update our e2e tests so they use the new way to suppress the first time dialog
- Add an e2e test to verify the controller properly update the store and updates the settings panel ui

#### Pull request checklist
- [x] Addresses an existing issue: completes WI # 1687239
- [x] Ran `yarn fastpass`


src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -71,4 +68,10 @@ export class AppController {
             } class high-contrast-theme`,
         );
     }
+
+    public async setTelemetryState(enableTelemetry: boolean): Promise<void> {
+        await this.app.webContents.executeJavaScript(
+            `window.insightsUserConfiguration.setTelemetryState(${enableTelemetry})`,
```
dbjorge: This might be happening too early, since it's going before we wait for a view to be visible; maybe better would be for it to involve some sort of `waitUntil` that verifies `window.insightsUserConfiguration` is defined first?

src/tests/electron/common/dismiss-telemetry-opt-in-dialog.ts
```diff
@@ -1,16 +0,0 @@
...
-// Licensed under the MIT License.
-import { Application } from 'spectron';
-import { DEFAULT_ELECTRON_TEST_TIMEOUT_MS } from 'tests/electron/setup/timeouts';
-import { popupPageElementIdentifiers } from 'tests/end-to-end/common/element-identifiers/popup-page-element-identifiers';
-import { Client } from 'webdriverio';
-
-export async function dismissTelemetryOptInDialog(app: Application): Promise<void> {
-    const webDriverClient: Client<void> = app.client;
-
-    await webDriverClient.waitForVisible(
```
dbjorge: It would be nice to move this to a test that verifies it has the expected "suppress telemetry" behavior rather than removing it outright, but if you'd rather leave that for a separate PR that's understandable

smoralesd: I though about that to, like: "uh... we don't have an actual test to check the first time dialog shows up the first time and that we can dismiss it properly". I'll add a test on a separated PR.

src/tests/electron/common/view-controllers/app-controller.ts
smoralesd: that's a good point. I'll work on it.

smoralesd: Updated. Is this close to what you had in mind?

dbjorge: Yes, this seems about right to me!

PR #2313, opened 2020-03-06
refactor(side-panel): add side panel actions to open settings panel

smoralesd:
#### Description of changes

Currently, **AI-Web** has 4 "side panels" (**AI-Android** is only using the Settings panel at this point): 
- Settings
- Preview Features
- Scoping
- Add a Failure Instance
- Info & Examples. 

All of the side panels except Add a Failure Instance are shown/hidden by updating the state on the `DetailsViewStore` by invoking specific actions:
- openPreviewFeatures
- openScopingPanel
- openContentPanel
- openSettingsPanel

There are also 4 actions to close each of those panels.

This could be improved by handling the side panel state in a more generic way, for example, by having a unique open side panel action which takes the type of panel we want to show. The current show panel logic ensures that only one panel is shown at most at any time, which means, using one action to opening is 100% feasible without changing the store logic to much.

This PR is the first one in the path of applying this idea. It only change how we open the settings panel, setting the precedent of how to update the other open panel logic. It also does most of the piping we need in order to do this refactor making the next refactors potentially smaller.

**Note**
Along the same line, we may want to change the state values we keep on the store, from this:
```javascript
currentPanel: {
    isPreviewFeaturesOpen: false,
    isScopingOpen: false,
    isContentOpen: false,
    isSettingsOpen: false,
}
```
to something less verbose (like just storing the type of the panel we are showing or `'None'` if we are not showing any panel).

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/background/actions/details-view-actions.ts
```diff
@@ -8,6 +8,5 @@ export class DetailsViewActions {
         DetailsViewRightContentPanelType
     >();
     public readonly getCurrentState = new Action();
-    public readonly openSettingsPanel = new Action();
```
JGibson2019: Why do we no longer need an action for opening side panel? 

Also for the `closeSettingsPanel` action, should it be renamed to `closeSidePanel`?

smoralesd: Not yet. `closeSettingsPanel` only allows to close that specific panel. As mention in the description, this is the first PR in a series of smaller refactors in order to change how we open/close panel. At this point, I didn't want to change any other action expect for how we open the settings panel.

PR #2315, opened 2020-03-06
refactor: scan results and test

smoralesd:
#### Description of changes

- Simplify testing by using `it.each` and lodash `set` function to make custom `rawData` objects
- Take advantage of typescript optional chaining to get rid of the try-catch blocks

**Note** I ended up explicitly returning `null` on most of the getter as the default case. Not doing so causes the getter to return `undefined` instead and I'm not sure I want to just "break" the current contract. But I can see it that we want that instead.

#### Pull request checklist
- [x] Ran `yarn fastpass`


smoralesd:
> I think the null returns are fair given client usage in
> 
> https://github.com/microsoft/accessibility-insights-web/blob/f6d9ae39f588335fab74f064b03db3e241d774fc/src/electron/platform/android/scan-results-to-platform-data.ts#L9
> 
> even though the comparison would work both ways.

I agree in general. Then again, working on a bug fix I found this ([here](https://github.com/microsoft/accessibility-insights-web/blob/master/src/DetailsView/details-view-container.tsx#L232)):
```typescript
this.props.deps !== null &&
this.props.deps.storesHub !== null &&
```
The `!==` is `false` when the left side is `undefined`.

I'm just being overcautious (and lazy I admit) here.

Issue #1713, opened 2019-11-19
Update prettier config to change line width to 100


Shobhit1:
## Describe the feature

The idea is to have line width more compatible with the `GitHub` diff view and we feel that default `prettier line width of 80 is best suited for that.
So, using [Prettier Override](https://prettier.io/docs/en/configuration.html#configuration-overrides) as suggested by @dbjorge and @smoralesd .

Creating this issue to track all the PRs that are going to be part of this, since we want to incrementally change the whole code base.

smoralesd:
Since this is an engineering task, I'm closing it myself.

PR #2321, opened 2020-03-09
chore: line width for src/tests/unit/tests/electron folder

smoralesd:
#### Description of changes

Remove` src/tests/unit/tests/electron` folder from prettier override for the line width value.

#### Pull request checklist
- [x] Addresses an existing issue: partially addresses #1713
- [x] Ran `yarn fastpass`

### DIFF
--- a/prettier.config.js
+++ b/prettier.config.js
@@ -11,7 +11,6 @@ module.exports = {
             files: [
                 'src/content/**/*',
                 'src/tests/unit/tests/DetailsView/**/*',
-                'src/tests/unit/tests/electron/**/*',
                 'src/tests/unit/tests/injected/**/*',
             ],
             options: {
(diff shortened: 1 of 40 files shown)


prettier.config.js
```diff
@@ -11,7 +11,6 @@ module.exports = {
             files: [
                 'src/content/**/*',
                 'src/tests/unit/tests/DetailsView/**/*',
-                'src/tests/unit/tests/electron/**/*',
```
JGibson2019: So to clarify, you aren't "adding the `src/tests/unit/tests/electron` folder to prettier override" but actually removing `src/tests/unit/tests/electron` folder from the prettier override? Or are the folders in the `files` array exempt from the prettier override, so removing the folder would make sure that the line width is in line with the prettier override changes we want to see?

smoralesd: You're right: I'm removing the folder so it doesn't follow the override. Originally, when we started, we were operating the other way around: adding paths to the override, but at the point we added somewhere around 60% of the code base, Dan made the change, making everything follow the line width we want to have, and adding every path we were missing to the override. I being copy/pasting the description as the work is pretty much the same. 
Good catch.

PR #2328, opened 2020-03-10
fix: handle 'color' attribute for dialogs in HC mode

karanbirsingh:
#### Description of changes

Currently the text in unified/electron dialogs does not change color in when high contrast mode is enabled. Two examples include the telemetry permissions dialog and the device disconnected dialog. These render in a layer inside body and adjacent to #root-container. 

This PR includes one approach to fix this by targeting all .ms-Fabric descendants of body in addition to body itself. It feels a bit odd to me that a change in device-connect-view.scss would affect the device disconnected popup - would like to better understand what approach is best.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/electron/views/device-connect-view/device-connect-view.scss
+++ b/src/electron/views/device-connect-view/device-connect-view.scss
@@ -5,8 +5,7 @@
 @import '../../../reports/components/outcome.scss';
 @import '../../../common/components/cards/fix-instruction-color-box.scss';
 
-body.ms-Fabric,
-body .ms-Fabric {
+.ms-Fabric {
     color: $primary-text;
 }
 
--- a/src/electron/views/device-connect-view/device-connect-view.scss
+++ b/src/electron/views/device-connect-view/device-connect-view.scss
@@ -5,7 +5,8 @@
 @import '../../../reports/components/outcome.scss';
 @import '../../../common/components/cards/fix-instruction-color-box.scss';
 
-body.ms-Fabric {
+body.ms-Fabric,
+body .ms-Fabric {
     color: $primary-text;
 }
 


src/electron/views/device-connect-view/device-connect-view.scss
```diff
@@ -5,7 +5,8 @@
 @import '../../../reports/components/outcome.scss';
 @import '../../../common/components/cards/fix-instruction-color-box.scss';
 
-body.ms-Fabric {
+body.ms-Fabric,
+body .ms-Fabric {
```
smoralesd: so... I guess the general strategy looks fine to me. I do have this questions/comments:
1. Did you check the telemetry opt-in popup, specifically the text, to check the color property? Are we just inheriting the color value from a parent? There may be something there worth exploring.
2.  I curious about having this both selectors instead of just `.ms-Fabric`. Having only `.ms-Fabric` seems equivalent to me, less verbose though.

karanbirsingh: @smoralesd 
1. Yes - modal dialogs include [Layer components](https://github.com/OfficeDev/office-ui-fabric-react/blob/c3be8bc3aba78ae432b13fe53c930d92c664d801/packages/office-ui-fabric-react/src/components/Layer/Layer.base.tsx#L88), which wrap [Fabric components](https://github.com/OfficeDev/office-ui-fabric-react/wiki/The-Fabric-Component). The Fabric component has .ms-Fabric applied, so its specificity overrides the body.ms-Fabric selector. 
2. That seems reasonable!

Looking closer, it seems that there is precendent for a .ms-Fabric -> primaryText or neutral-100 rule in [common.scss](https://github.com/microsoft/accessibility-insights-web/blob/f9eb1aa98ea48c323e825da161a7d8fb559324cd/src/common/styles/common.scss#L30), which is being used in some web consumers.

smoralesd: so... by point 1. I didn't mean to point to the class being use, but directly to the color value, and how it's resolve. That is indeed related to the class but I have seen some office fabric components to "just" inherit the color value. Office fabric may do this because they add `.ms-Fabric` to the body, apply a color value to the body, and then any descendant element with `color: inherit` will get it's color value from the hierarchy (parent, grand-parent and up top) which in some cases, means the body element. Anyway, I'm not sure this specific point is something we need to investigate much further unless you're curious about it.

karanbirsingh: I can try rephrasing - OF has default values for .ms-fabric. While we overrode these using body.ms-fabric, the dialogs in question render inside a <Fabric> component that includes its own more specific .ms-fabric class. This negates the value we've set in body.ms-fabric. I think the PR is close to what we've done in popup.css, although long-term I'm unsure why the default-theme-palette we have isn't overriding the default color values applied by .ms-fabric. I think this will take longer to delve into.

Issue #2324, opened 2020-03-09
"No Content Available" page breaks when refreshed


pownkel:
## Describe the bug

The "no content available" page that appears in the details view when the target page changes or is closed should be able to be refreshed. When it is refreshed, it turns into a blank page. This appears to be a regression since our last release.

**To Reproduce**

1. Start an assessment on any web page
2. Close the target page
3. The details view should change to say "no content available"
4. Refresh the details view

## Expected behavior

The page still says "no content available" after refreshing, instead of being blank.

## Context (please complete the following information)

- OS Name & Version: Windows 10 version 1909 Build 18363.657
- AI-Web Version & Environment: Insider 2.15.0
- Browser Version: 80.0.3987.122 (Official Build) (64-bit)
- Target Page: www.bing.com

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

Yes

## Additional context

This appears to be a regression since our last release.

Reproducing this in the dev build produces the following stacktrace:

```
Uncaught TypeError: Cannot read property 'textContent' of null
    at Header (webpack-internal:///./src/common/components/header.tsx:5)
    at renderWithHooks (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:14825)
    at mountIndeterminateComponent (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:17505)
    at beginWork (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:18629)
    at HTMLUnknownElement.callCallback (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:237)
    at invokeGuardedCallback (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:292)
    at beginWork$1 (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:23234)
    at performUnitOfWork (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:22188)
    at workLoopSync (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:22161)
webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:19560 

 

The above error occurred in the <Header> component:
    in Header (created by InteractiveHeader)
    in InteractiveHeader (created by DetailsViewContainer)
[...]

PR #2329, opened 2020-03-10
refactor: clean up details view container props/deps

smoralesd:
#### Description of changes

Currently, the `DetailsViewContainerProps` contains props that are actually deps. This make the testing and refactoring (in general of `DetailsViewContainer` and related component children) harder than it should be.

In this PR: moving all props that are deps to the deps prop.

**Notes** 
- this refactor will in time help solving #2324
- there is a lot of more clean up we could do, but right now seems like a rabbit hole, or something I would like to iron out by chunks

#### Pull request checklist
- [x] Addresses an existing issue: related to #2324
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/details-view-container.tsx
+++ b/src/DetailsView/details-view-container.tsx
@@ -52,6 +52,14 @@ export type DetailsViewContainerDeps = {
     getDetailsSwitcherNavConfiguration: GetDetailsSwitcherNavConfiguration;
     getCardViewData: GetCardViewData;
     getCardSelectionViewData: GetCardSelectionViewData;
+    issuesSelection: ISelection;
+    clickHandlerFactory: DetailsViewToggleClickHandlerFactory;
+    scopingActionMessageCreator: ScopingActionMessageCreator;
+    inspectActionMessageCreator: InspectActionMessageCreator;
+    issuesTableHandler: IssuesTableHandler;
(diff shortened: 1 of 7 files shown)


src/DetailsView/details-view-container.tsx
```diff
@@ -164,9 +157,9 @@ export class DetailsViewContainer extends React.Component<DetailsViewContainerPr
         return (
             <DetailsViewOverlay
                 deps={deps}
-                previewFeatureFlagsHandler={this.props.previewFeatureFlagsHandler}
-                scopingActionMessageCreator={this.props.scopingActionMessageCreator}
-                inspectActionMessageCreator={this.props.inspectActionMessageCreator}
+                previewFeatureFlagsHandler={this.props.deps.previewFeatureFlagsHandler}
```
waabid: Seems like we can just use `deps` here instead of `this.props.deps`

smoralesd: do you mean changing the current
```typescript
const { deps } = this.props;
...
deps={deps}
```
for
```typescript
deps={this.props.deps}
```
?

smoralesd: Will take care of this on a subsequent PR (there are more places we may want to look for this same pattern so...)

PR #2344, opened 2020-03-12
fix: "No Content Available" page breaks when refreshed 

smoralesd:
#### Description of changes

The "No Content Available" view breaks because when we refresh the details view of a non-existing target page (like in the case the user close the target page but not the details view) we are passing null dependencies to the renderer and some of the components we use for that view do not handle null dependencies.

The approach was to simplify how we handle the refresh + no target page case by not reusing the `DetailsViewRenderer` and instead create a new, more specific renderer called `NoContentAvailableViewRender` along with a new component to handle this content.

**Note**
When high contrast mode is on, refreshing the details view bring the no content available view back with the default theme. While this is not ideal, I checked our extension to a point before this issue was introduce and **this was the actual behavior**. I decided no to change it for now as doing so will be more complex and it will require more time and I'd like to push this fix out as soon as we can manage.

There is more work to do that I'd like to defer to future PR's:
- We should add end to end test to this case as this is at least the second time I know of were we have broke the No Content Available view
- We may be able to respect the high contrast mode being on when the user refresh the page
- Clean up the `DetailsViewContainer`, there may be a case we don't need anymore or we could leverage the new `NoContentAvailableView` component to simplify some of the internal logic

**default theme**
gif showing the issue repro steps being applied. Default theme is on. Refreshing the details view brings back the No Content Available view

**high contrast theme**
gif showing the issue repro steps being applied. High contrast theme is on. Refreshing the details view brings back the No Content Available view

#### Pull request checklist
- [x] Addresses an existing issue: #2324
- [x] Ran `yarn fastpass`


src/DetailsView/details-view-initializer.ts
```diff
@@ -449,6 +455,19 @@ if (isNaN(tabId) === false) {
...
 }
 
-function createNullifiedRenderer(doc, render): DetailsViewRenderer {
-    return new DetailsViewRenderer(null, doc, render, documentElementSetter);
+function createNullifiedRenderer(
+    doc: Document,
+    render: typeof ReactDOM.render,
+): NoContentAvailableViewRenderer {
+    // using an instance of an actual store (instead of a StoreProxy) so we can get the default state.
+    const store = new UserConfigurationStore(null, new UserConfigurationActions(), null);
```
waabid: would be nice to not have to do this ... but I don't really have a better suggestion.

smoralesd: I think we can improve this... but I try to balance complexity against having the fix as soon as possible. We'll get there.

PR #2345, opened 2020-03-12
refactor: debug tools to use WithStoreSubscription

smoralesd:
#### Description of changes

Currently, the main react component in the debug tools is in charge off listening to the stores and re-render when there are updates. 

This PR refactors this behavior by leveraging `WithStoreSubscription` and adding a new component in charge of being the top-level container component.

**Notes** there are no actual changes on the UI/UX 

#### Pull request checklist
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/debug-tools/components/stores-tree.tsx
+++ b/src/debug-tools/components/stores-tree.tsx
@@ -1,8 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { BaseStore } from 'common/base-store';
-import { StoreActionMessageCreator } from 'common/message-creators/store-action-message-creator';
-import { forEach, isEmpty } from 'lodash';
+import { NamedFC } from 'common/react/named-fc';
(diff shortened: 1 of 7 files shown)


src/debug-tools/components/stores-tree.tsx
```diff
@@ -42,98 +35,88 @@ export const columns: IColumn[] = [
...
+    }
 
-        let instanceCount: number = 0;
+    const compact = true;
+    const selection = new Selection();
+    const selectionMode = SelectionMode.none;
 
-        forEach(global, (storeState, storeKey) => {
-            const stateKeys = Object.keys(storeState);
+    const groups = [];
```
waabid: These probably should have types, no?

smoralesd: So... for `groups` we can use `IGroup` and I updated the PR with that.
But for `items`... office fabric defines items as `any[]` which makes sense but then again, not sure if it makes sense to add a type for `items` as this const is local and private.

PR #2357, opened 2020-03-16
feat(threat-model): remove usage of dangerouslySetInnerHtml on popup view

smoralesd:
#### Description of changes

As part of our threat model follow up, we need to handle our usages of `dangerouslySetInnerHtml`. That means remove them or somehow account for them (in the sense of: let's not add more usages and make it clear where and why we are using it).

In this case, it was possible to remove the usage of `dangerouslySetInnerHtml` and replace it with "normal" react code.

**Note** there is no actual change on the UI/UX

#### Pull request checklist
- [x] Addresses an existing issue: part of WI # 1697429
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/popup/components/popup-view.tsx
+++ b/src/popup/components/popup-view.tsx
@@ -192,16 +192,26 @@ export class PopupView extends React.Component<PopupViewProps> {
             <div className="ms-Fabric unsupported-url-info-panel">
                 {this.renderDefaultHeader()}
                 <div className="ms-Grid main-section">
-                    <div
-                        className="launch-panel-general-container"
-                        dangerouslySetInnerHTML={{
-                            __html: DisplayableStrings.urlNotScannable.join('</br>'),
-                        }}
-                    />
+                    <div className="launch-panel-general-container">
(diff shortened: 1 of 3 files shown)


src/popup/components/popup-view.tsx
```diff
@@ -192,17 +192,27 @@ export class PopupView extends React.Component<PopupViewProps> {
...
 
+    private getUrlNotScannableMessage = () => {
+        let keyIndex = 0;
+        const result = DisplayableStrings.urlNotScannable.reduce(
+            (r, current) =>
+                r.concat(<span key={keyIndex++}>{current}</span>, <br key={keyIndex++}></br>),
+            [],
+        );
+
+        result.pop();
```
waabid: I feel like this code could read a bit better; I'm not following why this is necessary? Maybe a comment would help here?

smoralesd: I would love to have a better algorithm to do this... but for now, this is the best I could come up with. I'll put a comment on why we need the `.pop()` call just before returning.

smoralesd: Done.

PR #2370, opened 2020-03-19
feat(threat-model): create LinkScriptProvider to add event listener on target page links

lisli1:
#### Description of changes

Created LinkScriptProvider that adds an event listener to the "click" event on elements with the `target-page-link` id in the generated HTML report. The event handler prompts the browser to generate a confirmation dialog, where selecting OK opens the target page in a new tab, and selecting Cancel prevents the target page from being opened.

Added relevant unit tests. There is no unit test for `getDefaultAddListenerForLink` because the snapshot text representation is a bit confusing (see `collapsible-script-provider.tsx` for similar issue).

Note: To keep this PR small, I will add the event listeners to the target page links in future PRs. As a result, there are currently no changes to UI.

#### Pull request checklist

- [x] Partially addresses an existing issue: # 1521197
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/report-sections/link-script-provider.tsx
+++ b/src/reports/components/report-sections/link-script-provider.tsx
@@ -1,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export const addEventListenerForLink = (doc: Document) => {
+    const targetPageLink = doc.getElementById('target-page-link');
+
+    targetPageLink.addEventListener('click', event => {
+        const result = confirm(
+            'Are you sure you want to navigate away from the Accessibility Insights report?\n' +
+                'This link will open the target page in a new tab.\n\nPress OK to continue or ' +
+                'Cancel to stay on the current page.',
+        );
+
+        if (result === false) {
+            event.preventDefault();
+        }
+    });
+};
+
+export const getAddListenerForLink = (code: string | Function): string =>
+    `(${String(code)})(document)`;
+
+export const getDefaultAddListenerForLink = (): string =>
+    getAddListenerForLink(addEventListenerForLink);
--- a/src/reports/components/report-sections/link-script-provider.tsx
+++ b/src/reports/components/report-sections/link-script-provider.tsx
@@ -1,11 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export const addEventListenerForLink = (doc: Document) => {
+export type ConfirmType = (message?: string) => boolean;
+
+export const addEventListenerForLink = (doc: Document, confirmCallback: ConfirmType) => {
     const targetPageLink = doc.getElementById('target-page-link');
 
     targetPageLink.addEventListener('click', event => {
-        const result = confirm(
+        const result = confirmCallback(
             'Are you sure you want to navigate away from the Accessibility Insights report?\n' +
                 'This link will open the target page in a new tab.\n\nPress OK to continue or ' +
                 'Cancel to stay on the current page.',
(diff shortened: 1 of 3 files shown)


src/reports/components/report-sections/link-script-provider.tsx
```diff
@@ -0,0 +1,24 @@
...
+                'Cancel to stay on the current page.',
+        );
+
+        if (result === false) {
+            event.preventDefault();
+        }
+    });
+};
+
+export const getAddListenerForLink = (code: string | Function): string =>
```
smoralesd: Besides the name, this is the same function we already have on 'src\reports\components\report-sections\collapsible-script-provider.tsx'. Should be good to extract this to it's own file and reuse on both places.

src/tests/unit/tests/reports/components/report-sections/link-script-provider.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    getAddListenerForLink,
+} from 'reports/components/report-sections/link-script-provider';
+import { It, Mock, MockBehavior, Times } from 'typemoq';
+
+describe('LinkScriptProvider', () => {
+    describe('getAddListenerForLink', () => {
+        it('matches content', () => {
+            const code = 'this is test code';
+
+            const result = getAddListenerForLink(code);
```
smoralesd: If we reuse the function from 'src\reports\components\report-sections\collapsible-script-provider.tsx', this test should be on a different place.

src/tests/unit/tests/reports/components/report-sections/link-script-provider.test.ts
```diff
@@ -0,0 +1,37 @@
...
+            const targetPageLinkMock = Mock.ofType<HTMLElement>();
+            const docMock = Mock.ofType<Document>(undefined, MockBehavior.Strict);
+            docMock
+                .setup(doc => doc.getElementById('target-page-link'))
+                .returns(() => targetPageLinkMock.object);
+
+            addEventListenerForLink(docMock.object);
+
+            targetPageLinkMock.verify(
+                link => link.addEventListener('click', It.is(isFunction)),
```
smoralesd: This `verify` probes we are adding and event listener to the proper element on the dom.
We are missing a test when we actually click that element, triggering the listener function. 
Doing so will likely make you change how we are using `confirm` as mention on the comment I left on the actual usage of `confirm`

src/reports/components/report-sections/link-script-provider.tsx
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export const addEventListenerForLink = (doc: Document) => {
+    const targetPageLink = doc.getElementById('target-page-link');
+
+    targetPageLink.addEventListener('click', event => {
+        const result = confirm(
```
smoralesd: Here, `confirm` is being use as a global dependency which is not ideal. One way to solve this is to do the same you are doing with the `doc: Document` param. That way, you can mock the function on the unit tests.

src/reports/components/report-sections/link-script-provider.tsx
smoralesd: On a second though, we may need a slightly different function to consider how we pass the reference to the `confirm` function.

src/tests/unit/tests/reports/components/report-sections/link-script-provider.test.ts
lisli1: Changed the function slightly, will be leaving the tests separate

src/reports/components/report-sections/link-script-provider.tsx
lisli1: Changed the function slightly to include a reference to `confirm`

Issue #1107, opened 2019-08-20
Scan fails when a page does not have a body tag


haonliu:
## Describe the bug

Scan fails when a page does not have a body tag

**To Reproduce**

1. Create a page with the following html: 
 ````
<!DOCTYPE html>
<html>
<head>
<title>Page Title</title>
</head>

<h1>Hello world</h1>

</html>
````
2. Open AI-web
3. Run automated checks
4. See error in the console

## Expected behavior

The scan should be able to run.

## Extension (please complete the following information)

- OS: any
- Version any
- Browser Version any
- Target Page N/A

**Comment by dbjorge** [comment]
I don't think we intended to close this or tag it as needs author feedback, looks like this was just lost in upkeep.

I found that the repro steps from the original bug didn't work as-is; Chrome normalized the repro content as having a body tag. However, using content of `<frameset></frameset>` does cause the issue, where running a fastpass hangs with a scanning spinner and the injected page shows the following error:

```
injected.bundle.js:97 Uncaught (in promise) TypeError: Cannot read property 'appendChild' of null
    at n.create (injected.bundle.js:97)
    at ae.initialize (injected.bundle.js:97)
    at ae.initialize (injected.bundle.js:97)
    at Object.<anonymous> (injected.bundle.js:97)
    at Object.<anonymous> (injected.bundle.js:97)
    at n (injected.bundle.js:2)
    at Object.<anonymous> (injected.bundle.js:97)
    at n (injected.bundle.js:2)
    at injected.bundle.js:2
    at injected.bundle.js:2
```

To clarify on @ferBonnin 's last question, this is our limitation, not axe-core's; I verified on the above minimal test page and the internal page that the same issue was recently re-raised for that reproduces the issue that axe-core and axe's chrome extension work fine.

ferBonnin:
@dbjorge no. It needed a reply from Wahaj.
.@waabid is this our limitation or axe-core's?

dbjorge:
@ferBonnin, I already clarified in the reopening comment: this is our issue, axe-core and axe's web extension both work fine on all repro variants.

waabid:
Sorry, I must have missed this from before! Dan is correct: we should be able to function even without a body element. My guess is we make this assumption in a few places and can likely replace it with html element (but this is just an educated guess).

ferBonnin:
per conversation with Peter, there are 2 issues:

1. we hang
1. we don't scan framesets.

Suggestion for this is (in parallel)

[...]

PR #2373, opened 2020-03-19
fix: assume html element exists instead of body element

waabid:
#### Description of changes

I didn't rename variables as a part of this yet to make this easier to understand for the PR. Also ran a bit of a smoke test on my machine.

#### Pull request checklist

- [n/a] Addresses an existing issue: #1107 
- [n/a] Ran `yarn fastpass`

### DIFF
--- a/src/injected/visualization/center-position-calculator.ts
+++ b/src/injected/visualization/center-position-calculator.ts
@@ -31,7 +31,7 @@ export class CenterPositionCalculator {
         }
 
         const myDocument = this.drawerUtils.getDocumentElement();
-        const body = myDocument.querySelector('body');
+        const body = myDocument.querySelector('html');
 
         const bodyStyle = this.windowUtils.getComputedStyle(body);
         const docStyle = this.windowUtils.getComputedStyle(myDocument.documentElement);
@@ -80,7 +80,7 @@ export class CenterPositionCalculator {
         );
 
         const myDocument = this.drawerUtils.getDocumentElement();
-        const body = myDocument.querySelector('body');
+        const body = myDocument.querySelector('html');
 
         const bodyStyle = this.windowUtils.getComputedStyle(body);
         const docStyle = this.windowUtils.getComputedStyle(myDocument.documentElement);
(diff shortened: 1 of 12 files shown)


src/injected/visualization/center-position-calculator.ts
```diff
@@ -80,7 +80,7 @@ export class CenterPositionCalculator {
         );
 
         const myDocument = this.drawerUtils.getDocumentElement();
-        const body = myDocument.querySelector('body');
+        const body = myDocument.querySelector('html');
 
         const bodyStyle = this.windowUtils.getComputedStyle(body);
```
smoralesd: so now... we'll get the computed style from the `<html>` element... does that just work the same as getting the computed styles from the `<body>` element

src/injected/visualization/root-container-creator.ts
```diff
@@ -9,6 +9,6 @@ export class RootContainerCreator {
         this.htmlElementUtils.deleteAllElements(`#${id}`);
         const root = document.createElement('div');
         root.id = id;
-        this.htmlElementUtils.querySelector('body').appendChild(root);
+        this.htmlElementUtils.querySelector('html').appendChild(root);
```
smoralesd: Does adding a `<div>` element directly to the `<html>` element "just work"??? I'm curious about how chrome and edge handle this. Care to elaborate on your findings?

waabid: Yes, since the html element is just another element. Since we don't really need to be on the body anyways, this isn't an issue.

devaradhanm: but is it allowed as per standards though? the browsers may be doing lots of things to suppress errors in the page. did u check the valid html children tags?

devaradhanm: i tested the validity here - https://validator.w3.org/nu/#textarea

with this content:
```
<!DOCTYPE html>
<html lang="en">
<head>
<title>title 1</title>
</head>

<body>
hello
</body>

<div>hi
</div>

</html>
```

smoralesd: Posting the result of that validation:

image

Error: Stray start tag div.
From line 11, column 1; to line 11, column 5
↩</body>↩↩<div>hi↩</d

PR #2382, opened 2020-03-23
feat(threat-model): introduce target page inspector

smoralesd:
#### Description of changes

As part of our threat model follow up we need to assess our usage of javascript `eval`(which we don't have any and we do have a tslint rule to catch this). While looking at our code base, I found we are using `chrome.devtools.inspectedWindow.eval` (this allow the user to inspect an element we marked as failing an automated check, directly from the target page's details dialog).

While we are using this `eval` API only to call `inspect`, we are implementing this in a way we can pass any script to `eval` (not only `inspect`) which is not necessary and also a opens the door for some security risks.

In this PR:
- Change the API so the user cannot pass just any script; the new API is specific for inspect, the user can pass a selector
- Run some basic sanitization on the input (the selector) as a malicious attack can still try to use this user input to try and run arbitrary code.

**Discussion**
This new API is not bullet proof; it's not close to be perfect neither. This is a starting point in case we start hearing about security concerns with concrete examples on how to exploit this on our extension. 

#### Pull request checklist
- [x] Addresses an existing issue: part of WI # 1697428
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/Devtools/target-page-inspector.ts
+++ b/src/Devtools/target-page-inspector.ts
@@ -1,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export class TargetPageInspector {
+    constructor(private readonly inspectedWindow: typeof chrome.devtools.inspectedWindow) {}
+
+    public inspectElement(selector: string, frameUrl: string): void {
+        const sanitizedSelector = this.sanitizeSelector(selector);
+
+        const script = `inspect(document.querySelector(${sanitizedSelector}))`;
+
+        this.inspectedWindow.eval(script, { frameURL: frameUrl } as any);
+    }
+
+    private sanitizeSelector = (selector: any): string => {
+        if (typeof selector !== 'string') {
+            throw new Error('selector is not a string');
+        }
+
+        // stringify the string selector allow us to handle quote escaping
+        return JSON.stringify(selector);
+    };
+}
--- a/src/Devtools/target-page-inspector.ts
+++ b/src/Devtools/target-page-inspector.ts
@@ -16,7 +16,7 @@ export class TargetPageInspector {
             throw new Error('selector is not a string');
         }
 
-        // stringify the string selector allow us to handle quote escaping
+        // handles scaping the quotes safely
         return JSON.stringify(selector);
     };
 }
--- a/src/tests/unit/tests/DevTools/target-page-inspector.test.ts
+++ b/src/tests/unit/tests/DevTools/target-page-inspector.test.ts
@@ -60,6 +60,7 @@ describe('TargetPageInspector', () => {
         'calls eval through the inspected window, with selector = %s',
         actualSelector => {
             // we need to define a inspect function so we can actually evaluate the script
+            // tslint:disable-next-line: no-unused-variable
             function inspect(): void {
                 // no op on purpose
             }
--- a/src/tests/unit/tests/DevTools/target-page-inspector.test.ts
+++ b/src/tests/unit/tests/DevTools/target-page-inspector.test.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { TargetPageInspector } from 'Devtools/target-page-inspector';
-import { IMock, It, Mock, Times } from 'typemoq';
+import { IMock, It, Mock, MockBehavior, Times } from 'typemoq';
 
 describe('TargetPageInspector', () => {
     type InspectedWindow = typeof chrome.devtools.inspectedWindow;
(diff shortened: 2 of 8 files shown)


src/Devtools/target-page-inspector.ts
```diff
@@ -0,0 +1,22 @@
...
+
+        this.inspectedWindow.eval(script, { frameURL: frameUrl } as any);
+    }
+
+    private sanitizeSelector = (selector: any): string => {
+        if (typeof selector !== 'string') {
+            throw new Error('selector is not a string');
+        }
+
+        // stringify the string selector allow us to handle quote escaping
```
dbjorge: Grammar seems a little confusing here. Maybe just `// Handles escaping quotes safely`?

src/tests/unit/tests/DevTools/target-page-inspector.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        inspectedWindowMock.verify(
+            inspected => inspected.eval(It.isAny(), It.isAny()),
+            Times.never(),
+        );
+
+        expect(act).toThrow('selector is not a string');
+    });
+
+    it('properly handles quotes', () => {
+        const unsafeSelector = `body'); alert('xss`; // this is an attempt to run arbitrary js code
```
dbjorge: Probably would be a good idea for this to include a few other variants (double-quotes, backticks, backslashes)

src/tests/unit/tests/DevTools/target-page-inspector.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        'a[lang|="en"]',
+        'li:nth-child(3n)',
+        'div:not(.awesome)',
+        'div::after',
+        '#result;button',
+    ];
+
+    it.each(safeSelectors)('calls eval on the target page with selector = %s', safeSelector => {
+        const expectedScript =
+            'inspect(document.querySelector(' + JSON.stringify(safeSelector) + '))';
```
dbjorge: I feel like it's a little disingenuous for the test to invoke `JSON.stringify`; I think I'd lean more towards pinning the expected results (I think it'd be most readable to use the table template string form of `it.each`, but snapshots would also be ok)

src/tests/unit/tests/DevTools/target-page-inspector.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    it('properly handles quotes', () => {
+        const unsafeSelector = `body'); alert('xss`; // this is an attempt to run arbitrary js code
+
+        const expectedScript =
+            'inspect(document.querySelector(' + JSON.stringify(unsafeSelector) + '))';
+
+        testSubject.inspectElement(unsafeSelector, testFrameUrl);
+
+        inspectedWindowMock.verify(
+            inspected => inspected.eval(expectedScript, { frameURL: testFrameUrl } as any),
```
dbjorge: This doesn't feel like it's testing the behavior we really care about (that it doesn't escape out of the string); I wonder if better would be to do something like:
* Replace the `alert` with a throw statement in the `unsafeSelector`
* define `inspect` and `document.querySelector` in the context of the test (as no-ops)
* Have the test literally eval the string passed to `inspected.eval`, such that the test would fail (via the `throw` in `unsafeSelector`) if the quoting weren't correct

src/tests/unit/tests/DevTools/target-page-inspector.test.ts
dbjorge: Alternately, maybe an approach similar to the below suggestion for the unsafe cases would be a more valuable test case:

* define inspect and document.querySelector in the context of the test (as verifiable mocks)
* have the test literally eval the string passed to inspected.eval
* verify that the document.querySelector mock gets called with the original selector

src/Devtools/target-page-inspector.ts
smoralesd: Updated.

src/tests/unit/tests/DevTools/target-page-inspector.test.ts
smoralesd: I followed this approach for both safe and unsafe selector. More details on the other comment.

src/tests/unit/tests/DevTools/target-page-inspector.test.ts
smoralesd: This is a very good suggestion! The test went from "the bare minimum" to be a very strong way to probe we are approaching this ok.

I did find some difficulties defining our dependencies in the test context: I tried to use global mocks (literally `GlobalMock`) but I encountered some weird issues from typemoq (an error with: "cannot override querySelector"). The weirdest part was, the first selector (from the array, since I'm using ìt.each` showed this error but the test passed, while the second selector both, showed the error and also the test failed. It didn't matter which selector I put first and second place. 

In the end, I chose to use a "normal" mock for the `querySelector` function which I use to override `document.querySelector` (there is an `afterEach` where I make sure the original `querySelector` is assign back to the `document` object)
In the end, I

src/tests/unit/tests/DevTools/target-page-inspector.test.ts
```diff
@@ -35,19 +50,48 @@ describe('TargetPageInspector', () => {
...
-
-        testSubject.inspectElement(safeSelector, testFrameUrl);
-
-        inspectedWindowMock.verify(
-            inspected => inspected.eval(expectedScript, { frameURL: testFrameUrl } as any),
-            Times.once(),
-        );
-    });
+    it.each(selectors)(
+        'calls eval through the inspected window, with safe selector = %s',
```
dbjorge: I think the word "safe" in this string is a leftover?

smoralesd: Updated.

src/tests/unit/tests/DevTools/target-page-inspector.test.ts
smoralesd: I updated the "injection attempt" selectors after breaking the code a little bit (to make sure the selector actually throw the error). I'm not sure what to do with backticks and backslashes though.

smoralesd: Updated.

PR #2397, opened 2020-03-26
refactor: multiplexing telemetry client

smoralesd:
#### Description of changes

Currently, we use some sort of strategy pattern when it comes to our telemetry clients: if we build without an app insights instrumentation key, we use a `NullTelemetryClient`, which do not send any telemetry to external services; if we do have an instrumentation key (as in the case of a released build) we use an `AppInsightsTelemetryClient` to actually send telemetry.

In both cases, the telemetry client uses a `TelemetryLogger` so we can log any event to the background console (mostly for debug/validation purposes).

This works just fine! But, I'm planning on moving the "local console telemetry" to our debug tools page. In this case, our debug tools will be just another "sink" for telemetry events, one that we may always want to send events to, while leaving the app insights sink with the same current behavior (base on the instrumentation key being present or not).

~I decided to go for a chain of responsibility pattern, where the chain will always have the "to debug tools" telemetry client (this is not implemented right now) and the app insights client will be only added to the chain if we have the instrumentation key for it to work properly.~

I decided to follow @dbjorge comment and implement a simple container to pass the telemetry client calls to all of them (a simple array and simple iteration through it).

On this PR:
- Introduce the `MultiplexingTelemetryClient`, to wrap the original `TelemetryClient` implementation (so no changes on those classes), and still allow to create the chain
- Update the `TelemetryClientProvider` to use the new chain of responsibility
- Rename `NullTelemetryClient` to `ConsoleTelemetryClient` to better reflects its responsibility
- Remove the usage of the `TelemetryLogger` from `AppInsightsTelemetryClient`; the main reason to do this is: the chain always have the `ConsoleTelemetryClient` as it's first link of the chain.

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/background/telemetry/telemetry-client-provider.ts
```diff
@@ -36,11 +37,17 @@ export const getTelemetryClient = (
         installDataGenerator,
     );
 
+    const nullTelemetryClient = new ConsoleTelemetryClient(coreTelemetryDataFactory, logger);
```
dbjorge: `consoleTelemetryClient`?

smoralesd: Updated

src/background/telemetry/chainned-telemetry-client.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { TelemetryClient } from 'background/telemetry/telemetry-client';
+
+export class ChainnedTelemetryClient implements TelemetryClient {
```
dbjorge: I like the idea of a multiplexing telemetry client, but I think maybe the data structure would be more straightforward and the most complex bit of new code could be removed by doing something like this instead of the linked list:

```ts
export class MultiplexingTelemetryClient implements TelemetryClient {
    constructor(private readonly wrappedClients: TelemetryClient[]);

    public enableTelemetry(): void {
        this.wrappedClients.forEach(c => c.enableTelemetry());
    }

    public disableTelemetry(): void {
        this.wrappedClients.forEach(c => c.disableTelemetry());
    }

    public trackEvent(name: string, properties?: Object): void {
        this.wrappedClients.forEach(c => c.trackEvent(name, properties));
    }
}
```

smoralesd: Like it. Will update my PR with this changes in a little bit.

smoralesd: Updated.

src/tests/unit/tests/background/telemetry/telemetry-client-provider.test.ts
```diff
@@ -38,7 +38,7 @@ describe('TelemetryClientProvider', () => {
             Mock.ofType<StorageAdapter>().object,
         );
 
-        expect(result).toBeInstanceOf(AppInsightsTelemetryClient);
+        expect(result).toBeInstanceOf(MultiplexingTelemetryClient);
```
dbjorge: These test cases should probably be testing the number/types of the wrapped clients, rather than the type of the wrapping client, no?

smoralesd: I get your point... although, that will either: break encapsulation or not really be a unit test, right?
I pushed a change with the first option (break encapsulation by accessing private member). Thoughts?

dbjorge: I agree that this isn't great for encapsulation, but I also think what you're doing here is no better or worse than the original test, so approving the PR on that basis since everything else looks fine. One option might be to have the factory be responsible for outputting `TelemetryClient[]` rather than `TelemetryClient` and make it the caller's job to wrap it in the `MultiplexingTelemetryClient` ctor

PR #2399, opened 2020-03-26
feat(threat-model): create NewTabLinkWithConfirmationDialog

lisli1:
#### Description of changes

- Created a new component `NewTabLinkWithConfirmationDialog` that uses `NewTabLink` and adds a script that opens a confirmation dialog, in which selecting OK opens the target page in a new tab, and selecting Cancel prevents the target page from being opened.
- The component generates a unique id for each link, which is used to find the corresponding element in the report
- Switched `NewTabLink` out for `NewTabLinkWithConfirmationDialog` on target page links in automated checks and assessment reports
- Added unit tests `new-tab-link-confirmation-dialog.test.tsx`
- Excluded specific functions from code coverage to make eval and snapshot testing work
- Updated integration.test.ts and snapshot to keep html in script tags
- Note: The dialog has the same text and looks the same in both reports.

Screenshot of the resulting confirmation dialog on new Edge:
image

Co-authored-by: @dbjorge 

#### Pull request checklist

- [x] Addresses an existing issue: # 1521197
- [x] Ran `yarn fastpass`


src/reports/components/new-tab-link-confirmation-dialog.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { NamedFC } from 'common/react/named-fc';
+import { getId, ILinkProps } from 'office-ui-fabric-react';
+import * as React from 'react';
+
+export const NewTabLinkWithConfirmationDialog = NamedFC<ILinkProps>(
+    'NewTabLinkWithConfirmationDialog',
+    props => {
+        const id = getId('new-tab-link-with-confirmation-dialog__'); // generates something like new-tab-link-with-confirmation-dialog__123
+
+        // tslint:disable: react-no-dangerous-html
```
smoralesd: I think this needs to be:
```typescript
// tslint:disable-next-line:react-no-dangerous-html
```
(notice the `next-line` in the comment)

Also (but not for this PR), I also had to disable this same rule on other places around the report. Should we think on a better strategy so we can avoid grandfathering new usages of this prop?

src/tests/unit/jest.config.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
         '<rootDir>/src/**/*.{ts,tsx}',
         '!<rootDir>/src/tests/**/*',
         '!<rootDir>/src/**/*.d.ts',
+        '!<rootDir>/src/reports/components/new-tab-link-confirmation-dialog.tsx',
```
smoralesd: this seems odd? care to add a comment explaining why we don't want to collect coverage from this? (I have a guess on the reason we need this exception but it would be nice to be a little more explicit)

src/tests/unit/tests/reports/components/new-tab-link-confirmation-dialog.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+        };
+        const child = <div>test-child</div>;
+
+        const testSubject = shallow(
+            <NewTabLinkWithConfirmationDialog {...props}>{child}</NewTabLinkWithConfirmationDialog>,
+        );
+
+        const testSubjectNewTabLink = testSubject.find(NewTabLink);
+
+        expect(testSubjectNewTabLink.prop('href')).toEqual(props.href);
```
smoralesd: Would it be better to check for all props except the id here? Something like
```typescript
const { id, ...testableProps } = testSubjectNewTabLink.props(); // this "discards" the id from the props, but keeps all the others

expect(testableProps).toMatchSnapshot(); // in-code assertion should work just fine too
```

src/tests/unit/tests/reports/components/new-tab-link-confirmation-dialog.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+            window.confirm = confirmMock.object;
+        });
+
+        afterEach(() => {
+            document.getElementById = originalGetElementById;
+            window.confirm = originalConfirm;
+        });
+
+        it('is added to the link', () => {
+            getElementByIdMock.reset();
```
smoralesd: is this call to `.reset()` necessary here? I see on line 56 we are creating a new instance before each test:
```ts
getElementByIdMock = Mock.ofType<GetElementById>(undefined, MockBehavior.Strict);
```

src/tests/unit/tests/reports/components/new-tab-link-confirmation-dialog.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+            targetPageLinkMock.verifyAll();
+            getElementByIdMock.verifyAll();
+        });
+
+        it('calls confirm function with expected text', () => {
+            const eventMock = Mock.ofType<Event>();
+            eventMock.setup(event => event.preventDefault());
+
+            const expectedConfirmationMessage =
```
smoralesd: Would it be worth it to try using snapshot for this value? Maybe on the mock setup code on line 105, something like:
```typescript
confirmMock
    .setup(handler => handler(It.isAnyString())
    .callback(message => expect(message).toMatchSnapshot())
    .verifiable(Times.once());
```

src/reports/components/new-tab-link-confirmation-dialog.tsx
lisli1: Would you be okay with the comment looking like this:
```
{/* tslint:disable-next-line: react-no-dangerous-html */}
```

I saw `tslint:disable` without `-next-line` in `results-container.tsx` too

I agree that we should try to come up with a better strategy for this in a separate PR

src/tests/unit/jest.config.js
lisli1: Added `// excluding the following file to omit code cov in generated script for running eval in corresponding test`. Is it sufficient?

src/tests/unit/tests/reports/components/new-tab-link-confirmation-dialog.test.tsx
lisli1: In line 59 (in `beforeEach`), we set up `getElementIdMock` in a more general way, without specifying the id. Do you think it would be better to move the setup to each individual test, so we don't have to reset the mock setup here? Or should we set up the mock such that we use the specific id each time?

src/tests/unit/tests/reports/components/new-tab-link-confirmation-dialog.test.tsx
lisli1: Updated according to suggestion

src/tests/unit/tests/reports/components/new-tab-link-confirmation-dialog.test.tsx
lisli1: Updated

src/reports/components/new-tab-link-confirmation-dialog.tsx
smoralesd: So, using `// tslint:disable:react-no-dangerous-html` (without `-next-line`) disables the rule for the rest of the file. I think in general we should avoid using that on our code base. The usage you found on `results-container.tsx` should be fixed.
Using `// tslint:disable-next-line:react-no-dangeours-html` does what you should expect: disable that rule only on the next line thus forcing us to be explicit when we want to do something "against the rules" (this may come handy if at any time we'd like to have a little script to count all the instances of this exceptions we have).

As per the comment syntax, I see what you mean and that's fine. The comment style should not matter as long as we can tell tslint to ignore that rule on that specific line.

src/tests/unit/jest.config.js
smoralesd: so... does the use of `eval` messes with the code coverage? If that's what's happening here, I think that should be part of the comment.

src/tests/unit/tests/reports/components/new-tab-link-confirmation-dialog.test.tsx
smoralesd: so if that setup on line 59 does not apply to all the tests under this `describe` maybe it should not be on the `beforeEach` ??

src/tests/unit/jest.config.js
dbjorge: It's the other way around; the use of code coverage messes with the eval in the test (the product code is functionally evaling the compiled version of the source, which in the jest case includes coverage-injected functions that don't actually exist at runtime)

dbjorge: It would be nice to look into whether it's possible to exclude just the specific function from coverage rather than the whole file, though (not sure if there's any sort of special comment syntax for that)

src/tests/unit/tests/reports/components/new-tab-link-confirmation-dialog.test.tsx
lisli1: Moved setup out of beforeEach

src/tests/unit/jest.config.js
lisli1:
```
// excluding the following file to omit code cov in generated script, which would contain
// code cov injected functions that interfere with running eval
```
How's this?

lisli1: According to [this](https://codewithhugo.com/jest-exclude-coverage/), it's possible to ignore a specific line or function

src/reports/components/new-tab-link-confirmation-dialog.tsx
lisli1: Changed for both files

Issue #1503, opened 2019-10-17
Perf issue when inspecting elements and using Assessment


ferBonnin:
## Describe the bug
Inspect elements using Chrome dev tools becomes slugish after opening Assessment

**To Reproduce**

1. Go to [Accessible university demo website](https://www.washington.edu/accesscomputing/AU/before.html) 
2. Go to inspect and select a few elements using hover and clicking on the list of elements
3. Open Assessment, go to any requirement that has a visual helper
4. Use inspect again, selecting and navigating elements becomes slower.

## Expected behavior

No perf impact 

## Extension (please complete the following information)

- OS: [e.g. Windows/Mac] Windows but repros in Mac too
- Version [e.g. 1.140.1] 2.10.0 
- Browser Version [e.g Chrome 71.0] Chrome 77
- Target Page [e.g www.bing.com] any

## Additional context
Reported by a user

ferBonnin:
Per conversation with Peter, @waabid can you please take a look?

ferBonnin:
perf is much better 👍

PR #2408, opened 2020-03-27
fix: perf issue in dev tools caused by duplicated color styles

dbjorge:
#### Description of changes

This PR fixes perf issues when trying to use Chromium's dev tools (particularly, the "inspect element" view) with our web extension. The issue was that every individual CSS module that imported `colors.scss` was resulting in its own copy of all the color variables being included in the final bundled css files (see https://github.com/webpack-contrib/sass-loader/issues/145); `detailsview.css` had dozens of copies, each of which was being considered and then rendered as "overriden" in the dev tools UI for every element, which was causing lots of slowdown.

This PR resolves that issue by splitting out the parts of `colors.scss` that define actual CSS `--var` statements from the parts that define the SCSS variables our modules use directly. The SCSS variables remain in place in `colors.scss` (so no change to most of the consumers is required), but the `--var` styles are moved into a separate `/common/styles/root-level-only/color-definitions.scss` file that is intended to be imported only by root-level scss files. I then updated the root-level SCSS files (covering popup/details view/guidance pages/injected/reports/ui package) to directly reference the new root-level-only `.scss`.

I did a similar split for `common.scss`, though most of the actual perf issue came from `colors.scss`.

I verified that there didn't seem to be any styling regressions in:
* Injected page dialog or highlight boxes
* Details view
* Popup
* Guidance pages
* Generated reports (fastpass and assessment)
* UI package
* Unified

#### Pull request checklist

- [x] Addresses an existing issue: #1503
- [x] Ran `yarn fastpass`


src/common/styles/colors.scss
```diff
@@ -268,8 +101,8 @@ $card-border: var(--card-border);
 $card-footer-border: var(--card-footer-border);
 $button-hover: var(--button-hover);
 
-// consistent colors
-$always-white: #ffffff;
-$always-black: #161616;
+// Consistent colors that don't vary with high contrast mode. Use sparingly.
```
waabid: 👍

Issue #1966, opened 2020-01-10
Visualizations are drawn for nodes that are not within the screenshot


waabid:
## Describe the bug

When running the electron application, errors that are not on the screenshot are still visualized.

**To Reproduce**

1. Connect the electron application on android emulator (personally using Pixel 2)
2. Go to Bing Search app.
3. Scan the app.

## Expected behavior

These visualizations should not exist.

## Screenshots

![image](https://user-images.githubusercontent.com/32555133/72116528-54dc6280-32ff-11ea-9a76-8545d00b4d82.png)

## Context (please complete the following information)

Using android 27 on Pixel 2.

## Are you willing to submit a PR?

Yes

ferBonnin:
Per conversation with Peter, at the very least we should not highlight things that are not in the screenshot. partial highlights are ok

ferBonnin:
@waabid can you please validate and close this?

PR #2410, opened 2020-03-28
refactor: how visualization highlight status is calculated

waabid:
#### Description of changes

Why? It's possible, in the electron app that we we will sometimes have elements that are not within the screenshot at all. This causes #1966. Ideally, in this case we want to show this as an "unavailable highlight" as per discussion offline with PM team. In the future, we may also have something similar in the web (where an element might disappear).

What? Instead of only having highlighted ids in the card selection view data, include all result statuses to their respective highlight status.

How? So there's a few ways to do this:
1. Do this in the card selection store: this would also let something in the web product "set" the highlight status if it were to change. The problem you run into with this is that not all information might be present to evaluate whether a result instance is unavailable for highlight when an action is invoked.
2. Do this in the unified scan result store: this would be fine for when we are dealing with screenshots because we can calculate when something is or isn't available upon scan. This however, is not the case with web apps. As a result, it would violate our principle of only modifying that store upon a user scanning.
3. Do this when we calculate the card selection view data. **This is the solution chosen in this PR.** This allows access to all the information necessary from the stores and be re-calculated as the store information changes, allowing support for future scenarios in web, if we so choose. Every other view is dependent on what we generate hence the comprehensive changes in this PR but it does give us flexibility. Implicitly we were already deciding what to highlight and what not highlight here before: we should move all said logic to this component isolating our complex selection/highlight behavior as best as we can.

Some notes about this implementation/PR in particular:
1. Instead of returning an array of highlighted result ids, the new view model has a map of the result id to it's highlight status.
2. I cleaned up the implementation of 'getCardView' as it was a bit hard to parse, since this logic is a bit confusing inherently. As a result, I added comments to outline the different cases.
3. Since we require platformInfo to calculate this availability in electron, it made sense to depend on the unified scan result object rather than each individual property. Happy to change this if there is disagreement on front (can use a Pick interface). Lots of the changes are due to this (and the fact that card view model data is used in many places).
4. In the getUnavailableHighlightStatus, I wanted flexibility such that we can have other ways of calculating what caused an element/result to be unavailable. This means it's not binary (i.e. it's available or not). This is why the function returns null when it is available. Another idea, which might be better, would be to introduce a super type that includes 'available' and all result highlight types, which is only exposed to consumers of this method.

#### Pull request checklist

- [x] Addresses an existing issue: #1966 
- [x] Ran `yarn fastpass`


src/common/get-card-selection-view-data.ts
```diff
@@ -1,65 +1,147 @@
...
     viewData.selectedResultUids = getOnlyResultUidsFromSelectedCards(
-        storeData.rules,
+        cardSelectionStoreData.rules,
         viewData.expandedRuleIds,
     );
 
-    viewData.highlightedResultUids = viewData.selectedResultUids.length
-        ? viewData.selectedResultUids
-        : getAllResultUidsFromRuleIdArray(storeData.rules, viewData.expandedRuleIds);
+    if (viewData.selectedResultUids.length) {
```
dbjorge, line 84: I think it would be a little more clear to say `if (viewData.selectedResultUids.length > 0) {` (I never trust my rememberance of javascript's falsiness rules and had to test it to convince myself this was right)

src/common/get-unavailable-highlight-status.ts
```diff
@@ -0,0 +1,41 @@
...
+} from 'common/types/store-data/unified-data-interface';
+
+import { HighlightState } from 'common/components/cards/instance-details-footer';
+
+import { BoundingRectangle } from 'electron/platform/android/scan-results';
+
+export type GetUnavailableHighlightStatus = (
+    result: UnifiedResult,
+    platformInfo: PlatformData,
+) => HighlightState;
```
dbjorge, line 17: I found it confusing that the function signature suggests it should be able to return `'unavailable' | 'visible' | 'hidden'`, but the actual behavior of all implementations is to return `'unavailable' | null`

dbjorge: Since the only callsite is relying on the `'unavailable` | null` behavior, I'd suggest just updating this to return a boolean instead and rename stuff here accordingly; I think it would make the callsite, type signature, and implemenations all more clear

src/tests/unit/tests/common/get-card-selection-view-data.test.ts
```diff
@@ -29,109 +38,228 @@ describe('getCardSelectionStoreviewData', () => {
...
+            sampleUid2: 'visible',
+            sampleUid3: 'visible',
+            sampleUid4: 'visible',
+        });
+        expect(viewData.expandedRuleIds).toEqual([]);
+        expect(viewData.selectedResultUids).toEqual([]);
+        expect(viewData.visualHelperEnabled).toEqual(true);
+    });
+
+    test('all rules collapsed, visual helper enabled, some results unavailable', () => {
```
dbjorge, line 91: It would be good to add test cases that verify the precedence of `unavailable` vs `hidden` in the "some rules expanded" and "some rules selected" states

src/common/get-unavailable-highlight-status.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    return null;
+};
+
+function hasValidBoundingRectangle(
+    boundingRectangle: BoundingRectangle,
+    viewPort: ViewPortProperties,
+): boolean {
+    return !(boundingRectangle.left > viewPort.width || boundingRectangle.top > viewPort.height);
```
dbjorge, line 38: Do we know for sure that `boundingRectangle.left` and `boundingRectangle.top` will always be non-negative?

src/common/get-card-selection-view-data.ts
```diff
@@ -1,65 +1,147 @@
...
+    cardSelectionStoreData: CardSelectionStoreData,
+    unifiedScanResultStoreData: UnifiedScanResultStoreData,
+    getUnavailableHighlightStatus: GetUnavailableHighlightStatus,
 ): CardSelectionViewData => {
     const viewData = getEmptyViewData();
 
-    if (!storeData) {
+    if (
+        isEmpty(cardSelectionStoreData) ||
+        isEmpty(unifiedScanResultStoreData) ||
```
dbjorge, line 38: I think can be simplified a little (and reduced to one line) by replacing `isEmpty(unifiedScanResultStoreData) || unifiedScanResultStoreData.results == null` with `unifiedScanResultStoreData?.results == null`

src/common/get-card-selection-view-data.ts
```diff
@@ -1,65 +1,147 @@
...
     );
 
-    viewData.highlightedResultUids = viewData.selectedResultUids.length
-        ? viewData.selectedResultUids
-        : getAllResultUidsFromRuleIdArray(storeData.rules, viewData.expandedRuleIds);
+    if (viewData.selectedResultUids.length) {
+        viewData.selectedResultUids.forEach(resultUid => {
+            viewData.resultsHighlightStatus[resultUid] = 'visible';
+        });
+        viewData.resultsHighlightStatus = {
```
dbjorge, line 88: I agree that the `resultsHighlightStatus` object is a nicer representation for consumers of the view data, but I think the implementation of this method has become too complex (88 lines). I tried out some options for simplifying and I think my favorite was with separating the concerns of identifying the visible UIDs vs consolidating the "hidden, visible, unavailable" lists into the combined highlight status object. Consider:

```typescript
export const getCardSelectionViewData: GetCardSelectionViewData = (
    cardSelectionStoreData: CardSelectionStoreData,
    unifiedScanResultStoreData: UnifiedScanResultStoreData,
    getUnavailableHighlightStatus: GetUnavailableHighlightStatus,
): CardSelectionViewData => {
    const allRules = cardSelectionStoreData.rules;
    const viewData = getEmptyViewData();

    if (isEmpty(cardSelectionStoreData) || unifiedScanResultStoreData.results == null) {
        return viewData;
    }

    viewData.visualHelperEnabled = cardSelectionStoreData.visualHelperEnabled || false;
    viewData.expandedRuleIds = getExpandedRuleIds(allRules);
    viewData.selectedResultUids = getSelectedResultUids(allRules, viewData.expandedRuleIds);

    const allResultUids = getAllResultUids(allRules);
    const unavailableResultUids = getUnavailableResultUids(
        unifiedScanResultStoreData,
        getUnavailableHighlightStatus,
    );

    let visibleResultUids: string[];
    if (!cardSelectionStoreData.visualHelperEnabled) {
        visibleResultUids = [];
    } else if (viewData.selectedResultUids.length > 0) {
        visibleResultUids = viewData.selectedResultUids;
    } else {
        visibleResultUids = getAllResultUidsForRuleIds(allRules, viewData.expandedRuleIds);
    }

    viewData.resultsHighlightStatus = getHighlightStatusByResultUid(
        allResultUids,
        visibleResultUids,
        unavailableResultUids,
    );

    return viewData;
};

// ...

function getUnavailableResultUids(
    unifiedScanResultStoreData: UnifiedScanResultStoreData,
    getUnavailableHighlightStatus: GetUnavailableHighlightStatus,
): string[] {
    const { platformInfo } = unifiedScanResultStoreData;

    return unifiedScanResultStoreData.results
        .filter(
            result =>
                result.status === 'fail' &&
                getUnavailableHighlightStatus(result, platformInfo) === 'unavailable',
        )
        .map(result => result.uid);
}

function getHighlightStatusByResultUid(
    allResultUids: string[],
    visibleResultUids: string[],
    unavailableResultUids: string[],
): ResultsHighlightStatus {
    const result = {};
    for (const resultUid of allResultUids) {
        result[resultUid] = 'hidden';
    }
    for (const resultUid of visibleResultUids) {
        result[resultUid] = 'visible';
    }
    for (const resultUid of unavailableResultUids) {
        result[resultUid] = 'unavailable';
    }
    return result;
}
```


```

File: src/common/get-card-selection-view-data.ts (hunk 1 above)
Comment (waabid) [reply src/common/get-card-selection-view-data.ts:84]: Agreed. I actually did have that at some point; I'll re-add it! 😄

File: src/common/get-unavailable-highlight-status.ts (hunk 2 above)
Comment (waabid) [reply src/common/get-unavailable-highlight-status.ts:17]: I was definitely favoring having other potential statuses because I figured we'd want to mention the reason (and this is consumed by the view as is) ... I guess on second-thought it might not be a bad idea to simplify the statuses (i.e. keep them as they are with unavailable, visible, hidden) and then later make it an object providing additional information about the reason.

That would require the view not directly depending on this value in the future, which is probably a good thing as well.

File: src/common/get-unavailable-highlight-status.ts (hunk 4 above)
Comment (waabid) [reply src/common/get-unavailable-highlight-status.ts:38]: Good call! I don't know: but it feels like this function/component should not make that assumption!

File: src/common/get-card-selection-view-data.ts (hunk 6 above)
Comment (waabid) [reply src/common/get-card-selection-view-data.ts:88]: Agreed.

Thanks for the code suggestions as well; let me look into making this a bit simpler/clearer.

File: src/common/is-result-highlight-unavailable.ts [hunk 7]
```diff
@@ -25,17 +22,22 @@ export const getUnavailableHighlightStatusUnified: GetUnavailableHighlightStatus
...
     ) {
-        return 'unavailable';
+        return true;
     }
 
-    return null;
+    return false;
 };
 
 function hasValidBoundingRectangle(
```
dbjorge, line 31: Do we want to be showing rectangles if they are only partially within the viewport? Right now this function is saying "yes" for rectangles that are partially occluded to the right/bottom but "no" for rectangles partially occluded to the left/top; feels like the answer should be consistent

waabid: I was thinking about that as well: based off of previous discussions with Fer, it seemed like we were fine with visualizations that were at least partially in-frame going to the right. I guess that would make it inconsistent here ... let me check with Fer to get an explicit condition we're comfortable with moving forward.

PR #2437, opened 2020-04-01
fix: using document.body instead of querying for body selector

waabid:
#### Description of changes

This is an alternative to the PR (#2373) that using the html element. This is safer than querying for the body selector as it [also includes frameset elements](https://developer.mozilla.org/en-US/docs/Web/API/Document/body).

#### Pull request checklist

- [x] Addresses an existing issue: #1107
- [x] Ran `yarn fastpass`
- [] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`

### DIFF
--- a/src/injected/details-dialog-handler.ts
+++ b/src/injected/details-dialog-handler.ts
@@ -141,7 +141,7 @@ export class DetailsDialogHandler {
 
         const modal = shadowRoot.querySelector('.insights-dialog-main-override-shadow');
         if (modal != null) {
-            this.htmlElementUtils.getBody();
+            this.htmlElementUtils.getBody().classList.add('insights-modal');
             modal.addEventListener('click', ev => {
                 if (modal === ev.target) {
                     this.closeWindow(shadowRoot);
--- a/src/injected/details-dialog-handler.ts
+++ b/src/injected/details-dialog-handler.ts
@@ -218,7 +218,7 @@ export class DetailsDialogHandler {
         if (dialogContainer) {
             dialogContainer.parentNode.removeChild(dialogContainer);
         }
-        const body = this.htmlElementUtils.querySelector('body');
+        const body = this.htmlElementUtils.getBody();
         body.classList.remove(...['insights-modal']);
     }
 }
--- a/src/injected/details-dialog-handler.ts
+++ b/src/injected/details-dialog-handler.ts
@@ -141,7 +141,7 @@ export class DetailsDialogHandler {
 
         const modal = shadowRoot.querySelector('.insights-dialog-main-override-shadow');
         if (modal != null) {
-            document.body.classList.add('insights-modal');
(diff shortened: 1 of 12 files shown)


src/injected/details-dialog-handler.ts
```diff
@@ -141,7 +141,7 @@ export class DetailsDialogHandler {
 
         const modal = shadowRoot.querySelector('.insights-dialog-main-override-shadow');
         if (modal != null) {
-            document.body.classList.add('insights-modal');
+            this.htmlElementUtils.getBody();
```
dbjorge, line 144: This should still be adding to the classList, right?

waabid: Ugh, yes! My bad! I'll fix it!

src/injected/visualization/root-container-creator.ts
```diff
@@ -9,6 +9,6 @@ export class RootContainerCreator {
         this.htmlElementUtils.deleteAllElements(`#${id}`);
         const root = document.createElement('div');
         root.id = id;
-        this.htmlElementUtils.querySelector('body').appendChild(root);
+        this.htmlElementUtils.getBody().appendChild(root);
```
dbjorge, line 12: The intention is for this to do the right thing on a page that uses `<frameset>` instead of `<body>`, right? I don't think making it work is quite this simple; MDN doesn't document `<frameset>` as supporting any child that isn't a `<frame>` element, and in local testing on lastest stable new edge, it doesn't look like `<frameset><div>test</div></frameset>` renders the `<div>`

dbjorge: (same comment on most of the other usages that involve mutating/adding children to `document.body`)

waabid: This doesn't fully "fix" the frameset scenario; it's just a safer/better way of doing what we already did that also happens to not completely crash/break the app in the frameset scenario. As in, this would still not be ideal in the frameset scenario but it won't completely break (i.e. it'll finish scanning and show results).

I think the "right thing" here likely requires some UX work and I figured this was an improvement/mitigation of some sorts.

dbjorge: I appreciate wanting to break up the full solution into parts, but I think it would be better to work out how to get the scanner to actually scan across `<frame>` boundaries before making the change to avoid the scanner crashing; I think having the scanner obviously not work is vastly superior to having it *look like* it works but actually not be scanning the content of `<frame>`s and giving false negatives.

dbjorge: Per offline conversation, my bad, I misremembered how we interacted with this the last time this issue came up and I tried making this PR myself; it looks like we don't have the false negatives after all.

PR #2462, opened 2020-04-08
chore: Remove EnvironmentInfoProvider and use DataTool instead

AhmedAbdoOrtiga:
#### Description of changes
- In unified render we use only DataTool to get app and engine info, in web, we use EnvironmentInfoProvider, and actually EnvironmentInfo is almost duplicate of DataTool.
- So we removed  EnvironmentInfoProvider, like that we can use the same report footer object for android.
- Refactored all files that use EnvironmentInfoProvider.
- Created a new footer object for unified.

**Note: The PR can't be broken down in smaller PRssince almost all these files use EnvironmentInfoProvider**

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/reports/components/report-sections/report-section-factory.tsx
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 import { FailedInstancesSectionDeps } from 'common/components/cards/failed-instances-section';
 import { FixInstructionProcessor } from 'common/components/fix-instruction-processor';
-import { EnvironmentInfo } from 'common/environment-info-provider';
+import { EnvironmentInfo, EnvironmentInfoProvider } from 'common/environment-info-provider';
```
smoralesd, line 5: is `EnvironmentInfo` still used after this changes?

src/reports/package/footer-text-for-service.tsx
```diff
@@ -2,18 +2,28 @@
 // Licensed under the MIT License.
 import * as React from 'react';
 
+import { EnvironmentInfoProvider } from 'common/environment-info-provider';
 import { ReportSectionFactory } from 'reports/components/report-sections/report-section-factory';
 import { ToolLink } from 'reports/components/report-sections/tool-link';
+import { NamedFC } from 'common/react/named-fc';
 
+export type FooterTextProps = { environmentInfoProvider: EnvironmentInfoProvider };
```
smoralesd, line 10: This is duplicated from 'src/reports/components/report-sections/footer-text.tsx', let's avoid this by defining the type in one place only and importing where we need to use it.

src/reports/package/footer-text-for-service.tsx
```diff
@@ -2,18 +2,28 @@
...
 
-    const footerText = ({ environmentInfo: { axeCoreVersion, browserSpec } }) =>
-        <>
-            This automated checks result was generated using the {serviceName}{' '}
-            that helps find some of the most common accessibility issues. The scan was
-            performed using axe-core {axeCoreVersion} and {browserSpec}. For a complete
-            WCAG 2.1 compliance assessment please visit{' '}<ToolLink />.
-        </>;
+    const footerText = NamedFC<FooterTextProps>(
+        'FooterText',
```
smoralesd, line 14: This component has the same 'displayName` as the original 'FooterText'. One of the things this 'displayName' is useful for is debugging, where having different names can help you identify the actual component being use to render some part of the dom. Not terribly important, but something to consider.

src/tests/unit/tests/reports/components/report-sections/footer-text.test.tsx
```diff
@@ -1,19 +1,21 @@
...
 import { FooterText } from 'reports/components/report-sections/footer-text';
 
 describe('FooterText', () => {
     it('renders', () => {
-        const environmentInfo: EnvironmentInfo = {
-            extensionVersion: '1.2.3',
-            browserSpec: 'dummy browser version 1.0.1',
-            axeCoreVersion: '4.5.6',
-        };
+        const environmentInfoProvider = new EnvironmentInfoProvider(
```
smoralesd, line 10: This should be a mock.
The main reason we were not using a mock before (with the `EnvironmentInfo`) is because there was not behavior, i.e. `EnvironmentInfo` is a simple type, a container of properties with not public/private methods.
By changing the dependency to be of type `EnvironmentInfoProvider` we are now using an actual class (even if we have an interface for it), at which point we should be using a mock, not the actual implementation. The reasoning here is: a unit test should isolate the test subject from all it's dependencies implementation details, in this case, the `EnvironmentInfoProvider` is such a dependency for the `FooterText`.

src/tests/unit/tests/reports/components/report-sections/footer-text.test.tsx
```diff
@@ -1,19 +1,21 @@
...
+            '1.2.3',
+            'dummy browser version 1.0.1',
+            '4.5.6',
+        );
 
-        const footerWrapper = shallow(<FooterText {...{ environmentInfo }} />);
+
+
+        const footerWrapper = shallow(<FooterText {...{ environmentInfoProvider }} />);
```
smoralesd, line 18: nitpick: this could be a little nicer by creating a props object:
```ts
const props: FooterTextProps = {
    environmentInfoProvider,
};

const footerWrapped = shallow(<FooterText {...props} />);
```
This way, the next time we change the `FooterText`'s properties, the line where we shallow render it will not change.

src/tests/unit/tests/reports/components/report-sections/report-body.test.tsx
```diff
@@ -26,11 +27,11 @@ describe('ReportBody', () => {
             pageUrl,
             description: 'test description',
             scanDate: new Date('2019-05-29T19:12:16.804Z'),
-            environmentInfo: {
-                axeCoreVersion: 'axe-core-version',
-                browserSpec: 'browser-spec',
-                extensionVersion: 'extension-version',
-            },
+            environmentInfoProvider: new EnvironmentInfoProvider(
```
smoralesd, line 30: Same here as in the other test: this should be a mock, not an instance of the actual class 'FooterText' has a dependency on.

src/tests/unit/tests/reports/package/footer-text-for-service.test.tsx
```diff
@@ -1,21 +1,21 @@
...
 import { FooterTextForService } from 'reports/package/footer-text-for-service';
 
 describe('FooterTextForService', () => {
     it('renders', () => {
-        const environmentInfo: EnvironmentInfo = {
-            extensionVersion: '1.2.3',
-            browserSpec: 'dummy browser version 1.0.1',
-            axeCoreVersion: '4.5.6',
-        };
+        const environmentInfoProvider: EnvironmentInfoProvider = new EnvironmentInfoProvider(
```
smoralesd, line 10: Same here as in the other test: this should be a mock, not an instance of the actual class 'FooterText' has a dependency on.

src/tests/unit/tests/reports/package/footer-text-for-service.test.tsx
```diff
@@ -1,21 +1,21 @@
...
+            '1.2.3',
+            'dummy browser version 1.0.1',
+            '4.5.6',
+        );
 
         const FooterText = FooterTextForService('ClientService');
 
-        const footerWrapper = shallow(<FooterText {...{ environmentInfo }} />);
+        const footerWrapper = shallow(<FooterText {...{ environmentInfoProvider }} />);
```
smoralesd, line 18: same nitpick here: create a `props` const to hold this values so this line doesn't change unnecessarily in the future.

src/tests/unit/tests/reports/report-html-generator.test.tsx
```diff
@@ -40,11 +40,11 @@ describe('ReportHtmlGenerator', () => {
         const getGuidanceTagsStub: GetGuidanceTagsFromGuidanceLinks = () => [];
 
         const sectionFactoryMock = Mock.ofType<ReportSectionFactory>();
-        const environmentInfo: EnvironmentInfo = {
-            axeCoreVersion,
-            browserSpec,
+        const environmentInfoProvider: EnvironmentInfoProvider = new EnvironmentInfoProvider(
```
smoralesd, line 43: Same here as in the other test: this should be a mock, not an instance of the actual class 'FooterText' has a dependency on.

src/reports/package/footer-text-for-service.tsx
```diff
@@ -2,18 +2,27 @@
...
-        <>
-            This automated checks result was generated using the {serviceName}{' '}
-            that helps find some of the most common accessibility issues. The scan was
-            performed using axe-core {axeCoreVersion} and {browserSpec}. For a complete
-            WCAG 2.1 compliance assessment please visit{' '}<ToolLink />.
-        </>;
+    const footerTextForService = NamedFC<FooterTextForServiceProps>(
+        'FooterText',
+        ({ environmentInfoProvider }) => {
+            const environmentInfo = environmentInfoProvider.getEnvironmentInfo();
```
waabid, line 15: We shouldn't be using `getEnvironmentInfo' but rather `getToolData`.

src/reports/package/footer-text-for-service.tsx
```diff
@@ -2,18 +2,27 @@
...
+            return (
+                <>
+                    This automated checks result was generated using the {serviceName}{' '}
```
waabid, line 18: Please update the text here from what is in the Figma.

src/tests/unit/tests/reports/package/footer-text-for-service.test.tsx
```diff
@@ -1,21 +1,38 @@
...
         };
 
         const FooterText = FooterTextForService('ClientService');
 
-        const footerWrapper = shallow(<FooterText {...{ environmentInfo }} />);
+        const footerWrapper = shallow(<FooterText {...footerTextForServiceProps} />);
         expect(footerWrapper.getElement()).toMatchSnapshot('footer');
     });
+
+    afterEach(() => {
```
waabid, line 35: I know this works but if I remember correctly, we decided to not verify within the afterEach blocks?

@smoralesd can maybe add to that.

smoralesd: Good catch. That's right, it's better not to break the AAA steps from the unit test, all of the steps should explicitly be part of the test itself.

src/reports/package/footer-text-for-service.tsx
```diff
@@ -2,18 +2,27 @@
 // Licensed under the MIT License.
 import * as React from 'react';
 
-import { ReportSectionFactory } from 'reports/components/report-sections/report-section-factory';
+import { EnvironmentInfoProvider } from 'common/environment-info-provider';
+import { NamedFC } from 'common/react/named-fc';
 import { ToolLink } from 'reports/components/report-sections/tool-link';
 
+export type FooterTextForServiceProps = { environmentInfoProvider: EnvironmentInfoProvider };
```
smoralesd, line 9: Original comment here:
`This is duplicated from 'src/reports/components/report-sections/footer-text.tsx', let's avoid this by defining the type in one place only and importing where we need to use it.`
The general idea is not about the name, it's about the how this components are related. Both are instances being use on `ReportSectionFactory.FooterText` with type `ReactFCWithDisplayName<Pick<SectionProps, 'environmentInfo'>>`; that's why the original code used `ReportSectionFactory['FooterText']`. That way, both comment refer to the same type, if you change the type, compilation will tell you you broke the two components. By creating a new type here (with a different or the same name but having the same properties for now) you blind yourself to this scenario, i.e. if something change on the type we are using on the other component, the compiler won't tell you you broke this component. You will only find out about the error when running the code. I don't think we have enough test coverage to catch that, since that's not something you can catch easily using unit tests and I'm pretty sure we don't have end to end test for exporting reports from AI-Web and the report package (I may be wrong).
My suggestion will be import `FooterTextProps` from `src\reports\components\report-sections\footer-text.tsx` that way we only have one type for the props object.

src/reports/package/footer-text-for-service.tsx
smoralesd: Reopen and add a new comment on the updated file.

src/reports/package/footer-text-for-service.tsx
smoralesd: I don't see any changes on the `displayName` since I left the original comment (that would be `'FooterText'`). I'm not sure   it's a good idea to keep both components using the same `displayName` and considering the change is easy enough, I rather have it updated.

src/tests/unit/tests/reports/components/report-sections/report-body.test.tsx
```diff
@@ -18,6 +19,16 @@ describe('ReportBody', () => {
         const getScriptStub = () => '';
         const getGuidanceTagsStub = () => [];
         const fixInstructionProcessorMock = Mock.ofType(FixInstructionProcessor);
+        const environmentInfoProviderMock = Mock.ofType(EnvironmentInfoProvider);
+        const environmentInfo: EnvironmentInfo = {
```
smoralesd, line 23: I'm not sure this is needed on this test. All `ReportBody` is doing with this props is passing them down to the appropriate child components. There is not a call to `environmentInfoProvider.getEnvironmentInfo()` in this class. Can you try remove this and the mock setup and see if that still work without other changes?

src/tests/unit/tests/reports/report-html-generator.test.tsx
```diff
@@ -40,11 +40,16 @@ describe('ReportHtmlGenerator', () => {
         const getGuidanceTagsStub: GetGuidanceTagsFromGuidanceLinks = () => [];
 
         const sectionFactoryMock = Mock.ofType<ReportSectionFactory>();
+        const environmentInfoProviderMock = Mock.ofType(EnvironmentInfoProvider);
         const environmentInfo: EnvironmentInfo = {
```
smoralesd, line 44: This case seems similar to the test for `ReportBody`: I don't see any call to `environmentInfoProvider.getEnvironmentInfo()` inside `ReportHtmlGenerator`, which means, there is no need to set up a call to `getEnvironmentInfo()` on the mock, neither we need the `environmentInfo` anymore.

PR #2467, opened 2020-04-08
feat(android-report): refactor scan timestamp into UnifiedScanResultStore

pownkel:
#### Description of changes

Store scan timestamp in UnifiedScanResultStore and use this timestamp when generating report instead of the one from VisualizationScanResultData.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/platform/android/scan-results.ts
```diff
@@ -60,6 +60,10 @@ export class ScanResults {
         return this.rawData?.axeContext?.axeMetaData?.axeVersion || 'no-version';
     }
 
+    public get analysisTimestamp(): string {
+        return this.rawData?.axeContext?.axeMetaData?.analysisTimestamp || null;
```
smoralesd, line 64: this doesn't seem to have unit test

src/common/date-provider.ts
```diff
@@ -4,7 +4,11 @@ import { utc } from 'moment';
 
 export class DateProvider {
     public static getDateFromTimestamp(timestamp: string): Date {
-        return new Date(timestamp);
+        const dateNumber = Number(timestamp);
```
smoralesd, line 7: At first read, I got a little confused with this code and when one and the other scenario happen. I got a better understanding by looking at the unit tests. I think it's like this: on scenario is when we get a string formatted as a date and the other one is wen we get and actual timestamp (like a plain number, usually a lot of digits).
Would it be better if we have some private helper functions with names that help to better understand why are we checking what we are checking?

PR #2469, opened 2020-04-09
feat(android-report): uses the automated check report as is unified behind feature flag

waabid:
#### Description of changes

The title generally covers the changes: we decided to do minimal changes to report to get it in the app first (behind a feature flag) and this makes that happen.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/views/automated-checks/automated-checks-view.tsx
+++ b/src/electron/views/automated-checks/automated-checks-view.tsx
@@ -6,6 +6,7 @@ import { GetCardViewData } from 'common/rule-based-view-model-provider';
 import { CardSelectionStoreData } from 'common/types/store-data/card-selection-store-data';
 import { CardsViewModel } from 'common/types/store-data/card-view-model';
 import { DetailsViewStoreData } from 'common/types/store-data/details-view-store-data';
+import { FeatureFlagStoreData } from 'common/types/store-data/feature-flag-store-data';
 import { UnifiedScanResultStoreData } from 'common/types/store-data/unified-data-interface';
 import { UserConfigurationStoreData } from 'common/types/store-data/user-configuration-store';
 import { CardsView, CardsViewDeps } from 'DetailsView/components/cards-view';
@@ -24,6 +25,8 @@ import { DeviceDisconnectedPopup } from 'electron/views/device-disconnected-popu
 import { ScreenshotView } from 'electron/views/screenshot/screenshot-view';
 import { ScreenshotViewModelProvider } from 'electron/views/screenshot/screenshot-view-model-provider';
 import * as React from 'react';
+import { ReportGenerator } from 'reports/report-generator';
+
 import * as styles from './automated-checks-view.scss';
 import { CommandBar, CommandBarDeps } from './components/command-bar';
 import { HeaderSection } from './components/header-section';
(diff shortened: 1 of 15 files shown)


src/electron/views/automated-checks/automated-checks-view.tsx
```diff
@@ -70,17 +78,26 @@ export class AutomatedChecksView extends React.Component<AutomatedChecksViewProp
                 unifiedScanResultStoreData,
                 cardSelectionViewData.highlightedResultUids,
             );
+            reportHTML = deps.reportGenerator.generateFastPassAutomatedChecksReport(
+                deps.getCurrentDate(),
+                unifiedScanResultStoreData.targetAppInfo.name,
+                null,
```
smoralesd, line 84: not for this PR (most likely): this `null` values here, does it make sense to rearrange the params and make this values optional or otherwise refactor how we pass data to this functions? I think the first null is related to the url, so I think you already have this in mind. Just leaving a comment so we keep this under our radar.

src/electron/views/automated-checks/automated-checks-view.tsx
```diff
@@ -70,17 +78,26 @@ export class AutomatedChecksView extends React.Component<AutomatedChecksViewProp
                 unifiedScanResultStoreData,
                 cardSelectionViewData.highlightedResultUids,
             );
+            reportHTML = deps.reportGenerator.generateFastPassAutomatedChecksReport(
```
smoralesd, line 81: I'm not sure this is the best place to generate the report. It looks to me this will generate the report every time we render the `automated-checks-view` component without the user input which may or may not impact performance. Is this the strategy we are using on AI-Web?

src/electron/views/renderer-initializer.ts
```diff
@@ -351,6 +362,23 @@ getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch).then(
 
         const documentManipulator = new DocumentManipulator(document);
 
+        const reportHtmlGenerator = new ReportHtmlGenerator(
+            UnifiedReportSectionFactory,
+            new ReactStaticRenderer(),
+            { extensionVersion: '1.1', browserSpec: "doesn't exist", axeCoreVersion: '0.16' },
```
smoralesd, line 368: I guess this may have conflicts with the other PR that refactor how we pass environment info, right? Just a heads up in case you forgot.

src/electron/views/automated-checks/automated-checks-view.tsx
```diff
@@ -6,6 +6,7 @@ import { GetCardViewData } from 'common/rule-based-view-model-provider';
 import { CardSelectionStoreData } from 'common/types/store-data/card-selection-store-data';
```
waabid, line 6: Changes in this file are hardly ideal but this class is in a strong need of a re-factor that I did not want to do as a part of this PR.

src/electron/views/automated-checks/automated-checks-view.tsx
waabid: No longer there 😄

src/electron/views/automated-checks/automated-checks-view.tsx
waabid: Agreed. Made the change to consume report export component which was more than I wanted to do but we also get the dialog now 😄

src/electron/views/automated-checks/automated-checks-view.tsx
waabid: Actually still kind of there but yes, we need to refactor this in this/next week or so.

PR #2487, opened 2020-04-13
feat(android-report): created/re-factored header section for unified report 

AhmedAbdoOrtiga:
#### Description of changes
- Added a new header section for unified report.
- Updated HeaderSection in ReportSectionFactory to: 
  `HeaderSection: ReactFCWithDisplayName<SectionProps> | ReactFCWithDisplayName;`

**Web:**
image

**Unified**
image

#### Pull request checklist

- [x] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`


src/reports/components/report-sections/report-section-factory.tsx
```diff
@@ -36,7 +36,7 @@ export type ReportSectionFactory = {
     HeadSection: ReactFCWithDisplayName;
     BodySection: ReactFCWithDisplayName;
     ContentContainer: ReactFCWithDisplayName;
-    HeaderSection: ReactFCWithDisplayName<SectionProps>;
+    HeaderSection: ReactFCWithDisplayName<SectionProps> | ReactFCWithDisplayName;
```
waabid, line 39: Is this change really necessary? As in, do the typings break if that's not the case?

waabid: I'm just thinking an empty set of props should be a subset of SectionProps so as a result, this should just work, no?

AhmedAbdoOrtiga: actually it's not necessary

src/electron/views/report/unified-header-section.tsx
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License.
+import { NamedFC } from 'common/react/named-fc';
+import { androidAppTitle } from 'content/strings/application';
+import { BrandWhite } from 'icons/brand/white/brand-white';
+import * as React from 'react';
+
+export const UnifiedHeaderSection = NamedFC('UnifiedHeaderSection', () => {
+    return (
+        <header>
+            <div className="report-header-bar">
```
waabid, line 11: A couple things here:
1. I'd try to use css modules for all styles here.
2. I would also remove the ms-* styles as we should be clear about what styles we're using and why.

src/common/styles/fonts.scss
```diff
@@ -79,6 +79,13 @@ $semiBoldFontFamily: make-font-family($semiBoldFonts) !default;
     line-height: 24px;
 }
 
+@mixin text-style-header-s {
```
waabid, line 82: the name should probably be headers instead of header-s

waabid:
Could you include a screenshot for web as well?

src/common/styles/fonts.scss
AhmedAbdoOrtiga: this style is mainly for the text font style, so I think it should be as it is, same for text-style-title-* and text-style-body-*

PR #2492, opened 2020-04-14
feat(android-report): minor improvements to passed section

waabid:
#### Description of changes

A couple changes:
1. Rules are not rendered as links when there is no url in passed/not-applicable section.
2. Rules that do not have guidance will not have empty parenthesis where that guidance link would go.

These changes apply to Web and the not applicable section as well. However, since we do not show that yet and both changes will not realistically be shown in Web, it doesn't "really" apply, if that makes sense. As in, no user should see them but since they are shared components, they changes are there.

Before on the left, After on the right:
image

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/report-sections/full-rule-header.tsx
+++ b/src/reports/components/report-sections/full-rule-header.tsx
@@ -6,9 +6,8 @@ import { NewTabLink } from 'common/components/new-tab-link';
 import { GetGuidanceTagsFromGuidanceLinks } from 'common/get-guidance-tags-from-guidance-links';
 import { NamedFC } from 'common/react/named-fc';
 import { CardRuleResult } from 'common/types/store-data/card-view-model';
-import { kebabCase } from 'lodash';
+import { isEmpty, kebabCase } from 'lodash';
 import * as React from 'react';
-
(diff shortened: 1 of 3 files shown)


src/reports/components/report-sections/full-rule-header.tsx
```diff
@@ -44,21 +43,29 @@ export const FullRuleHeader = NamedFC<FullRuleHeaderProps>('FullRuleHeader', pro
...
-                    aria-label={`rule ${ruleId}`}
-                    aria-describedby={ariaDescribedBy}
-                >
-                    {ruleId}
-                </NewTabLink>
-            </span>
+        const displayedRule = ruleUrl ? (
+            <NewTabLink
+                href={ruleUrl}
+                aria-label={`rule ${ruleId}`}
```
karanbirsingh, line 49: I see that this PR did not introduce the aria-attributes here (they were already there), but I am curious about why they differ from other places we render the rule links (like [here](https://github.com/microsoft/accessibility-insights-web/blob/3459929849613cd7403afd8b568594a99a950361/src/common/components/cards/rule-resources.tsx#L38)). Is one approach better to follow?

PR #2493, opened 2020-04-14
feat(android-report): add connected device to details section

pownkel:
#### Description of changes

Created a new component for the scan details in component report, which includes device/app information:

electron-report-scan-details

This also required a bit of additional refactoring to ReportGenerator.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/views/report/unified-details-section.tsx
```diff
@@ -0,0 +1,43 @@
...
+            <span className="icon" aria-hidden="true">
+                {icon}
+            </span>
+            <span className="screen-reader-only">{label}</span>
+            <span className={css('text', contentClassName)}>{content}</span>
+        </li>
+    );
+
+    const scanDateUTC: string = toUtcString(scanDate);
+    const showCommentRow = !!description && description !== '';
```
karanbirsingh, line 30: Would `!!description` work here without the additional empty string check?

src/electron/views/report/unified-details-section.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    return (
+        <div className="scan-details-section">
+            <h2>Scan details</h2>
+            <ul className="details-section-list">
+                {createListItem(<UrlIcon />, 'connected device name:', targetAppInfo.device)}
```
karanbirsingh, line 36: I wonder if the `UrlIcon` should be replaced with something else for the connected device case

src/electron/views/report/unified-details-section.tsx
```diff
@@ -0,0 +1,43 @@
...
+                {createListItem(<DateIcon />, 'scan date:', scanDateUTC)}
```
karanbirsingh, line 37: Do we need the ":" here?

src/electron/views/report/unified-details-section.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { css } from '@uifabric/utilities';
+import { CommentIcon } from 'common/icons/comment-icon';
+import { DateIcon } from 'common/icons/date-icon';
+import { UrlIcon } from 'common/icons/url-icon';
+import { NamedFC } from 'common/react/named-fc';
+import * as React from 'react';
+import { DetailsSectionProps } from 'reports/components/report-sections/details-section';
+
+export const UnifiedDetailsSection = NamedFC<DetailsSectionProps>('DetailsSection', props => {
+    const { targetAppInfo, description, scanDate, toUtcString } = props;
```
karanbirsingh, line 12: For the android case, do we have anything in `targetAppInfo.name` that would be helpful to display? The code seems to reference `device`, but not name

src/electron/views/report/unified-details-section.tsx
waabid: I think we can use isEmpty here.

src/electron/views/report/unified-details-section.tsx
waabid: I'm a bit confused as well since your screenshot has the device name and the app name?

src/electron/views/report/unified-details-section.tsx
pownkel: I was following what we were doing in the web report. I'll check with Fer if that's needed.

src/common/types/store-data/unified-data-interface.ts
```diff
@@ -37,6 +37,7 @@ export interface PlatformData {
 export interface TargetAppData {
     name: string;
     url?: string;
+    device?: string;
```
waabid, line 40: This seems like it should not be here ... maybe have it at the ScanMetadata level (or another better fitting object that has this)?

waabid: I'm just thinking/saying that since target app data should contain data about the targeted application whereas device would be above that. If we do want to put this here I would heavily recommend re-naming to something else that seems to encompass all concepts?

src/electron/views/report/unified-details-section.tsx
waabid, line 36: As far as I can tell, the only difference between this and the other details-section is this one line that mentions the targetAppInfo. That likely means this is a good opportunity to re-factor and extract that first line, re-use the component, and avoid duplication.

src/electron/views/report/unified-details-section.tsx
pownkel: She said it's helpful for separation of terms, so we should keep it in

src/electron/views/report/unified-details-section.tsx
pownkel: So it turns out when we connect we're concating the device name with the app name and storing the whole thing as connectedDevice. I changed it so TargetAppInfo stores the device name from UnifiedScanResultStore and we concat that with the app name here instead (using connectedDevice from the DeviceStore was causing other problems)

src/reports/components/report-sections/make-details-section-fc.tsx
```diff
@@ -0,0 +1,71 @@
...
+                <span className="icon" aria-hidden="true">
+                    {icon}
+                </span>
+                <span className="screen-reader-only">{label}</span>
+                <span className={css('text', contentClassName)}>{content}</span>
+            </li>
+        );
+
+        const scanDateUTC: string = toUtcString(scanDate);
+        const showCommentRow = !!description && description !== '';
```
waabid, line 44: We can use isEmpty here.

src/reports/components/report-sections/make-details-section-fc.tsx
```diff
@@ -0,0 +1,71 @@
...
+        const urlOrDeviceItemInfo: ScanDetailInfo = getUrlOrDeviceItemInfo(
```
waabid, line 45: The general idea here is great; it lets us specify the content of this list item in separate components that can target either unified or web.

However, there is a bit of an issue in this class particularly: this class is generic and will be utilized in both scenarios. The device name however is only for the unified scenario, so this is making an assumption of what data we have and the reason this works fine _now_ is JS will be comfortable with null properties. However, someone coming back to this code later might not understand that deviceName might sometimes be null and that this is not a good place to touch it in any significant way.

Ideally, in this case, we shouldn't touch device name because this generic component doesn't _really_ have to know that exists or not, right? All it's doing is passing it to the `getUrlOrDeviceItemInfo` method. To resolve this, we simply don't spread the scanMetadata at this level, since that object contains properties that might not be available in all scenarios for this **generic** component.

That means modifying the `SectionProps` such that it contains the 'scanMetadata` property rather than being combined with the `ScanMetaData` type.

waabid: I'd also suggest re-naming the `getUrlOrDeviceItemInfo` method. As far as this class is concerned it doesn't matter what is returned by that method as long as it has the expected return types. If you want to be more specific, ideally we'd capture broadly what is happening here. Maybe something like: `getDisplayedScanTargetInfo` or something?

src/reports/components/report-sections/make-details-section-fc.tsx
```diff
@@ -0,0 +1,69 @@
...
+    'scanMetadata' | 'description' | 'scanDate' | 'toUtcString'
+>;
+
+export type ScanDetailInfo = {
+    label: string;
+    content: string | JSX.Element;
+};
+
+export function makeDetailsSectionFC(
+    getDisplayedScanTargetInfo: (scanMetaData: ScanMetaData) => ScanDetailInfo,
```
waabid, line 24: You're using this type in a few places; might not be a bad idea to name it and have it separately.

PR #2495, opened 2020-04-14
refactor: rename exportReport feature flag

pownkel:
#### Description of changes

The feature flag on web to enable exporting to CodePen currently has the same name as the feature flag to enable the export report button in electron. As a result, turning on this feature flag in electron automatically also enables exporting to CodePen, which we do not want. This PR renames the CodePen feature flag from exportReport to exportReportOptions

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/electron/common/unified-feature-flags.test.ts
+++ b/src/tests/unit/tests/electron/common/unified-feature-flags.test.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { FeatureFlagDefaultsHelper } from 'common/feature-flag-defaults-helper';
+import { getAllFeatureFlagDetails } from 'common/feature-flags';
 import { FeatureFlagStoreData } from 'common/types/store-data/feature-flag-store-data';
 import {
     getAllFeatureFlagDetailsUnified,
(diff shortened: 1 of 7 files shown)


karanbirsingh:
We had a similar issue in our assessment test steps & added [this unit test](https://github.com/microsoft/accessibility-insights-web/blob/3459929849613cd7403afd8b568594a99a950361/src/tests/unit/tests/assessments/assessments-style-cop.test.ts#L81) to prevent it from happening again. Maybe something like that would be helpful here too, particularly if the web & electron feature flags are in different places

src/tests/unit/tests/electron/common/unified-feature-flags.test.ts
```diff
@@ -82,4 +83,18 @@ describe('FeatureFlagsTest', () => {
...
+
+    test('web and electron feature flags with matching names should have the same purpose', () => {
+        const unifiedDetails = getAllFeatureFlagDetailsUnified();
+        const webDetails = getAllFeatureFlagDetails();
+
+        unifiedDetails.forEach(unifiedDetail => {
+            const matchingWebDetail = webDetails.find(
+                webDetail => webDetail.id === unifiedDetail.id,
+            );
+            if (matchingWebDetail !== undefined) {
```
waabid, line 95: Any reason we can't be more assertive here? Like comparing the values of the objects entirely?

pownkel: Added an assert on displayableDescription as well, but I don't think we would need to require the other values in the object (defaultValue, isPreviewFeature, and forceDefault) to be the same between web and electron.

PR #2500, opened 2020-04-15
feat(android-report): Add Time to unified report name

AhmedAbdoOrtiga:
#### Description of changes
- In AIWeb while saving the generated report, the browser automatically append a number to the file name if there's a file with the same name.
- This is not the case in electron.
-  We agreed to append the time to the unified report file name.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/views/report/unified-report-name-generator.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { ReportNameGenerator } from 'reports/report-name-generator';
+
+export class UnifiedReportNameGenerator extends ReportNameGenerator {
```
waabid, line 5: We should favor composition over inheritance and I feel like this case lends itself well to composition. Let's discuss offline.

src/reports/report-generator.ts
```diff
@@ -7,13 +7,14 @@ import { FeatureFlagStoreData } from 'common/types/store-data/feature-flag-store
...
 
+import { UnifiedReportNameGenerator } from 'electron/views/report/unified-report-name-generator';
 import { AssessmentReportHtmlGenerator } from './assessment-report-html-generator';
 import { ReportHtmlGenerator } from './report-html-generator';
 import { ReportNameGenerator } from './report-name-generator';
 
 export class ReportGenerator {
     constructor(
-        private reportNameGenerator: ReportNameGenerator,
+        private reportNameGenerator: ReportNameGenerator | UnifiedReportNameGenerator,
```
waabid, line 17: This shouldn't be necessary if you use an interface here instead (since they are depending on the same method and types).

src/electron/views/report/unified-report-name-generator.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { ReportNameGeneratorBuilder } from 'reports/report-name-generator-builder';
+
+export class UnifiedReportNameGenerator {
+    constructor(
+        private readonly reportNameGeneratorBuilder: ReportNameGeneratorBuilder = new ReportNameGeneratorBuilder(),
```
waabid, line 7: nit: We usually avoid making the object itself responsible for creating things it's dependent on. In this case, I understand it's basically going to move this line to the initializers but we should try to remain consistent.

waabid: The same applies to the other generator.

waabid: This hasn't yet been resolved, so opening it again.

AhmedAbdoOrtiga: it's just for setting the default value, since the class is kinda static, plus the constructor takes no parameters and needs no special initiating.
I'll close for now, we can update it in another PR tho

PR #2503, opened 2020-04-15
refactor: use clearer naming for export format vs service

dbjorge:
#### Description of changes

This PR is strictly refactoring, no functionality changes.

Before this PR, we have types for the following:

* `ExportResultsType`: is it an assessment report or a fastpass report
* `ExportFormat`: is the user downloading the report as a file, or sending the report to codepen

I found this confusing when rereading the code, since the former concept feels much more like a "format" to me than the latter concept. The types' naming were also a little inconsistent with other types in this space, which most commonly used use `ReportExport` as the prefix rather than `ExportResults` or `Export` (eg, the `report-export` top level folder name for the `ReportExportService` that `CodePenReportExportService` implements).

This PR renames those two types to `ReportExportFormat` and `ReportExportServiceKey`, respectively. It updates the related variable names, **with the intentional exception of the property name sent to telemetry**, to avoid breaking correlatability with past telemetry.

There are also a few more tightly scoped renames in the `code-pen-report-export-service`:
* Renamed files from `codepen` to `code-pen` prefix to match type `CodePen`-like type names ("CodePen" is the canonical way to capitalize the service)
* Updated a `const` from PascalCase to camelCase for consistency
* Used `type` instead of `interface`
* Standardized a few other report export service related types on `ReportExport*` instead of `Export*` type names

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/report-export/types/report-export-service.ts
```diff
@@ -1,18 +1,18 @@
...
     description: string;
-}
+};
 
-export interface ReportExportService {
-    key: ExportFormat;
+export type ReportExportService = {
+    key: ReportExportServiceKey;
     displayName: string;
-    exportForm: React.ComponentType<ExportFormProps> | null;
```
karanbirsingh, line 17: Sanity-checking that the union type removal is intentional - AFAIK this doesn't affect behavior given that we don't seem to use `strictNullChecks`. Is that right?

dbjorge: That is right, but also a bigger hammer than necessary here; we only have one implementation of a service and it doesn't ever set this to null :)

PR #2511, opened 2020-04-17
chore: fix mislabeled dependecies

dbjorge:
#### Description of changes

I audited our list of dependencies and devDependencies in the root package.json; since we don't publish an npm package using this package file, the distinction doesn't matter technically, but we use the distinction during policy/compliance tasks to keep track of what we "distribute" vs not for the sake of legal requirements, so we want the lists to be accurate for that purpose.

Specifically:
* `electron` was listed as a devDependency but is actually a dependency
* `7zip-bin`, `electron-download`, `react-devtools`, and `unzipper` were listed as dependencies, but are not actually distributed in web or unified. `react-devtools` is a special case; it is bundled with dev builds of the extension but not release builds. We don't distribute built forms of the dev builds, so it's a devDependency for license obligation purposes.

#### Pull request checklist

- [x] Ran `yarn fastpass`


dbjorge:
(note: PR build has passed, it's just not reporting correctly to github)

PR #2515, opened 2020-04-19
tools: telemetry viewer

smoralesd:
#### Description of changes

Adding a new section (not hook up yet though) to be able to visualize telemetry events from the extension, displaying them on a list (`<DetailsList>` to be more precise).

**no telemetry messages yet**
01 - no telemetry messages yet

**telemetry message on a list**
02 - with telemetry messages on display

**Future work**
- There is a missing styling on the content border. I think we do have a small (5px maybe) border around the content (except for the header itself). I'd like to adjust that styling as part of a future hook up PR
- Hook up the view into the debug tools container so it actually shows up. I didn't want to include that part in this PR as it will introduce a new UI piece (probably a nav bar so the user can navigate between the different sections)
- Filtering events
- Better visualization for custom properties
- Start/stop registering events: currently, the telemetry messages are stored on an array on the viewer's state. I was a little concerned about RAM usage as there is no limit at the amount of messages the state will keep. I though it may be nice to add a start/stop feature and/or maybe put a maximum size to the array but left both options out for now

#### Pull request checklist
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/debug-tools/components/telemetry-viewer/telemetry-common-fields.scss
+++ b/src/debug-tools/components/telemetry-viewer/telemetry-common-fields.scss
@@ -1,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+.container {
+    display: grid;
+    grid-template-columns: 7em auto;
+    column-gap: 1em;
+
+    > strong {
+        justify-self: end;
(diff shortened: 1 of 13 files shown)


dbjorge:
Merging now having verified that the PR build succeeded (just failed to report to github)

Issue #2418, opened 2020-03-30
Automated checks automatically selected after start over assessment


smoralesd:
## Describe the bug

After the user selects **start over assessment**, automated checks gets selected.

**To Reproduce**

1. navigate to any website
1. start a new assessment
1. select any test but automated checks (neither overview), e.g. Keyboard
1. use the start over assessment option
1. Observe: the selected test gets updated from what the user selected before and now the automated checks is selected

## CodePen repro example

N/A

## Expected behavior

Not sure what the expected behavior should be here but I do see a couple of options:
1. The selected test is left unchanged
1. The selection is update to **Overview**

## Screenshots

![01 - automated checks automatically selected after start over](https://user-images.githubusercontent.com/2837582/77967444-fd181c80-7299-11ea-9912-d7bde3f6cbec.gif)

## Context (please complete the following information)

- OS Name & Version: Windows Version 10.0.18363 Build 18363
- AI-Web Version & Environment: Prod 2.15.0 & Insider 2.16.0
- Browser Version: Google Chrome  80.0.3987.149 (Official Build) (64-bit)
- Target Page: N/A

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

Yes

## Additional context

N/A

ferBonnin:
Per conversation with Peter, let's fix this

ferBonnin:
Validated with Canary 2020.4.27.1939

PR #2522, opened 2020-04-21
fix: go to overview when restarting assessment

waabid:
#### Description of changes

simply sets the right content panel to be overview when starting over the entire assessment.

startovertooverview

#### Pull request checklist

- [x] Addresses an existing issue: #2418 
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/actions/details-view-action-message-creator.ts
+++ b/src/DetailsView/actions/details-view-action-message-creator.ts
@@ -500,12 +500,18 @@ export class DetailsViewActionMessageCreator extends DevToolActionMessageCreator
 
     public startOverAllAssessments = (event: React.MouseEvent<any>): void => {
         const telemetry = this.telemetryFactory.fromDetailsView(event);
-        const payload: BaseActionPayload = {
+        const setDetailsViewRightContentPanelPayload: DetailsViewRightContentPanelType = 'Overview';
+        const startOverAllAssessmentsActionPayload: BaseActionPayload = {
             telemetry: telemetry,
         };
+
+        this.dispatcher.dispatchMessage({
+            messageType: Messages.Visualizations.DetailsView.SetDetailsViewRightContentPanel,
+            payload: setDetailsViewRightContentPanelPayload,
+        });
         this.dispatcher.dispatchMessage({
             messageType: Messages.Assessment.StartOverAllAssessments,
-            payload,
+            payload: startOverAllAssessmentsActionPayload,
         });
     };
 
--- a/src/tests/unit/tests/DetailsView/actions/details-view-action-message-creator.test.ts
+++ b/src/tests/unit/tests/DetailsView/actions/details-view-action-message-creator.test.ts
@@ -380,11 +380,15 @@ describe('DetailsViewActionMessageCreatorTest', () => {
             source: TelemetryEventSource.DetailsView,
         };
 
-        const expectedMessage = {
+        const expectedMessageToStartOverAllAssessments = {
             messageType: Messages.Assessment.StartOverAllAssessments,
             payload: {
                 telemetry,
             },
+        };
+        const expectedMessageToSetDetailsViewRightContentPanel = {
+            messageType: Messages.Visualizations.DetailsView.SetDetailsViewRightContentPanel,
+            payload: 'Overview',
         };
 
         telemetryFactoryMock.setup(tf => tf.fromDetailsView(event)).returns(() => telemetry);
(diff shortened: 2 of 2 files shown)


src/DetailsView/actions/details-view-action-message-creator.ts
```diff
@@ -500,12 +500,18 @@ export class DetailsViewActionMessageCreator extends DevToolActionMessageCreator
...
+        const setDetailsViewRightContentPanelPayload: DetailsViewRightContentPanelType = 'Overview';
+        const startOverAllAssessmentsActionPayload: BaseActionPayload = {
             telemetry: telemetry,
         };
+
+        this.dispatcher.dispatchMessage({
+            messageType: Messages.Visualizations.DetailsView.SetDetailsViewRightContentPanel,
+            payload: setDetailsViewRightContentPanelPayload,
+        });
         this.dispatcher.dispatchMessage({
```
dbjorge, line 512: The ordering in which these 2 messages are processed is important, right (to avoid flashing the UI)? Does that suggest it would make more sense to handle the ordering on the background page rather than the details view page, to avoid depending on Chromium maintaining the ordering reliably across the message boundary?

waabid: That was my first thought as well. I decided against it due to the changes that it would necessitate since the assessment store update and the view update are in separate contexts (global vs tab).

Your comment made me re-investigate that solution and there might be something that we can do ... although it involves the details view action creator invoking assessmentActions.

waabid: Okay, yeah. I am pretty confident that we'd require some pretty heavy finagling to work in the background side.

I had some brief optimism that I might be able to do something when I noticed the actionsHub in `TabContextFactory` seemed to have `AssessmentActions` inside. Unfortunately, those are not the same actions that the global assessment store is hooked up to, and as a result, nothing happens if you invoke them. This also begs the question why do we have them inside the non-global actions hub (but I guess that's clean-up for the future 😞).

Effectively speaking, we'd need access to tab-specific details view actions within the assessment action creator or vice versa. I think the latter makes more sense: give access to the tab stores to enact changes on the global level since their life-cycle extends beyond the tab context. Doing so would require:

1. exposing the global actions hub from global context.
2. making the target page controller depend on the actions such that it can be passed to each tab context at creation.
3. in the tab context factory, the details view action creator* needs to be depend on the assessment actions and then enact what we want inside on the message received.

* this will likely need to be re-factored since it's being used in unified, which does not have assessment anything. 

**Unless** I'm missing something, this would be the approach to take. Personally, I'd go with what I have in this branch as is. Actually, I found there to be no flickering even with the statements switched and if we have users run into it, then that time might be a good idea to enact this kind of change.

Let me know what you think!

(Also: there's an approach to listen to the same assessment message in detailsviewactioncreator, but that completely loses any sense of ordering, so I figured that was a no-go)

waabid: Okay, yeah. I am pretty confident that we'd require some pretty heavy finagling to work in the background side.

I had some brief optimism that I might be able to do something when I noticed the actionsHub in `TabContextFactory` seemed to have `AssessmentActions` inside. Unfortunately, those are not the same actions that the global assessment store is hooked up to, and as a result, nothing happens if you invoke them. This also begs the question why do we have them inside the non-global actions hub (but I guess that's clean-up for the future 😞).

Effectively speaking, we'd need access to tab-specific details view actions within the assessment action creator or vice versa. I think the latter makes more sense: give access to the tab stores to enact changes on the global level since their life-cycle extends beyond the tab context. Doing so would require:

1. exposing the global actions hub from global context.
2. making the target page controller depend on the actions such that it can be passed to each tab context at creation.
3. in the tab context factory, the details view action creator* needs to be depend on the assessment actions and then enact what we want inside on the message received.

* this will likely need to be re-factored since it's being used in unified, which does not have assessment anything. 

**Unless** I'm missing something, this would be the approach to take. Personally, I'd go with what I have in this branch as is. With that said, I found there to be no flickering even with the statements switched and if we run into flickering issues in the future, then that time might be a good idea to enact this kind of change.

Let me know what you think!

(Also: there's an approach to listen to the same assessment message in detailsviewactioncreator, but that completely loses any sense of ordering, so I figured that was a no-go)

dbjorge: Thanks for the write-up! I'm okay with keeping as-is for now and reassessing if we see flickering behavior.

PR #2532, opened 2020-04-22
refactor: use scanMetadata in FailedInstancesSection

pownkel:
#### Description of changes

This will remove FailedInstancesSection's dependency on targetAppInfo, which is duplicated in SectionProps and should eventually be removed. Because FailedInstancesSection is also used in CardsView, we also need to pass ScanMetadata to CardsView in other places it is used.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/details-view-body.tsx
```diff
@@ -114,7 +114,6 @@ export class DetailsViewBody extends React.Component<DetailsViewBodyProps> {
 
     private renderRightPanel(): JSX.Element {
         const rightPanelProps: RightPanelProps = {
```
waabid, line 116: This const doesn't seem to be needed anymore.

PR #2541, opened 2020-04-23
fix(android-report): sets minimum width for export dialog to 320px

waabid:
#### Description of changes

Sets a min width to match web. Looks like this:

image

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/electron/views/renderer-global-styles.scss
+++ b/src/electron/views/renderer-global-styles.scss
@@ -15,4 +15,5 @@
 .insights-dialog-main-override {
     max-width: 25%;
     width: 25%;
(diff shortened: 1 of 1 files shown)


waabid:
> This is larger than the min-width of the whole application (400px), no? That doesn't feel right, and won't be right in web either in the near future when we'll need it to be able to scale down to 320px width for the reflow features.

Where is the 400px width declared btw? Good point about the 320px; I will switch to that! 😄

dbjorge:
Reopening PR to pick up PR build from updated devops organization

Issue #2380, opened 2020-03-23
Info & Examples panel close button is misaligned


dbjorge:
## Describe the bug

(split out from [this comment](https://github.com/microsoft/accessibility-insights-web/issues/2256#issuecomment-596767677) in #2256)

The "close" button in our Info & Examples panels renders up above the title for the panel, inconsistent with the standard fabric layout our other panels use. It should ideally be laid out more similarly to the other panels (vertically aligned with the panel header).

**To Reproduce**

1. Start an assessment in web
2. Select any test/requirement
3. Select "info & examples"
4. Observe panel layout:

![image](https://user-images.githubusercontent.com/376284/77356277-38977180-6d03-11ea-908e-fbf7d925db37.png)

## Expected behavior

Close button and panel title should be at same vertical alignment

## Context (please complete the following information)

- OS Name & Version: Windows 10 1903 18362.720
- AI-Web Version & Environment: Canary Version 2020.3.20.1613
- Browser Version: Edge stable Version 80.0.361.69
- Target Page: irrelevant

ferBonnin:
per conversation with Peter, makes sense to pick up

dbjorge:
Didn't end up completing this before next feature started, so moving back to ready for work. The bits of work I had in progress are at https://github.com/dbjorge/accessibility-insights-web/tree/content-panel-header

jalkire:
Fix merged in PR #2572

dbjorge:
Confirmed

PR #2572, opened 2020-04-28
fix: Move Info & Examples title from content to panel header

jalkire:
#### Description of changes
Issue #2380 (Info & Examples panel close button is misaligned) calls out that the close button in the info & examples panel is positioned inconsistently with our other panels. This happens because the titles for the info & examples are included in the content of the panel rather than as the header. To address the issue, this PR moves all info & examples titles to the header of the panel. `ContentCreatorWithTitle` and `ContentPageConfiguration` are introduced to accomplish this. Credit to @dbjorge for initial investigation work and the concept of the change!

Note that this PR changes a lot of infoAndExamples files. The GitHub diff shows these files as heavily modified. This is because separating the title out changes the indentation level of the entire file. The only intended/noteworthy changes to those files occur in the first 10 or so lines, where the titles are pulled out of the content.

Before on left, after on right. 
Unscrolled:
image
Scrolled:
image

#### Pull request checklist

- [x] Addresses an existing issue: #2380
- [x] Ran `yarn fastpass`


src/content/test/parsing/keyboard-bookmarklet-instructions.tsx
```diff
@@ -1,157 +1,164 @@
...
-                        In Chrome, type <Markup.Term>Ctrl + Shift + o</Markup.Term> or <Markup.Term>⌘ + Option + b</Markup.Term>.
-                    </li>
-                    <li>
-                        A new browser tab called <Markup.Term>Bookmarks</Markup.Term> will open.
-                    </li>
-                </ol>
-            </li>
+export const keyboardBookmarkletInstructions = createWithTitle(
+    {
+        pageTitle: '',
```
dbjorge, line 8: This seems off

src/views/content/content.scss
```diff
@@ -14,6 +14,16 @@
     }
 }
 
+.content-header {
+    font-size: 28px;
```
dbjorge, line 18: Maybe better would be to create a `@mixin text-style-title-l` for this in `fonts.scss`

src/tests/unit/tests/views/content/__snapshots__/content-panel.test.tsx.snap
```diff
@@ -1,23 +1,23 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`ContentPanel renders closed 1`] = `
-"<StyledPanelBase isOpen={false} onDismiss={[undefined]} type={3} isLightDismiss={true} closeButtonAriaLabel=\\"Close panel\\">
+"<StyledPanelBase isOpen={false} onDismiss={[undefined]} type={3} isLightDismiss={true} closeButtonAriaLabel=\\"Close panel\\" headerText={[undefined]} headerTextProps={{...}} headerClassName=\\"content-header\\">
```
dbjorge, line 4: It would be nice to update the test to pass a more representative value for the headerText. Maybe also nice would be to somehow write a test case verifying that all the infoAndExample content does actually specify a pageTitle, but not sure how challenging that would be to implement.

dbjorge:
(plus need to update e2e snapshots)

src/content/test/parsing/keyboard-bookmarklet-instructions.tsx
jalkire: This is not info and examples content, so I have undone all changes to the file. Looks like it was the only one that didn't belong. Good catch!

src/views/content/content.scss
jalkire: Introduced and used `text-style-title-l` mixin

src/tests/unit/tests/views/content/__snapshots__/content-panel.test.tsx.snap
jalkire: Updated the test to pass in a value for pageTitle/headerText. I'm thinking about ways to verify the content--I'll open a new PR if I come up with anything

PR #2581, opened 2020-04-29
feat(electron upgrade): Remove remote usage

DaveTryon:
#### Description of changes

Electron's `remote` class comes with a set of issues, as described in [this blog post](https://github.com/electron/electron/issues/21408) and [this video](https://www.youtube.com/watch?v=0lb7AxaucZI&list=PLhSaNgmcIodCuiyY5fHIH-dIm9FhrNj9X&index=9&t=0s), and the recommendation is to discontinue using `remote`. We're currently using `remote` to get the app version from the main process, and to give the renderer process an object to access the main process's `BrowserWindow` object.

The core change is to explicitly route all of the code that used `remote` through Electron's IPC layer. I've pulled all of the awareness of the IPC into 2 classes--one in the main process and one in the renderer process. These 2 classes, taken together, replace the way we were previously using `remote`.

One note--our flux layer explicitly prevents nested calls to `Action.invoke()` within a given scope, but there's no way that a caller can specify any scope other than the default value. I added a way for a caller of `Action.invoke()` to specify a non-default scope. This seemed consistent with the way the code was structured, but it seems like something that needs to get broader buy-off from the team.

I've validated on Windows, but since this code involves some Mac-specific behavior (maximize vs. full screen), someone with a Mac will need to validate that it's still working as expected.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/main/main.ts
```diff
@@ -56,6 +58,13 @@ const createWindow = () => {
         nativeHighContrastModeListener.startListening();
     });
 
+    mainWindowRendererMessageHandlers = new MainWindowRendererMessageHandlers(
+        app,
+        mainWindow,
+        ipcMain,
+    );
+    mainWindowRendererMessageHandlers.startListening();
```
dbjorge, line 66: Should this be waiting to set up listeners until the IPC_MAIN_WINDOW_INITIALIZED_CHANNEL_NAME event, like the native HC listener works?

DaveTryon: Curiously, that didn't work. I think it's because the getVersion() call is synchronous and causing a deadlock. The main window object has been created at this point, so it's safe to install the listeners before the window is fully initialized.

src/electron/ipc/ipc-renderer-shim.ts
```diff
@@ -0,0 +1,74 @@
...
+    public readonly unmaximizeEvent = new Action<void>();
+    public readonly enterFullScreenEvent = new Action<void>();
+
+    // Call these methods to send data FROM renderer process
+    public initializeWindow = (): void => {
+        this.ipcRenderer.send(IPC_MAIN_WINDOW_INITIALIZED_CHANNEL_NAME);
+    };
+
+    public getVersion = (): string => {
+        return this.ipcRenderer.sendSync(IPC_APP_VERSION_CHANNEL_NAME);
```
karanbirsingh, line 52: It doesn't seem like this is worse than remote, but I wonder if we're able to use invoke (async) instead - or whether we should treat it similar to `getPersistedData` in `renderer-initializer.ts`. There's a warning [here](https://www.electronjs.org/docs/api/ipc-renderer#ipcrenderersendsyncchannel-args) about `sendSync` blocking the renderer process like the remote module did. In this case the async signature would bubble up to its usage in `scan-controller.ts`.

DaveTryon: I'm not a fan of the sync option, but it was the "least bad" option I could find. Making the call async means changing the signature of the `AppDataAdapter` interface, and that is used in lots of places, including many that are unrelated to Electron. It's not persisted data, either, since it's a function of the bits instead of the user store. I had hoped to make it async but to find a way to block the calling thread (like .NET's `Thread.Wait` method), but that concept isn't supported in JavaScript.

src/common/flux/action.ts
```diff
@@ -12,21 +12,22 @@ export class Action<TPayload> {
     private listeners: ((payload: TPayload) => void)[] = [];
     private scope: string = 'DEFAULT_SCOPE';
 
-    public invoke(payload: TPayload): void {
-        if (Action.executingScopes[this.scope]) {
+    public invoke(payload: TPayload, scope: string = null): void {
+        const activeScope = scope ? scope : this.scope;
```
karanbirsingh, line 16: This can use https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

src/electron/ipc/ipc-renderer-shim.ts
karanbirsingh: (chatted offline) seems like options are:
- move version into InsightsConfiguration or a store since it's a constant value
- make this method async and update all its usages

It'd be nice to explore more but doesn't need to be part of this PR given we already were sync here

src/tests/unit/tests/electron/ipc/ipc-renderer-shim.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    });
+
+    afterEach(() => {
+        ipcRendererMock.verifyAll();
+    });
+
+    it('do nothing at construction', () => {});
+
+    describe('tests after initialization', () => {
+        const initialzationMessages = [
```
karanbirsingh, line 35: small typo, initializationMessages

src/common/flux/action.ts
DaveTryon: Done

src/tests/unit/tests/electron/ipc/ipc-renderer-shim.test.ts
DaveTryon: Fixed

src/electron/ipc/ipc-channel-names.ts
```diff
@@ -3,3 +3,13 @@
 
 export const IPC_MAIN_WINDOW_INITIALIZED_CHANNEL_NAME = '60dc5587-7f98-4216-92d5-7026875b2ee4';
 export const IPC_MESSAGE_CHANNEL_NAME = '9a436d48-61e9-4c34-849d-3c15ef06f2c9';
+export const IPC_APP_VERSION_CHANNEL_NAME = '48987ad6-02d1-4297-8ab8-57e3a94c83f2';
+export const IPC_BROWSERWINDOW_MAXIMIZE_CHANNEL_NAME = '4edeb851-a52d-4653-acf9-38da8f0048bc';
```
devaradhanm, line 7: not related to this PR. but i am curious why we decided to use guids instead of user friendly ids (like maximize-window etc)

src/electron/main/main-window-renderer-message-handlers.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export class MainWindowRendererMessageHandlers {
+    public constructor(
+        private readonly app: App,
+        private readonly mainWindow: BrowserWindow,
+        private readonly ipcMain: IpcMain,
+    ) {}
+
+    public startListening(): void {
+        this.ipcMain.on(IPC_APP_VERSION_CHANNEL_NAME, this.onGetVersionMessage);
```
devaradhanm, line 24: How about something like this:
```
type EventListener = { eventName: string, callback: Function};
private ipcMainCallbacks: EventListener[]= [];
private mainWindowCallbacks : EventListener[]= [];

public startListening(): void {
   ipcMainCallbacks = [
       { eventName: IPC_APP_VERSION_CHANNEL_NAME, callback: this.onGetVersionMessage},
       { eventName: IPC_BROWSERWINDOW_MAXIMIZE_CHANNEL_NAME, callback: this.onMaximizeMessage},
       // ...
       ]; 
   mainWindowCallbacks = [
      {eventName: 'maximize', callback: this.onMaximizeEvent}
      //...
      ];
   ipcMainCallbacks .forEach(listener => {
     this.ipcMain.on(listener.eventName, listener.callback);
   });
   mainWindowCallbacks .forEach(listener => {
     this.mainWindow.on(listener.eventName, listener.callback);
   });
}
public stopListening(): void {
  ipcMainCallbacks .forEach(listener => {
     this.ipcMain.removeListener(listener.eventName, listener.callback);
  });
  mainWindowCallbacks .forEach(listener => {
     this.mainWindow.removeListener(listener.eventName, listener.callback);
  });
}
}

```

src/electron/ipc/ipc-renderer-shim.ts
```diff
@@ -0,0 +1,74 @@
...
+    IPC_MAIN_WINDOW_INITIALIZED_CHANNEL_NAME,
+} from 'electron/ipc/ipc-channel-names';
+
+export class IpcRendererShim {
+    public constructor(private readonly ipcRenderer: IpcRenderer) {}
+
+    private readonly invokeScope: string = 'IpcRendererShim';
+
+    public initialize(): void {
+        this.ipcRenderer.on(IPC_BROWSERWINDOW_MAXIMIZE_CHANNEL_NAME, this.onMaximize);
```
devaradhanm, line 24: We have two types of events. 
1.) Event 1 - renderer process sends request to main process to maximize.
2.) Event 2 - Main process invokes window.maximize. After window is maximized, main process gets notified & it calls window.webContents.send to notify renderer process that maximize event has occured.

Currently, in this PR both Event 1 & Event 2, shares the same name. Event though they are used for different puposes. (1st event to submit maximize request & second event to notify maximize event has occured.) So, may be better to have different names to avoid confusion.

src/electron/main/main-window-renderer-message-handlers.ts
DaveTryon: Done

src/electron/ipc/ipc-renderer-shim.ts
DaveTryon: Done. Channels originating from the renderer start IPC_FROMRENDERER. Channels originating from the browserWindow start IPC_FROMBROWSERWINDOW. Also updated the private methods to be more specific in their names

src/electron/ipc/ipc-channel-names.ts
```diff
@@ -1,5 +1,22 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export const IPC_MAIN_WINDOW_INITIALIZED_CHANNEL_NAME = '60dc5587-7f98-4216-92d5-7026875b2ee4';
 export const IPC_MESSAGE_CHANNEL_NAME = '9a436d48-61e9-4c34-849d-3c15ef06f2c9';
```
devaradhanm, line 4: P2: might be better to group the event names

something like:
export var ipcEvents = {
   RENDERER_GET_APP_VERSION_CHANNEL_NAME : '',
   ...
}

devaradhanm: never mind. i see all the event names here are for ipc from renderer

DaveTryon: I'll reserve this for a separate PR

Issue #2147, opened 2020-02-12
Electron app main window always opens full-screen


LiLoDavis:
## Describe the bug
The primary window always opens full screen, and the window's **Restore down** button always shrinks it to a small, fixed size. As a consequence, getting the app window to the preferred size is a two-step process.  

**To Reproduce**
1. Launch the app.
2. Start testing.
3. Use the **Restore down** button.

## Expected behavior
The primary window should open to its most recent size and position.

## Context (please complete the following information)
Accessibility Insights for XXXXXXX - Canary
2020.211.7
2/11/2020

## Are you willing to submit a PR?
No

## Did you search for similar existing issues?
Yes

ferBonnin:
Per conversation with Peter, let's take a look at this

lisli1:
Started working on this here: https://github.com/microsoft/accessibility-insights-web/pull/2584

Need to update to use IpcRendererShim and also save the most recent position of the window, as mentioned in Dave's comment on the above PR.

DaveTryon:
@LiLoDavis, This fix is now live in the Canary build. Please regress at your convenience. Thanks!

LiLoDavis:
Looks great!

PR #2584, opened 2020-04-30
fix: electron main window always maximizing

lisli1:
#### Description of changes

The expected behavior from #2147 is to maximize the main window the first time. On subsequent runs of the app, the main window should open at its most recent size (i.e. if the window is last resized to 600 x 400, it should open at 600 x 400).

To address the issue, this PR does the following:
- Removes the `windowFrameActionCreator.maximize()` call in `DeviceConnectFooter`, which was causing the main window to always open maximized
- Adds `lastWindowSize` into `UserConfigurationStoreData` to store the last/most recent window size
- In `WindowStateActionCreator`, `setRoute` sets the size of the main window based on the `lastWindowSize` from `UserConfigurationStore` or maximizes the main window if `lastWindowSize` is null.
- Gets and saves the size of the window in `WindowFrameListener` on 'resize', but not when the `routeId` is deviceConnectView. Checking `routeId` is necessary because disconnecting a device, rescanning, and reconnecting a device will reopen the deviceConnectView at its fixed size, which is not the `lastWindowSize` that we want to save.

Note: Updated some tests and snapshots but still need tests for:
- UserConfigurationActionCreator
- UserConfigurationStore
- UserConfigMessageCreator
- UserConfigurationController

#### Pull request checklist

- [x] Ran `yarn fastpass`


DaveTryon:
As we discussed offline, this will be impacted by changes that went into #2581. I'd suggest listening for the `resize` and `moved` events in MainWindowRendererMessageHandlers, and pass the post-event bounding rectangle as a parameter, when then gets saved to the store by a method in WindowStateActionCreator. That preserves the existing relationships with the stores and ought to work on Windows, Mac, and Linux

PR #2596, opened 2020-04-30
refactor: explicitly download electron mirror

karanbirsingh:
#### Description of changes
This PR simplifies how our build pipeline handles the mirrored electron dependency. The contents of node_modules/electron/dist and node_modules/electron-chromedriver/bin are replaced.

Before this PR:
- mirrored electron is downloaded once in windows e2e tests and again in the electron-builder step
- on mac/linux, e2e tests run against the original version of electron rather than the mirrored electron we're bundling with
- e2e tests are aware of node_modules electron vs mirrored electron

After this PR:
- mirrored electron is downloaded only once, and electron-builder is given a path to the folder
- on all platforms, e2e tests run against the mirrored electron we're bundling with
- e2e tests can always use the imported electron path because electron (original or mirrored) is only ever in node_modules
- we pull in mirrored chromedriver as well

Once this PR is merged, I will fix up this [codecs e2e test](https://github.com/karanbirsingh/accessibility-insights-web/tree/codecs) and put it up for review. That test is unblocked now since we download the mirrored framework separately from creating the installer. Earlier these steps were atomic so we could not hook in anywhere.

Other notes:
- I'm unable to remove the original 'yarn install' because electron's & electron-chromedriver's install scripts handle the mirror environment variables differently. This has been fixed in master (electron-chromedriver now uses the same mechanism as electron to install), but we can't consume it yet. Once we can, then the explicit download script can go away and yarn install w/ appropriate env vars should be sufficient.
- I switched to extract-zip instead of unzipper because on MacOS the entire file contents weren't being extracted (similar to https://github.com/ZJONSSON/node-unzipper/issues/165). extract-zip is used by electron's install script
- sample installer [here](https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_build/results?buildId=10052&view=artifacts&type=publishedArtifacts)

#### Pull request checklist

- [ ] Ran `yarn fastpass`


pipeline/scripts/download-electron-mirror.js
```diff
@@ -1,26 +1,51 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 const pkg = require('../../package.json');
-const download = require('electron-download');
```
dbjorge, line 4: Can we drop the electron-download dev dependency?

karanbirsingh: Yes, I think so, good catch. electron-chromedriver uses it but we should be able to remove it from our explicit devDependency list. Will try

pipeline/scripts/download-electron-mirror.js
```diff
@@ -1,26 +1,51 @@
...
-        } else {
-            console.log('zipPath= ', zipPath);
-            fs.createReadStream(zipPath).pipe(unzipper.Extract({ path: 'drop/electron-local' }));
-        }
-    },
-);
+/*
+This script replaces existing electron & chromedriver modules
+with mirror dependencies specified by ELECTRON_MIRROR_VAR
+and ELECTRON_CUSTOM_DIR_VAR (see @electron/get).
```
dbjorge, line 12: It might be nice to also note why we do this (something like "We use this to avoid bundling non-freely-redistributable media codecs in our release builds. The version of Electron published to npm includes these as part of Chromium; our release builds use a Microsoft-maintained build of Electron that removes those codecs.")

karanbirsingh: Thanks for the suggested wording, have added it

package.json
```diff
@@ -57,6 +57,8 @@
         "with:mock-service-for-android": "npm-run-all --race --parallel serve:mock-service-for-android"
     },
     "devDependencies": {
+        "7zip-bin": "^5.0.3",
```
karanbirsingh, line 60: yarn add/remove moved 7zip-bin to the top. This seems right based on sorting

PR #2598, opened 2020-05-01
feat(electron upgrade): Revise getVersion

DaveTryon:
#### Description of changes

#2581 replaced `remote` usage with IPC calls, but it exposed a synchronous IPC call that we want to replace. Since the appVersion is a constant during the lifetime of the process, we're moving it into the app's config file that is accessible from all processes. This allows us to remove the blocking IPC call. The change has a few parts:
1. Add the `unifiedAppVersion` field to the config file. The default value is '0.0.0' unless it's overridden by the config file
2. Update `ElectronAppDataAdapter.getVersion`() to read the value from the config file (instead of calling the `ipcRendererShim`). Update its unit tests.
3. Update the config file generator to set the `unifiedAppVersion` field when building the config file. The value is the value set via the `unified-version` command line parameter. This parameter is specified only in the build pipeline, where it is set to $(Build.BuildNumber). Non-pipeline builds omit the `unified-version` parameter, so `getVersion()` simply returns the default value.
4. Remove all code associated with `ipcRendererShip.getAppVersion()`. This includes the channel name.
5. Defer calling `MainWindowRendererMessageHandlers.startListening()` until after the main window is fully initialized. We were unable to do this in #2581 due to the synchronous nature of the getVersion call.

Validation:
1. Ran `yarn build:unified:all`, then ensured that the `unifiedAppVersion` override was not added to any config files
2. Ran `yarn build:unified:all --unified-version=12.3.4`, then ensured that the `unifiedAppVersion` override was correctly added to all 4 unified config files (dev, canary, insider, production)
3. Added code (not included in the PR) to cause the renderer process to query the version. Verified that it returns '0.0.0' when `unifiedAppVersion` is not overridden in the config file, and that it returns the override value when `unifiedAppVersion` does exists in the config file.
4. Ran the app and ensured that all windows operations are as expected (and that no deadlock is introduced by deferring the call to `ipcRendererShip.startListening`

#### Pull request checklist

- [x] Ran `yarn fastpass`


Gruntfile.js
```diff
@@ -382,6 +386,13 @@ module.exports = function (grunt) {
             config.options.appInsightsInstrumentationKey = grunt.option(telemetryKeyIdentifier);
         }
 
+        // Add unifiedAppVersion value for electron-based products
+        if (config.options.productCategory === 'electron') {
```
karanbirsingh, line 390: flagging to say we might want to change the product category to 'unified' from 'electron' to be consistent (not part of this PR)

Gruntfile.js
```diff
@@ -444,7 +455,7 @@ module.exports = function (grunt) {
             `electron-builder.template.yaml`,
         );
 
-        const version = grunt.option('unified-version') || '0.0.0';
+        const version = getUnifiedVersion();
```
karanbirsingh, line 458: it feels odd to me that we check for existence above in line 392, but not here - I'd be fine with either as long as it's consistent

DaveTryon: You're correct. I had moved this into getUnifiedVersion, then pulled it out and forgot to restore this one.

PR #2637, opened 2020-05-07
fix: install rename in linux signing pipeline

pownkel:
#### Description of changes

Signing pipeline is currently failing because the rename command is not installed by default on the new ubuntu vm images

#### Pull request checklist

- [n/a] Ran `yarn fastpass`

### DIFF
--- a/pipeline/unified/channel/sign-release-package-linux.yaml
+++ b/pipeline/unified/channel/sign-release-package-linux.yaml
@@ -48,7 +48,7 @@ jobs:
                         }
                     ]
 
-          - script: 'rename "s/\.AppImage$/.sig/" $(System.DefaultWorkingDirectory)/detachedSignature/*.AppImage'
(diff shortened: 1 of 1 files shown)


pipeline/unified/channel/sign-release-package-linux.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
                         }
                     ]
 
-          - script: 'rename "s/\.AppImage$/.sig/" $(System.DefaultWorkingDirectory)/detachedSignature/*.AppImage'
+          - script: 'sudo apt-get install rename; rename "s/\.AppImage$/.sig/" $(System.DefaultWorkingDirectory)/detachedSignature/*.AppImage'
```
dbjorge, line 51: I think you can probably use bash to avoid installing a dependency for this with something like:

```yaml
- bash: 'for file in *.AppImage; do mv -- "${file}" "${file/%AppImage/sig}"; done'
```

If we do need to install a dependency, probably best to do it in a separate step to make it easier to separate apt-get failures vs rename failures in build failure logs/analytics

Issue #2521, opened 2020-04-20
Landmarks requirements confusing for target page with no (or multiple) main landmarks


dbjorge:
## Describe the bug

The "How to test" info for the Landmarks > Primary content requirement reads:

> The visual helper for this requirement highlights the page's main landmark.
>
> 1. In the target page, examine the main landmark to verify that it contains all of the page's primary content.
> 2. Record your results:
>   a. If you find any failures, select Fail, then add them as failure instances.
>   b. Otherwise, select Pass.

The Getting Started text for landmarks includes this paragraph:

> Landmarks are not required, but if you use them, you must use them correctly. Also, if you add Landmarks, you must have a main Landmark.

...however, it's not clear to me that this is up to date (I'd have expected it to fail an automated check for "contains landmarks but no main landmark", but it doesn't, maybe it's one of the best-practice landmark rules now?), and it's easy to miss this while going through the Primary content requirement, and the "How to test" text seems to be written assuming that there will always be a main landmark. I was assessing a page without landmarks today and was confused for a while about whether I was supposed to mark this "pass" or "fail".

Same feedback applies to the "No repeating content" requirement's text.

I think it would be nice to change the How to test text to something that makes it more clear whether the user should mark pass or fail for target pages that don't have exactly 1 main landmark. Leaving the exact text change for @ferBonnin  / @LiLoDavis to suggest.

If we want these to be marked pass, it would probably be ideal for the requirement to auto-pass if there is no main element, similar to how the Landmark roles requirement auto-passes if there are no landmarks found.

While we're here it would also be nice to fix the inconsistent use of `<Markup>` tags with the term "main" in these requirements.

**To Reproduce**

1. Create a local test html file with the following content (can't use codepen because its visual wrappers have landmarks in them):
  ```html
    <html lang="en">
        <head>
            <title>Test page</title>
        </head>
        <body>
            <h1>Content</h1>
            <p>Test content.</p>
        </body>
    </html>
  ```
2. Start an AIWeb assessment against this page (will need to enable local file usage if not already enabled)
3. Observe that automated checks all pass
4. Go to "Landmark" test
5. Observe confusing text/behavior of "Primary content" and "No repeating content" requirements
6. (note: the example html above has no landmarks at all, but the same repro steps apply to a page with some non-main landmarks but no main landmarks, or to a page with multiple main landmarks)

## Expected behavior

It should be clear to a user without any background in landmarks usage what they need to do to assess a page with:
* No landmarks
[...]

PR #2644, opened 2020-05-08
fix: auto-pass main landmark requirements for no-landmark pages

dbjorge:
#### Description of changes

#2521 tracks an issue where it wasn't clear what the expectations on a user were for testing the Landmarks "primary content" and "no repeating content" requirements on pages with no landmarks.

The proposed fix for that was a text update + making the requirements pass automatically on pages that have no landmarks.

This fix involves a significant bit of new behavior. We wanted to *keep* the 2 requirements in question as manual requirements, not assisted requirements, since it's possible for a page to have no main landmarks and for a user to need to manually enter a failure not associated with any instance we might show in an instance list. However, this is the first time we're assigning auto-pass behavior to a manual test; in all other cases where we support auto-pass, it works by looking for an assisted requirement with no matching instances.

This new behavior is especially challenging because the set of instances required for auto-passing ("are there **any** landmarks on the page?") is different from the set of instances required for visualization ("what are all the **main** landmarks on the page?").

To support this, I've added 2 new features to `AssessmentStore`'s `Requirement` processing:
* A new `isVisualizationSupportedForResult` property which `Requirement`s can use to indicate that certain results returned by their analyzers should not support being visualized (by default, all results support visualization)
* A new `getInitialManualTestStatus` property which manual `Requirement`s can use to infer an initial pass/fail state after scanning for results (by default, a manual requirement is put in the UNKNOWN state regardless of analysis results)

These are used by the updated requirements by updating their analyzers to use the `unique-landmark` rule to look for *all* landmarks instead of the old behavior of looking only for main landmarks, use `isVisualizationSupportedForResult` to only visualize the main landmarks (like before), and use `getInitialManualTestStatus` to set a "PASS" status if the scan completes and there are no instances with the landmark role data from the `unique-landmark` rule.

New behavior:

**no landmarks: "Primary content" and "No repeating content" automatically pass**
screen recording of no-landmarks interactions

**only non-main landmarks: "Primary content" and "No repeating content" don't auto-pass and show no matching instances to visualize**
screen recording of banner-landmark-only interactions

**Mix of main and non-main landmarks: "Primary content" and "No repeating content" don't auto-pass and visualize only the main landmark**
screen recording of mixed-landmarks interactions

#### Pull request checklist

- [x] Addresses an existing issue: #2521
- [x] Ran `yarn fastpass`


dbjorge:
Per discussion from Karan and Fer: we think as a followup from this work, we'd want to look into similar autopass behavior for `Sequence > CSS positioning` and `Semantics > CSS content` when they have nothing available to visualize. Leaving them out for the sake of this PR since it's big enough already and already has 1 motivating example. @ferBonnin will double check whether there are any other requirements that might similarly benefit.

dbjorge:
I think this new behavior would benefit from an e2e test validating it; I have it mostly implemented already, but again, leaving it for a separate PR to try to keep the scope of this one more limited, it's already quite large

ferBonnin:
> @ferBonnin will double check whether there are any other requirements that might similarly benefit.

checked and right now, those two are the only two requirements that will benefit from this at the moment.

src/assessments/landmarks/test-steps/primary-content.tsx
```diff
@@ -10,19 +11,48 @@ import { AnalyzerConfigurationFactory } from '../../common/analyzer-configuratio
...
+                        The page has exactly one <Markup.CodeTerm>main</Markup.CodeTerm> landmark,
+                        and
+                    </li>
+                    <li>
+                        The <Markup.CodeTerm>main</Markup.CodeTerm> landmark contains all of the
+                        page's primary content.
+                    </li>
+                </ol>
+                <p>
+                    Exception: If a page has nested <Markup.CodeTerm>document</Markup.CodeTerm> or{' '}
```
karanbirsingh, line 50: I considered suggesting we share some of the markup between primary-content and no-repeating-content because a good amount of the text is identical, but decided against it - it doesn't buy us much

src/background/stores/assessment-store.ts
```diff
@@ -432,12 +434,38 @@ export class AssessmentStore extends BaseStoreImpl<AssessmentStoreData> {
...
 
+    private applyInitialManualTestStatus(
+        assessmentData: AssessmentData,
+        testStepName: string,
+        testType: VisualizationType,
+        getInitialManualTestStatus: (InstanceIdToInstanceDataMap) => ManualTestStatus,
+    ): void {
+        const originalStatus = assessmentData.manualTestStepResultMap[testStepName].status;
+        if (originalStatus !== ManualTestStatus.UNKNOWN) {
+            return; // Never override an explicitly set status
```
karanbirsingh, line 460: could this have any effect on assessments in progress? It seems not - if a status is set, we won't override it, and otherwise, it's a change we want to make (unknown -> pass)

src/scanner/custom-rules/landmark-rule.ts
```diff
@@ -1,15 +0,0 @@
...
-// Licensed under the MIT License.
-import { RuleConfiguration } from '../iruleresults';
-
-const landmarkCheckId: string = 'unique-landmark';
-
-export const landmarkConfiguration: RuleConfiguration = {
-    checks: [],
-    rule: {
-        id: 'main-landmark',
-        selector: '[role=main], main',
```
karanbirsingh, line 11: In the rule we have removed, we check for the HTML technique too instead of just role. It looks like `doesResultHaveMainRole` doesn't have that - are we still picking up on <main> elements?

dbjorge: Good question! Yes, we are (the gifs in the description are testing against `<main>` and `<banner>` elements rather than explicit role attributes) - the information in the `role` checkResult data is populated in the `unique-landmark.ts` custom rule implementation, which uses an explicit role element if present but falls back to using `axe.commons.aria.implicitRole(element)` when an explicit role attribute is not present.

dbjorge: Adding a comment to clarify

src/background/stores/assessment-store.ts
dbjorge: I think the scan complete callback that triggers this logic only happens after scans happen, so in general I'd expect it to only affect an assessment in progress if that assessment either hadn't already gotten to the requirements in question or if that assessment explicitly restarts Landmarks; I think in either case, it's desireable for the new logic to take effect

PR #2654, opened 2020-05-10
fix: 1.4.10 Reflow is not solely about text

patrickhlauke:
Normatively, 1.4.10 is scoped to all content, not just text. While the principal concern of the SC is indeed text, it applies in equal measure to other types of content/mixed content.
Admittedly, the normative wording is a bit wooly/omits certain extra scenarios (i.e. that the primary concern is that blocks of text are legible without tow-dimensional scrolling, rather than an outright
ban on two-dimensional scrolling ... see for instance https://github.com/w3c/wcag/issues/668), but one things it does not do is explicitly limit this to just the text portion of content.

> Success Criterion 1.4.10 Reflow (Level AA): *Content* can be presented without loss of information or functionality [...]
(emphasis on "Content", not "Text" or "Text content")

https://www.w3.org/WAI/WCAG21/Understanding/reflow.html

#### Pull request checklist

- [n/a] Ran `yarn fastpass`

### DIFF
--- a/src/assessments/text-legibility/test-steps/reflow.tsx
+++ b/src/assessments/text-legibility/test-steps/reflow.tsx
@@ -9,7 +9,7 @@ import * as content from 'content/test/text-legibility/reflow';
 import { ManualTestRecordYourResults } from '../../common/manual-test-record-your-results';
 
 const reflowDescription: JSX.Element = (
-    <span>Text content must be visible without having to scroll in two dimensions.</span>
+    <span>Content must be visible without having to scroll in two dimensions.</span>
 );
 
 const reflowHowToTest: JSX.Element = (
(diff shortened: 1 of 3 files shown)


dbjorge:
You can ignore the web-e2e-reliability failures in the linux jobs; they're an unrelated break introduced by a recent version update to the Docker base image the linux environment uses. I'll be fixing it shortly in a separate PR.

LiLoDavis:
> Normatively, 1.4.10 is scoped to all content, not just text. While the principal concern of the SC is indeed text, it applies in equal measure to other types of content/mixed content.
> Admittedly, the normative wording is a bit wooly/omits certain extra scenarios (i.e. that the primary concern is that blocks of text are legible without tow-dimensional scrolling, rather than an outright
> ban on two-dimensional scrolling ... see for instance [w3c/wcag#668](https://github.com/w3c/wcag/issues/668)), but one things it does not do is explicitly limit this to just the text portion of content.
> 
> > Success Criterion 1.4.10 Reflow (Level AA): _Content_ can be presented without loss of information or functionality [...]
> > (emphasis on "Content", not "Text" or "Text content")

I agree.

PR #2662, opened 2020-05-11
feat(nav-bar-content): create new links for reflow left nav

waabid:
#### Description of changes

This creates a new "duplicate" function on the left nav link builder to create the links for reflow (links now have tests, requirements, and getting started).

Ideally, I'd like to re-factor the left nav link builder but felt that the old link building didn't require it (i.e. fastpass) but this does. Additionally, I don't know how much I want to break up the responsibility of creating the link objects themselves: it's largely just composition and there are few branches of logic. I'd like to move out some of inner small components to separate modules though, as they allow for their own logic/have own responsibility.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/DetailsView/components/left-nav/__snapshots__/left-nav-link-builder.test.tsx.snap
+++ b/src/tests/unit/tests/DetailsView/components/left-nav/__snapshots__/left-nav-link-builder.test.tsx.snap
@@ -86,6 +86,392 @@ exports[`LeftNavBuilder buildAssessmentTestLinks should build links for assessme
 />
 `;
 
+exports[`LeftNavBuilder buildReflowAssessmentTestLinks should build links for assessments: getting started nav link render 1`] = `<GettingStartedNavLink />`;
+
+exports[`LeftNavBuilder buildReflowAssessmentTestLinks should build links for assessments: getting started nav link render 2`] = `<GettingStartedNavLink />`;
+
+exports[`LeftNavBuilder buildReflowAssessmentTestLinks should build links for assessments: requirement nav link render 1`] = `
+<TestViewLeftNavLink
+  link={
+    Object {
+      "displayedIndex": "-1.1",
+      "forceAnchor": true,
+      "iconProps": Object {
+        "className": "hidden",
+      },
+      "index": 1,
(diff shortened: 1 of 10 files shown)


src/tests/unit/tests/DetailsView/components/left-nav/__snapshots__/left-nav-link-builder.test.tsx.snap
```diff
@@ -86,6 +86,392 @@ exports[`LeftNavBuilder buildAssessmentTestLinks should build links for assessme
 />
 `;
 
+exports[`LeftNavBuilder buildReflowAssessmentTestLinks should build links for assessments: getting started nav link render 1`] = `<GettingStartedNavLink />`;
+
+exports[`LeftNavBuilder buildReflowAssessmentTestLinks should build links for assessments: getting started nav link render 2`] = `<GettingStartedNavLink />`;
+
+exports[`LeftNavBuilder buildReflowAssessmentTestLinks should build links for assessments: requirement nav link render 1`] = `
+<TestViewLeftNavLink
```
waabid, line 94: These massive snapshots are produced because we do `getElement` which renders/surfaces the passed properties as well. Since we validate the link objects ourselves, I almost prefer `debug` in this situation to eliminate additional coupling.

Would appreciate thoughts here.

JGibson2019: I personally prefer `debug()` for this situation

pownkel: I agree, debug() makes sense here since we validate the link objects anyway

waabid: Well. I looked into using debug but unfortunately, it always seems to render the actual component (i.e. instead of <TestViewLeftNavLink> you see the actual html for that component).

I think this might be a part of the re-factoring work in all honesty: extract out the render functions to be separate components/modules that are not a part of this builder.

PR #2665, opened 2020-05-12
refactor: rename isEnabled

pownkel:
#### Description of changes

The isEnabled property on Assessment view is a little ambiguous as to what it refers to. This PR will rename it to selectedRequirementIsEnabled so it's somewhat clearer.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/reflow-assessment-view.tsx
+++ b/src/DetailsView/components/reflow-assessment-view.tsx
@@ -5,6 +5,7 @@ import * as React from 'react';
 import {
     AssessmentViewUpdateHandler,
     AssessmentViewUpdateHandlerDeps,
+    AssessmentViewUpdateHandlerProps,
 } from 'DetailsView/components/assessment-view-update-handler';
 import { AssessmentTestResult } from '../../common/assessment/assessment-test-result';
 import { Tab } from '../../common/itab';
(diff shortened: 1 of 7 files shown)


src/DetailsView/components/reflow-assessment-view.tsx
```diff
@@ -28,7 +28,7 @@ export interface ReflowAssessmentViewProps {
     currentTarget: Tab;
     prevTarget: PersistedTabInfo;
     assessmentTestResult: AssessmentTestResult;
-    isEnabled: boolean;
+    selectedRequirementIsEnabled: boolean;
```
JGibson2019, line 31: Out of curiosity, why was `isEnabled`/`selectedRequirementIsEnabled` added to this component if it's value is not being used by the component at all? This should probably be removed and added when the view actually uses the value.

pownkel: it's used by the updateHandler, and will most likely be needed by RequirementView when we add that here. I'll take it out for now though (or make it clearer that it goes to updateHandler).

JGibson2019: I got that it's used by the updateHandler! If it's not used in the component currently though, it might make more sense to remove it from the props now and add it back once the ReflowAssessmentView is using it.

PR #2670, opened 2020-05-12
feat(android-bug-filing): Add class name as identifier

AhmedAbdoOrtiga:
#### Description of changes
- For unified rule result bug filing, we don't have css selectors like web.
- We wanted to use the class name as identifiers.
- Class name will be part of the bug filing info.

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/electron/platform/android/scan-results-to-unified-results.ts
```diff
@@ -90,7 +90,10 @@ function createUnifiedResult(
         ruleId: ruleInformation.ruleId,
         status: getStatus(ruleResult.status),
         descriptors: getDescriptors(viewElementLookup[ruleResult.axeViewId]),
-        identifiers: null,
+        identifiers: {
+            identifier: viewElementLookup[ruleResult.axeViewId]?.className,
+            conciseName: viewElementLookup[ruleResult.axeViewId]?.className,
```
jalkire, line 95: Are we ok with both of these as duplicate values or would it be better to set one blank/null?

AhmedAbdoOrtiga: The reason I'm using both is later in bug filing we expect both, in unified-result-to-issue-fling-data.ts line 29, 30, otherwise it will create null error

PR #2686, opened 2020-05-14
feat(mobile-rules-guidance): update common components to use LinkComponent

haonliu:
#### Description of changes

update components to use LinkComponent so that in electron links will open in a new browser window.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/components/guidance-links.tsx
```diff
@@ -1,14 +1,18 @@
...
 import * as React from 'react';
 import { HyperlinkDefinition } from 'views/content/content-page';
+
 import { NamedFC } from '../react/named-fc';
-import { NewTabLink } from './new-tab-link';
 
 export interface GuidanceLinksProps {
     links: HyperlinkDefinition[];
     classNameForDiv?: string;
+    LinkComponent?: LinkComponentType;
```
dbjorge, line 15: I think making this optional makes it easier to write the same class of bug (forgetting to use the electron-specific link in the electron app) easier to do in the future. I think it looks reasonably contained to make it non-optional and pipe it in everywhere; it looks like there are only 3 other places where `<GuidanceLinks>` is being used, one of which already has a `LinkComponent` dep in place and the other 2 of which already have `deps` chains in place.

PR #2687, opened 2020-05-15
feat(manual-auto-pass): add auto-pass for CSS content and CSS positioning

dbjorge:
#### Description of changes

Per discussion in #2644, the new manual auto-pass behavior we added for some of the Landmarks requirements would also be nice to have for the Semantics > CSS content and Sequence > CSS positioning requirements when their visualization analyzers find no relevant elements.

This PR generalizes the "auto-pass if no landmarks" behavior to a "auto-pass if no results relevant to the requirement" behavior and adds it in to the 2 CSS requirements. To do this, it updates the assessment store to only pass instance data with results relevant to the current requirement when it calls a requirement's getInitialManualTestStatus function.

I considered an alternative approach of leaving assessment store as-is and creating a factory function for each requirement to specify its own `autoPassIfNoResultsFor(key)` rather than a single `autoPassIfNoResults`; I liked that because it avoided adding more complexity to `assessment-store`, which is already a very complex class, but I felt that it was better for data encapsulation if requirements were systematically prevented from seeing/depending on other requirements' data, and I thought the encapsulation was more valuable.

gif of 2 CSS requirements auto-passing on a page with no matching instances

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/unit/tests/assessments/auto-pass-if-no-landmarks.test.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
karanbirsingh, line 1: this file can be renamed (Still references landmarks)

PR #2689, opened 2020-05-15
feat(android-bug-filing): Use tool data for issue filing

AhmedAbdoOrtiga:
#### Description of changes

In this [PR](https://github.com/microsoft/accessibility-insights-web/pull/2462), I wanted to replace environment info provider with tool data since it's duplicate information, and also I wanted to get things ready for android issue filing, but I received feedback about the number of files is big so I had to close the PR and haven't gotten slack time since to revisit it.

In the android issue filing now, we faced the same issue again, we need to use the same data structure in web and unified, toolData is a good option for that, so I tried to break down the previous PR into two pieces:
1- This PR which is using toolData for issue filing instead of envirmentInfoProvider
2- Later PR which will use toolData for React components like reportFooter [this could be done out of this feature scope]

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/details-view-initializer.ts
```diff
@@ -292,6 +294,14 @@ if (isNaN(tabId) === false) {
                 AxeInfo.Default.version,
             );
 
+            const toolData = createToolData(
```
AhmedAbdoOrtiga, line 297: The reason I used another variable here for toolData instead of using environmentInfoProvider.getToolData() is to make it easier to remove the environmentInfoProvider as mentioned in the PR description.

Note: you might find similar issue in other files in this PR

PR #2717, opened 2020-05-19
fix: updated Tab Stops toggle aria-label to match the visually displayed label

JGibson2019:
Added logic to visualization-toggle to set the ariaLabel to have the same value as the label displayed on the visualization toggle. Previously, we always set the aria-label value to the visualization name; having parity between the visually rendered label and ariaLabel makes the most sense. If no label is present in the toggle props, the aria-label value will default to the visualization name.

Tested the changes using the repro steps in the Trusted Tester bug; used NVDA and Narrator in Edge Insider

#### Description of changes
Before changes in this PR:
- displayed label: "Show tab stops"
- aria-label: "Tab stops" (since that's the visualization name)

After this PR:
- displayed label: "Show tab stops"
- aria-label: "Show tab stops" 

#### Pull request checklist

- [X] Addresses an existing issue: Bug 1720359: A11y_Accessibility Insights for Web_Tab stops_Label in name: Narrator/NVDA is reading Label "Show tab stops" as "tab stops"
- [X] Ran `yarn fastpass`

### DIFF
--- a/src/common/components/visualization-toggle.tsx
+++ b/src/common/components/visualization-toggle.tsx
@@ -26,7 +26,7 @@ export class VisualizationToggle extends React.Component<VisualizationToggleProp
             className: this.props.className,
             onText: 'On',
             offText: 'Off',
-            ariaLabel: this.props.visualizationName,
+            ariaLabel: this.props.label ? this.props.label : this.props.visualizationName,
             componentRef: this.props.componentRef,
             onBlur: this.props.onBlur,
             onFocus: this.props.onFocus,
--- a/src/common/components/visualization-toggle.tsx
+++ b/src/common/components/visualization-toggle.tsx
@@ -26,7 +26,7 @@ export class VisualizationToggle extends React.Component<VisualizationToggleProp
             className: this.props.className,
             onText: 'On',
             offText: 'Off',
-            ariaLabel: this.props.label ? this.props.label : this.props.visualizationName,
+            ariaLabel: this.props.label ?? this.props.visualizationName,
             componentRef: this.props.componentRef,
             onBlur: this.props.onBlur,
             onFocus: this.props.onFocus,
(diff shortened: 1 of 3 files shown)


src/common/components/visualization-toggle.tsx
```diff
@@ -26,7 +26,7 @@ export class VisualizationToggle extends React.Component<VisualizationToggleProp
             className: this.props.className,
             onText: 'On',
             offText: 'Off',
-            ariaLabel: this.props.visualizationName,
+            ariaLabel: this.props.label ? this.props.label : this.props.visualizationName,
```
dbjorge, line 29: You can simplify this a little bit with a [null coalescing operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing):

```suggestion
            ariaLabel: this.props.label ?? this.props.visualizationName,
```

src/tests/unit/tests/common/components/visualization-toggle.test.tsx
```diff
@@ -74,7 +74,7 @@ describe('VisualizationToggleTest', () => {
             className: props.className,
             onText: 'On',
             offText: 'Off',
-            ariaLabel: props.visualizationName,
+            ariaLabel: props.label ? props.label : props.visualizationName,
```
dbjorge, line 77: It's usually best to avoid repeating conditional logic between a bit of code and its unit tests; it makes it easy to accidentally be "testing" behavior that isn't necessarily right despite matching the test.

This looks like it's probably an older test case from before we used snapshot testing; I'd probably lean towards removing this helper method entirely, updating the tests to use snapshots instead for pinning the rendered object under test, and make sure there are test cases that cover the different combinations of "label specified vs not" and "visualizationName specified vs not"

src/common/components/visualization-toggle.tsx
JGibson2019: Made the change in the most recent commit

src/tests/unit/tests/common/components/visualization-toggle.test.tsx
JGibson2019: Updated this in my last commit; lmk what you think of the new changes!

Issue #2545, opened 2020-04-23
AI Android: High Contrast hover state color not changing for certain components


JGibson2019:
## Describe the bug
For certain buttons, such as "Start Over" the color of the component is not changing to blue when the user hovers over. 

**To Reproduce**

1. Go to the Insider build of Accessibility Insights for Android desktop application
2. Connect to a port (either through an emulator or through using mock of the Android service)
3. Hover over the "Start over" icon at the top of the automated checks view

## Expected behavior

You expect for the "Start over" button to turn blue to adhere to UX standards for High Contrast hover behavior

Example: "Export result" appears blue in hover and white in the default state
![image](https://user-images.githubusercontent.com/19158512/80157312-bc4eb300-857a-11ea-96a3-d65f941567e6.png)

![image](https://user-images.githubusercontent.com/19158512/80157295-b22cb480-857a-11ea-849c-b444126d6d33.png)

## Screenshots

"Start over" appears white in all states
![image](https://user-images.githubusercontent.com/19158512/80157269-a4772f00-857a-11ea-9d65-00ebbaa9efec.png)

## Context (please complete the following information)

- OS Name & Version: 
- AI-Web Version & Environment:  Accessibility Insights for Android 2020.422.21
- Browser Version: 
- Target Page:

ferBonnin:
Per conversation with Peter, let's fix this.

PR #2735, opened 2020-05-20
fix: make electron command bar button stylings match web

pownkel:
#### Description of changes

Currently, some of the button appearances in the command bar of the electron app are inconsistent:
cmd bar old
The icon for start over is black like the text, but the icon for export report is blue. In web, all button text is black and icons are blue in the default state. Per a conversation with Fer, these buttons should match the web stylings.

The different stylings also causes some issues in high contrast mode, as described in #2545. In addition to the text color not changing when hovering, buttons also appear whited out in the active state:
cmd bar high contrast old active

To fix this, this pr will remove our custom styles from the command bar buttons and use the default office fabric styles like we currently do in web. As a result, all button icons will be blue:
cmd bar
Hover and active states will appear the same (below shows the command bar with the start over button hovered):
cmd bar hover
High contrast mode will appear the same in default state:
cmd bar high contrast
Button text will correctly change color when hovered:
cmd bar high contrast hover
And the active state is no longer whited out:
cmd bar high contrast active
These changes are all consistent with what we currently have in the web extension.

#### Pull request checklist

- [x] Addresses an existing issue: #2545
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/views/automated-checks/components/command-bar.scss
+++ b/src/electron/views/automated-checks/components/command-bar.scss
@@ -12,28 +12,6 @@
     }
 }
 
-// specificity required to override fabric style
-button.menu-item-button {
-    color: $primary-text;
-
-    .button-icon {
-        color: $primary-text;
-        line-height: 16px;
-    }
-
-    &:hover {
-        @include buttonColors($neutral-alpha-4);
-    }
-
-    &:active {
-        @include buttonColors($neutral-alpha-8);
-    }
-
-    &:focus::after {
-        outline: $secondary-text solid 1px !important;
-    }
-}
-
 .command-bar {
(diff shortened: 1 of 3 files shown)


jalkire:
>And the active state is no longer whited out:

In the last screenshot, the circle arrow icon on the left is pretty hard to see. Is that as designed? It's definitely better than the before photo either way, just wanted to double check!

pownkel:
> In the last screenshot, the circle arrow icon on the left is pretty hard to see. Is that as designed? It's definitely better than the before photo either way, just wanted to double check!

Discussed this with Fer, it seems the contrast here is an accessibility bug currently existing in web as well. I'll file this as a separate bug.

Issue #1875, opened 2019-12-17
Exported report does not show icons when opened in IE


JGibson2019:
## Describe the bug
The report does not show icons when opened in Internet Explorer

Here is a picture of how the report looks in IE
![image](https://user-images.githubusercontent.com/19158512/70955318-41521a80-2025-11ea-83a1-537af18438a8.png)

**To Reproduce**

1. Go to 'www.google.com'
2. Click on the Accessibility Insights extension
3. Select 'FastPass'
4. In FastPass, Automated Checks, select 'Export Result'
5. Save the exported result
6. Open the exported file with IE
7. See error

## Expected behavior

I expect the icons to show in the report, like they do in Chrome (pictured below)
![image](https://user-images.githubusercontent.com/19158512/70955343-5a5acb80-2025-11ea-929c-8200cc45f2b1.png)

## Context (please complete the following information)

- OS Name & Version: Windows 10
- AI-Web Version & Environment:  Canary 2019.12.16.1956
- Browser Version:  Internet Explorer
- Target Page:  www.google.com

ferBonnin:
Per conversation with Peter, lets time box the investigation for this and find out if its simply that what we are using is not supported by IE or if there is an easy tweak we can do to support it. In documentation for Web, we do mention that we work for Chrome and Edge insider only.

dbjorge:
#2737 fixes the most glaring usability issues. Most other issues stem from our use of CSS variables (which IE doesn't support) for colors. Investigating a possible update to improve how we manage CSS colors such that we might be able to support this in the future (in a separate PR from #2737).

ferBonnin:
@peterdur per dbjorge comment above looks like we have fixed the bigger usability issues and fixing the CSS colors will require more investigation since IE doesn't support it at the moment. are we ok with closing this issue with a partial fix?

peterdur:
@ferBonnin here's what the report currently looks like in IE:
![image](https://user-images.githubusercontent.com/1752950/89329838-3bf12100-d644-11ea-8be3-8218fe6af688.png)
There's some regression in that most of the icons are no longer visible; since the earlier screenshot we moved to using CSS variables more extensively around icons for high contrast support purposes - see #3064 for more details (and really we probably should be using CSS variables even more, which would make the colored circles and icons disappear entirely). 
Even without the icons, though, the report is navigable and all the instance information can be accessed. I think this meets an appropriate bar for Internet Explorer support. I'm OK with closing this issue as-is.

PR #2737, opened 2020-05-21
fix: improve compatibility of fastpass exported reports with IE

dbjorge:
#### Description of changes

This is a partial fix for #1875. It fixes two issues:

* Our logo had an explicit `height` CSS property but only an implied `width` property, which IE misinterprets as "100%" instead of "scaled down by the same factor as the height". This caused the logo to take up the entire screen width and push the "Accessibility Insights for Web" text completely off-screen. We fix this by providing a `width` as well as a `height` (the logo image is a square, so we just use the same dimension as the height).
* The scripts we injected failed to run in IE because they were defined in terms of arrow-functions, which IE does not support. Updating the source of the injected functions to use `function() {}` style instead of `() => {}` style fixes the issue and makes expand/collapse behavior and our target page confirmation dialog work in IE.

This is not a complete fix for IE compat; there are still lots of issues with colors (due to IE not supporting CSS variables) and some other more minor formatting fixes. But it at least gets it to a point where the report is basically usable in IE.

Before (broken header bar, can't expand rules, no dialog on target page link):
image

After (more normal header bar, can expand/collapse rules, confirmation dialog for target page link):
image

#### Pull request checklist

- [x] Addresses an existing issue: #1875 (partially)
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/report-sections/collapsible-script-provider.tsx
+++ b/src/reports/components/report-sections/collapsible-script-provider.tsx
@@ -1,16 +1,21 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-// the following comment is to exclude the function from code coverage and prevent code cov from
-// injecting functions that affect snapshot tests
+// Note: the source of this function's body is stringified and injected into the report.
+//
+// The use of function() {} syntax over arrow functions is important for IE compat (see #1875).
+//
+// The "istanbul ignore next" excludes the function from code coverage to prevent code cov from
+// injecting functions that interfere with eval in the unit tests.
+//
 /* istanbul ignore next */
-export const addEventListenerForCollapsibleSection = (doc: Document) => {
+export const addEventListenerForCollapsibleSection = function (doc: Document): void {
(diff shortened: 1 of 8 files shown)


src/reports/components/report-sections/collapsible-script-provider.tsx
```diff
@@ -1,16 +1,21 @@
...
 /* istanbul ignore next */
-export const addEventListenerForCollapsibleSection = (doc: Document) => {
+export const addEventListenerForCollapsibleSection = function (doc: Document): void {
     const collapsibles = doc.getElementsByClassName('collapsible-container');
 
     for (let index = 0; index < collapsibles.length; index++) {
         const container = collapsibles.item(index);
         const button = container.querySelector('.collapsible-control');
-        button.addEventListener('click', () => {
+        button.addEventListener('click', function (): void {
```
jalkire, line 18: Is there any way we could test for/enforce using function() syntax in this file?

dbjorge: Great idea, added!

PR #2772, opened 2020-05-29
fix: use throttle instead of timeout for scroll/resize re-drawing

waabid:
#### Description of changes

A couple changes I looked into here:
1. Re-drawing more frequently on scroll/re-size (60fps).
2. Using throttle instead of manually creating/managing timeouts. Throttle effectively can handle that for us making this simpler and ensuring that we're not calling this method too frequently.

Notes: I'm very interested to see how this performs on other people's machines and hence added the second review required label. I think we can safely make the 2nd change but I'm open to increasing the throttle time for a less perf heavy task. In my experience, resizing manually did yield some noticeable lag however scrolling was indistinguishable. I also noticed some lag in resizing regardless of the visualization or not so ... 🤷 

Here's the difference:

Before:
bingheadingtimeoutredraw

After:
bingheadingthrottleredraw

#### Pull request checklist

- [.] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [.] (UI changes only) Verified usability with NVDA/JAWS

### DIFF
--- a/src/tests/unit/tests/injected/visualization/drawer.test.ts
+++ b/src/tests/unit/tests/injected/visualization/drawer.test.ts
@@ -13,7 +13,6 @@ import { DrawerUtils } from '../../../../../injected/visualization/drawer-utils'
 import { DrawerConfiguration, Formatter } from '../../../../../injected/visualization/formatter';
 import { HighlightBoxDrawer } from '../../../../../injected/visualization/highlight-box-drawer';
 import { NonTextComponentFormatter } from '../../../../../injected/visualization/non-text-component-formatter';
-import { itIsFunction } from '../../../common/it-is-function';
 import { TestDocumentCreator } from '../../../common/test-document-creator';
 
 describe('Drawer', () => {
@@ -37,6 +36,8 @@ describe('Drawer', () => {
 
         shadowUtilsMock = Mock.ofType(ShadowUtils);
         shadowUtilsMock.setup(x => x.getShadowContainer()).returns(() => shadowContainer);
+
+        jest.useFakeTimers();
     });
 
     test('eraseLayout called when initialize', () => {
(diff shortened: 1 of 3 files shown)


src/tests/unit/tests/injected/visualization/drawer.test.ts
```diff
@@ -877,35 +877,24 @@ describe('Drawer', () => {
...
-        windowUtilsMock.setup(x => x.clearTimeout(It.isAny())).verifiable(Times.never());
-        windowUtilsMock
-            .setup(x => x.setTimeout(itIsFunction, HighlightBoxDrawer.recalculationTimeout))
-            .returns(() => timeOutId)
-            .verifiable();
+        const drawMock = Mock.ofInstance(() => {});
+        drawMock.setup(draw => draw()).verifiable(Times.exactly(2));
+        (testSubject as any).draw = drawMock.object;
 
+        // call once, run all timers, call again; should yield two draw calls.
```
dbjorge, line 888: It would be nice for there to also be a test case that without `runAllTimers` in between, the throttling behavior causes `draw` to only execute once.

dbjorge:
One lint failure and a suggestion testing, but otherwise the code itself looks good! Will try it out to see the perf impact shortly

dbjorge:
Tried it out locally. Saw no visible difference on simpler test pages, and notable improvement on a more complex page (cnn.com); in particular, it did a much better job of handling floating header/footers, where the old version required a second after scrolling to "catch up" and the new version was not quite keeping up real-time but was staying much closer, more like 50ms behind the scrolling.

Looks great to me, pending the minor updates from earlier!

src/tests/unit/tests/injected/visualization/drawer.test.ts
waabid: Good call! I actually did have a test for that ... my concern was that it might potentially introduce flakiness to the tests as I was seeing inconsistent results. However, that was before I was using the mock timers and I think with them, it should be consistently failing.

PR #2787, opened 2020-06-02
feat(android-setup): initial android setup UI component

karanbirsingh:
#### Description of changes

This PR adds android setup UI dialog infrastructure behind a feature flag. Users with the feature flag off should see no difference in behavior. Each android setup step will have its own component tsx that serves as the device-connect-body. The mapping between the current step and the component to show is facilitated by android-setup-step-container. 

The actual UI added behind the feature flag is a placeholder text tag.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/views/device-connect-view/components/device-connect-view-container.tsx
```diff
@@ -26,34 +31,48 @@ export type DeviceConnectViewContainerProps = {
...
+                disableJSXElement={productionDeviceConnectView(props)}
+            ></FlaggedComponent>
         );
     },
 );
+
+const productionDeviceConnectView = (props: DeviceConnectViewContainerProps) => {
+    return (
+        <div className={deviceConnectView}>
+            <WindowTitle
```
karanbirsingh, line 56: The WindowTitle will be shared, although the page title might be different(?). Need to decide what's best

src/electron/views/device-connect-view/components/device-connect-view-container.tsx
```diff
@@ -26,34 +31,48 @@ export type DeviceConnectViewContainerProps = {
...
+            </WindowTitle>
+            <div className={mainContentWrapper}>
+                <DeviceConnectBody
+                    deps={props.deps}
+                    viewState={{
+                        deviceConnectState: props.deviceStoreData.deviceConnectState,
+                        connectedDevice: props.deviceStoreData.connectedDevice,
+                    }}
+                />
+                <TelemetryPermissionDialog
```
karanbirsingh, line 71: Should the telemetry dialog be part of one of the setup step tsx components?

DaveTryon: I don't think so. Imagine a future version where the user needs to explicitly choose and android scanning option--we'd probably still want the telemetry dialog to be the first thing to appear

src/electron/views/device-connect-view/components/android-setup/android-setup-step-component-provider.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { CommonAndroidSetupStepProps } from 'electron/views/device-connect-view/components/android-setup/android-setup-types';
+import * as React from 'react';
+
+export class AndroidSetupStepComponent extends React.Component<CommonAndroidSetupStepProps> {
+    constructor(props: CommonAndroidSetupStepProps) {
+        super(props);
+    }
+
+    public render(): JSX.Element {
+        const Component = this.props.deps.androidSetupStepComponentProvider[
```
DaveTryon, line 12: This is an interesting pattern. Is it something that we already use in other parts of the code?

JGibson2019: I also haven't seen this in our code before; do you have a link to the doc/book you saw this pattern in originally @karanbirsingh? Curious about seeing it in other places!

src/electron/views/device-connect-view/components/device-connect-view-container.tsx
JGibson2019: When you say it might be different, this means it would change for the different steps of the setup? Or do you mean something else?

src/electron/views/device-connect-view/components/android-setup/android-setup-step-component-provider.tsx
dbjorge: We use a (more complicated) version of this pattern for the `ContentPageComponent`s that define the various guidance and "info & example" page/panel content; you can see it in `/src/views/content/content-page.tsx` (though you can ignore most of its implementation, most of it is mechanisms for registering/looking up components that aren't relevant to what we're doing here). The simplified version of what that does is:

```ts
// definitions in content-page.ts
export type ContentPageComponent = React.FC<ContentPageProps> /* ... plus some omitted complexity */;
export type ContentReference = string /* ... plus some omitted complexity */;
export interface ContentProvider {
    contentFromReference(reference: ContentReference): ContentPageComponent;
    // ...
}

// usage in render() of content-panel.tsx
const ContentPage = deps.contentProvider.contentFromReference(content);
return (
    <Panel { ...irrelevantPanelProps }>
        <div className="content">
            <ContentPage deps={deps} />
        </div>
    </Panel>
);
```

src/electron/views/device-connect-view/components/android-setup/android-setup-step-component-provider.tsx
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { CommonAndroidSetupStepProps } from 'electron/views/device-connect-view/components/android-setup/android-setup-types';
+import * as React from 'react';
+
+export class AndroidSetupStepComponent extends React.Component<CommonAndroidSetupStepProps> {
```
dbjorge, line 6: I think this class name might be a bit confusing; I feel like it implies that it should represent one component of an android setup step, rather than the thing responsible for selecting the appropriate component based on input. `AndroidSetupStepDispatcher`? `CurrentAndroidSetupStep`? `AndroidSetupStepContainer`? `AndroidSetupWizard`?

src/electron/views/device-connect-view/components/device-connect-view-container.tsx
dbjorge: +1. I would lean towards just wrapping the `DeviceConnectBody` in the `FlaggedComponent` and leaving the `WindowTitle` and `TelemetryPermissionDialog` as-is regardless of the flag

src/electron/views/device-connect-view/components/android-setup/android-setup-types.ts
```diff
@@ -0,0 +1,26 @@
...
+    deps: AndroidSetupPageDeps;
+};
+
+export type AndroidSetupPageDeps = {
+    androidSetupActionCreator: AndroidSetupActionCreator;
+    androidSetupStepComponentProvider: AndroidSetupStepComponentProvider;
+    LinkComponent: LinkComponentType;
+};
+
+export type AndroidSetupStepComponentProvider = Partial<
```
dbjorge, line 24: Would maybe be good to leave a note here clarifying that this `Partial` should be removed once feat(android-setup) is more complete

src/electron/views/device-connect-view/components/device-connect-view-container.tsx
karanbirsingh: At first I thought it might change for the different steps of the setup, but that doesn't seem right anymore. So we will probably be able to leave the WindowTitle out of the FlaggedComponent as Dan suggests

src/electron/views/device-connect-view/components/device-connect-view-container.tsx
```diff
@@ -26,28 +31,30 @@ export type DeviceConnectViewContainerProps = {
...
-                <WindowTitle
-                    pageTitle={'Connect to your Android device'}
-                    deps={props.deps}
-                    windowStateStoreData={props.windowStateStoreData}
-                >
-                    <HeaderIcon invertColors deps={props.deps} />
-                </WindowTitle>
                 <div className={mainContentWrapper}>
-                    <DeviceConnectBody
+                    <WindowTitle
```
dbjorge, line 44: Was moving the WindowTitle inside the mainContentWrapper div intentional?

karanbirsingh: nope, thanks for catching!

PR #2791, opened 2020-06-03
feat(android-setup): Add core state machine logic

RobGallo:
#### Description of changes

Given a set of steps (states), the job of the state machine is to link a step to a set of actions. When an action is invoked on the state machine, the action is invoked for the current step. The step itself may fire a callback to transition to another step and the process repeats.

This implementation is designed to link specific Action<> invocations to the actions contained in the steps. In this way, a store which uses the state machine can forward action calls, and consistency with the standard actions mechanism can be maintained.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/electron/platform/android/setup/state-machine/state-machine.test.ts
+++ b/src/tests/unit/tests/electron/platform/android/setup/state-machine/state-machine.test.ts
@@ -4,7 +4,6 @@
 import { Action } from 'common/flux/action';
 import { StateMachine } from 'electron/platform/android/setup/state-machine/state-machine';
 import { StateMachineSteps } from 'electron/platform/android/setup/state-machine/state-machine-steps';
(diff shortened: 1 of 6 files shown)


src/tests/unit/tests/electron/platform/android/setup/state-machine/state-machine.test.ts
```diff
@@ -0,0 +1,291 @@
...
+type MartiniStepId = 'gin' | 'vermouth' | 'olives';
+
+class MartiniActions {
+    public shake = new Action<void>();
+    public stir = new Action<void>();
+    public garnish = new Action<string>();
+}
+
+type MartiniSteps = StateMachineSteps<MartiniStepId, MartiniActions>;
+type MartiniStepTransition = (stepId: MartiniStepId) => void;
```
JGibson2019, line 18: 🍸 :) 

But also @RobGallo I've not seen the Martini pattern with state machines before. Do you know where I could read more about implementing this/look at other examples?

Issue #2758, opened 2020-05-26
Don't show "no content available" page if target page URL changes if all url permissions granted


waabid:
## Is your feature request related to a problem? Please describe.

Every time you switch the target page to another website, the details view shuts down. It does this because of activeTab permissions. However, if I set the permissions for allUrls to be granted, the "reactivation of Accessibility Insights for Web" is not required, and so I shouldn't see this page.

## Describe the desired outcome
If allUrl permissions are granted, then I shouldn't see the "no content available" page when the target page is changed to a different url/host.

## Describe alternatives you've considered
n/a

## Additional context
The no content available page should still be shown when the target page is closed or if the user doesn't allow all url permissions (and the target page url is changed).

PR #2793, opened 2020-06-03
feat:Avoid no content available page when URL permissions granted

jlperkins:
#### Description of changes

Changing behavior of the automated checks in FastPass when all URL permissions are granted to avoid showing the "no content available" page.  Addresses issue #2758 . Added variables to see if permissions were enabled and if the origin of the URL had changed. Amended unit tests as needed. 

#### Pull request checklist

- [x] Addresses an existing issue: #2758 
- [x] Ran `yarn fastpass`


jlperkins:
avoided
no content available
These are images of the details view as it is now when permissions are enabled and the target page is navigated away from, and then the "no content available" page that it previously showed before these changes.

jlperkins:
no permissions, no content available
Gif of user running Automated Checks and then navigating away from the target page to another site.  The "no content available" page is displayed as a result, which is the expected behavior.

jlperkins:
yes permissions, avoid no content available
Gif of user running Automated Checks and then clicks the "give Accessibility Insights additional permissions" link in the message about iframes.  User selects "allow" on the popup about permissions that appears.  Then, the user navigates away from the target page.  Instead of the "no content available" page, the Automated Checks page remains visible and a message about the target page having been changed is displayed, which is the expected behavior, since permissions were granted.

src/DetailsView/details-view-container.tsx
```diff
@@ -85,13 +86,17 @@ export interface DetailsViewContainerState {
...
+    permissionsStateStoreData: PermissionsStateStoreData;
 }
 
 export class DetailsViewContainer extends React.Component<DetailsViewContainerProps> {
     private initialRender: boolean = true;
 
     public render(): JSX.Element {
-        if (this.isTargetPageClosed()) {
+        if (
+            !this.hasStores() ||
```
haonliu, line 97: NIT: would be good to extract this condition into a separate function as well.

PR #2797, opened 2020-06-04
feat(android-setup): Implement AppiumServiceConfigurator

DaveTryon:
#### Description of changes

This adds the `AppiumServiceConfigurator` class, which implements the `AndroidServiceConfigurator` interface using appium-adb as our layer to interact with ADB. it also adds the `AndroidServiceConfiguratorFactory` to build it with the path to the Android SDK root. The code has been validated against both an emulator and a physical device. The physical device returned slightly different responses than the emulator, so it's possible that newer Android versions may behave unexpectedly. We're using appium-adb's handling as much as possible to _minimize_ this risk, but I don't know of a way that we can completely _eliminate_ it.

The unit tests mock out appium-adb's `ADB` class.

#### Pull request checklist

- [x] Addresses an existing issue: [1724327](https://mseng.visualstudio.com/1ES/_workitems/edit/1724327)
- [x] Ran `yarn fastpass`


src/tests/unit/tests/electron/platform/android/appium-service-configurator-factory.test.ts
```diff
@@ -0,0 +1,54 @@
...
+                expect(error.message.includes(nonExistentPath)).toBeTruthy();
+                expect(error.message.includes(ANDROID_HOME)).toBeFalsy();
+                expect(error.message.includes(ANDROID_SDK_ROOT)).toBeFalsy();
+            }
+        }
+    });
+
+    // Appium-adb's built-in behavior is to use the environment variables if
+    // they're set, and without a good way to override them, we just skip the tests.
+    function canTestBeRun(): boolean {
```
dbjorge, line 44: Rather than have the unit test depend on the environment, I'd suggest injecting `ADB.createADB` as a constructor parameter in the class under test so the unit tests can mock it

src/tests/unit/tests/electron/platform/android/appium-service-configurator-factory.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    const nonExistentPath = './This/Path/Does/Not/Exist';
+    const ANDROID_HOME = 'ANDROID_HOME';
+    const ANDROID_SDK_ROOT = 'ANDROID_SDK_ROOT';
+    const errorsFromAdb = 3;
+
+    it('getServiceConfigurator uses default if no path is provided', async () => {
+        if (canTestBeRun()) {
+            const factory = new AppiumServiceConfiguratorFactory();
+            expect.assertions(errorsFromAdb);
+            try {
```
dbjorge, line 16: I think the more typical way for us to test that exceptions are thrown is to use `expect(() => someCall()).toThrowError('error message');`. For a promise, there's also a more explicit `.rejects` option.

Combined with the other feedback on this test file, I think I'd suggest someting like https://gist.github.com/dbjorge/b90e27cea6e6d4d7150545f26379a3de

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,144 @@
...
+                const output: Array<DeviceInfo> = [];
+                Object.values(detectedDevices).forEach(device => output.push(device as DeviceInfo));
+                resolve(output);
+            } catch (error) {
+                reject(error);
+            }
+        });
+    };
+
+    private AddDevices(
```
dbjorge, line 47: We usually prefer `camelCase` to `PascalCase` for functions

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export class AppiumServiceConfigurator implements AndroidServiceConfigurator {
+    private readonly adb: ADB;
+
+    constructor(adb: ADB) {
+        this.adb = adb;
+    }
+
+    public getConnectedDevices = (): Promise<Array<DeviceInfo>> => {
+        return new Promise<Array<DeviceInfo>>(async (resolve, reject) => {
```
dbjorge, line 26: There's a few places in this file where you're using a pattern like

```ts
public someAsyncMethod = (): Promise<OutputT> => {
    return new Promise<OutputT>(async (resolve, reject) => {
        try {
            const output = await doWork();
            resolve(output);
        } catch (error) {
            reject(error);
        }
    });
}
```

You can simplify this a lot by just making the method itself async; the language can deal with most of the wrapping you're currently doing by hand:

```ts
public someAsyncMethod = async (): Promise<OutputT> => {
    return await doWork();
}
```

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: Yup, muscle memory from C#

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,144 @@
...
+            }
+        });
+    };
+
+    public setTcpForwarding = (deviceId: string): Promise<void> => {
+        return new Promise<void>((resolve, reject) => {
+            try {
+                const port: number = 62442;
+                this.adb.setDeviceId(deviceId);
+                this.adb.forwardPort(port, port);
```
dbjorge, line 137: `forwardPort` is asynchronous; we should be waiting for it before resolving

src/tests/unit/tests/electron/platform/android/appium-service-configurator.test.ts
```diff
@@ -0,0 +1,303 @@
...
+        const expectedPackageInfo: PackageInfo = { versionCode } as PackageInfo;
+        adbMock.setup(m => m.setDeviceId(emulatorId)).verifiable(Times.once());
+        adbMock
+            .setup(m => m.getPackageInfo(servicePackageName))
+            .returns(() => expectedPackageInfo)
+            .verifiable(Times.once());
+
+        const info = await testSubject.getPackageInfo(emulatorId);
+        expect(info.versionCode).toBe(versionCode);
+        expect(info.versionName).toBeUndefined();
```
dbjorge, line 158: Is there a case where this would be expected? I would have expected us to treat this as an error

src/tests/unit/tests/electron/platform/android/appium-service-configurator.test.ts
```diff
@@ -0,0 +1,303 @@
...
+            .verifiable(Times.once());
+
+        const permissionInfo: PermissionInfo = await testSubject.getPermissionInfo(emulatorId);
+
+        expect(permissionInfo.screenshotGranted).toBe(false);
+
+        adbMock.verifyAll();
+    });
+
+    it('getPermissionInfo, service is running but has no screenshot permission', async () => {
```
dbjorge, line 212: Test name seems off?

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: Yup!

src/tests/unit/tests/electron/platform/android/appium-service-configurator.test.ts
DaveTryon: versionName is optional, but I agree that we should always be setting it

src/tests/unit/tests/electron/platform/android/appium-service-configurator.test.ts
DaveTryon: Copy/paste error

DaveTryon: Copy/paste error

src/electron/platform/android/appium-service-configurator-factory.ts
```diff
@@ -0,0 +1,47 @@
...
+        return ADB.createADB(parameters);
+    };
+}
+
+export class AppiumServiceConfiguratorFactory implements AndroidServiceConfiguratorFactory {
+    public async getServiceConfigurator(sdkRoot: string): Promise<AndroidServiceConfigurator> {
+        return this.getServiceConfiguratorTestable(sdkRoot, new ProductionAdbCreator());
+    }
+
+    public async getServiceConfiguratorTestable(
```
dbjorge, line 32: It's not necessary to have a separate "testable" verssion like this; our normal pattern is to leave the responsibility of composing the "real" dependency to the top-level initializer that's composing the dependency tree, and just have most classes implement only functionality that requires injected dependencies. I think the usual pattern would look something like:

```ts
// appium-service-configurator-factory.ts
export type AdbCreator = typeof ADB.createADB;
export class AppiumServiceConfiguratorFactory implements AndroidServiceConfiguratorFactory {
    public constructor(private readonly createAdb: AdbCreator) { }
    // in impl, use "await this.createAdb(parameters)"
}

// unit tests
let mockCreateAdb: IMock<AdbCreator>;
it('invokes createAdb as expected', async () => {
    mockCreateAdb.setup(m => m(expectedParam)).returns(() => Promise.resolve(stubAdbInstance)).verifiable();
    const testSubject = new AppiumServiceConfiguratorFactory(mockCreateAdb);
    // ...
});

// production initializer
const androidServiceConfiguratorFactory = new AppiumServiceConfiguratorFactory(ADB.createADB);
```

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,99 @@
...
+        this.adb.setDeviceId(id);
+        const friendlyName = await this.adb.getModel();
+        return { id, isEmulator, friendlyName };
+    }
+
+    public getPackageInfo = async (deviceId: string): Promise<PackageInfo> => {
+        this.adb.setDeviceId(deviceId);
+        const info: PackageInfo = await this.adb.getPackageInfo(servicePackageName);
+        const output: PackageInfo = {};
+        if (info?.versionCode) {
```
dbjorge, line 61: As a rule, I don't recommend relying on how javascript decides whether a non-boolean is "truthy" or not, it leads to some surprising behavior a lot of the time (eg, as-written, if versionCode is 0 or versionName is an empty string, those will both propagate as `undefined` from this function)

I think you can avoid that issue and simplify the code by using this instead:

```ts
return {
    versionCode: info?.versionCode,
    versionName: info?.versionName,
};
```

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,99 @@
...
+        Object.values(detectedDevices).forEach(device => output.push(device as DeviceInfo));
+        return output;
+    };
+
+    private async addDevices(
+        devices: Array<AdbDevice>,
+        isEmulator: boolean,
+        detectedDevices: any,
+    ): Promise<void> {
+        for (let loop = 0; loop < devices.length; loop++) {
```
dbjorge, line 42: A foreach loop would match the intent a little more closely (`for (const device of devices)` is the syntax)

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        const output: Array<DeviceInfo> = [];
+        Object.values(detectedDevices).forEach(device => output.push(device as DeviceInfo));
+        return output;
+    };
+
+    private async addDevices(
+        devices: Array<AdbDevice>,
+        isEmulator: boolean,
+        detectedDevices: any,
```
dbjorge, line 40: This can be given a type (either `{ [udid: string]: DeviceInfo }` or `DictionaryStringTo<DeviceInfo>`, whichever you find more readable)

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public getConnectedDevices = async (): Promise<Array<DeviceInfo>> => {
+        // First add devices that are flagged as emulator, then
+        // add devices that aren't flagged as emulators
+        const detectedDevices = {};
+        await this.addDevices(await this.adb.getConnectedEmulators(), true, detectedDevices);
+        await this.addDevices(await this.adb.getConnectedDevices(), false, detectedDevices);
+
+        const output: Array<DeviceInfo> = [];
```
dbjorge, line 32: You don't need the intermediate array, you can just `return Object.values(detectedDevices);`

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export class AppiumServiceConfigurator implements AndroidServiceConfigurator {
+    private readonly adb: ADB;
+
+    constructor(adb: ADB) {
+        this.adb = adb;
+    }
+
+    public getConnectedDevices = async (): Promise<Array<DeviceInfo>> => {
+        // First add devices that are flagged as emulator, then
```
dbjorge, line 26: Now that the implementation has been simplified I don't think this comment is adding any value; the code already expresses this readably

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -0,0 +1,99 @@
...
+type AdbDevice = {
+    udid: string;
+};
+
+const servicePackageName: string = 'com.microsoft.accessibilityinsightsforandroidservice';
+
+export class AppiumServiceConfigurator implements AndroidServiceConfigurator {
+    private readonly adb: ADB;
+
+    constructor(adb: ADB) {
```
dbjorge, line 21: You can avoid some of the boilerplate by declaring the field within the ctor (`public constructor(private readonly adb: ADB) { }`)

src/tests/unit/tests/electron/platform/android/appium-service-configurator-factory.test.ts
DaveTryon: Done

src/tests/unit/tests/electron/platform/android/appium-service-configurator-factory.test.ts
DaveTryon: Done

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: Thanks! Much nicer in this format :)

src/tests/unit/tests/electron/platform/android/appium-service-configurator.test.ts
DaveTryon: I updated this to make versionName and versionCode both optional, so that we can distinguish between "the package isn't installed" and "the package is installed but we can't get useful version info". In the former case, we would just install the package, while in the latter case we would uninstall the existing package before installing the new one

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: Done

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: I agree, and originally wrote this using foreach. Unfortunately, foreach doesn't understand the await syntax and causes the function to exit without waiting. I'll add a comment to that effect

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: Done

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: Removed

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: Nice!

src/electron/platform/android/appium-service-configurator-factory.ts
DaveTryon: Let's talk about this. It feels wrong to have the appium-adb dependency leak all the way to the initializer.

src/electron/platform/android/appium-service-configurator.ts
dbjorge: Were you perhaps trying to uses a loop like `devices.forEach((device) => { /* stuff */ });`? It's true that that wouldn't work as-is, but you should be able to use the statement form of foreach loop with async/await (`for (const device of devices) { /* stuff */ }`).

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: `DictionaryStringTo` is closest to what I'm used to

src/electron/platform/android/appium-service-configurator-factory.ts
dbjorge: Discussed offline and agreed on a compromise between these approaches

src/electron/platform/android/appium-service-configurator.ts
DaveTryon: Thanks--updated

PR #2801, opened 2020-06-04
feat(android-setup): add adb location to user config

karanbirsingh:
#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/common/types/store-data/user-configuration-store.ts
```diff
@@ -11,6 +11,8 @@ export interface UserConfigurationStoreData {
 
     bugService: string;
     bugServicePropertiesMap: IssueFilingServicePropertiesMap;
+
+    adbLocation: string;
```
dbjorge, line 15: I would do `string | null` with a comment about what `null` implies; it's not technically necessary since we don't have strict null checks enabled, but I think it makes it more clear to the reader.

PR #2803, opened 2020-06-04
Add CodeQL security scanning

jhutchings1:
Hi, I'm a PM on the GitHub security team. This repository is eligible to try the new [GitHub Advanced Security code scanning](https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning) beta.

Code scanning runs a static analysis tool called CodeQL which scans your code at build time to find any potential security issues. We've tuned the set of queries to be only the most severe, most precise issues. We'll show alerts in the security tab, and we'll show alerts for any net new vulnerabilities on pull requests as well. We've tried to make this super developer friendly, but we'd love your feedback as we work through the beta. 

If you're interested in trying it out, you can merge this pull request to set up the Actions workflow. You can also get this set up yourself in any additional repositories in this organization by following these [instructions](https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,52 @@
+name: "Code scanning - action"
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 19 * * 0'
+
+jobs:
+  CodeQL-Build:
+
+    # CodeQL runs on ubuntu-latest and windows-latest
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+      
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      # Override language selection by uncommenting this and choosing your languages
+      # with:
+      #   languages: go, javascript, csharp, python, cpp, java
+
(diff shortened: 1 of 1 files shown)


dbjorge:
Thanks for the contribution! Our team discussed this a little among ourselves. We're interested in the idea of this tool, but we'd like to confirm that it is successfully able to detect issues before merging; it's not clear from the initial successful results whether our codebase is actually squeaky-clean vs needing a bit of extra attention to the build step for the tool to scan properly (I notice that the auto-builder task logs suggest that it is no-op'ing, which makes me suspicious it might be the latter).

We're a little hesitant to add the maintenance/reliability burden of a second CI system for the sake of one tool (our other builds use Azure Pipelines rather than GitHub actions), but we're willing to consider it if the tool demonstrates value. In particular, we'd love to see an example of the tool catching an issue (eg, try this configuration in a branch with an artificially introduced example failure).

jhutchings1:
@dbjorge Thanks for your questions on this! Your project is a JS/Typescript project, so it's a "scanned" language which doesn't require compilation. As a result, it's basically bulletproof on all but the very biggest projects. 

In terms of the analysis capabilities, CodeQL is the same analysis engine that MSFT is running internally for validation of most projects as part of the SDL. It's super capable, and you can see a very interesting demo of the capabilities [in our announcement presentation from GitHub Satellite](https://www.youtube.com/watch?v=58N0_0HCDPE). What this is testing testing here is a new runtime for that same engine. We've tuned the default set of queries very tight, so we expect that we'll only find a small number of issues on a medium sized project like this. We did find a warning in your code around some of your string sanitization. The most important reason you'd enable something like this is to watch for new security concerns over time. We'll pop an alert in a pull request if a developer makes a mistake that could introduce a security vulnerability.

dbjorge:
We discussed this with our engineering team in our weekly "engineering discussion" yesterday to make sure everyone would understand what was going on with the new checks. Folks were excited to give it a try, and unanimously agreed that it would be a good idea to enable it. Thanks again, Justin!

PR #2814, opened 2020-06-08
feat(android-setup):UI for detect devices step

JGibson2019:
#### Description of changes
Creating the UI that lets users know we're scanning for devices. I also created a reusable component that can be utilized for other steps where we need the "Cancel" and "Start testing" buttons with a spinner; the Cancel button invokes the cancel action for adb setup experience, and the spinner label text will be customized to the situation. Added unit tests for new components added

Image of the new UI; spinner with "Scanning for devices...", an enabled Cancel button and disabled Start testing button. 
image

#### Pull request checklist

- [N/A] Addresses an existing issue: #0000
- [X] Ran `yarn fastpass`


src/electron/views/device-connect-view/components/android-setup/android-setup-spinner-layout.tsx
```diff
@@ -0,0 +1,42 @@
...
+        return (
+            <>
+                <Spinner
+                    className={androidSetupSpinner}
+                    size={SpinnerSize.large}
+                    label={props.spinnerLabel}
+                    role="alert"
+                    aria-live={props['aria-live']}
+                />
+                <DeviceConnectFooter
```
dbjorge, line 34: I think we decided in the feature chat that we should use "Next" rather than the `DeviceConnectFooter`'s "Start testing" for the disabled bottom-right button. You could achieve this by updating the `DeviceConnectFooter` to take a a label param; however, I chose not to do that for the `AndroidSetupPromptLayout`, since I thought repeating a bit of the impl now would make it easier to remove the obsolete components along with the feature flag at the end of the feature.

That said, it does seem like it would be a good idea to share the wrapping `<main>` element and the footer button layout between the spinner and prompt layouts; maybe sharing a "base layout" between this class and the prompt layout would be the way to go? I'm imagining splitting out an `AndroidSetupBaseLayout` from part of the current `AndroidSetupPromptLayout` such that this `AndroidSetupSpinnerLayout` might look something like:

```ts
return (<AndroidSetupBaseLayout {...props}>
    <Spinner
        className={androidSetupSpinner}
        size={SpinnerSize.large}
        label={props.spinnerLabel}
        role="alert"
        aria-live="polite" />
</AndroidSetupBaseLayout>);
```

JGibson2019:
Closing since we're creating different shared components across views

PR #2830, opened 2020-06-09
feat(android-setup): Add prompt install service

DaveTryon:
#### Description of changes

This PR adds the dialog that prompts the user to install the service on the selected device. It also includes the `DeviceDescription` component that will be used in multiple dialogs. The `DeviceDescription` component will need to be revised when we reach the final dialog that displays the running application in addition to the device description.

Screenshot with a device:
image

Screenshot with an emulator:
image

#### Pull request checklist

- [x] Addresses an existing issue: [1724344](https://mseng.visualstudio.com/1ES/_workitems/edit/1724344)
- [x] Ran `yarn fastpass`


DaveTryon:
Temporarily closing to rework with flexbox

src/electron/views/device-connect-view/components/android-setup/device-description.tsx
```diff
@@ -0,0 +1,28 @@
...
+import * as styles from './device-description.scss';
+
+export interface DeviceDescriptionProps extends DeviceMetadata {}
+
+export const DeviceDescription = NamedFC<DeviceDescriptionProps>('DeviceDescription', props => {
+    const iconName: string = props.isEmulator ? 'Devices3' : 'CellPhone';
+    const iconAriaLabel: string = props.isEmulator ? 'Emulator' : 'Device';
+
+    return (
+        <React.Fragment>
```
dbjorge, line 17: You can omit the `<React.Fragment>` wrapper here; it's the longhand way to write `<>` in a tsx file, but from a render function, it's only necessary if you need to have render return multiple root-level values

src/electron/views/device-connect-view/components/android-setup/device-description.scss
```diff
@@ -0,0 +1,15 @@
...
+@import '../../../../../common/styles/fonts.scss';
+
+.content {
+    display: flex;
+    line-height: 19px;
+    margin-top: 24px;
+    margin-bottom: 24px;
+}
+
+.iconContent {
```
dbjorge, line 12: nit: class names are usually `kebab-cased` (the build process will normalize it to `camelCase` for typescript to use)

src/electron/views/device-connect-view/components/android-setup/prompt-install-service-step.tsx
```diff
@@ -0,0 +1,52 @@
...
+                disabled: true,
+                onClick: null,
+            },
+        };
+
+        const descriptionProps: DeviceDescriptionProps = props.androidSetupStoreData.selectedDevice;
+
+        return (
+            <AndroidSetupStepLayout {...layoutProps}>
+                <React.Fragment>
```
dbjorge, line 43: Same as earlier comment, the `<React.Fragment>` wrapper isn't really doing anything here; I'd recommend either using a `<p>` tag instead or just omitting it

src/electron/views/device-connect-view/components/android-setup/device-description.scss
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+@import '../../../../../common/styles/fonts.scss';
+
+.content {
+    display: flex;
+    line-height: 19px;
+    margin-top: 24px;
```
dbjorge, line 8: I think the vertical alignment between the icon and the text is just slightly off from the figma; I think the exact match for the figma would be

```css
align-items: center;
height: 48px;
margin-top: 16px;
margin-bottom: 16px;
```

You might also consider handling the margins in the step's CSS rather than the DeviceDescription's (generally, spacing between sibling elements should be the responsibility of the component that's composing those sibling elements, especially since we know we'll want to reuse this in a different step with different margin requirements), but if you'd rather leave that for the prompt-choose-device UI PR to handle, that's okay too

src/electron/views/device-connect-view/components/android-setup/device-description.tsx
DaveTryon: Removed

src/electron/views/device-connect-view/components/android-setup/device-description.scss
DaveTryon: Seems weird to have different names in the style and in the code, and looking at the existing code I don't see it used a whole lot. Let's discuss this, and maybe make it a point for an engineering discussion.

DaveTryon: changed

src/electron/views/device-connect-view/components/android-setup/prompt-install-service-step.tsx
DaveTryon: Still had to keep <> and </> around the text or it wouldn't compile, but removed the React.Fragment

src/electron/views/device-connect-view/components/android-setup/device-description.scss
DaveTryon: marginTop and marginBottom are now passed in via the DeviceDescriptionProps

DaveTryon:
@dbjorge, changes have been made as requested. Please let me know if there are comments that I misunderstood 😵

PR #2865, opened 2020-06-11
feat(android-setup): Create three state machine steps

RobGallo:
#### Description of changes

The new steps are: 

- prompt-locate-adb
-     prompt-connect-to-device
-     detect-devices
    

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/electron/platform/android/setup/steps/detect-adb.ts
```diff
@@ -7,6 +7,6 @@ export const detectAdb: AndroidSetupStepConfig = deps => ({
     actions: {},
     onEnter: async () => {
         const detected = await deps.hasAdbPath();
-        deps.stepTransition(detected ? 'detect-devices' : 'prompt-locate-adb');
+        deps.stepTransition(detected ? 'prompt-connect-to-device' : 'prompt-locate-adb');
```
DaveTryon, line 10: Our golden path includes finding ADB by environment variables. When that's the case, hasAdbPath will return false but we should attempt to get the devices anyway. We should only go to prompt-locate-adb after detectDevices has failed. Not part of the changes of this PR, and it may also be included in future work, but I wanted to call it out while I was thinking about this path

DaveTryon, line 10: Clarification of logic: Our transition to detect-devices should be independent of hasAdbPath, because detect-devices is able to use environment variables that aren't in our store. Flow should be:

1: get adbPath (null is OK)
2: Call detectDevices(adbPath)
3: If error (ADB wasn't found), prompt for ADB path. Update ADB path, then restart at step 1
4. Apply logic based on how many devices were found. 0 => prompt to connect, 1 => select it, 2+ => choose device

PR #2884, opened 2020-06-16
feat(android-setup): Connect steps to adb

DaveTryon:
#### Description of changes
Add code to provide the ADB-based implementation of the steps dependencies. This code also integrates the AndroidServiceApkLocator and adds the initialization code. There are no UI changes in this PR.

One caveat: The `semver-comparer` class is intended to be temporary and is added to unblock the feature. If we have a more standardized way of comparing 2 semver strings, I'll use it. If not, I'll do the work needed to bring this class up to production levels. At the moment it has no unit tests and doesn't cover any semver values with decorators.

#### Pull request checklist

- [x] Addresses an existing issue: [1724353](https://mseng.visualstudio.com/1ES/_workitems/edit/1724353)
- [x] Ran `yarn fastpass`


src/tests/unit/tests/electron/platform/android/appium-service-configurator.test.ts
```diff
@@ -200,13 +213,13 @@ describe('AppiumServiceConfigurator tests', () => {
...
+        apkLocatorMock.verifyAll();
     });
 
     it('getPermissionInfo, propagates error', async () => {
-        const expectedMessage = 'Something bad happened';
-        adbMock.setup(m => m.setDeviceId(emulatorId)).verifiable(Times.once());
+        const expectedMessage = 'Thrown during getPermissionInfo';
         adbMock
-            .setup(m => m.shell(['dumpsys', 'accessibility']))
+            .setup(m => m.setDeviceId(emulatorId))
```
DaveTryon, line 222: I intentionally moved the exception 'upstream', from the shell call to the setDeviceId call. Did this in several test cases for consistenty

src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    public hasExpectedServiceVersion = async (): Promise<boolean> => {
+        try {
+            const installedVersion: string = await this.getInstalledVersion();
+            if (installedVersion) {
+                return installedVersion === (await this.getTargetVersion());
+            }
+        } catch (error) {
+            console.log(error);
```
DaveTryon, line 55: This tends to clutter up the test. Is there a recommended way to just silently eat an exception? When I remove the console.log call, I get warnings about having an empty block. Not quite sure how to handle this...

dbjorge: The usual way we handle that is to inject a `Logger` dependency rather than invoking `console` directly, that way you can just stub it as a no-op logger for the error cases

DaveTryon: Sweet! Will do

src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -0,0 +1,108 @@
...
+            this.serviceConfig = await this.configFactory.getServiceConfigurator(adbLocation);
+            return true;
+        } catch (error) {
+            this.logger.log(error);
+            return false;
+        }
+    };
+
+    public setAdbPath = (path: string): void => {
+        this.configStore.getState().adbLocation = path;
```
RobGallo, line 40: Is it okay to update a property on the state data of a store directly and externally? Should emitChange be called when the data is updated?

src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -0,0 +1,108 @@
...
+    };
+
+    public getDevices = async (): Promise<DeviceInfo[]> => {
+        return await this.serviceConfig.getConnectedDevices();
```
RobGallo, line 44: I believe you don't need to await this promise. As it is, this function isn't really async.

src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -0,0 +1,108 @@
...
+    };
+
+    public setSelectedDeviceId = (id: string): void => {
+        this.selectedDeviceId = id;
+    };
+
+    public hasExpectedServiceVersion = async (): Promise<boolean> => {
+        try {
+            const installedVersion: string = await this.getInstalledVersion();
```
RobGallo, line 53: If we need to await this promise, this function shouldn't be async. We can just return the value.

src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -0,0 +1,108 @@
...
+            if (installedVersion) {
+                return installedVersion === (await this.getTargetVersion());
```
RobGallo, line 55: Isn't this where the SemVer comparer should go?

src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -0,0 +1,108 @@
...
+            }
+        } catch (error) {
+            this.logger.log(error);
+        }
+        return false;
+    };
+
+    public installService = async (): Promise<boolean> => {
```
RobGallo, line 63: Didn't we decide it is okay for this function to throw?

DaveTryon: I was holding that for a separate PR

src/electron/platform/android/setup/live-android-setup-deps.ts
DaveTryon: I made it a private method instead of a class (on the assumption that we'll eventually use a library here instead of our method)

src/electron/platform/android/setup/live-android-setup-deps.ts
DaveTryon: Added the plumbing through the messageCreator

src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -0,0 +1,128 @@
...
+import {
+    AndroidServiceConfigurator,
+    AndroidServiceConfiguratorFactory,
+    DeviceInfo,
+    PackageInfo,
+    PermissionInfo,
+} from 'electron/platform/android/android-service-configurator';
+import { AndroidSetupDeps } from 'electron/platform/android/setup/android-setup-deps';
+
+export class LiveAndroidSetupDeps implements AndroidSetupDeps {
```
dbjorge, line 17: I think it's a little confusing for a "deps" type like this to contain actual business logic and state; I worry that it feels a little too much like a god-class in terms of the number of orthogonal bits of logic it contains. I think it would be easier for a future reader to understand if we tried to contain actual business logic in classes whose names more closely match that logic's responsibilities; for example:

* I think the state machine steps are probably where I'd first think to look for decisions about "should some particular type of failure result in an error state transition vs silently logging"; I think that's a more natural home for most of the try/catch/log handling
* I think if I were looking for where logic about installing the service or configuring its permissions might live, the `AndroidServiceConfigurator` seems like the most natural home; that's probably where I'd put these implementations of `hasExpectedServiceVersion`, `installService`, `hasExpectedPermissions`
* I think for "this state has to coordinate behavior of storing some persistent state vs connecting to a device", probably the state machine logic seems like the most appropriate place for that coordination to happen; I think that's where I'd put coordinating L31/L32 of `hasAdbPath`, for example

That set of changes would leave you with a `Deps` type that is only responsible for composing the different dependencies, rather than *also* implementing some of those dependencies; it would look more like all the other Deps types where there's just a `type` for Deps that contains properties and no logic, without having a separate interface and impl class.

PR #2907, opened 2020-06-17
chore: ensure the error explanation is always logged

waabid:
#### Description of changes

By logging in once, we ensure that the error explanation we have is always logged. Previously, due to the async nature, this was not guaranteed.

Please regard the latest playground release to validate.

#### Pull request checklist

- [ ] Ran `yarn fastpass`


deploy/Gruntfile.js
```diff
@@ -49,10 +49,13 @@ module.exports = function (grunt) {
                     e.errors.itemError[0] &&
                     e.errors.itemError[0].error_code === 'ITEM_NOT_UPDATABLE'
                 ) {
+                    const errorExplanation =
+                        'Cannot publish due to extension not being updatable. This is likely due to a previous deployment that is pending review. As such, marking this as partially successful.\n';
+
                     grunt.log.write(
-                        'Cannot publish due to extension not being updatable. This is likely due to a previous deployment that is pending review. As such, marking this as partially successful.',
+                        errorExplanation,
```
dbjorge, line 56: Are you sure this is logging both lines? The [grunt.log docs](https://gruntjs.com/api/grunt.log) don't mention supporting multiple parameters to write at all, and my reading of the [grunt.log source](https://github.com/gruntjs/grunt-legacy-log/blob/master/index.js) suggests that it would treat the first param as a format string and subsequent params as format string arguments, but it looks like this code is assuming it will concatenate them

PR #2911, opened 2020-06-17
feat(android-setup): Add code to set the app name from the device in the setup store

RobGallo:
#### Description of changes

The app name is required for the start testing screen, and it can't be set until after the permissions have been granted. It also can't be set in the onEnter function for the start testing step because the onEnter method is asynchronous, and the UI will therefore be visible possibly before the app name is set.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/electron/platform/android/setup/steps/detect-permissions.ts
+++ b/src/electron/platform/android/setup/steps/detect-permissions.ts
@@ -6,18 +6,7 @@ import { AndroidSetupStepConfig } from 'electron/platform/android/setup/android-
 export const detectPermissions: AndroidSetupStepConfig = deps => ({
     actions: {},
     onEnter: async () => {
-        deps.setApplicationName(); // init
-
         const detected = await deps.hasExpectedPermissions();
-
-        if (detected === false) {
-            deps.stepTransition('prompt-grant-permissions');
-            return;
-        }
-
-        // device is good to go; so we can get the current app name
-        const appName = await deps.getApplicationName();
-        deps.setApplicationName(appName);
-        deps.stepTransition('configuring-port-forwarding');
+        deps.stepTransition(detected ? 'configuring-port-forwarding' : 'prompt-grant-permissions');
     },
 });
--- a/src/electron/platform/android/setup/steps/detect-permissions.ts
+++ b/src/electron/platform/android/setup/steps/detect-permissions.ts
@@ -6,7 +6,18 @@ import { AndroidSetupStepConfig } from 'electron/platform/android/setup/android-
 export const detectPermissions: AndroidSetupStepConfig = deps => ({
     actions: {},
     onEnter: async () => {
+        deps.setApplicationName(); // init
+
         const detected = await deps.hasExpectedPermissions();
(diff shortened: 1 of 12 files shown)


src/electron/platform/android/setup/steps/detect-permissions.ts
```diff
@@ -6,7 +6,18 @@ import { AndroidSetupStepConfig } from 'electron/platform/android/setup/android-
 export const detectPermissions: AndroidSetupStepConfig = deps => ({
     actions: {},
     onEnter: async () => {
+        deps.setApplicationName(); // init
+
         const detected = await deps.hasExpectedPermissions();
-        deps.stepTransition(detected ? 'configuring-port-forwarding' : 'prompt-grant-permissions');
+
```
DaveTryon, line 12: The deviceFetcher actually makes an HTTP call to the service on port 62442, which means that it needs to occur _after_ the port forwarding has been handled. As such, this is in the wrong step. I'd suggest putting it into the configuring-port-forwarding step.

src/electron/platform/android/setup/steps/detect-permissions.ts
```diff
@@ -6,7 +6,18 @@ import { AndroidSetupStepConfig } from 'electron/platform/android/setup/android-
...
+        if (detected === false) {
+            deps.stepTransition('prompt-grant-permissions');
+            return;
+        }
+
+        // device is good to go; so we can get the current app name
+        const appName = await deps.getApplicationName();
```
DaveTryon, line 19: The deviceFetcher actually makes an HTTP call to the service on port 62442, which means that it needs to occur after the port forwarding has been handled. As such, this is in the wrong step. I'd suggest putting it into the configuring-port-forwarding step.

(detecting permissions occurs through ADB, not through a service call)

src/tests/unit/tests/electron/platform/android/setup/live-android-setup-deps.test.ts
```diff
@@ -428,4 +452,17 @@ describe('LiveAndroidSetupDeps', () => {
 
         verifyAllMocks();
     });
+
+    it('getApplicationName returns epty string on error', async () => {
```
DaveTryon, line 456: typo
```suggestion
    it('getApplicationName returns empty string on error', async () => {
```

src/tests/unit/tests/electron/platform/android/setup/steps/detect-permissions.test.ts
```diff
@@ -23,17 +23,15 @@ describe('Android setup step: detectPermissions', () => {
...
+        depsMock.setup(m => m.stepTransition('configuring-port-forwarding'));
 
         const step = detectPermissions(depsMock.object);
         await step.onEnter();
 
         depsMock.verifyAll();
     });
 
-    it('onEnter transitions to prompt-install-service as expected', async () => {
+    it('onEnter transitions to prompt-grant-permissions as expected', async () => {
```
DaveTryon, line 34: For symmetry with "on success", maybe "on failure"?
```suggestion
    it('onEnter transitions to prompt-grant-permissions on failure', async () => {
```

PR #2927, opened 2020-06-20
refactor(android-setup): Refactor business logic into one class

DaveTryon:
#### Description of changes
One piece of feedback that didn't get included in #2884 was that business logic was spread between the `LiveAndroidSetupDeps` class and the `AppiumServiceConfigurator` class, making it hard to follow and potentially maintain. The suggestion was to create a clearer distinction of responsibilities so that the business logic was encapsulated into a single class, if possible. This PR does the following:

- The `AndroidServiceConfigurator` class is now the class that centralizes the business logic associated with the service.
- The `AppiumAdbWrapper` class is a thin wrapper around ADB methods. It's abstracted behind the `AdbWrapper` class so that we can swap out the appium-adb layer without changing the business logic.
- The `LiveAndroidSetupDeps` class interacts with the `AndroidServiceConfigurator` class and never touches an `AdbWrapper` implementation.

Things _not_ included in this PR:

- We discussed moving the store interactions upstream from the deps, possibly into the steps. This PR doesn't try to make that change, since that ought to be discussed and coordinated with @RobGallo.
- @RobGallo  and I have discussed allowing the deps to throw during calls to `installService` and `setTcpForwarding`. I've left them non-throwing for now (that change will be simpler to manage in a separate PR)

All unit tests have been updated. Tested by running various scenarios with an android device.

#### Pull request checklist

- [x] Addresses an existing issue: (PR feedback from #2884)
- [x] Ran `yarn fastpass`


DaveTryon:
Closing/reopening to jump-start stuck checks

src/electron/platform/android/setup/live-android-service-setup-business-logic.ts
```diff
@@ -0,0 +1,119 @@
...
+// Licensed under the MIT License.
+
+import { AndroidServiceApkLocator } from 'electron/platform/android/android-service-apk-locator';
+import {
+    AndroidServiceConfigurator,
+    DeviceInfo,
+    PackageInfo,
+} from 'electron/platform/android/android-service-configurator';
+
+export interface AndroidServiceSetupBusinessLogic {
```
dbjorge, line 11: What problem is this interface solving? There's only one implementation, and it's exactly the same code to mock it for a test with or without the extra interface in between, no? Why the extra layer?

src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -31,7 +25,7 @@ export class LiveAndroidSetupDeps implements AndroidSetupDeps {
     public hasAdbPath = async (): Promise<boolean> => {
         try {
             const adbLocation = this.configStore.getState().adbLocation;
-            this.serviceConfig = await this.configFactory.getServiceConfigurator(adbLocation);
+            this.businessLogic = await this.businessLogicFactory.getBusinessLogic(adbLocation);
```
dbjorge, line 28: This is pretty confusing to read; I don't think it's intuitive that something named `getBusinessLogic()` would throw based on whether ADB is findable or not.

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -2,24 +2,19 @@
...
 import { DictionaryStringTo } from 'types/common-types';
 
 type AdbDevice = {
     udid: string;
 };
 
-const servicePackageName: string = 'com.microsoft.accessibilityinsightsforandroidservice';
-
-export const defaultAdbPortNumber: number = 62442;
 export class AppiumServiceConfigurator implements AndroidServiceConfigurator {
```
dbjorge, line 16: I think this should probably be renamed since it no longer contains any logic related to configuring our service, specifically; maybe `AdbAdapter` or `AppiumAdbAdapter`?

src/electron/platform/android/setup/live-android-service-setup-business-logic.ts
DaveTryon: Muscle memory from C# mocking libraries. Will remove the extra layer

PR #2933, opened 2020-06-22
feat(android-setup): choose host port dynamically

dbjorge:
#### Description of changes

This PR uses the `portfinder` NPM package to have our service configurator dynamically choose a free host port to bind to, rather than unilaterally choosing the same port hardcoded into the service.

The logic to use portfinder is pretty straightforward; the more complicated change was the corresponding update to `removeTcpForwarding`. I didn't want to break the existing behavior of being able to clean up old forwarded ports from past app launches, so rather than updating it to accept a number parameter, it instead uses ADB `getForwardList` to query for *all* host ports that are mapped to the hardcoded service port, and removes the forwarding entry for each of those.

This change should be transparent to an end user except that it eliminates a class of errors that could occur from port reuse during setup; no visual change.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -83,13 +90,41 @@ export class AppiumServiceConfigurator implements AndroidServiceConfigurator {
...
         this.adb.setDeviceId(deviceId);
-        await this.adb.removePortForward(defaultAdbPortNumber);
+        const applicableHostPorts = await this.getHostPortsForwardedToDevicePort(servicePortNumber);
+        for (const applicableHostPort of applicableHostPorts) {
+            await this.adb.removePortForward(applicableHostPort);
+        }
+    };
+
+    // Assumes setDeviceId was already invoked
+    private getHostPortsForwardedToDevicePort = async (devicePort: number): Promise<number[]> => {
```
DaveTryon, line 112: Do we need to worry about the case where multiple instances of the client run simultaneously? There's no theoretical reason that it wouldn't work at the moment, and we do nothing to prevent it. Cleaning up all ports redirected to the device port would impact this multiple instance scenario. As soon as any of the instances exits, all of the instances lose their port forwarding. That feels a little bit aggressive--I think I'd prefer to see us remove just the port that we added within this instance.

src/tests/unit/tests/electron/platform/android/appium-service-configurator-factory.test.ts
```diff
@@ -32,6 +38,7 @@ describe('AppiumServiceConfiguratorFactory tests', () => {
 
         adbCreatorMock.verifyAll();
```
DaveTryon, line 39: 3 mocks is usually my "threshold" for moving the verifyAll calls into a method called verifyAllMocks

src/electron/platform/android/appium-service-configurator.ts
dbjorge: That's a fair point, I hadn't considered running multiple copies of the app. Will update to have it take a hostPort param instead

src/tests/unit/tests/electron/platform/android/appium-service-configurator-factory.test.ts
dbjorge: done

src/electron/platform/android/appium-service-configurator.ts
```diff
@@ -83,13 +90,41 @@ export class AppiumServiceConfigurator implements AndroidServiceConfigurator {
...
     public setTcpForwarding = async (deviceId: string): Promise<number> => {
+        const hostPort = await this.portFinder({
+            port: servicePortNumber,
+            stopPort: servicePortNumber + 100,
+        });
+
         this.adb.setDeviceId(deviceId);
-        await this.adb.forwardPort(defaultAdbPortNumber, defaultAdbPortNumber);
-        return defaultAdbPortNumber;
+        await this.adb.forwardPort(hostPort, servicePortNumber);
```
DaveTryon, line 99: If I'm reading this correctly, it looks like we're going to assign a new port each time we execute this step. If someone were to go several times through the UX via the cancel button, that could accumulate a lot of ports. I wonder if we ought to limit ourselves to only using 1 port at a time--maybe remove a previously forwarded port (within this instance) before forwarding another one? That might also allow us to be less aggressive when removing forwarded ports

dbjorge: Agreed

PR #2945, opened 2020-06-23
feat(android-setup): add telemetry for steps in android-setup process

karanbirsingh:
#### Description of changes

This PR adds telemetry events for each step in the android-setup process. The events are triples of <prevStep, currentStep, duration of last step> and are fired upon reaching a new step.

Implementation notes:
- It looks like the electron events don't use any telemetry-data-factory. It's not clear to me whether we want to separate electron vs. extension data factories out like we have for the event names. I've created the event in the class itself like scan-controller does
- The first event intentionally sends a prevDuration of 0 and prevStep of null. This works out well because of the wait-for-start step.
- Because app-initialized is sent after rendering, it's likely some of these events will be sent before app-initialized
- we are using the performance api instead of datetimes

Sample log:

```
eventName:  DeviceSetupStep
prevStep: "null"
newStep: "wait-to-start"
prevDuration: "0"

eventName:  DeviceSetupStep 
prevStep: "wait-to-start"
newStep: "detect-adb"
prevDuration: "9.814999997615814"

eventName:  **AppInitialized** 

eventName:  DeviceSetupStep ; customProperties
prevStep: "detect-adb"
newStep: "detect-devices"
prevDuration: "1507.8150001354516"

eventName:  DeviceSetupStep ; customProperties:  
prevStep: "detect-devices"
newStep: "detect-service"
prevDuration: "2077.504999935627"

eventName:  DeviceSetupStep ; customProperties:  
prevStep: "detect-service", 
newStep: "detect-permissions", 
prevDuration: "1106.2749999109656"

eventName:  DeviceSetupStep ; customProperties:  
prevStep: "detect-permissions", 
newStep: "configuring-port-forwarding", 
prevDuration: "1390.3350001201034"

eventName:  DeviceSetupStep ; customProperties:  
prevStep: "configuring-port-forwarding", 
newStep: "prompt-connected-start-testing", 
prevDuration: "754.3299999088049"
```

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/tests/unit/tests/electron/platform/android/android-setup-telemetry-sender.test.ts
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    prevStep: 'detect-devices',
+                    newStep: 'detect-permissions',
+                    prevDuration: 15,
+                },
+            ],
+        );
+    });
+
+    it('does not send event if step has not changed', () => {
```
dbjorge, line 39: Maybe extend this test case to test "state 1, state 1, state 2" so it can verify whether the "state 1 -> state 2" event tracks duration correctly?

karanbirsingh: good idea, have updated

src/tests/unit/tests/electron/platform/android/android-setup-telemetry-sender.test.ts
```diff
@@ -0,0 +1,111 @@
...
+            androidSetupStoreMock.object,
+            telemetryEventHandlerMock.object,
+            getCurrentMsMock.object,
+        );
+
+        steps.forEach(currentStepId =>
+            androidSetupStoreMock
+                .setup(m => m.getState())
+                .returns(_ => ({ currentStepId }))
+                .verifiable(Times.exactly(steps.length)),
```
karanbirsingh, line 76: this seemed odd to me - it seems that the default (Times.once()) doesn't 'stack' like the return record-and-replay behavior does. This [codepen](https://codepen.io/karanbirsingh/pen/dyGRMJG?editors=0010) shows a repro. If the arguments in setup() are different, or the mock isn't strict, this goes away

dbjorge: This looks like the same issue as https://github.com/florinn/typemoq/issues/94, but it kind of looks like typemoq is a dead project, so it seems unlikely that it's worth following up on :(

PR #2955, opened 2020-06-24
feat:(needs-review) Add scan for needs review

jlperkins:
#### Description of changes

No UI changes, adding in scan behavior for needs review.  Proper scan triggered from details view or ad hoc panel, and incomplete results returned as a unified result, marked as 'fail' for now so they are visible in cards. 

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/ad-hoc-visualizations/needs-review/visualization.tsx
```diff
@@ -28,15 +31,25 @@ export const NeedsReviewAdHocVisualization: VisualizationConfiguration = {
...
+        provider.createRuleAnalyzerUnifiedScanForNeedsReview({
+            rules: [
+                'aria-input-field-name',
+                'color-contrast',
+                'td-headers-attr',
+                'th-has-data-cells',
+            ],
+            resultProcessor: (scanner: ScannerUtils) => scanner.getFailingInstances,
+            telemetryProcessor: (telemetryFactory: TelemetryDataFactory) =>
+                telemetryFactory.forIssuesAnalyzerScan,
```
waabid, line 43: We should probably have a function that refers to NeedsReview here instead of Issues for the telemetry.

src/injected/selector-map-helper.ts
```diff
@@ -69,6 +69,7 @@ export class SelectorMapHelper {
                 VisualizationType.Landmarks,
                 VisualizationType.TabStops,
                 VisualizationType.Color,
+                VisualizationType.NeedsReview,
```
waabid, line 72: We don't need to do this right now but I wonder if we can just use an if else instead at the top and get rid of this function. Basically have the logic:

```
if this test is an assement
    // do something for assessments
else
    // call getFilteredSelectorMap
```

src/ad-hoc-visualizations/needs-review/visualization.tsx
jlperkins: got it, added in telemetry specific to needs review

src/injected/selector-map-helper.ts
jlperkins: Made a note for future work to adjust this

Issue #2869, opened 2020-06-12
Codebase should use strict null checks


dbjorge:
Per 6/12/20, there was broad (but not unanimous) agreement among the engineering team that we would like to enable typescript's `strictNullChecks: true` config in our codebase; most participants felt that the errors that could be prevented with this particular check outweighed the engineering cost of retrofitting the code to allow it.

To enable this, we'll be following the same strategy that VS Code did; they outline the migration strategy they used in great detail in their blog post [Strict null checking Visual Studio Code](https://code.visualstudio.com/blogs/2019/05/23/strict-null). To summarize it as it applies to our codebase:

1. PR #2835 adds a new `tsconfig.strictNullChecks.json` that performs null checking on the set of already-strictNullChecks-clean files
2. PR #2835 sets up stay-clean verification in `yarn fastpass` and PR/CI builds
3. We've added a widget to our team standup dashboard to track progress (for now, it needs to be manually updated as progress is made)
4. To contribute an incremental improvement:
    1. Clone [dbjorge/aiweb-strict-null-check-migration-tools](https://github.com/dbjorge/aiweb-strict-null-check-migration-tools) and `npm install`
    2. From the web repo, run `node /path/to/aiweb-strict-null-check-migration-tools/index.js .`
    3. Pick any file(s) it suggests, add it manually to the `files` listing in `tsconfig.strictNullChecks.js`
    4. Run `yarn null:check --watch` and fix any issues it's finding
    5. (optional) use `node /path/to/aiweb-strict-null-check-migration-tools/autoAdd.js .` to try to auto-add any transitively fixed files to `tsconfig.strictNullChecks.js`
    6. Repeat from 2-5, periodically sending PRs with incremental work

PR #2957, opened 2020-06-24
chore(strict-null): add strict null check migration tools to repo

dbjorge:
#### Description of changes

This integrates the strict null checking tools that I migrated from VS Code's as an experiment in https://github.com/dbjorge/aiweb-strict-null-check-migration-tools into the `/tools/` folder of our repo, to make it easier for others to use them.

As part of the migration, I eliminated the separate `package.json` machinery for the tools and had them just re-use the same `node_modules` as the main project (in particular, re-using the correct typescript version). I also added entry point run-scripts for the tools to our root-level package.json, so the source for the tools should be an implementation detail for most purposes of actually running null check migration work; usage is just `yarn null:find` and `yarn null:autoadd` from the root, to match the previously-added `yarn null:check`.

I've updated #2869 with instructions for how to use these scripts to contribute to the strict null check migration work.

#### Pull request checklist

- [x] Addresses an existing issue: tools to help address #2869 
- [x] Ran `yarn fastpass`


tools/strict-null-checks/auto-add.js
```diff
@@ -0,0 +1,62 @@
...
+                    fs.writeFileSync(tsconfigPath, JSON.stringify(originalConfig, null, '\t'));
+                }
+
+                child.stdout.removeListener('data', listener);
+                resolve();
+            }
+        };
+        child.stdout.on('data', listener);
+
+        fs.writeFileSync(tsconfigPath, JSON.stringify(newConfig, null, '    '));
```
karanbirsingh, line 60: how does this work in concert with line 48? It seems like only one would be needed, but I might be missing something

tools/strict-null-checks/auto-add.js
```diff
@@ -0,0 +1,62 @@
...
+
+        // Config on accept
+        const newConfig = Object.assign({}, originalConfig);
+        newConfig.files = Array.from(
+            new Set(originalConfig.files.concat('./' + relativeFilePath).sort()),
+        );
+
+        const listener = data => {
+            const textOut = data.toString();
+            // console.log('  ' + textOut);
```
karanbirsingh, line 42: can remove(?)

tools/strict-null-checks/auto-add.js
dbjorge: Agreed, 48 is unnecessary

tools/strict-null-checks/auto-add.js
dbjorge: Yep

PR #2966, opened 2020-06-25
feat(android-setup): Remove dynamic port forwarding on exit

DaveTryon:
#### Description of changes
The current code dynamically maps a port (via ADB) to allow the client to communicate with the service. It makes no attempt to release that port when the client exits, so the port is left in an unfortunate state. This change allows us to hook the shutdown process in a way that allows us to remove the port forwarding under all "normal" scenarios. This code will not be executed if the process terminates abnormally.

  These are the key changes:
1. A new class `AndroidPortCleaner`, remembers ports forwardings that are added and removed. This class also listens for an event to know that the app is closing, and uses the current `ServiceConfigurator` to clean up any "orphaned" ports.
2. The `PortCleaningAndroidServiceConfigurator` is a decorator that wraps the existing `AndroidServiceConfigurator` object. It chains everything through, and keeps the `AndroidPortCleaner` class updated as things change. It has a corresponding factory that wraps the previous factory.
3. The main process now watches for the `close` event and notifies the renderer process of the change. This gets picked up by the `AndroidPortCleaner`, which removes the ports. After the ports are removed, the closeWindow action is triggered again, and the second attempt to close actually shuts down the process. This adds 1 extra round-trip between the main process and the renderer, but it's not human-perceivable. It also ensures that we can catch all app closing scenarios, including those that bypass the renderer (Alt+F4 and closing with the task bar probably being the most common).

There's one unusual thing: when removing the orphaned ports, we call into appium-adb, which calls into ADB. All of these calls are async/await so the timing should work, but it didn't work in practice. Appium-adb logged that it was calling ADB, but the the port forwarding didn't actually get removed. To work around this, I added another call into ADB--this time we call ADB to get to list the currently mapped ports (which will include from all processes, not just ours). We don't actually need this data, but the fact that we're making the second call allows the first ADB process to run to completion. 

#### Pull request checklist

- [x] Addresses an existing issue: [1732837](https://mseng.visualstudio.com/1ES/_workitems/edit/1732837)
- [x] Ran `yarn fastpass`


src/electron/platform/android/setup/android-service-configurator.ts
```diff
@@ -79,17 +83,35 @@ export class AndroidServiceConfigurator {
             stopPort: servicePortNumber + 100,
         });
 
-        return await this.adbWrapperMock.setTcpForwarding(
+        const assignedPort: number = await this.adbWrapperMock.setTcpForwarding(
             this.selectedDeviceId,
             hostPort,
             servicePortNumber,
         );
+
```
DaveTryon, line 91: This approach feels wrong. I'm going to refactor this to use a decorator

src/tests/unit/tests/electron/platform/android/setup/android-port-cleaner.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        const port: number = 7;
+        testSubject.setServiceConfig(serviceConfigMock.object);
+        testSubject.addPort(port);
+        testSubject.removePort(port);
+        testSubject.removePort(port);
+
+        ipcRendererShim.fromBrowserWindowClose.invoke(null);
+    });
+
+    // it('orphaned ports get removed on action invocation', () => {
```
karanbirsingh, line 78: will this be removed or recommented?

src/tests/unit/tests/electron/platform/android/setup/port-cleaning-service-configurator-factory.test.ts
```diff
@@ -0,0 +1,60 @@
...
+            .verifiable(Times.once());
+        portCleanerMock
+            .setup(m => m.setServiceConfig(It.isAny()))
+            // .callback(config => (attachedServiceConfig = config as ServiceConfigurator))
+            .verifiable(Times.once());
+
+        // const finalServiceConfig: ServiceConfigurator =
+        await testSubject.getServiceConfigurator(expectedAdbLocation);
+
+        // TODO - DHT: Why do these generate function.asymmetricMatch?
```
karanbirsingh, line 54: flagging TODO

src/electron/platform/android/appium-adb-wrapper.ts
```diff
@@ -78,4 +78,9 @@ export class AppiumAdbWrapper implements AdbWrapper {
         this.adb.setDeviceId(deviceId);
         await this.adb.removePortForward(localPort);
     };
+
+    public listForwardedPorts = async (deviceId: string): Promise<string[]> => {
```
karanbirsingh, line 82: is the device id needed here?

src/electron/platform/android/setup/android-port-cleaner.ts
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under the MIT License.
+
+import { Logger } from 'common/logging/logger';
+import { IpcRendererShim } from 'electron/ipc/ipc-renderer-shim';
+import { ServiceConfigurator } from 'electron/platform/android/setup/android-service-configurator';
+import { DictionaryStringTo } from 'types/common-types';
+
+export class AndroidPortCleaner {
+    private serviceConfig: ServiceConfigurator;
+    private readonly portMap: DictionaryStringTo<number> = {};
```
karanbirsingh, line 11: would a Set be better here if we only need inclusion?

PR #2967, opened 2020-06-25
fix(android-setup): fix aria-label on adb path text field

dbjorge:
#### Description of changes

Fluent UI has its own `ariaLabel` prop for `TextField`s, but not its own `aria-describedby`, so the casing has to be different for applying them (oops). React does provide an `aria-label` (which is why this wasn't a compile error), but Fluent ignores it in favor of its own property.

NVDA now reads the control (when empty) as "Enter a path  edit  Select the folder containing adb. We'll use it to connect to your device.  Enter a path...  blank" - that is, it repeats the label and placeholder text. Would like confirmation from @ferBonnin / @RobGallo on whether this feels reasonable or if it's too redundant

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/views/device-connect-view/components/android-setup/folder-picker.tsx
+++ b/src/electron/views/device-connect-view/components/android-setup/folder-picker.tsx
@@ -42,7 +42,7 @@ export const FolderPicker = NamedFC<FolderPickerProps>(
                 </p>
                 <div className={styles.pickerControl}>
                     <TextField
-                        placeholder="Enter a path..."
+                        placeholder="adb path"
                         ariaLabel="Enter a path"
                         aria-describedby={instructionsId}
                         className={styles.textField}
--- a/src/electron/views/device-connect-view/components/android-setup/folder-picker.tsx
+++ b/src/electron/views/device-connect-view/components/android-setup/folder-picker.tsx
@@ -43,7 +43,7 @@ export const FolderPicker = NamedFC<FolderPickerProps>(
                 <div className={styles.pickerControl}>
                     <TextField
                         placeholder="Enter a path..."
-                        aria-label="Enter a path"
+                        ariaLabel="Enter a path"
                         aria-describedby={instructionsId}
                         className={styles.textField}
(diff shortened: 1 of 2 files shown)


ferBonnin:
@dbjorge the NVDA text definitely feels redundant. 
Enter a path edit Select the folder containing adb. We'll use it to connect to your device. Enter a path... blank

cc @RobGallo do you agree?

Are there any options to at least not read "enter a path" twice?

dbjorge:
@ferBonnin We don't have an option to add `aria-hidden` on the placeholder text, since it's not a custom control, just a `placeholder` attribute on the `input` control; we don't control its rendering. Removing the `aria-label` probably also isn't an option (even if we ignored the programmatic labelling requirement, the placeholder text only helps if the box is empty). Other cases in the app where we have a text input that uses placeholder text (eg, the "Comment" box when adding a failure instance in a manual assessment step) also have the same double-reading of the label + placeholder, though they usually have less verbose descriptions.

I think the only realistic option here would be to either remove the placeholder text or update the text/label to be less redundant (eg, maybe use an `aria-label` of "adb path" rather than "enter a path")

ferBonnin:
> I think the only realistic option here would be to either remove the placeholder text or update the text/label to be less redundant (eg, maybe use an `aria-label` of "adb path" rather than "enter a path")
if we remove the placeholder text, how would it read? changing the placeholder text to _adb path_ also sounds like a good idea.

dbjorge:
With the `placeholder` attribute removed and the label applied, it reads "Enter a path  edit  Select the folder containing adb. We'll use it to connect to your device.  blank". Note that removing the `placeholder` attribute also visually removes the "Enter a path..." placeholder text, ie, when the text field is empty it looks like this:

image

dbjorge:
Updated version per discussion with no path entered reads like this in NVDA:

> Enter a path  edit  Select the folder containing adb. We'll use it to connect to your device.  adb path  blank

...and looks like this:

screenshot of updated UI

PR #2991, opened 2020-06-29
feat(android-setup): placeholder e2e test with adb feature flag on

karanbirsingh:
#### Description of changes

This is a placeholder e2e test that has the feature flag on for adb. One approach we can follow is to add new e2e tests here before we turn the feature flag off entirely. The test assumes that the app will get to prompt-locate-adb or prompt-connect-to-device, which are reasonable assumptions for a dev machine without an emulator running. This assumption/test can be removed once we have more of the adb mock in.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/tests/electron/common/view-controllers/app-controller.ts
+++ b/src/tests/electron/common/view-controllers/app-controller.ts
@@ -74,6 +74,14 @@ export class AppController {
         );
     }
 
+    public async setFeatureFlag(flag: string, enabled: boolean): Promise<void> {
+        await this.waitForfeatureFlagInitializer();
+        const action = enabled ? 'enable' : 'disable';
+        await this.app.webContents.executeJavaScript(
+            `window.featureFlagsController.${action}Feature('${flag}')`,
+        );
+    }
+
     private async waitForUserConfigurationInitializer(): Promise<void> {
         await this.client.waitUntil(
             async () => {
@@ -85,6 +93,20 @@ export class AppController {
             },
             DEFAULT_WAIT_FOR_ELEMENT_TO_BE_VISIBLE_TIMEOUT_MS,
             'was expecting window.insightsUserConfiguration to be defined',
+        );
+    }
+
+    private async waitForfeatureFlagInitializer(): Promise<void> {
+        await this.client.waitUntil(
+            async () => {
+                const executeOutput = await this.client.executeAsync(done => {
+                    done((window as any).featureFlagsController != null);
+                });
+
+                return executeOutput.status === 0 && executeOutput.value === true;
+            },
+            DEFAULT_WAIT_FOR_ELEMENT_TO_BE_VISIBLE_TIMEOUT_MS,
+            'was expecting window.featureFlagsController to be defined',
         );
     }
 }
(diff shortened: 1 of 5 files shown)


src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -87,4 +95,18 @@ export class AppController {
             'was expecting window.insightsUserConfiguration to be defined',
         );
     }
+
+    private async waitForfeatureFlagInitializer(): Promise<void> {
+        await this.client.waitUntil(
+            async () => {
+                const executeOutput = await this.client.executeAsync(done => {
+                    done((window as any).featureFlagsController != null);
```
karanbirsingh, line 103: I tried to share this method between both window objects (featureFlagsController, insightsUserConfiguraiton) e.g. `(window as any)[property]` but the inner scope of executeAsync didn't have access to the passed in variable. Suggestions appreciated

src/tests/electron/tests/automatic-android-setup.test.ts
```diff
@@ -0,0 +1,38 @@
...
+import { createApplication } from 'tests/electron/common/create-application';
+import { AppController } from 'tests/electron/common/view-controllers/app-controller';
+import { DeviceConnectionDialogController } from 'tests/electron/common/view-controllers/device-connection-dialog-controller';
+describe('AutomaticAndroidSetup', () => {
+    let app: AppController;
+    let deviceConnectionController: DeviceConnectionDialogController;
+
+    beforeEach(async () => {
+        app = await createApplication({ suppressFirstTimeDialog: true });
+        await app.setFeatureFlag(UnifiedFeatureFlags.adbSetupView, true);
```
dbjorge, line 15: Would be good to add setting this `false` to the existing device connect view e2es, to prevent them breaking when we flip the flag's default

src/tests/electron/tests/automatic-android-setup.test.ts
```diff
@@ -0,0 +1,38 @@
...
+        expect(await app.getTitle()).toBe(
+            'Accessibility Insights for Android - Connect to your Android device',
+        );
+    });
+
+    it('first dialog has more info link', async () => {
+        const selector = getAutomationIdSelector(moreInfoLinkAutomationId);
+        await deviceConnectionController.client.waitForExist(selector);
+        const text = await deviceConnectionController.client.getText(selector);
+        expect(text.startsWith('How do I')).toBeTruthy();
```
dbjorge, line 36: We usually prefer `toBe(true)` to `toBeTruthy()`, unless there's some compelling reason to use the less specific assertion

src/tests/electron/common/view-controllers/app-controller.ts
dbjorge: You need to pass along any variables you want via parameters to the inner function and arguments to `executeAsync`, like this:

```ts
await client.executeAsync((innerFoo, innerBar, done) => {
    done(functionUsing(innerFoo, innerBar));
}, foo, bar);
```

See [webdriverio's executeAsync docs](https://webdriver.io/docs/api/browser/executeAsync.html)

karanbirsingh: Should have thought to look there, the docs are super clear. Thanks!

PR #2996, opened 2020-06-30
Chore: Add null checks

RobGallo:
#### Description of changes

__browser-message-broadcaster-factory.ts__

Since it does seem valid that tab.id may be null, simply add a null check guard around the code which uses it.

__interpreter.ts __

The change to message.ts fixes a possible null assignment in interpreter.ts on line 18.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/extension-details-view-controller.ts
+++ b/src/background/extension-details-view-controller.ts
@@ -27,7 +27,9 @@ export class ExtensionDetailsViewController implements DetailsViewController {
 
         const tab = await this.browserAdapter.createTabInNewWindow(this.getDetailsUrl(targetTabId));
 
-        this.tabIdToDetailsViewMap[targetTabId] = tab.id;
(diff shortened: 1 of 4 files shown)


src/background/extension-details-view-controller.ts
```diff
@@ -27,7 +27,9 @@ export class ExtensionDetailsViewController implements DetailsViewController {
 
         const tab = await this.browserAdapter.createTabInNewWindow(this.getDetailsUrl(targetTabId));
 
-        this.tabIdToDetailsViewMap[targetTabId] = tab.id;
+        if (tab?.id != null) {
```
dbjorge, line 30: I think it would make more sense for this case to throw an error explicitly; if we aren't able to assign the tab-to-view map entry, it means that the controller is in a bad state and isn't safe to use in future calls.

PR #3000, opened 2020-06-30
feat(android-setup): use mock-adb in e2e tests

karanbirsingh:
## Description of changes

This PR uses the mock-adb executable to use the new adb-setup-view flow in automated checks. 
- merged https://github.com/microsoft/accessibility-insights-web/pull/3005 to unblock this PR
- Dan merged #3008 (mock-adb to be updated by Dan so that [this line](https://github.com/microsoft/accessibility-insights-web/blob/2bcc09a0080678726a15201689058e46b6b65963/src/electron/platform/android/setup/steps/configuring-port-forwarding.ts#L20) doesn't execute until the test server is fully set up; once that fix is in, the tests on this PR should pass and it can be reviewed)

See penultimate commit (before removing debugging info) if extra logs are needed when testing remote.

#### Pull request checklist

- [ ] Ran `yarn fastpass`


DaveTryon:
I'm not sure why the testServer isn't exiting as expected, but something seems off in the concept of having a hardcoded port in the config file. The whole reason we added portFinder was to allow us to be dynamic, and if we're in a cloud environment, there's no guarantee that port 62442 is open (for example, a test from a different PR might be running concurrently). I think we may need to consider some way to support multiple ports. The "brute force" way is to map every number the app will use (62442 to 62541) as a separate entry in the config file. The much nicer way would be to use some sort of pattern matching

src/tests/miscellaneous/mock-adb/app/bin.js
```diff
@@ -61,7 +61,7 @@ function main() {
 
     result.input = process.argv;
     result.inputCommand = inputCommand;
-    fs.writeFileSync(outputFile, JSON.stringify(result, null, '    '));
+    fs.writeFileSync(outputFile, JSON.stringify(result, null, '    '), { flag: 'a' });
```
karanbirsingh, line 64: not necessary but was helpful for me when debugging

PR #3005, opened 2020-07-01
feat(android-setup): wait for port forwarding removal before application exits

karanbirsingh:
#### Description of changes

This fixes the issue outlined in #3000 - before this PR, running `npx cross-env ANDROID_HOME=drop/mock-adb electron drop/electron/unified-dev/product/bundle/main.bundle.js` twice will lead to port-forwarding-failure the second time (See #3000 for debugging commands).

After this PR the application can be run multiple times and should clean up the port each time robustly.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/electron/ipc/ipc-event-action.ts
+++ b/src/electron/ipc/ipc-event-action.ts
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export class IpcEventAction {
+    private listeners: (() => Promise<void>)[] = [];
+
+    public async invokeAsync(): Promise<void> {
+        for (const listener of this.listeners) {
+            await listener();
+        }
+    }
+
+    public addAsyncListener(listener: () => Promise<void>): void {
+        this.listeners.push(listener);
+    }
+}
--- a/src/electron/platform/android/setup/android-port-cleaner.ts
+++ b/src/electron/platform/android/setup/android-port-cleaner.ts
@@ -15,7 +15,7 @@ export class AndroidPortCleaner {
     ) {}
 
     public initialize(): void {
-        this.ipcRendererShim.fromBrowserWindowClose.addListener(this.removeRemainingPorts);
+        this.ipcRendererShim.fromBrowserWindowClose.addAsyncListener(this.removeRemainingPorts);
     }
 
     public setServiceConfig = (serviceConfig: ServiceConfigurator): void => {
@@ -37,7 +37,6 @@ export class AndroidPortCleaner {
                 if (p) {
                     try {
                         await this.serviceConfig.removeTcpForwarding(p);
-                        await this.serviceConfig.listForwardedPorts(); // Allows ADB to complete
                     } catch (error) {
                         this.logger.log(error);
                     }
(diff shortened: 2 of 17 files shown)


src/electron/platform/android/setup/android-port-cleaner.ts
```diff
@@ -37,7 +37,6 @@ export class AndroidPortCleaner {
                 if (p) {
                     try {
                         await this.serviceConfig.removeTcpForwarding(p);
-                        await this.serviceConfig.listForwardedPorts(); // Allows ADB to complete
```
JGibson2019, line 40: Why did you remove the call to list the forwarded ports here?

src/electron/ipc/ipc-event-action.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export class IpcEventAction {
```
DaveTryon, line 3: This should have unit tests, but the rest looks good to me

src/electron/ipc/ipc-event-action.ts
```diff
@@ -0,0 +1,15 @@
...
+    private listeners: (() => Promise<void>)[] = [];
+
+    public async invokeAsync(): Promise<void> {
```
DaveTryon, line 6: This may need to be an arrow function to get the correct `this`. Unit tests will make that clear

src/electron/platform/android/setup/android-port-cleaner.ts
karanbirsingh: good question - the comment & prior PR show it was intended to try helping ADB commands fully complete before the app closed, but it turns out this behavior was due to our missing async listeners & not because of appium itself; we don't need the list for the production flow

src/electron/ipc/ipc-event-action.ts
karanbirsingh: I think it is OK, I've compared it to the methods I see in [Action](https://github.com/microsoft/accessibility-insights-web/blob/master/src/common/flux/action.ts), can you let me know if I'm missing something in particular?

src/electron/ipc/ipc-event-action.ts
dbjorge, line 3: I think this approach seems reasonable, but I think it would make more sense to call this an `AsyncAction`; I don't think what it's actually doing is really related to `IpcEvent`s except insofar as that happens to be how it's consumed

src/electron/ipc/ipc-event-action.ts
DaveTryon: If unit tests are happy and it enables the E2E tests, I'm good :)

src/tests/unit/tests/electron/ipc/async-action.test.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { AsyncAction } from 'electron/ipc/async-action';
+import { Mock, Times } from 'typemoq';
+
+describe('IpcEventAction', () => {
```
karanbirsingh, line 6:
```suggestion
describe('AsyncAction', () => {
```

PR #3010, opened 2020-07-02
Chore: Add null checks

RobGallo:
#### Description of changes

These are all pretty straight forward.

### DIFF
--- a/src/common/components/body-class-modifier.tsx
+++ b/src/common/components/body-class-modifier.tsx
@@ -17,7 +17,7 @@ export type BodyClassModifierProps = {
 // body classes office fabric maintains.
 export class BodyClassModifier extends React.Component<BodyClassModifierProps> {
     public render(): JSX.Element {
-        return null;
+        return <></>;
     }
 
     public componentDidMount(): void {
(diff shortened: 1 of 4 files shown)


src/common/components/body-class-modifier.tsx
```diff
@@ -17,7 +17,7 @@ export type BodyClassModifierProps = {
 // body classes office fabric maintains.
 export class BodyClassModifier extends React.Component<BodyClassModifierProps> {
     public render(): JSX.Element {
-        return null;
+        return <></>;
```
RobGallo, line 20: If there is a constant somewhere for returning an empty React node/fragment, please let me know.

dbjorge: `null` is canonical; I think the better fix here would actually be to expand the return type domain (if you look at the `React.Component` typings, `render` is actually allowed to return a `ReactNode`, which `null` is a valid option for; the compiler allows us to use `JSX.Element` without complaint because it is a subtype of `ReactNode`, but so is `JSX.Element | null`)

tsconfig.strictNullChecks.json
```diff
@@ -7,6 +7,10 @@
         "./src/background/browser-message-broadcaster-factory.ts",
         "./src/background/extension-details-view-controller.ts",
         "./src/background/interpreter.ts",
+
```
dbjorge, line 10: nit:; extra blank line here intended?

PR #3027, opened 2020-07-06
test: sample multiple-devices android setup e2e test

karanbirsingh:
#### Description of changes

This PR includes an initial implementation of the 'multiple-devices' dialog e2e test in Android setup. The most open question is how these e2e tests should be organized. In this PR several operations that could be validated in the same individual test are separated for readability; however, this uses more time because the application restarts between each test. Alternatively, a single test could contain more of the flow to save time, but debugging would be more difficult.

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/tests/electron/tests/android-setup-multiple-devices.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        expect(await dialog.isEnabled(getAutomationIdSelector(rightFooterButtonAutomationId))).toBe(
+            true,
+        );
+        await dialog.client.click(getAutomationIdSelector(rightFooterButtonAutomationId));
+        await dialog.waitForSelector(getAutomationIdSelector('detect-service-content'));
+    });
+
+    it('selecting rescan goes to detect-devices', async () => {
+        await dialog.client.click(getAutomationIdSelector('rescan'));
+        await dialog.waitForSelector(getAutomationIdSelector('detect-devices-content'));
```
DaveTryon, line 53: Since the detect-devices step is very short, is there any chance that we'll go through it without triggering the wait condition?

src/tests/electron/tests/android-setup-multiple-devices.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        expect(await dialog.isEnabled(getAutomationIdSelector(closeId))).toBe(true);
+        expect(await dialog.isEnabled(getAutomationIdSelector(nextId))).toBe(false);
+        expect(await dialog.isEnabled(getAutomationIdSelector('rescan'))).toBe(true);
+        const devices = await dialog.client.$$(
+            getAutomationIdSelector(deviceDescriptionAutomationId),
+        );
+        expect(devices.length).toBe(3);
+    });
+
+    it('choosing a device changes next button state, selecting next goes to detect-service', async () => {
```
DaveTryon, line 42: Should we consider a test that ensures that the selected device is properly chosen in  future steps? Possibly mock a failure while installing the service then check the device? Just a thought...

src/tests/electron/tests/android-setup-multiple-devices.test.ts
karanbirsingh: We can modify the waitForInterval property if we need - https://webdriver.io/docs/options.html#waitforinterval. I was wondering about this myself and two runs of the unified-e2e tests didn't show this issue. My instinct is to leave it as-is and change later if needed.

src/tests/electron/tests/android-setup-multiple-devices.test.ts
karanbirsingh: I think it's an interesting idea, I wonder if the mock-adb command would fail if the same device weren't requested later

DaveTryon: `workingDeviceComands`, which builds the config file, enumerates through all 3 devices to ensure that they would all work. That being said, if the test were to start with `multiple-devices`, then overlay it with responses to cause mock-adb to throw for the devices that aren't being selected, then we could distinguish between the 2 cases. If we end up in detect-permissions, then it used the correct device. If we end up in prompt-install-service, then we know that the correct device was not used.

src/tests/electron/tests/android-setup-multiple-devices.test.ts
DaveTryon: Makes sense.

Issue #2988, opened 2020-06-29
Unified: new setup process UI


ferBonnin:
## Describe the bug

A few UI changes from the new setup dialogs:
1.	On the “choose which device to use” screen, the list of connected devices appears to be indented quite a bit. It would be nicer if the device list was flush-left like the other places where we show a device.

![Choose which device to use screen](https://user-images.githubusercontent.com/9062104/86054244-a8a64980-ba0e-11ea-930d-5bbb06a9b419.png)

1.	On the “choose which device to use” screen and the “connected and ready to go!” screen, there appears to be a bit more vertical space between the title element and the content than on the other screens.
**more spacing**
![choose which device and connected screens](https://user-images.githubusercontent.com/9062104/86054396-f0c56c00-ba0e-11ea-9506-36c4a2fa93ae.png)
**ideal**
![port forwarding and let us install screens](https://user-images.githubusercontent.com/9062104/86054442-02a70f00-ba0f-11ea-802d-81e91ddb37e4.png)

**Fix**
•	I’m not sure why there’s extra space on the “choose which device to use” screen. It should match the “Let us install the latest version” picture, since they have the same combination of {title without help link} + {content text}. 
•	For the “connected and ready to go” screen, there’s probably some extra spacing that’s being introduced because of the hit area of the device tile. I’ve seen different approaches to fixing this. If it’s only ever going to be a one-off case, we can just create an ad-hoc spacing fix. If there are other cases where the devices will show up without text between them and the title, we could create some logic around it so that whenever this combination appears, it’s properly spaced.
•	In either case, the ideal state is that, while in the rest state, the spacing appears visually to be the same as the comparison examples.

## Context (please complete the following information)

- OS Name & Version:  Windows 10
- AI-Web Version & Environment:  Unified Canary

## Are you willing to submit a PR?

No

## Did you search for similar existing issues?

Yes

## Additional context

Found by Nate

AhmedAbdoOrtiga:
@ferBonnin The android setup feature is still open, do you think this issue should be addressed as a part of it so it'd be included in the release validation?
@DaveTryon what do you think?

ferBonnin:
@AhmedAbdoOrtiga no, we talked about this and we will take this outside of the feature.

dbjorge:
@ferBonnin The extra indent on the choose your device screen is because that empty space is where the checkbox renders when the device is selected. What do you want to do about the checkbox?

ferBonnin:
[...]

PR #3067, opened 2020-07-10
fix(android-setup): Address UI feedback from Nate

DaveTryon:
#### Description of changes

Nate pointed out some minor variations between the figma and the implementation of the android setup:
1. The intent of the "choose device" step was to auto-select the first item in the list when we enter the dialog. Actual implementation requires the user to choose a device, which makes the initial list of devices see to be offset far more than is expected.
2. Vertical spacing in the "choose device step looks slightly larger than other steps
3. Vertical spacing in the "start testing" step looks slightly larger than other steps

The changes are:
1. In the "start testing step", reduce the vertical spacing between the header and the body by 2 logical pixels.
2. In the choose-device step, the vertical space above the device count is set to 24px by the `margin-top` property of the `layoutContent` style. I tried unsuccessfully to override this style, and ultimately changed the paragraph to a div and explicitly set both the top and bottom margins via a new style. Updated the snapshot.
3. Auto-select the first item in the "choose device" step. Note that this _always_ picks the first item, even if you're looping back to this display after starting with a different device. Updated the tests and the snapshot.

I'd normally have 3 PRs for this, but since 3 issues were flagged in a single issue, there's a single PR. Also, there were several "false starts" in the intermediate commits that should be ignored.

Screenshot of old "start testing" step (on left) and new "start testing" step (on right). The new version decreases the spacing above the device by 2 logical pixels (although you have to look very closely to notice it):
image

Screenshot of old "choose-device" step on initial state (on left) and new "choose-device" step on initial state (on right). The new version selects the first item in list, enables the Next button automatically, and decreases the spacing above "2 Android devices or emulators connected" by 2 logical pixels (although you have to look very closely to notice it):
image

#### Pull request checklist

- [x] Addresses an existing issue: #2988 
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/views/device-connect-view/components/android-setup/prompt-choose-device-step.tsx
+++ b/src/electron/views/device-connect-view/components/android-setup/prompt-choose-device-step.tsx
@@ -47,7 +47,9 @@ export class PromptChooseDeviceStep extends React.Component<
             headerText: 'Choose which device to use',
             children: (
                 <>
-                    <p>{devices.length} Android devices or emulators connected</p>
+                    <div className={styles.deviceCount}>
+                        {items.length} Android devices or emulators connected
+                    </div>
                     <DefaultButton
                         data-automation-id={'rescan'}
                         text="Rescan"
--- a/src/electron/views/device-connect-view/components/android-setup/prompt-choose-device-step.tsx
+++ b/src/electron/views/device-connect-view/components/android-setup/prompt-choose-device-step.tsx
@@ -27,7 +27,9 @@ export class PromptChooseDeviceStep extends React.Component<
     private selection: ISelection;
     constructor(props) {
         super(props);
-        this.state = { selectedDevice: null };
+
+        const listItems = this.getListItems();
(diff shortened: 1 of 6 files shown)


src/electron/views/device-connect-view/components/android-setup/prompt-choose-device-step.tsx
```diff
@@ -99,4 +104,19 @@ export class PromptChooseDeviceStep extends React.Component<
...
+
+    private getListItems(): any {
+        const devices: DeviceInfo[] = this.props.androidSetupStoreData.availableDevices;
+        const items = devices.map(m => ({ metadata: m }));
+        return items;
+    }
+
+    private selectItemZeroInList(items: any): void {
+        if (items.length > 0) {
+            this.selection.setChangeEvents(false, true);
```
DaveTryon, line 116: The setChangeEvents lines here are to prevent the selection's change handler from getting invoked during the ctor. I originally omitted these lines, but the got errors because the component hadn't yet been mounted. I'll see if I can refactor this a bit

PR #3080, opened 2020-07-14
feat(android-setup-e2e): Enable mock-adb customization, add 2 tests

DaveTryon:
#### Description of changes
We have good coverage of the "happy path", but need to be able to force scenarios that take us off that path. This PR includes the following:
- Add 4 new methods that allow tests to specify the deviation from the happy path. These methods are     `simulateNoDevices`, `simulateServiceNotInstalled`, `simulateServiceLacksPermissions`, and `simulatePortForwardingError`. The methods can be applied individually or together to any of the existing `mockAdbConfigs` objects.
- As a proof of concept, use the new methods to create the E2E tests for the `prompt-connect-to-device` and `prompt-install-service` dialogs.
- Populate the steps needed to install the service. These steps were previously omitted because the happy path detects that the service exists, so there is no need to install it. Some of the changes to add the install steps are very brittle and will break the next time we take a new drop of the service. To unblock parallel test development, I'm proceeding with the PR with a work item to allow mock-adb to use regular expressions so that we can remove this brittleness. This PR will be part of this feature, so it will be fixed before any work is scheduled to occur on the service.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/electron/tests/android-setup-prompt-connect-to-device.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { AppController } from 'tests/electron/common/view-controllers/app-controller';
+import {
+    commonAdbConfigs,
+    setupMockAdb,
+    simulateNoDevices,
+} from '../../miscellaneous/mock-adb/setup-mock-adb';
+
+describe('Android setup - prompt-connect-to-device ', () => {
+    const defaultDeviceConfig = commonAdbConfigs['multiple-devices'];
+    const detectButtonId = 'detect-device';
```
karanbirsingh, line 20: can we import `detectDeviceAutomationId` from `prompt-connect-to-device-step.tsx` instead of redefining it here?

src/tests/miscellaneous/mock-adb/common-adb-configs.js
```diff
@@ -23,24 +66,25 @@ function workingDeviceCommands(deviceIds, port) {
     }
 
     for (const id of deviceIds) {
-        output[`-s ${id} devices`] = cloneDeep(output.devices);
+        output[`-s ${id} ` + devicesCommandMatch] = cloneDeep(output.devices);
```
karanbirsingh, line 69: we can likely use interpolation (like the id variable) for the Match variables too; this might be nice so we're not using both ways (interpolation and concatenation) to build the string

something like `-s ${id} ${devicesCommandMatch}`

src/tests/miscellaneous/mock-adb/setup-mock-adb.d.ts
```diff
@@ -22,3 +22,7 @@ export async function setupMockAdb(config: MockAdbConfig): Promise<void>;
 
 export type CommonAdbConfigName = 'single-device';
```
karanbirsingh, line 23: we should update this too in this or a future PR, looks like we are missing some

src/tests/electron/tests/android-setup-prompt-install-service.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { AppController } from 'tests/electron/common/view-controllers/app-controller';
+import {
+    commonAdbConfigs,
+    setupMockAdb,
+    simulateServiceNotInstalled,
+} from '../../miscellaneous/mock-adb/setup-mock-adb';
+
+describe('Android setup - prompt-install-service ', () => {
+    const defaultDeviceConfig = commonAdbConfigs['single-device'];
+    const installButtonId = 'install';
```
karanbirsingh, line 20: can we import `installAutomationId` from `prompt-install-service-step.tsx` here?

src/tests/electron/tests/android-setup-prompt-install-service.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        const [closeId, nextId] = [leftFooterButtonAutomationId, rightFooterButtonAutomationId];
+        expect(await dialog.isEnabled(getAutomationIdSelector(closeId))).toBe(true);
+        expect(await dialog.isEnabled(getAutomationIdSelector(nextId))).toBe(false);
+        expect(await dialog.isEnabled(getAutomationIdSelector(installButtonId))).toBe(true);
+    });
+
+    it('install button triggers installation', async () => {
+        await setupMockAdb(defaultDeviceConfig);
+        await dialog.client.click(getAutomationIdSelector(installButtonId));
+        await dialog.waitForDialogVisible('prompt-grant-permissions');
```
karanbirsingh, line 46: Not an issue, just leaving a comment that I was initially confused that this works (checking for prompt-grant-permissions even on the happy path), and then I saw the code always proceeds to this step. I considered suggesting going to detect-permissions first in production but it doesn't make a meaningful difference because of the short time window.

src/tests/miscellaneous/mock-adb/common-adb-configs.js
```diff
@@ -9,6 +9,49 @@ const successfulTestServerContentPath = path.join(
...
+        stdout: '10',
+    };
+    output[`-s ${id} help`] = {
+        stdout: '--streaming: force streaming APK directly into Package Manager',
+    };
+    output[`-s ${id} features`] = {
+        stdout:
+            'abb_exec\nfixed_push_symlink_timestamp\nabb\nstat_v2\napex\nshell_v2\nfixed_push_mkdir\ncmd',
+    };
+    output[`-s ${id} shell ls -t -1 /data/local/tmp/appium_cache 2>&1 || echo _ERROR_`] = {
```
karanbirsingh, line 33: I didn't notice this at all when I ran locally. Is it because appium-adb didn't log it as dbug ADB, or are there commands that differ from emulator to emulator?

src/tests/miscellaneous/mock-adb/common-adb-configs.js
DaveTryon: Fixed

src/tests/miscellaneous/mock-adb/setup-mock-adb.d.ts
DaveTryon: Fixed

src/tests/electron/tests/android-setup-prompt-install-service.test.ts
DaveTryon: Fixed

src/tests/electron/tests/android-setup-prompt-install-service.test.ts
DaveTryon: Yeah, I noticed that too. In the customer scenario, they'd never have time to grant permissions after we just installed the service, though

src/tests/miscellaneous/mock-adb/common-adb-configs.js
DaveTryon: It probably has to do with the version installed. This is was I was seeing with the A4 tablet

src/tests/electron/tests/android-setup-prompt-connect-to-device.test.ts
DaveTryon: Fixed

PR #3093, opened 2020-07-16
feat(android-setup-e2e): refactor scan-for-accessibility issues, rename variable

karanbirsingh:
#### Description of changes

Refactored scan-for-accessibility-issues slightly so that:
- we don't restart the application between HC on/off
- easier to call from within another Jest test (e.g. when we test the detect-permissions dialog)

In all usages I found we set HC on/off before scanning for issues.

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/tests/electron/tests/android-setup-locate-adb.test.ts
```diff
@@ -45,14 +45,8 @@ describe('Android setup - locate adb', () => {
...
-        async highContrastMode => {
-            await app.setHighContrastMode(highContrastMode);
-            await app.waitForHighContrastMode(highContrastMode);
-
-            const violations = await scanForAccessibilityIssues(dialog);
-            expect(violations).toStrictEqual([]);
-        },
-    );
+    it('should pass accessibility validation in both contrast modes', async () => {
+        await scanForAccessibilityIssues(app, true);
```
DaveTryon, line 49: We're still repeating the pattern of scan(app, true) scan(app, false). Would it make sense to move the true/false logic into a helper method like `scanForAccessibilityIssuesInAllModes(app)`, and call that from the tests. If a future change added a new HC mode, none of the tests would need to be updated.

src/tests/electron/tests/first-time-dialog.test.ts
```diff
@@ -21,14 +21,8 @@ describe('first time dialog', () => {
         }
     });
 
-    it.each([true, false])(
-        'should pass accessibility validation with highContrastMode=%s',
-        async highContrastMode => {
-            await app.setHighContrastMode(highContrastMode);
-            await app.waitForHighContrastMode(highContrastMode);
-
-            const violations = await scanForAccessibilityIssues(dialog);
```
DaveTryon, line 30: dialog is still defined on line 10 but never used, which is causing lint to fail :(

PR #3107, opened 2020-07-17
feat(android-setup-e2e): Remove old workflow,  step 1

DaveTryon:
#### Description of changes
Removing the old workflow will touch lots of files (mostly deletions), so we're splitting it up into multiple PR's. This first PR removes the `adbSetupView` feature flag and updates code that _directly_ uses it. We basically treat the flag as always being true, then do whatever cleanup makes sense. If it was assumed to be true, just remove the reference. If it was assumed to be false, remove the code/test case. If it was expected to change, remove the code/test case.

This leaves a lot of old classes and their tests/snapshots in place. Those will be removed in Phase 2, which will be almost exclusively deleted files.

#### Pull request checklist

- [x] Addresses an existing issue: [1748956](https://mseng.visualstudio.com/1ES/_workitems/edit/1748956)
- [x] Ran `yarn fastpass`


src/tests/unit/tests/electron/platform/android/setup/android-setup-start-listener.test.ts
```diff
@@ -33,70 +28,26 @@ describe('AndroidSetupStartListener', () => {
         },
     );
 
-    it('sends ready-to-start when initial state is not ready but changed state is', () => {
```
karanbirsingh, line 36: I wonder if we still want these test cases, just triggered with different changes w/o adbFeatureFlag

DaveTryon: Restored, triggering on a state change

PR #3111, opened 2020-07-20
fix: update detect service dialog text

karanbirsingh:
#### Description of changes

This PR updates outdated text in the step. It had the wrong text from the detect devices step.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/electron/tests/android-setup-prompt-install-service.test.ts
+++ b/src/tests/electron/tests/android-setup-prompt-install-service.test.ts
@@ -12,7 +12,6 @@ import { AndroidSetupViewController } from 'tests/electron/common/view-controlle
 import { AppController } from 'tests/electron/common/view-controllers/app-controller';
 import {
     commonAdbConfigs,
-    delayAllCommands,
     setupMockAdb,
     simulateServiceInstallationError,
     simulateServiceNotInstalled,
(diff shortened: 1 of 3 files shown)


src/tests/electron/tests/android-setup-prompt-install-service.test.ts
```diff
@@ -35,25 +38,30 @@ describe('Android setup - prompt-install-service ', () => {
...
 
     it('install button triggers installation, prompts correctly on failure', async () => {
+        await dialog.waitForDialogVisible('prompt-install-service');
         await setupMockAdb(simulateServiceInstallationError(defaultDeviceConfig));
         await dialog.client.click(getAutomationIdSelector(installAutomationId));
         await dialog.waitForDialogVisible('prompt-install-failed');
     });
 
-    it('should pass accessibility validation in both contrast modes', async () => {
+    it('spinner & prompt dialogs pass accessibility validation in both contrast modes', async () => {
```
DaveTryon, line 62: Slowing all of the tests when we only need to slow this one feels awkward. What if we left the existing tests as written, then did the test on rescan, something like this??

await setupMockAdb(delayAllComands(2500, simulateServiceNotInstaled(defaultDeviceConfig));
await dialog.client.click(installAutomationId);
await app.openAndroidSetupView('detect-service');
await scanForAccessibilityIssuesInAllModes(app);
await diaog.waitForDialogVisible(prompt-install-service);
await scanForAccessibilityIssuesInAllModes(app);

karanbirsingh: after discussing offline I will make a new test class for the spinner alone to avoid making this test class too complex / slower than it needs to be

PR #3150, opened 2020-07-23
feat(needs-review): Resource link update

jlperkins:
#### Description of changes
This PR hides 'BEST PRACTICE' when other WCAG links are present.  We do not want to display BEST PRACTICE and other links side by side; when a rule has only a best practice link as guidance, we display it, but if other links are present, we hide it.  

It is useful to keep a best practice link stored in the guidance for a rule even if other links are present, because it is used elsewhere in our code (more specifically, to determine whether a rule is included or excluded from an Automated checks scan).  However, it is not useful to display a best practice link when other links are present.  This PR leaves the link stored but does not display it when other links are present.  

#### Pull request checklist

- [x] Addresses an existing issue: [This work item](https://mseng.visualstudio.com/DefaultCollection/1ES/_workitems/edit/1751592)
- [x] Ran `yarn fastpass`

image
(Image of an experimental rule with just WCAG link showing; previously, this rule displayed both BEST PRACTICE and the WCAG link)


src/common/components/guidance-links.tsx
```diff
@@ -21,11 +21,16 @@ export const GuidanceLinks = NamedFC('GuidanceLinks', (props: GuidanceLinksProps
...
-        return links.map((link, index) => {
-            return renderLink(link, index, links.length);
+        const linksToRender = getLinksWithoutBestPracticeWhenWCAGPresent();
+        return linksToRender.map((link, index) => {
+            return renderLink(link, index, linksToRender.length);
         });
     };
 
+    const getLinksWithoutBestPracticeWhenWCAGPresent = (): HyperlinkDefinition[] => {
+        return links.length === 1 ? links : links.filter(link => link.text !== 'Best Practice');
```
waabid, line 31: Can't you just reference the `BestPractice` constant here instead?

waabid: Same in the test.

jlperkins: You're right, I'll do that

PR #3160, opened 2020-07-27
fix: Persist main window size in results view

DaveTryon:
#### Description of changes
As called out in #2147, the android client always opens results view maximized, which isn't ideal if you're trying to do something like create screenshots for documentation. #2584 was an initial attempt to address this, but was shelved at the time because other things were more urgent. This PR starts with #2584, updates it to use the Ipc layer, and expands it to remember more than just the last size. In total, this change adds the following behavior:

- Persist the maximized state (a boolean) of the app to `userConfigurationStore.windowWasMaximized` when the app exits
- Persist the last non-maximized bounds (a Rectangle) of the main window to `userConfigurationStore.lastWindowBounds`
- When opening results view, do the following:
  - If `userConfigurationStore.lastWindowBounds` exists, use it to set the main window's bounds
  - Maximize the main window unless `userConfigurationStore.windowWasMaximized` exists and is false

The net result is that the initial results view will be maximized, and that subsequent results views will resume the state that was in place when the app last exited.

One comment: Adding fields to the `UserConfigurationStoreData` class triggered changes in several other test classes that happen to use `UserConfigurationStoreData`, but that don't care about these fields. That feels like an anti-pattern to me, so I'll probably make a separate PR to see if I can clean things up. For this PR, though, I went ahead and retained the existing ugliness.

#### Pull request checklist

- [x] Ran `yarn fastpass`


DaveTryon:
Closing in favor of #3164 (which also fixes fullScreen on mac)

PR #3164, opened 2020-07-28
fix: Persist main window size in results view

DaveTryon:
#### Description of changes
As called out in #2147, the android client always opens results view maximized, which isn't ideal if you're trying to do something like create screenshots for documentation. #2584 was an initial attempt to address this, but was shelved at the time because other things were more urgent. This PR starts with #2584, updates it to use the Ipc layer, and expands it to remember more than just the last size. In total, this change adds the following behavior:

- Persist the window state ('normal', 'maximized', or 'full-screen') of the results view mainWindow to `userConfigurationStore.lastWindowState` when the app changes state in results view
- Persist the last 'normal' bounds (a Rectangle) of the results view mainWindow to `userConfigurationStore.lastWindowBounds`
- When opening results view, do the following:
  - If `userConfigurationStore.lastWindowBounds` exists, use it to set the main window's bounds
  - Based on `userConfigurationStore.lastWindowState`, either do nothing ('normal'), or enter full screen ('full-screen') or maximize (anything else).

The net result is that the normal window bounds and window state will be persisted, and that subsequent results views will resume the state that was in place when the app last exited.

Validated on both Windows and Mac (which has both 'maximized' and 'full-screen' states).

One comment: Adding fields to the `UserConfigurationStoreData` class triggered changes in several other test classes that happen to use `UserConfigurationStoreData`, but that don't care about these fields. That feels like an anti-pattern to me, so I'll probably make a separate PR to see if I can clean things up. For this PR, though, I went ahead and retained the existing ugliness.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/types/store-data/user-configuration-store.ts
```diff
@@ -14,10 +18,18 @@ export interface UserConfigurationStoreData {
...
+    lastWindowBounds: Rectangle;
 }
 
 export interface IssueFilingServicePropertiesMap {
     [service: string]: IssueFilingServiceProperties;
 }
 
 export type IssueFilingServiceProperties = Object;
+
+export interface WindowSize {
```
DaveTryon, line 32: This should be removed

src/tests/unit/tests/electron/flux/action-creator/window-state-action-creator.test.ts
```diff
@@ -79,4 +93,87 @@ describe(WindowStateActionCreator, () => {
...
+            setRouteActionMock.setup(s => s.invoke(testPayload)).verifiable(Times.once());
+            windowStateActionsMock
+                .setup(actions => actions.setRoute)
+                .returns(() => setRouteActionMock.object)
+                .verifiable(Times.once());
+        });
+
+        afterEach(() => {
+            setRouteActionMock.verifyAll();
+            windowFrameActionCreatorMock.verifyAll();
```
lisli1, line 115: Should we put these verify statements in each of the test cases? To my understanding, the afterEach block is more for cleanup

DaveTryon: Failures in afterEach will still cause the specific test to fail, so there's no functional difference. I prefer to group the common code into common places, so that the specific unit tests are short and targeted to the exact scenario. I'll put a post in the chat to see if anyone has really strong opinions either way.

src/tests/unit/tests/electron/window-management/window-frame-updater.test.ts
```diff
@@ -32,6 +32,12 @@ describe(WindowFrameUpdater, () => {
             testSubject.initialize();
         });
 
+        it('invokes enterFullScreen', () => {
+            ipcRendererShimMock.setup(b => b.maximizeWindow()).verifiable(Times.once());
+
+            windowFrameActions.maximize.invoke(null);
```
lisli1, line 38: Based on the test name, it looks like this should say enterFullScreen instead of maximizeWindow/maximize

DaveTryon: Fixed copy/paste error

src/tests/unit/tests/electron/flux/action-creator/window-state-action-creator.test.ts
DaveTryon: Moved them into setRouteNonDeviceViewCore, which is already the helper method where all code converges

PR #3165, opened 2020-07-28
fix: separate report export button from dialog

pownkel:
#### Description of changes

Currently, we render the report export button and dialog as one component, which assumes that the button will always be visible when we need the dialog to render. This prevents the "..." menu from closing after the export button is clicked and the dialog is visible.

Refactors:
- Create separate button and dialog components to use instead of ReportExportComponent
- Render the dialog at a higher level in DetailsViewCommandBar
- Store the dialog's open/close state in DetailsViewCommandBar's component state
- Refactor ReportExportComponentFactory into ReportExportDialogFactory
- Add functions to determine whether the button should be hidden (for example, in fastpass before an automated checks run has completed)

UI changes:
- "..." menu now closes after clicking the export report button
- After closing the dialog, focus returns to either the "..." button or the export report button (whichever is visible)
- Behavior when the details view is refreshed or when the view changes between assessment and fastpass is unchanged
- Behavior of the start over button is also unchanged

export button refactors

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/details-view-command-bar.tsx
```diff
@@ -115,11 +133,43 @@ export class DetailsViewCommandBar extends React.Component<DetailsViewCommandBar
...
+        );
+        if (showButton) {
+            return (
+                <ReportExportButton
+                    showReportExportDialog={this.showReportExportDialog}
+                    buttonRef={ref => (this.exportDialogCloseFocus = ref)}
+                />
+            );
+        }
+        return null;
```
waabid, line 163: I think we usually prefer the other way around. i.e.

``` if (!showButton) { return null }
return button here
```

src/DetailsView/components/should-show-report-export-button.ts
```diff
@@ -0,0 +1,27 @@
...
+import { CommandBarProps } from 'DetailsView/components/details-view-command-bar';
+
+export type ShouldShowReportExportButton = (props: CommandBarProps) => boolean;
+
+export function shouldShowReportExportButtonForAssessment(props: CommandBarProps): boolean {
+    return true;
+}
+
+export function shouldShowReportExportButtonForFastpass(props: CommandBarProps): boolean {
+    const scanResult = props.visualizationScanResultData.issues.scanResult;
```
waabid, line 14: Would be good to make the change here to not use the scan result store but rather the unified store.

src/tests/unit/tests/DetailsView/components/report-export-dialog-factory.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+        cardsViewData = null;
+        deps = {
+            detailsViewActionMessageCreator: detailsViewActionMessageCreatorMock.object,
+            getCurrentDate: () => theDate,
+            reportGenerator: reportGeneratorMock.object,
+            getDateFromTimestamp: value => theDate,
+        } as DetailsViewCommandBarDeps;
+    });
+
+    function getProps(): ReportExportDialogFactoryProps {
```
waabid, line 76: You could just set the props as a variable in the beforeEach.

src/DetailsView/components/export-dialog.tsx
```diff
@@ -24,6 +24,7 @@ export interface ExportDialogProps {
     reportExportFormat: ReportExportFormat;
     onExportClick: () => void;
     featureFlagStoreData: FeatureFlagStoreData;
+    afterDismissed?: () => void;
```
waabid, line 27: How come this is optional?

src/DetailsView/components/report-export-button.tsx
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { InsightsCommandButton } from 'common/components/controls/insights-command-button';
+import { NamedFC } from 'common/react/named-fc';
+import { IButton, IRefObject } from 'office-ui-fabric-react';
+import * as React from 'react';
+
+export interface ReportExportButtonProps {
+    showReportExportDialog: () => void;
+    buttonRef?: IRefObject<IButton>;
```
waabid, line 10: Should this be optional?

src/DetailsView/components/export-dialog.tsx
pownkel: This is only used for setting focus after the dialog is closed. In the old ReportExportComponent (which is still used in electron) we don't do any work after dismissing the dialog. Updating state to reflect that the dialog has closed is handled in onClosed, which runs just before before the dialog closes, but since the buttons are not focusable before the dialog closes we can't use onClose to set focus.

src/DetailsView/components/report-export-button.tsx
pownkel: It's not strictly needed, since all we do is pass it through to another optional prop in the OF button component, and it's only useful if the caller wants to set focus manually.

waabid: Gotcha! didn't know the OF prop was optional.

PR #3171, opened 2020-07-29
test: E2E tests for reflow components

pownkel:
#### Description of changes

Add E2E test cases to run the A11ySelfValidator on new reflow UI components, including:
- Hamburger button and panel
- Command bar "..." button and menu
- Getting started pages

We found several A11y bugs in these components during feature validation that these E2E tests would have caught, so this will prevent regressions for those fixes.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
+++ b/src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
@@ -40,10 +40,7 @@ describe('Details View -> Assessment -> Reflow', () => {
             async highContrastMode => {
                 await browser.setHighContrastMode(highContrastMode);
 
-                const results = await scanForAccessibilityIssues(
-                    detailsViewPage,
-                    detailsViewSelectors.mainContent,
-                );
+                const results = await scanForAccessibilityIssues(detailsViewPage, '*');
                 expect(results).toHaveLength(0);
             },
         );
@@ -58,10 +55,7 @@ describe('Details View -> Assessment -> Reflow', () => {
                     detailsViewSelectors.commandBarMenuButtonExpanded(true),
                 );
 
-                const results = await scanForAccessibilityIssues(
-                    detailsViewPage,
-                    detailsViewSelectors.mainContent,
-                );
+                const results = await scanForAccessibilityIssues(detailsViewPage, '*');
                 expect(results).toHaveLength(0);
 
                 // Close menu before next test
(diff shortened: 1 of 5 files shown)


src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
```diff
@@ -0,0 +1,126 @@
...
+                    detailsViewSelectors.commandBarMenuButtonExpanded(true),
+                );
+
+                const results = await scanForAccessibilityIssues(
+                    detailsViewPage,
+                    detailsViewSelectors.mainContent,
+                );
+                expect(results).toHaveLength(0);
+
+                // Close menu before next test
```
dbjorge, line 67: It's best to avoid doing cleanup like this as part of a test case, because it won't get run if an exception is thrown earlier in the test body (eg, if an `expect()` fails).

Instead, I'd suggest using another nesting level of `describe` (you can do a `describe.each` if you need each case to do the setup separately), and putting the setup/cleanup in `beforeEach`/`afterEach` blocks

src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
```diff
@@ -0,0 +1,126 @@
...
+                    detailsViewSelectors.assessmentNavHamburgerButtonExpanded(true),
+                );
+
+                const results = await scanForAccessibilityIssues(
+                    detailsViewPage,
+                    detailsViewSelectors.mainContent,
+                );
+                expect(results).toHaveLength(0);
+
+                // Close panel before next test
```
dbjorge, line 116: See earlier comment

src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        beforeEach(async () => {
+            const width = narrowModeThresholds.collapseCommandBarThreshold - 1;
+            await detailsViewPage.setViewport(width, height);
+            await detailsViewPage.waitForSelector(detailsViewSelectors.commandBarMenuButton);
+        });
+
+        it.each([true, false])(
+            `should pass accessibility validation with high contrast mode=%s`,
+            async highContrastMode => {
+                await browser.setHighContrastMode(highContrastMode);
```
dbjorge, line 41: It helps with reliability to include an `await pageUnderTest.waitForHighContrastMode(highContrastMode);` after the set call (we should probably do some sort of separate refactor PR to encapsulate the repetitive pattern into some sort of test helper...)

dbjorge: (same comment everywhere else in this file where `setHighContrastMode` is called)

Issue #445, opened 2019-03-26
Firefox support?


mike-engel:
Hey y'all! I'm sorry if this is mentioned somewhere, but I couldn't find anything about support for other browsers (specifically firefox for me).

I know that the firefox addon api is similar to chrome, but I unfortunately don't know the specifics about what's different.

Thanks!

flyingUnderTheRadar:
@mike-engel - Currently Accessibility Insights for Web only supports Chrome. We haven't explored Firefox support.

mike-engel:
@flyingUnderTheRadar Are there any plans to support other browsers in the future?

flyingUnderTheRadar:
@mike-engel  - At this point of time there is no plan to support Firefox or any other browser in the near future.

flyingUnderTheRadar:
Need to update docs to reflect this so leaving it open.

ferBonnin:
Hey @mike-engel thanks for your interest in Accessibility Insights for Web!

After long conversations with the team we have determined that for now, the costs to create an extension for Firefox would be too significant for the team.
From early explorations looks like we would have to do considerable changes, to mention some:
•	Changes to the code to ensure that it works properly in Firefox.
•	There are likely to be some special cases around visualizations, accessibility, storage, updating that we would have to develop for.
•	To fully support this we would have to test regularly in Firefox and add it to our release cadence.
•	We would have to publish this to the Firefox application store and ensure that we aren’t broken by Firefox updates.
•	Our user base in the canary and insider builds would need to be sufficient robust to ensure that our roll through the tiers is stable.

We agree that there is value in your suggestion however, for now, we want to focus on continuing providing a great experience in our current extension and improving the capabilities in it.

Thanks!

digitarald:
Firefox DevTools team member here, who would be available for help. Accessibility audits are important to keep the web open and accessible to all; and developers struggle with understanding and addressing this topic. Your audit has some great ideas and can have a meaningful impact!

Having an product labeled "for Web" only available in Chrome would be an sad outcome for an [announcement](https://blogs.windows.com/windowsexperience/2018/12/06/microsoft-edge-making-the-web-better-through-more-open-source-collaboration/) that promised:

> Together we seek the best outcome for all people who use the web across many devices. […] Ultimately, we want to make the web experience better for many different audiences.

[...]

PR #3172, opened 2020-07-29
fix: Firefox compatibility for script/CSS injection into target page

dbjorge:
#### Description of changes

This PR fixes the largest remaining blocker for Firefox support; it enables ai-web to successfully inject code into the target page. With this PR, the extension is mostly-functional in Firefox, though there are still several known issues outstanding that it does not address (see #445).

screenshot of injected automated checks dialog rendering on a target page in Firefox

The root cause of the incompatibility was this note buried in the documentation for the file parameter of the [`browser.tabs.insertCSS`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/insertCSS) and [`browser.tabs.executeScript`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/executeScript) APIs:

> **`file`** | *Optional*
>    `string`. Path to a file containing the code to inject. In Firefox, relative URLs are resolved relative to the current page URL. In Chrome, these URLs are resolved relative to the extension's base URL. To work cross-browser, you can specify the path as an absolute URL, starting at the extension's root, like this: `"/path/to/stylesheet.css"`.

This issue took a while to track down because the error message you get from Firefox when you ask it to inject a CSS/JS file with an invalid file path contains no useful information; it just says `Error: an unexpected error occurred`. We happen to use it here in a context where we're using `.then`/`.catch` callback handling all the way through our call stack, rather than `async`/`await`, which further exacerbated debugging due to not having full stack information from the error.

Technically, the only updates strictly required here were adding the `/` prefixes to the `cssFiles` and `jsFiles` in `content-script-injector.ts` and to the link elements in `shadow-initializer.ts`. The remaining changes are:

* Updates to `ContentScriptInjector` to log errors during injection to `logger.error` with context about which tab/file had trouble
* Updates to `WebextensionBrowserAdapter` to fail-fast with a descriptive error if a relative path would be passed to `insertCSS`/`executeScript`

#### Pull request checklist

- [x] Addresses an existing issue: Part of #445
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/injector/content-script-injector.ts
+++ b/src/background/injector/content-script-injector.ts
@@ -1,15 +1,16 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
+import { Logger } from 'common/logging/logger';
 import { PromiseFactory } from 'common/promises/promise-factory';
 import { flatten } from 'lodash';
 
 export class ContentScriptInjector {
-    public static readonly jsFiles: string[] = ['bundle/injected.bundle.js'];
+    public static readonly jsFiles: string[] = ['/bundle/injected.bundle.js'];
 
     public static readonly cssFiles: string[] = [
-        'injected/styles/default/injected.css',
-        'bundle/injected.css',
+        '/injected/styles/default/injected.css',
+        '/bundle/injected.css',
     ];
 
     public static timeoutInMilliSec = 5e4;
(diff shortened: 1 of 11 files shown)


dbjorge:
Reopening to fix cla bot

src/background/injector/content-script-injector.ts
```diff
@@ -1,37 +1,37 @@
...
-        );
+        this.injectCssFilesInBackground(tabId);
+        const injectJs = this.injectJsFilesInOrder(tabId);
 
-        return this.promiseFactory.timeout(inject, ContentScriptInjector.timeoutInMilliSec);
+        await this.promiseFactory.timeout(injectJs, ContentScriptInjector.timeoutInMilliSec);
     }
 
-    private injectCssFilesConcurrently(tabId: number): void {
+    private injectCssFilesInBackground(tabId: number): void {
```
waabid, line 34: I think the language here could be a bit better: 'InBackground' seems to imply you're injecting files into background. Maybe 'FromBackground' might make more sense? Or simply `injectCssFilesToTab` since this is in background regardless.

dbjorge: Fair point! Here "In Background" was meant to imply that it is something that will run in the background in a fire-and-forget manner, didn't intend it to have a meaning related to the use of the background page. Will switch to something that doesn't use the overloaded term.

src/tests/unit/tests/background/content-script-injector.test.ts
```diff
@@ -10,36 +11,48 @@ describe('ContentScriptInjector', () => {
...
-        promiseFactoryMock
-            .setup(factory => factory.timeout(It.isAny(), It.isAny()))
-            .returns(() => Promise.reject('artificial timeout'));
+            promiseFactoryMock.verifyAll();
+        });
+
+        it('rejects if a timeout occurs', async () => {
+            promiseFactoryMock
+                .setup(factory => factory.timeout(It.isAny(), It.isAny()))
+                .returns(() => Promise.reject('artificial timeout'));
```
waabid, line 52: nit: would prefer the string here to be stored/used as a variable.

src/background/injector/content-script-injector.ts
waabid: Gotcha, I completely missed that. Makes sense now.

PR #3177, opened 2020-07-30
fix(firefox): fix details view gear menu options launching extra windows

dbjorge:
#### Description of changes

Previously, in Firefox, selecting "Settings" or "Preview Features" from the gear menu in the details page would always launch an extra new details view window, in addition to opening the panel it's supposed to.

The root cause is that the actions that handle those menu item presses are the same actions that handle the similar item presses in the popup page's gear menu; they choose whether to launch a new details view window or not as part of responding to the request by checking whether the ExtensionDetailsViewController already knows about a details view tab being open for a given target tab ID, and opening a new details view tab if not. However, the tracking mechanism the control uses for detecting those details view tabs being open is based on looking for tab-update events with URL properties that match the output of this function:

```typescript
private getDetailsUrlWithExtensionId(tabId: number): string {
        return `${this.browserAdapter.getRunTimeId()}${this.getDetailsUrl(tabId)}`;
}
```

This works in Chrome because the extension URL prefix is based on the runtime ID. However, in Firefox, the runtime ID and the extension URL prefix are not necessarily related to one another (in particular, they are completely distinct for unpacked extensions), so the controller never "notices" that a details view tab was opened; this means that it thinks it has to launch a new one every time.

This PR updates the logic to use `getUrl(...)` instead of trying to format the URL manually using `getRunTimeId()`. Since this was the only point where we actually used `getRunTimeId()`, I removed it from the `BrowserAdapter` interface.

#### Pull request checklist

- [x] Addresses an existing issue: part of #445
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/extension-details-view-controller.ts
+++ b/src/background/extension-details-view-controller.ts
@@ -25,7 +25,9 @@ export class ExtensionDetailsViewController implements DetailsViewController {
             return;
         }
 
-        const tab = await this.browserAdapter.createTabInNewWindow(this.getDetailsUrl(targetTabId));
+        const tab = await this.browserAdapter.createTabInNewWindow(
+            this.getRelativeDetailsUrl(targetTabId),
+        );
 
         if (tab?.id != null) {
             this.tabIdToDetailsViewMap[targetTabId] = tab.id;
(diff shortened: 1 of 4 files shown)


src/background/extension-details-view-controller.ts
```diff
@@ -52,20 +54,23 @@ export class ExtensionDetailsViewController implements DetailsViewController {
     };
 
     private hasUrlChange(changeInfo: chrome.tabs.TabChangeInfo, targetTabId): boolean {
-        return (
-            changeInfo.url != null &&
-            !changeInfo.url
-                .toLocaleLowerCase()
-                .includes(this.getDetailsUrlWithExtensionId(targetTabId).toLocaleLowerCase())
-        );
+        if (changeInfo.url == null) {
```
dbjorge, line 57: This function's logic hasn't changed, I just refactored it with clearer names for the intermediates so I could set breakpoints during debugging and didn't change it back after because I thought it looked more clear this way.

PR #3188, opened 2020-07-31
feat(needs-review): Avoid Needs review/Automated checks visualizations on at same time

jlperkins:
### Description of changes

This PR aims to disallow Needs review and Automated checks from both being toggled on in the Adhoc Tools Panel at the same time.  It does so by changing enableTest so that it is specified in the visualization configuration, which allows for specific disabling of needs review/automated checks when needed.  For example, in the needs review configuration, when enableTest runs, it enables Needs review and disables Automated checks.  

#### Pull request checklist

- [x] Addresses an existing issue: [This work item](https://mseng.visualstudio.com/DefaultCollection/1ES/_workitems/edit/1755749)
- [x] Ran `yarn fastpass`

toggle 2
gif of toggling Needs review and Automated checks toggles in the adhoc tools panel, and when each toggle is turned on the other is automatically toggled off


src/ad-hoc-visualizations/issues/visualization.tsx
```diff
@@ -33,7 +33,10 @@ export const IssuesAdHocVisualization: VisualizationConfiguration = {
         <AdhocIssuesTestView instancesSection={FailedInstancesSection} {...props} />
     ),
     getStoreData: data => data.adhoc.issues,
-    enableTest: (data, _) => (data.enabled = true),
+    enableTest: (data, _) => {
+        data.adhoc.issues.enabled = true;
+        data.adhoc[AdHocTestkeys.NeedsReview].enabled = false;
```
JGibson2019, line 38: Just out of curiosity, why couldn't you use `data.adhoc.needsReview.enabled = false`? Is there a specific reason you need to specify `[AdHocTestkeys.NeedsReview]`

src/ad-hoc-visualizations/needs-review/visualization.tsx
```diff
@@ -32,7 +32,10 @@ export const NeedsReviewAdHocVisualization: VisualizationConfiguration = {
         <AdhocIssuesTestView instancesSection={NeedsReviewInstancesSection} {...props} />
     ),
     getStoreData: data => data.adhoc.needsReview,
-    enableTest: data => (data.enabled = true),
+    enableTest: data => {
+        data.adhoc.needsReview.enabled = true;
+        data.adhoc[AdHocTestkeys.Issues].enabled = false;
```
JGibson2019, line 37: Same question here that I had in the issues visualization file

src/ad-hoc-visualizations/issues/visualization.tsx
jlperkins: Changed it so it's always using the enums directly.  This way if the enum ever changes, these will still work

PR #3192, opened 2020-08-03
test: migrate from puppeteer to playwright

dbjorge:
#### Description of changes

This PR is an experiment to try migrating our web e2e tests from Puppeteer to Playwright. We'll probably want to discuss it in Engineering Discussion before merging.

[Playwright](https://playwright.dev/) is a relatively new (released 1.0 in January) competitor to Selenium/Puppeteer/Cypress, maintained by Microsoft and developed by a team that includes several ex-Puppeteer team members. Its API surface is very similar to Puppeteer's (note that there are very few changes to individual tests required in this migration PR), but it offers several advantages that are relevant to our project compared to Puppeteer:

* Playwright's selectors pierce the shadow DOM by default (we needed to do some tricky workarounds for this with Puppeteer to test our target page visualizations)
* Playwright's selectors are generally more flexible; they support xpath based queries, innertext-based queries, and even innertext+regex queries. The latter two require more effort with Puppeteer.
* Playwright took a page out of Cypress's book and will, by default, wait for elements to be ["actionable"](https://playwright.dev/#version=v1.2.1&path=docs%2Factionability.md&q=) as a part of all its APIs which interact with elements on a page. This is a *big deal* for test reliability, and would allow us to obsolete a lot of our `Page` helpers that try to do similar actionability checks (but in a less complete/robust way than Playwright advertises).
* Playwright's error messages are dramatically more actionable than Puppeteer's. For example, when I was doing this migration PR, there were a few tests in `visualization-boxes.test.ts` that were failing when they tried to look for the `#insights-shadow-host .insights-highlight-container` selector. In Puppeteer, this would have shown up in Jest logs as something like "Element not found", and it would have been up to you to figure out which element based on the stack. In Playwright, you still have the same stack info, but you additionally get this message which explains not only which element wasn't found, but context about *why* it wasn't found (in this case, that it was in the DOM but not yet visible as of the timeout):
  ```
    TimeoutError: Timeout 5000ms exceeded during page.waitForSelector.
    ================ page.waitForSelector logs ================
    [api] waiting for selector "#insights-shadow-host #insights-shadow-container" to be visible
    [api]   selector resolved to hidden <div id="insights-shadow-container">…</div>
    ============================================================
    Note: use DEBUG=pw:api environment variable and rerun to capture Playwright logs.
  ```
* Playwright's Firefox support is more complete (it is mostly-supported, but extension support requires workarounds; see https://github.com/microsoft/playwright/issues/2644 and https://github.com/microsoft/playwright/issues/2874 - Puppeteer's support for Firefox is currently experimental and much less complete)
* Playwright maintains their own headful-browser-compatible docker base image, so we don't have to rely on Cypress's even though we aren't using Cypress
* Playwright is written in Typescript, so its typings are included with the main npm package and cannot become out of date like `@types/puppeteer`

#### Pull request checklist

- [x] Addresses an existing issue: part of enabling e2e tests for #445
- [x] Ran `yarn fastpass`


src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -111,14 +101,15 @@ export class Page {
     }
 
     public async bringToFront(): Promise<void> {
```
karanbirsingh, line 103: should we remove the method signature entirely?

src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -22,29 +24,17 @@ export type PageOptions = {
...
-                `Puppeteer.Page '${underlyingPage.url()}' emitted ${eventDescription}`,
+                `Playwright.Page '${underlyingPage.url()}' emitted ${eventDescription}`,
             );
         }
 
         function serializeError(error: Error): string {
             return `[Error]{name: '${error.name}', message: '${error.message}', stack: '${error.stack}'}`;
         }
 
-        underlyingPage.on('error', error => {
```
karanbirsingh, line 36: are we able to remove this because of https://github.com/microsoft/playwright/blob/master/docs/api.md#event-crash? or is there a different reason?

src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -155,132 +147,68 @@ export class Page {
...
 
     public async clickSelector(selector: string): Promise<void> {
         await this.screenshotOnError(async () => {
             await this.underlyingPage.waitForSelector(selector, {
                 timeout: DEFAULT_PAGE_ELEMENT_WAIT_TIMEOUT_MS,
             });
-            await this.underlyingPage.hover(selector);
-            await this.underlyingPage.waitFor(DEFAULT_CLICK_HOVER_DELAY_MS);
-            await this.underlyingPage.click(selector, { delay: DEFAULT_CLICK_MOUSEUP_DELAY_MS });
+            // await this.underlyingPage.hover(selector);
```
karanbirsingh, line 176: can we remove these?

src/tests/end-to-end/common/page-controllers/page.ts
dbjorge: Yes, that's the replacement. Playwright doesn't actually support the `error` event on its `Page` objects, but even if it did we'd want to use the per-API exceptions instead for more local error reporting.

Issue #2655, opened 2020-05-10
Rename "Text legibility" category


patrickhlauke:
## Is your feature request related to a problem? Please describe.
The current category name of "Text legibility" does not sufficiently/accurately convey what some of the tests/SCs are about.

High Contrast Mode, Orientation, Reflow (see https://github.com/microsoft/accessibility-insights-web/pull/2654) and Hover / focus content are not solely about text, but content in general.

Having these grouped under "text legibility" may lead auditors to the false impression that any problems under those tests relating to non-text content are ok/not applicable.

## Describe the desired outcome
Ideally, find a more suitable category name (and generalise the high-level description). Perhaps something like "Page adaptability", "Content adaptability" or similar. However, this admittedly makes "Contrast" and "Hover / focus content" stand out again/doesn't quite make them fit.

The Contrast test could logically be moved into the "Contrast" category.

"Hover / focus content" could potentially be placed under "Custom widgets"

ferBonnin:
Thanks for the suggestion @patrickhlauke!

@LiLoDavis thoughts for changing the title of the test?
Changing requirements from one test to another is not trivial work (specially considering what happens for users that have an assessment in progress) but we could change the test title to something that is more comprehensive

patrickhlauke:
if they were all to still stay within the same category, maybe (though a bit ugly) renaming to a combined "Content adaptability / text legibility"? hover/focus would still feel like the odd one out though to an extent...

LiLoDavis:
We'll need the test name to be short. At this time, the longest test name is **Automated checks**. 
How about **Viewable content**?

patrickhlauke:
> Changing requirements from one test to another is not trivial work

Going forward though, I think that would be the preferred option...but understand that it's quite a breaking change.

> We'll need the test name to be short. At this time, the longest test name is Automated checks.
How about Viewable content?

Doesn't completely capture some of the nuance (e.g. could argue that for "orientation", stuff is still "viewable", just that you potentially are forced to tilt your head on your hard-mounted device to see it properly), but as I can't come up with anything more suitable (closest I had was "Adaptable / viewable" but that still doesn't fit the current layout/length), I guess I could live with that for the time being.

(it does still feel odd at a very high level that text contrast would not be under contrast though...but maybe the more fundamental reshuffling of categories is something that could be considered for a future version)

ferBonnin:
[...]

PR #3200, opened 2020-08-04
fix: renaming Text legibility as Adaptable content

JGibson2019:
#### Description of changes
Renamed the text legibility assessment to be adaptable content per #2655 

Image showing Assessment 17 now reads "Adaptable content"
image

Image showing the new getting started view
image

#### Pull request checklist

- [X] Addresses an existing issue: #2655 
- [X] Ran `yarn fastpass`


src/assessments/adaptable-content/assessment.tsx
```diff
@@ -14,9 +14,9 @@ import { Reflow } from './test-steps/reflow';
 import { ResizeText } from './test-steps/resize-text';
 import { TextSpacing } from './test-steps/text-spacing';
 
-const key = 'textLegibility';
-const title = 'Text legibility';
-const { guidance } = content.textLegibility;
+const key = 'adaptableContent';
```
dbjorge, line 17: Be careful about backwards compatibility; it will probably be necessary to avoid changing whatever keys names appear in IndexedDB data for storing assessment results, otherwise any users with an assessment in progress at the time they receive the update with this change will end up with broken assessments.

JGibson2019: So should I keep the key as text legibility? Or is there another way for me to ensure backwards compatibility?

dbjorge: If I'm remembering correctly, the `key` is the important thing to maintain. The `storeDataKey` at L59 also makes me suspicious based on its name, but not so much based on the code using it.

I'd recommend testing it out to see what happens (and to validate any updates you might try making here):

* Check out master
* Start an assessment and fill in a few results for Text Legibility
* Open up the dev tools on the background page and look through the stored IndexedDB data for the assessment
* In particular, Ctrl+F through the whole contents for the terms "Text Legibility" or "textLegibility"; it's likely that anywhere it appears as a key for a dictionary (as opposed to a value) will need to be preserved
* Close chrome, checkout your feature branch, rebuild, open chrome
* Continue the assessment in progress and check IndexedDB again; see if the in-progress results persisted appropriately across the upgrade

Note that I'm making an assumption here that we would want in-progress results to persist across the rename; if you haven't already, it would be good to double check that assumption with @ferBonnin. We do sometimes intentionally invalidate in-progress results if there is a significant enough change to the requirements/text. It's also possible that we don't think users would care either way; if that's true, the consistency might be more valuable than a few users having to redo their Text Legibility results for one assessment.

An alternative would be to rekey the requirement for consistency and update the `InitialAssessmentStoreDataGenerator` to understand how to convert from the old key name to the new one when loading persisted data. I think that's probably a bigger hammer than this change warrants, but it's an option.

JGibson2019: Did this a couple of times to be sure (building master first, then my feature branch and then vice versa), and can confirm you're right! It will remove the assessment data for text-legibility/adaptable-content if you rename the key! @ferBonnin let me know if overriding in progress results for that one assessment (the data for other assessments remains in tact)

ferBonnin: Great catch Dan! @JGibson2019 we don't want our users to loose their data from the text legibility/adaptable content test.

JGibson2019: Ok tested that everything else is fine if we leave the key as `text-legibility`. Out of curiosity, will there ever be a point where we rename the key? Are there any other (maybe unintended) consequences of having the key name not match the assessment name? @dbjorge

dbjorge: If we were to make a different change in the future that dramatically changed most/all of the requirements, such that we would want to intentionally invalidate any existing results, it would make sense to change the key then. That sort of change is fairly rare; I would expect it to be unlikely to happen within the product's lifetime.

I think the main negative consequence of leaving the key as-is permanently are that it might make debugging/logging a little confusing in the future, but as long as the old key is the only place we retain `textLegibility`, I don't think it's a big deal; it's not hard to search the codebase and find what it really refers to.

If we decided we really wanted to change it for consistency's sake, you could do it by changing the key here but separately updating the point where we read in previously-persisted data with new logic to convert from the old key to the new one. That conversion logic could potentially be ripped out a few releases later. I think if there were bigger disadvantages to the inconsistency it would make sense to take this approach - it's not particularly difficult or dangerous, just adds a bit of dev cost to write the transformation logic and test it. I don't feel like it's worth the time in this case, since I don't think the inconsistency causes any real problems for us.

It would probably be a good idea to leave a comment next to the `key` definition explaining that the inconsistency is intentional for backcompat purposes

src/tests/end-to-end/tests/content/__snapshots__/guidance-content.test.ts.snap
```diff
@@ -1368,6 +1368,160 @@ exports[`Guidance Content pages rules/uniqueLandmark matches the snapshot 1`] =
...
+    class="content-container"
+  >
+    <div
+      class="content-left"
+    />
+    <div
+      class="content"
+    >
+      <h1>
+        Cannot find test/adaptableContent/contrast/infoAndExamples
```
dbjorge, line 1383: This snapshot looks off to me (this is probably why the e2e tests are failing)

JGibson2019: ^ yes this was why but rebuilding fixed it. Just pushed a commit with the update and it should be good now!

PR #3203, opened 2020-08-04
feat(needs-review): Selectively show 'Export result' button

jlperkins:
#### Description of changes

This is a PR to avoid showing the 'export result' button in the upper right hand corner of the details view in automated checks when the data that is used to generate the report is needs review data and not from an automated checks scan.  This is achieved by moving the logic of whether that button is shown to the visualization configuration and to where the button is rendered. 

#### Pull request checklist

- [x] Addresses an existing issue: [This work item](https://mseng.visualstudio.com/DefaultCollection/1ES/_workitems/edit/1755016)
- [x] Ran `yarn fastpass`

export report
(Gif of the details view, starting on a completed needs review scan.  The user switches to automated checks, where no 'export result' button is present.  The user runs the automated checks scan and the button appears.  The user navigates back to the needs review test and the button is once again gone.)


src/tests/unit/tests/DetailsView/components/details-view-command-bar.test.tsx
```diff
@@ -79,24 +93,30 @@ describe('DetailsViewCommandBar', () => {
...
-    });
-
-    test('renders with export button, without start over', () => {
-        testOnPivot(true, false);
-    });
-
-    test('renders without export button, with start over', () => {
-        testOnPivot(false, true);
-    });
+    test.each`
```
dbjorge, line 103: Nice!

jlperkins:
> I think the duplication is probably better than the overhead of a helper, given how many arguments the helper would need

Yeah, that was the thinking behind keeping the duplication.  And thanks for the catch on the link! I have just been using <a> tags for hyperlinks but you're right, it should be a NewTabLink. I'll make that change there and everywhere else I introduced links.

PR #3212, opened 2020-08-05
Chore: Add null checks

RobGallo:
#### Description of changes

Add more files to the list of null checked files and update code in cases that need modifications for null checking


src/injected/shadow-initializer.ts
```diff
@@ -38,7 +38,7 @@ export class ShadowInitializer {
 
         const shadowHostElement = this.createDivWithId('insights-shadow-host');
 
-        rootContainer.appendChild(shadowHostElement);
+        rootContainer?.appendChild(shadowHostElement);
```
dbjorge, line 41: This makes it so the method is a no-op if it can't do its job, where previously it would throw an error; I think throwing an error would probably be preferable if the root container can't be found

src/injected/shadow-utils.ts
```diff
@@ -7,9 +7,9 @@ export class ShadowUtils {
     constructor(htmlElementUtils: HTMLElementUtils) {
         this.htmlElementUtils = htmlElementUtils;
     }
-    public getShadowContainer(): HTMLElement {
+    public getShadowContainer(): HTMLElement | undefined {
         const shadowHost = this.htmlElementUtils.querySelector('#insights-shadow-host');
 
-        return shadowHost.shadowRoot.querySelector('#insights-shadow-container') as HTMLElement;
+        return shadowHost?.shadowRoot?.querySelector('#insights-shadow-container') as HTMLElement;
```
dbjorge, line 13: Every callsite where `getShadowContainer()` is called assumes that it will be returned as defined; based on that, I'd recommend having this throw an error rather than return undefined, since it doesn't introduce any new error paths and would improve the message in existing error paths (`cannot invoke method appendChild on undefined` turns into `Could not find shadow container element #insights-shadow-host`)

src/injected/client-utils.ts
```diff
@@ -41,10 +41,14 @@ export class ClientUtils {
     }
 
     public matchesSelector(element: ElementMatcher, selectorString: string): boolean {
```
dbjorge, line 43: I think we should just get rid of this method entirely; it's an IE9 polyfill for a method that Chromium has supported since 2014, and we already have other parts of the code using it without polyfilling. I think it would be fine to just update the only callsite (`heading-formatter.ts`) to invoke `element.matches(selector)` directly in place of `clientUtils.matchesSelector(element, selector)`.

(Alternately, we could add an `HtmlElementUtils` dependency in `heading-formatter` and use the existing non-polyfill wrapper there (`elementMatches`); I'm not sure that would add any value to `heading-formatter`, since it already depends directly on a variety of other `Element` methods/properties directly)

PR #3246, opened 2020-08-13
fix: Exclude color contrast test results (false positives)

DaveTryon:
#### Description of changes
We've had reports of false positives in the color contrasts test, so we're temporarily excluding the results from the output. This is intended to be a short-term change that will be reverted after we are able to improve the error rate of the results being returned from axe-android.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/common/components/cards/__snapshots__/rule-resources.test.tsx.snap
+++ b/src/tests/unit/tests/common/components/cards/__snapshots__/rule-resources.test.tsx.snap
@@ -3,7 +3,7 @@
 exports[`RuleResources renders with {
   url: 'test-url',
   guidanceLinks: [ [length]: 0 ],
-  linkComponent: [Function] {
+  linkComponent: [Function (anonymous)] {
     [length]: 1,
     [name]: '',
     displayName: 'ElectronExternalLink'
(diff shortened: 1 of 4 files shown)


src/tests/unit/tests/common/components/cards/__snapshots__/rule-resources.test.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`RuleResources renders with {
   url: 'test-url',
   guidanceLinks: [ [length]: 0 ],
```
DaveTryon, line 5: This file needed to be updated to pass on my machine. Not really sure why, since it seems unrelated to these changes

peterdur: Looks like the PR build unit tests are failing here. When I tried out this change on my machine, running unit tests generated new snapshots without the `(anonymous)`. I'm guessing something is out of sync? Was a `yarn install` missed, or is your `node --version` something other than `v12.16.3`?

DaveTryon:
Pausing this until I get the details sorted out

src/tests/unit/tests/common/components/cards/__snapshots__/rule-resources.test.tsx.snap
DaveTryon: Node version issue. Homebrew installed 14.X on my mac. It works now that I forced it back to 12.X

PR #3253, opened 2020-08-17
feat(axe-core-4.0.1): update approach for testing axe.commons

karanbirsingh:
#### Description of changes

Between axe 3 and 4, several modules in axe.commons.* [were refactored](https://github.com/dequelabs/axe-core/pull/2147). See below for details. This PR updates the way we mock axe.commons and test axe functionality so that:
- we add a utility function that can be used to mock axe.commons.* types*
- in some cases, we add missing usage of `getFlattenedTree` to ensure axe.commons.* works correctly in contexts outside of `axe.run`. 

These changes should pass in both axe 3 and axe 4. The current tests in master fail in axe 4.0.1.

details:
*The resulting axe.js we consume in 4.0.1 introduces (via webpack + es modules) `Object.defineProperty` with its default `configurable = false` behavior to populate the module's transitive imports. (See definition of `__webpack_require__.d` [here](https://cdn.jsdelivr.net/npm/axe-core@4.0.1/axe.js)). This contrasts with the pre-ES-module approach of simple assignment. Because `configurable` is set to false, the properties defined this way are [immutable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty). Tests that attempt to use `=` or `Object.defineProperty` (as typemoq does [here](https://github.com/florinn/typemoq/blob/0721689785fe533e68fe455280288980020c7eac/src/GlobalScope.ts#L46)) fail; for instance, `axe-utils.test.ts` fails with `1: TypeError: Cannot redefine property: accessibleText`. The workaround includes minimal changes to allow tests to continue running, including redefining `axe.commons` and switching some cases of mock-usage to integration-tests that use axe directly. Existing issues that discuss this in more detail are at https://github.com/webpack/webpack/pull/7132 (unmerged) and https://github.com/webpack/webpack/issues/6979

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/scanner/axe-utils.test.ts
+++ b/src/tests/unit/tests/scanner/axe-utils.test.ts
@@ -1,8 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import * as axe from 'axe-core';
-import { GlobalMock, GlobalScope, IGlobalMock, It, MockBehavior, Times } from 'typemoq';
+import { GlobalMock, GlobalScope, IGlobalMock, It, Mock, MockBehavior, Times } from 'typemoq';
 
+import { withAxeCommonsMocked } from 'tests/unit/tests/scanner/mock-axe-utils';
 import * as AxeUtils from '../../../../scanner/axe-utils';
 import { setAxeGlobalTreeTo } from '../../common/axe-internals';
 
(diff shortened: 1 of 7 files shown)


src/tests/unit/tests/scanner/custom-rules/unique-landmark-check.test.ts
```diff
@@ -59,13 +47,12 @@ describe('axe.Check: unique-landmark', () => {
 
     it('set role & label for html5 element', () => {
         fixture.innerHTML = `
-            <div id="sectionLabel">section landmark</sectionLabel>
+            <div id="sectionLabel">section landmark</div>
```
DaveTryon, line 50: Was this just a typo in the old test?

karanbirsingh: I think so

src/tests/unit/tests/scanner/custom-rules/css-content-rule.test.ts
```diff
@@ -98,16 +102,21 @@ describe('verify matches', () => {
...
         },
     );
 
     function testSemantics(elements: HTMLElement, expectedResult: boolean): void {
-        let result: boolean;
-
-        GlobalScope.using(getComputedStyleMock, axeVisibilityMock).with(() => {
-            result = cssContentConfiguration.rule.matches(elements, null);
-        });
+        const result = cssContentConfiguration.rule.matches(elements, null);
```
dbjorge, line 119: The helper is only 2 lines long now; I think just repeating it would make the tests more clear about what they're acting/asserting

src/tests/unit/tests/scanner/axe-utils.test.ts
```diff
@@ -28,34 +28,22 @@ describe('AxeUtils', () => {
...
                 accessibleTextMock
                     .setup(m => m(It.isValue(elementStub), isLabelledByContext))
                     .verifiable(Times.once());
-                AxeUtils.getAccessibleText(elementStub, isLabelledByContext);
-            });
-            accessibleTextMock.verifyAll();
-        });
+                withAxeCommonsMocked('text', { accessibleText: accessibleTextMock.object }, () => {
+                    AxeUtils.getAccessibleText(elementStub, isLabelledByContext);
+                });
```
dbjorge, line 44: Should there be a `accessibleTextMock.verifyAll()` in here?

PR #3254, opened 2020-08-17
feat(axe-core-4.0.1): Duplicate title string as aria-label for links

DaveTryon:
#### Description of changes

Axe-core 4.0.1 is generating the following error in 2 different places (the guidance link and the assessment completion link) when scanning the current extension:

```
Element does not have text that is visible to screen readers
aria-label attribute does not exist or is empty
aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
Element's default semantics were not overridden with role=\"presentation\"
Element's default semantics were not overridden with role=\"none\"", "selector": [".link-..."]}]}]
```

These links both have the title property set, so it seems odd to assert that the aria-label attribute must also be set (possible false positive in axe-core?). This PR simply sets the aria-label property to whatever the title property is set to, and updates the necessary snapshots.

There will be a separate PR (including an update for the version) in the axe-core-4.0.1 branch, but I created this against master to:
1. Collect this change into a single PR in case it proves to be a false positive in axe-core
2. Reduce the scope of the code change (and its potential impact) when we merge the axe-core-4.0.1 branch into master 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/nav-link-button.tsx
+++ b/src/DetailsView/components/nav-link-button.tsx
@@ -12,10 +12,12 @@ export interface NavLinkButtonProps extends INavButtonProps {
 
 export const NavLinkButton = NamedFC<NavLinkButtonProps>('NavLinkButton', props => {
     const link = props.link;
+    const titleAndAriaLabel = link.title || link.name;
     return (
         <Link
+            aria-label={titleAndAriaLabel}
             aria-expanded={link.isExpanded}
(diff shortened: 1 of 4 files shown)


dbjorge:
DIscussed offline with Dave; we think these are both false positives due to https://github.com/dequelabs/axe-core/issues/2459

PR #3257, opened 2020-08-18
feat(axe-core-4.0.1): Disable link-name rule (false positive)

DaveTryon:
#### Description of changes

Axe-core 4.0.1 has a false positive (documented at https://github.com/dequelabs/axe-core/issues/2459) that can trigger `link-name` errors on links that specify their text via the `target` attribute instead of the `aria-label` attribute. This is breaking our E2E tests and could expose users to false positives. To address these, we're temporarily classifying the `link-name` rule as a BestPractice and updating the E2E tests to expect the failures from axe-core.

We also explored using the `aria-label` attribute (augmenting or replacing the `target` attribute), but concluded that this was the better approach, on the assumption that the false positive would be fixed, hopefully soon.

#### Pull request checklist

- [x] Addresses an existing issue: [1759944](https://mseng.visualstudio.com/1ES/_workitems/edit/1759944)
- [x] Ran `yarn fastpass`
- [blocked pending #3253] Added/updated relevant unit test(s) (and ran `yarn test`)
- [blocked pending #3253] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`

### DIFF
--- a/src/tests/end-to-end/tests/details-view/headings.test.ts
+++ b/src/tests/end-to-end/tests/details-view/headings.test.ts
@@ -73,7 +73,7 @@ describe('Details View -> Assessment -> Headings', () => {
                     headingsPage,
                     detailsViewSelectors.mainContent,
                 );
-                expect(results).toHaveLength(0);
+                expect(results).toHaveLength(1);
             },
         );
     });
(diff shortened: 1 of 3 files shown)


src/tests/end-to-end/tests/details-view/headings.test.ts
```diff
@@ -73,7 +73,7 @@ describe('Details View -> Assessment -> Headings', () => {
                     headingsPage,
                     detailsViewSelectors.mainContent,
                 );
-                expect(results).toHaveLength(0);
+                expect(results).toHaveLength(1);
```
karanbirsingh, line 76: can we leave a comment in the file linking to the issue?

DaveTryon: Added

PR #3276, opened 2020-08-24
feat (compliance-web): Global readiness fixes

RobGallo:
#### Description of changes

- Use the official name for the es-419 language code: Spanish (Latin American)
- Remove the first parameter to Date.toLocaleTimeString so that the default language format will be used

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/formatted-date.tsx
+++ b/src/reports/components/formatted-date.tsx
@@ -1,9 +1,15 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import * as moment from 'moment';
+import { Globalization } from 'common/globalization';
+import * as Moment from 'moment';
 import * as React from 'react';
 
+export type FormattedDateDeps = {
+    globalization: Globalization;
+};
+
 export interface FormattedDateProps {
+    deps: FormattedDateDeps;
     date: Date;
 }
 
@@ -13,7 +19,8 @@ export class FormattedDate extends React.Component<FormattedDateProps> {
     }
 
     private formatDateTime(date: Date): string {
-        let momentDate = moment(this.props.date);
-        return momentDate.toLocaleString();
+        const moment = Moment.utc(this.props.date);
+        const localMoment = moment.locale(this.props.deps.globalization.languageCode);
+        return localMoment.format('L LTS [UTC]Z');
     }
 }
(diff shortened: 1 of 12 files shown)


src/reports/components/formatted-date.tsx
```diff
@@ -50,7 +50,7 @@ export class FormattedDate extends React.Component<FormattedDateProps> {
 
     private getTimeZone(date: Date): string {
         const timeString = date
-            .toLocaleTimeString('en-us', { timeZoneName: 'short' })
+            .toLocaleTimeString(undefined, { timeZoneName: 'short' })
```
dbjorge, line 53: Even with this change, I think this datetime formatting logic is a little suspect; it's almost-but-not-quite ISO8601 (it uses 12h instead of 24h). We already have a runtime dependency on `moment`, so I'm wondering if it wouldn't be both simpler and more correct to replace all of `formatDateTime` and its helpers with something like

```ts
const keepTimeZoneOffset = true;
const momentDate = moment(this.props.date);
return momentDate.toISOString(keepTimeZoneOffset);
```

src/reports/components/formatted-date.tsx
```diff
@@ -1,61 +1,19 @@
...
-        const timeString = date
-            .toLocaleTimeString('en-us', { timeZoneName: 'short' })
-            .replace(/\u200E/g, '');
-        return timeString.substr(timeString.lastIndexOf(' ') + 1);
-    }
-
-    private padStartWithZero(num: number, digits: number): string {
-        return padStart(num.toString(), digits, '0');
+        let momentDate = moment(this.props.date);
+        return momentDate.toLocaleString();
```
dbjorge, line 17: I'm pretty sure this isn't right. Moment doesn't document a `toLocalString()` implementation at all, so I think this is going to actually be picking up `Object.prototype.toLocaleString()`, which falls back to the moment object's plain `toString()` since it doesn't implement a `toLocaleString()`.

What you actually would want is to call `format`, passing it one of the "Localized format" options [it documents](https://momentjs.com/docs/#/displaying/format/).

Note that the default `'moment'` package only contains an english locale, not all locales; to get support for other locales, you'd want to change the import line to be `from 'moment/min/moment-with-locales'` rather than just `from 'moment'`.

Even with the correct package imported, Moment infers the local timezone but it does *not* infer the local language information by default. You would need to tell it explicitly which locale to use. The correct way to infer that from modern versions of Firefox/Chromium is `window.navigator.language`, though we'd typically use a `windowUtils` around the direct navigator access for mockability.

Be:

```ts
import * as moment from 'moment/min/moment-with-locales';

const momentDate = moment(this.props.date);
const localMomentDate = momentDate.locale(this.props.deps.windowUtils.language() /* returns window.navigator.language */);
const formattedDate = localMomentDate.format('LTS');
```

PR #3283, opened 2020-08-25
feat(compliance-web): Address security warnings

RobGallo:
#### Description of changes

Enabled security eslint rules for production code. Maintained disabling the below security eslint rules for code which is not released to production because of the extremely high false positive rate for the security rule set.

- (2) security/detect-non-literal-regexp
- (3) security/detect-non-literal-fs-filename
- (0) security/detect-unsafe-regex
- (0) security/detect-child-process
- (0) security/detect-eval-with-expression

Maintained the disabled rule 'security/detect-object-injection' because there are 295 warnings. Those will be addressed in another PR.

### DIFF
--- a/src/common/configuration/file-system-configuration.ts
+++ b/src/common/configuration/file-system-configuration.ts
@@ -13,25 +13,27 @@ import {
 import * as fs from 'fs';
 import * as path from 'path';
 
+export type ReadFileSync = () => Buffer;
+
 // Note: this is resolved relative to the bundled js file, not the src layout
-const defaultConfigJsonPath = path.join(__dirname, '../insights.config.json');
+const defaultConfigPath = path.join(__dirname, '../insights.config.json');
+// eslint-disable-next-line security/detect-non-literal-fs-filename
+const defaultReadFileSync: ReadFileSync = () => fs.readFileSync(defaultConfigPath);
 
 // Appropriate for contexts without a DOM but with access to the fs module
 // (eg, electron main process)
 export class FileSystemConfiguration implements ConfigAccessor, ConfigMutator {
-    private readonly configJsonPath: string;
+    private readonly readFileSync: ReadFileSync;
     public config: InsightsConfiguration;
 
-    constructor(configJsonPath?: string) {
-        this.configJsonPath = configJsonPath ?? defaultConfigJsonPath;
+    constructor(readFileSync?: ReadFileSync) {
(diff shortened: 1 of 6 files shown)


src/common/configuration/file-system-configuration.ts
```diff
@@ -29,6 +29,8 @@ export class FileSystemConfiguration implements ConfigAccessor, ConfigMutator {
 
     public reset(): ConfigMutator {
         try {
+            // The following warning is disabled because this.configJsonPath is a string literal whose value is set differently only for unit tests
+            // eslint-disable-next-line security/detect-non-literal-fs-filename
             const configJsonRawContent = fs.readFileSync(this.configJsonPath).toString();
```
dbjorge, line 34: I think this is actually somewhat legitimate; even if it isn't a problem on its own because we only use it intentionally with a constant parameter, it would allow a bad actor a path for escalate "property pollution of the config object" into "arbitrary file read".

If the path override is only used by unit tests anyway, probably the easiest fix is to just inject `fs.readFileSync` as the dependency in place of `configJsonPath` and update the tests replacing `testFilePath` parameters with either `() => fs.readFileSync(testFilePath)` or `() => Buffer.from(inlineTestFileContent)`. (I'd go with the former despite it being a unit test that touches the file system, better test of the real error handling responsibilities of the class)

src/issue-filing/common/http-query-builder.ts
```diff
@@ -43,6 +43,8 @@ export class HTTPQueryBuilder {
 
         const fullUrl = `${this.baseUrl}?${queryParameters}`;
 
+        // The following warning is thrown incorrectly because the call to 'truncate' is not on the file system
```
dbjorge, line 46: Would be good to link the bug tracking the false positive, https://github.com/nodesecurity/eslint-plugin-security/issues/54

src/popup/handlers/launch-panel-header-click-handler.ts
```diff
@@ -14,7 +14,11 @@ export class LaunchPanelHeaderClickHandler {
         if (item == null) {
             return;
         }
+
         const url: string = item.data;
+
+        // the following warning is thrown incorrectly because the call to 'open' is not on the file system
```
dbjorge, line 20: Would be good to link the bug tracking the false positive, https://github.com/nodesecurity/eslint-plugin-security/issues/54

Issue #3301, opened 2020-08-28
AI-Web > Assessment > Adaptable content > Contrast isn't showing instances


LiLoDavis:
## Describe the bug

In two places, the tool is supposed to show instances of the color-contrast rule that need review:

- FastPass > Needs review > color-contrast
- Assessment > Adapable content > contrast

However, the following pages are showing inconsistent results in the two place:
1. https://www.w3.org/WAI/standards-guidelines/aria/
2. https://www.factzoo.com/mammals/types-of-monkeys.htm
3. https://www.monkeyhappy.com/

**To Reproduce**
1. Go to one of the above pages.
2. Compare instances generated in FastPass vs. Assessment (see paths above).
3. All three pages generate multiple instances in FastPass but 0 instances in Assessment

## Expected behavior

Instances should be identical in FastPass and Assessment.

## Context (please complete the following information)

- OS Name & Version: Microsoft Windows 10 Enterprise 10.0.18363
- AI-Web Version & Environment: 2.21.1 Production
- Browser Version: Google Chrome Version 85.0.4183.83 (Official Build) (64-bit)
- Target Page: See above

## Are you willing to submit a PR?

No

## Did you search for similar existing issues?

Yes

pownkel:
This looks like it might have been caused by the recent upgrade to Axe 4.0

ferBonnin:
> This looks like it might have been caused by the recent upgrade to Axe 4.0
Interesting, how so?

pownkel:
We get the color contrast evaluate function directly and wrap it in another function, but in axe 4 they've changed how they store/access those functions, so our check is failing silently. We don't call the evaluate function directly for any other checks, so this shouldn't affect other rules.

ferBonnin:
@LiLoDavis this is now fixed in Canary

LiLoDavis:
Looks good to me.

PR #3304, opened 2020-08-31
fix: remove use of defaultConfig from Axe._audit

pownkel:
#### Description of changes

In Axe 4.0, the structure of check configurations changed so that the `evaluate` property could be set as either a function or as a string key that references an internal axe function. In `Axe._audit.defaultConfig.checks`, checks may still have a string in place of the evaluate function. We use `defaultConfig` to access the color-contrast evaluate function in our custom text-contrast check, but since that value is now a string instead of a function, the check was failing silently and causing a false negative for the Adaptable content > contrast requirement.

This PR removes uses of `defaultConfig` and instead uses `Axe._audit.checks` or `Axe._audit.rules`, since the checks found here do have the real evaluate function, instead of just a string. This fixes the false negative.

#### Pull request checklist

- [x] Addresses an existing issue: #3301
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/scanner/rule-to-links-mappings.test.ts
+++ b/src/tests/unit/tests/scanner/rule-to-links-mappings.test.ts
@@ -11,6 +11,7 @@ describe('ruleToLinkConfiguration', () => {
     const bestPracticeAxeRules: string[] = axe.getRules(['best-practice']).map(rule => rule.ruleId);
 
     it.each(allAxeRules)(`should have a mapping for axe rule %s`, rule => {
+        const config = ruleToLinkConfiguration;
         expect(ruleToLinkConfiguration[rule]).toBeDefined();
     });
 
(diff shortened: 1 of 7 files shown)


waabid:
Not sure how quickly we want to get this out so might not want to do this right now but it might make sense to maybe have some integration tests around this usage of axe.

src/tests/unit/tests/scanner/rule-to-links-mappings.test.ts
```diff
@@ -11,6 +11,7 @@ describe('ruleToLinkConfiguration', () => {
     const bestPracticeAxeRules: string[] = axe.getRules(['best-practice']).map(rule => rule.ruleId);
 
     it.each(allAxeRules)(`should have a mapping for axe rule %s`, rule => {
+        const config = ruleToLinkConfiguration;
```
RobGallo, line 14: The variable config appears not to be used.

waabid (approved):

PR #3319, opened 2020-09-03
fix: update lockfile for bl

waabid:
#### Description of changes

bl is a transitive dependency of a couple dependencies, hence I specified the appium-adb one, since that is the one that needs to be updated per dependabot.

#### Pull request checklist

- [x] Ran `yarn fastpass`


package.json
```diff
@@ -177,6 +177,7 @@
         "dot-prop": "^5.2.0",
         "kind-of": "^6.0.3",
         "minimist": "^1.2.3",
-        "yargs": "^15.3.1"
+        "yargs": "^15.3.1",
+        "appium-adb/**/bl": "^4.0.3"
```
dbjorge, line 181: I don't think a resolution is necessary here. I don't understand why dependabot couldn't do this automatically, since appium-adb/**/bl was already requesting bl@^4.0.1, which can be satisfied by 4.0.3 without a resolution.

This can be resolved by keeping the resolutions as-is and forcing yarn to update yarn.lock manually by deleting the bl@^4.0.1 entry and then running `yarn install`

dbjorge (changes requested):
see inline

package.json
waabid: gotcha! Updated!

dbjorge (approved):

dbjorge:
Merging on Wahaj's behalf since he's OOF

PR #3360, opened 2020-09-15
perf: ~9% bundle size reduction by replacing moment with luxon

dbjorge:
#### Description of changes

The Moment.js maintainers [announced today](https://momentjs.com/docs/) that they consider moment to be "a legacy project in maintenance mode".

This isn't necessarily a problem for us using it, but it prompted me to look into what alternatives existed. [Luxon](https://moment.github.io/luxon/), a newer library by some of the same developers as Moment, is the one that stuck out from Moment's "recommend alternatives" list, since I think its internationalization support uses a strategy that I think is more likely to stay maintained long-term than its peers (Luxon uses the [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) APIs built into modern browsers, rather than maintaining its own locale packages).

One of the notable features Moment will never support is tree-shaking, a feature of Webpack (and other bundlers) that allows for us to avoid including code we never call from libraries we only use small portions of. I wanted to see how big of an impact this made in practice for us, and it was significant enough to cause a measurable difference; several prod bundles get reduced by 9% when replacing Moment with Luxon. Bundle size translates pretty directly into time-to-interactive for the details view and time-to-scan when we inject code into the target page, so this will amount to a minor perf improvement for end users (probably in the 20-50ms range for most users)

Because Luxon is deferring to the standard browser format, the format does change a little from the old version Moment was producing (`01/01/2000 12:00:00 AM UTC+00:00` becomes `1/1/2000, 12:00:00 AM UTC`). I've already cleared this change with @ferBonnin.

The only significant negative from using Luxon would be that it relies on running in an environment where any locale you want to use is provided by the platform. This is true on all browser and electron versions we support and in Node 14, but in Node 12 there is only partial support (the `Intl` APIs are supported, but Node 12 is only packaged with support for the `en-us` locale, so the code doesn't crash but also doesn't localize anything). This has small but non-zero impact:

* We would need to migrate ai-service and its scan CLI package to Node 14 before they could add localization support (very unlikely, based on current roadmaps)
* There is a unit test in this PR that needs to be skipped until we migrate this repo to Node 14 for builds

There are still a few places in code that weren't using Moment for datetime formatting in the first place (`DateProvider`, notably); I've left updating those cases as out-of-scope for this PR, but it would be good to update them to work consistently with `FormattedDate`.

Prod bundle size deltas:
* `background.bundle.js`: `3792912` -> `3478656` (-9%)
* `debugTools.bundle.js`: `1756842` -> `1442207` (-22%)
* `detailsView.bundle.js`: `3980565` -> `3666312` (-9%)
* `devtools.bundle.js`: `114063` -> `114063` (0%)
* `injected.bundle.js`: `3759476` -> `3445218` (-9%)
* `insights.bundle.js`: `2642975` -> `2642828` (0%)
* `popup.bundle.js`: `3274890` -> `3274683` (0%)

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/debug-tools/components/telemetry-viewer/telemetry-messages-list.tsx
```diff
@@ -91,4 +91,4 @@ export const onRenderTimestamp = (
 
 export const defaultDateFormatter: DateFormatter = timestamp =>
     // untested line: moment is sensitive to the host machine time zone
```
peterdur, line 93: Dangling reference to moment in comment; is this still true for luxon?

src/tests/unit/tests/reports/components/formatted-date.test.tsx
```diff
@@ -1,23 +1,23 @@
...
+import { DateTime } from 'luxon';
 import { FormattedDate, FormattedDateProps } from 'reports/components/formatted-date';
 
 describe('FormattedDate', () => {
     describe('render', () => {
         test('end of last millennium', () => {
-            const date = moment.utc('1999-12-31T23:59:59').toDate();
-            testDate('en-us', date, '12/31/1999 11:59:59 PM UTC+00:00');
+            const date = DateTime.fromISO('1999-12-31T23:59:59', { zone: 'utc' }).toJSDate();
+            testDate('en-us', date, '12/31/1999, 11:59:59 PM UTC');
```
peterdur, line 10: (Existing issue) - is `expected` parameter to `testDate` redundant with the snapshot check?

src/common/date-provider.ts
```diff
@@ -17,13 +16,15 @@ export class DateProvider {
     }
 
     public static getUTCStringFromDate(date: Date): string {
-        return utc(date.toISOString()).format('YYYY-MM-DD h:mm A z');
+        const utcDateTime = DateTime.fromJSDate(date, { zone: 'utc' });
+        return utcDateTime.toFormat('YYYY-MM-DD h:mm A z');
```
peterdur: We recently had someone observe that UTC times are typically specified in 24-hour format; do we want to address that here?

dbjorge: I wanted to try to minimize the number of format changes that weren't strictly required for the library conversion. Personally, I agree with [Luxon's writeup](https://moment.github.io/luxon/docs/manual/formatting.html#consider-alternatives) that we should either be using ISO8601 or a system-default locale string and not using our own, but I wanted to separate the "moment to luxon conversion" PR from any "format change PRs" to the extent possible

src/tests/unit/tests/reports/components/formatted-date.test.tsx
dbjorge: Yes, I agree. I think it's easier to see whether the test is testing the right thing with the inline version than the snapshot, so removing the snapshot

src/debug-tools/components/telemetry-viewer/telemetry-messages-list.tsx
dbjorge: Yes, but will update the comment

dbjorge: (the format string is also wrong (see: not tested))

src/reports/components/formatted-date.tsx
```diff
@@ -19,8 +19,12 @@ export class FormattedDate extends React.Component<FormattedDateProps> {
     }
 
     private formatDateTime(date: Date): string {
-        const moment = Moment.utc(this.props.date);
-        const localMoment = moment.locale(this.props.deps.globalization.languageCode);
-        return localMoment.format('L LTS [UTC]Z');
+        const utcDateTime = DateTime.fromJSDate(date, { zone: 'utc' });
+        const localDateTime = utcDateTime.setLocale(this.props.deps.globalization.languageCode);
```
peterdur, line 23: The name `localDateTime` makes me think this has been converted to the local time zone, but that's not the case here; the _locale_ or culture/language has been specified.

waabid (approved):
LGTM

peterdur (approved):
Second approval

PR #3366, opened 2020-09-16
fix: add application icon for dev builds of unified

dbjorge:
#### Description of changes

Our electron builds already used branded application icons in blue/violet/red for prod/insider/canary builds, but were missing icons for dev builds. I had thought this was just a quirk of running electron without going through electron-builder, but it actually turns out that we just never generated electron-compatible icon sets for "gray" or "green" colors.

(In Electron 9 specifying an icon that doesn't exist changes behavior from "use default electron icon" to "crash on startup", so this preempts having to debug that crash when we next update electron)

We don't currently use green, but I added it anyway in case we do someday; the cost for me to add it now is way lower than the cost to re-learn how to generate these in the future.

**Before:**
screenshot of dev build on taskbar with default electron icon

**After:**
screenshot of dev build on taskbar with our "gray" icon

#### Pull request checklist

- [x] Ran `yarn fastpass`


karanbirsingh (approved):
validated (used InkScape for the icns files)

PR #3370, opened 2020-09-17
fix: unified cards view reflow at 320px width

dbjorge:
#### Description of changes

The PR addresses a few reflow issues with the unified cards UI at 320px effective width (the bug was reported at 1280x1024 in 400% zoom; per discussion with @ferBonnin, this PR enables the app to resize down to the equivalent 320x256@100%, since we need to support that anyway for 400% users and since it makes testing reflow changes much easier).

Particularly:

* The "expand/collapse" toggle and "visual helper" toggle can now be reflowed onto separate lines, rather than forcing a horizontal scrollbar in the "automated checks" panel
* The "rule resources" card text will now break across lines within words if a word cannot fit in the view (particularly notable for long rule names)
* The highlight status text in the card footer will now break across multiple lines if necessary to avoid the "more actions" button flowing off the right edge of the card (and the footer will stretch upwards to accomodate the reflowed text if necessary)

While I was working here, I also noticed that the breakpoint at which the cards reflow from two-column to one-column is wonky in unified; it decides to do that reflow based on a media query for the total width of the application, and the reflow point is designed for the assorted card views in web and reports where the cards are given the entire width of the screen to play with. On unified, where it has to contend with the screenshot panel, the reflow point doesn't work very well; it allows the content column of the cards to get way too small (3-4 characters wide) before reflowing down to one column. I plan to address this by switching the reflow mechanism from media-query-based to display:flex-on-the-table-rows-based, but that change is much more complicated/scary than these ones and isn't strictly necessary to address the trusted tester bug, so leaving it for a separate PR.

**Before:**
screenshot highlighting reflow issues fixed by this PR

**After:**
screenshot with highlights around changes described in PR

#### Pull request checklist

- [x] Addresses an existing issue: 1765540
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/components/cards/cards-visualization-modifier-buttons.scss
+++ b/src/common/components/cards/cards-visualization-modifier-buttons.scss
@@ -2,4 +2,5 @@
 // Licensed under the MIT License.
 .cards-visualization-modifiers-container {
     display: flex;
+    flex-wrap: wrap;
 }
--- a/src/common/components/cards/instance-details-footer.scss
+++ b/src/common/components/cards/instance-details-footer.scss
@@ -31,6 +31,7 @@
         label {
             color: $secondary-text;
             font-size: $fontSizeM;
+            word-break: break-word;
         }
     }
 }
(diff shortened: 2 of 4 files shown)


dbjorge:
FYI for @ferBonnin on visual changes

karanbirsingh (approved):

PR #3372, opened 2020-09-17
fix: improve result card reflow breakpoints

dbjorge:
#### Description of changes

Previously, the reflow behavior for cards (switching from "row headers to the left of content" to "row headers above content" at small widths) was driven using a media query over screen width. This worked OK in web and in reports, where cards are given the entire width of the screen to work with, but caused poor behavior in unified, where the cards are competing for screen width with the screenshot panel. In the most pathological width, you can see something like this:

**Before:**
screenshot showing cards using only 3-4 characters per column for card row content

This PR addresses this by redoing the reflow behavior for cards in terms of `display:flex` on the card rows, so the reflow behavior will be based on the width of the cards, not the width of the screen.

Per discussion with @ferBonnin, as part of this update, I've fixed up a few padding/margin sizes that were off from Figma's suggestions (in particular, we were using a width for the labels on android that was *way* larger than spec). With these changes, it was possible to combine the web/unified behavior and remove the unified-specific override by accepting web's padding being 4px higher than figma, so we did that (the result is still closer to the figma than the "before" case).

**I've reviewed each of these cases and encourage reviewer(s) to build the code locally and do the same:**
* Unified automated checks view
* Unified report exports
* Web fastpass automated checks
* Web fastpass needs review
* Web fastpass report exports
* Report package

##### Before:
unified-autochecks-before

##### After:
unified-autochecks-after

##### Before:
unified-report-before

##### After:
unified-report-after

##### Before:
web-autochecks-before

##### After:
web-autochecks-after

##### Before:
web-needsreview-before

##### After:
web-needsreview-after

#### Pull request checklist

- [n/a] Addresses an existing issue: continuation from #3370
- [x] Ran `yarn fastpass`


src/reports/components/instance-details.scss
```diff
@@ -51,80 +51,66 @@
...
-    .report-instance-table td {
-        display: block;
+    overflow-wrap: break-word;
+    word-break: break-word;
 
-        .label {
-            text-align: left;
+    .row {
+        display: flex;
+        flex-direction: row;
```
waabid, line 61: flex-direction is row by default so this is not necessary.

src/reports/components/instance-details.scss
```diff
@@ -51,80 +51,66 @@
...
+        flex-wrap: wrap;
+
+        // The padding on the parent and margins on the children are coupled; they
+        // are intended to produce the desired spacing regardless of whether the children
+        // are presented side-by-side or wrapped
+        padding-top: 2px;
+        padding-bottom: 14px;
+        padding-left: 0px;
```
waabid, line 69: I tried removing this line and it had no effect on the styling. We might be able to get rid of this.

src/reports/components/instance-details.scss
```diff
@@ -51,80 +51,66 @@
...
+        padding-right: 20px;
+        > * {
+            margin-top: 12px;
+            margin-bottom: 0px;
+            margin-left: 20px;
+            margin-right: 0px;
```
waabid, line 75: I tried removing this line as well and it also had no effect on the styling.

src/reports/components/instance-details.scss
```diff
@@ -51,80 +51,66 @@
...
+            padding: 0;
         }
 
-        .instance-list-row-content {
-            text-align: right;
+        border-bottom: 0.5px solid $neutral-10;
```
waabid, line 79: nit: we can also write this like so:

```
        + .row {
            border-top: 0.5px solid $neutral-10;
        }
```

Adjacent elements can have styles applied to them like ($selector + $selector) to apply styles to all but the first element (hence why border-top over border-bottom). Yes, the nittiest of nits but I always like sharing this cool selector usage. Link to some text on it: https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/

src/reports/components/instance-details.scss
```diff
@@ -51,80 +51,66 @@
...
+        color: $secondary-text;
+        font-size: 14px;
+        line-height: 20px;
+        align-items: flex-end;
+        display: flex;
     }
 
-    .report-instance-table .instance-list-row-content {
-        padding: 6px 20px 14px 20px;
+    .instance-list-row-content.content-snipppet {
```
waabid, line 112: I know this style was from before this change but figured it might be nice to fix the typo here.

```.content-snipppet``` should be ```.content-snippet```

waabid (approved):
Minor suggestions, otherwise LGTM

src/reports/components/instance-details.scss
dbjorge: Fixed (and moved to a more specific module for the one use site)

src/reports/components/instance-details.scss
dbjorge: I agree that this would work equally well, but I didn't use it because I think it's significantly harder for a reader to reason about, despite being shorter:

* if someone don't know the advanced selectors off the top of their head, "last-child" is a lot easier to infer the meaning of than "+"
* even if I do remember that "+" is an adjacent sibling selector, I have to remember whether it refers to the "left" or "right" sibling to understand if the css should be using -top or -bottom in a code review

src/reports/components/instance-details.scss
dbjorge: I agree that it isn't currently necessary; I included it (and the 0 margins a few lines down) because I found it more readable to be explicit (there are some elements that browsers will use a 1px padding/margin for by default, and this saves a reader/reviewer from having to think about that/look up whether it applies)

src/reports/components/instance-details.scss
dbjorge: see above

PR #3376, opened 2020-09-18
refactor: add explicit rule inclusion list, decoupled from 'Best Practice' designation

karanbirsingh:
#### Description of changes

This PR separates the concepts 'best practice' from 'should we include/exclude this rule'. It does so by:
- restoring the WCAG mappings to their 'true' values, e.g. `link-name` has valid WCAG mappings even though we want to exclude it
- replacing the `rule-sifter` files with a new `get-rule-inclusions` file that compiles a list of what rules we include/exclude based on explicit overrides, WCAG mappings, whether axe enables it, etc
- adding a snapshot file that describes the rules we include/exclude and why

This sets some groundwork so we can share the rules between projects.

Note: I had to edit changes to `integration.test.ts` manually. Generating the snapshot automatically made the right changes for `link-name`, but also brought in incorrect changes locally in untouched parts of the report (possibly due to line-endings or different node version). In particular, `collapsible-script-provider` and `new-tab-link-with-confirmation-dialog` had different whitespace.

Out of scope:
- figuring out whether we want to reuse Axe's best practice tags
- figuring out whether we should remove `audio-captions` explicit inclusion
- figuring out whether we should export Axe.Spec configuration overrides
- any new package work

[Reports.zip](https://github.com/microsoft/accessibility-insights-web/files/5285446/Reports.zip) contains
- PR: a report generated from the code in this PR
- Prod: a report generated from the code in Prod (same results as PR)
- CommentOutLinkNameExclusion: a report generated from the code in this PR, with `link-name` removed from the exclusion list. `link-name` results show up (edited the target page to force some)

#### Pull request checklist

- [x] Ran `yarn fastpass`

original thoughts:

> - in PRs like https://github.com/microsoft/accessibility-insights-web/pull/1269 and https://github.com/microsoft/accessibility-insights-web/pull/3257 the WCAG mappings were changed to ignore false positives. In #3257 in particular, we not only added `BestPractice` but also removed the original WCAG mappings entirely. When re-enabling this rule, we will have to re-release the reporter package to undo this; one could imagine having left the WCAG list alone & toggling the rule status elsewhere
> - it's unclear whether we mark rules as `BestPractice` to match [axe's list](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md#best-practices-rules) or whether it's because of a false positive. For instance, the rule `scrollable-region-focusable` (ignored bc of a false positive in #1269) is still disabled today (we might want to re-enable), but you wouldn't know it from the code https://github.com/microsoft/accessibility-insights-web/blob/9741e08c004b45aa81452f7bcf935e124a3cbb26/src/scanner/rule-to-links-mappings.ts#L26 Keeping these concepts separate would help readability
> - The service codebase maintains its [own exclusion lists](https://github.com/microsoft/accessibility-insights-service/blob/52005b03e8e6465614beea940ea779d6128578f6/packages/scanner-global-library/src/factories/rule-exclusion.ts) that are mostly in sync with our `BestPractice` lists, but they don't match completely; for instance, `no-autoplay-audio` is an axe-core experimental rule similar to `hidden-content` but we only exclude `hidden-content` in the service list, and not `no-autoplay-audio`. One idea is to export an enabled/disabled list in the reporter package so that consumers can decide whether to run those specific rules (rather than run all rules and use the reporter to 'hide' results)


src/scanner/get-rule-inclusions.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { BestPractice } from 'scanner/rule-to-links-mappings';
+import { DictionaryStringTo } from 'types/common-types';
+import { HyperlinkDefinition } from 'views/content/content-page';
+import * as Axe from 'axe-core';
+
+// Temporary workaround for issue https://github.com/dequelabs/axe-core/issues/2459
```
karanbirsingh, line 9: this file could house false positive info

src/scanner/get-rule-inclusions.ts
```diff
@@ -0,0 +1,29 @@
...
+const explicitlyExcludedRules: string[] = ['link-name'];
+
+export const getRuleInclusions = (
```
karanbirsingh, line 12: have to make sure Axe.configure() is called before this

src/tests/unit/tests/scanner/__snapshots__/get-rule-inclusions.test.ts.snap
```diff
@@ -0,0 +1,342 @@
...
+  "aria-valid-attr": Object {
+    "reason": null,
+    "status": "included",
+  },
+  "aria-valid-attr-value": Object {
+    "reason": null,
+    "status": "included",
+  },
+  "audio-caption": Object {
+    "reason": null,
```
karanbirsingh, line 66: `audio-caption` is disabled in axe. this gets pulled in because of https://github.com/microsoft/accessibility-insights-web/blame/6fd2816817db760818b35a3e711c02df3cc90c97/src/scanner/axe-rule-overrides.ts I'm exploring if these overrides are still relevant. If they are, we may be able to pull them in here so there's a reason for its explicit inclusion as well.

src/scanner/get-rule-inclusions.ts
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { BestPractice } from 'scanner/rule-to-links-mappings';
+import { DictionaryStringTo } from 'types/common-types';
+import { HyperlinkDefinition } from 'views/content/content-page';
+import { IRuleConfiguration } from 'scanner/iruleresults';
+
+export interface RuleIncluded {
```
dbjorge, line 9: Can probably be a little stronger in this type

`export type RuleIncluded = { status: 'included' } | { status: 'excluded', reason: string }`

src/tests/unit/tests/reports/package/__snapshots__/integration.test.ts.snap
```diff
@@ -9136,12 +9136,23 @@ exports[`report package integration with no issues 1`] = `
                     <span
                       class="guidance-links"
                     >
```
karanbirsingh, line 9138: This change can be compared with the source here before `link-name` was disabled: https://raw.githubusercontent.com/microsoft/accessibility-insights-web/f4ae57d42d1be911b055d327868a1975a8e615e0/src/tests/unit/tests/reports/package/__snapshots__/integration.test.ts.snap (ctrl-f for `link-name`)

src/scanner/get-rule-inclusions.ts
karanbirsingh: went with in-between `{​​​​​ status: 'included', reason?: string }​​​​​ | {​​​​​ status: 'excluded', reason: string }​​​​​`

dbjorge (approved):

Issue #3337, opened 2020-09-10
Text changes related to color-contrast 'needs review'


LiLoDavis:
## Describe the bug

The color-contrast rule can generate 'needs review' instances for elements that contain icons. We should tell users to ignore icons.

### FastPass > How to check
New first bullet, modified start of second bullet. All other text remains as is, including links.

> - If the instance is an icon or other non-text content, ignore it. This rule applies only to text.
> - If the instance is text, use Accessibility Insights for Windows (or the Colour Contrast Analyser if you're testing on a Mac) to manually verify that it has sufficient contrast compared to the background. If the background is an image or gradient, test an area where contrast appears to be lowest.
> 
> For detailed test instructions, see **Assessment** > **Adaptable content** > **Contrast**.

### Assessment > Adaptable content > Contrast > How to test
Add a new first step, and modify the second step. All other text remains as is, including links.

> For this requirement, Accessibility Insights for Web highlights instances of text where the contrast ratio can't be determined, typically because the background color is not uniform. You must manually verify the contrast for these instances. 
> 
> _Note: If no matching/failing instances are found, this requirement will automatically be marked as pass._
> 
> 1. Examine each instance in the target page to determine whether it is text. (Because icons are assessed in **Contrast** > **Graphics**, they can be marked as **Pass** in this test.)
> 1. Examine each text instance to identify an area where the text and background are most likely to have a low contrast ratio (e.g., white text on a light gray background).
> 1. Use Accessibility Insights for Windows to test the contrast at that area. (If you are testing on a Mac, you can use the Colour Contrast Analyser.)
> 1. Verify that each instance meets these contrast thresholds:
>     a. Regular text must have a ratio ≥ 4.5
>     b. Large text (18pt or 14pt+bold) must have a ratio ≥ 3.0.
> 1. Record your results:
>  a. Select **Fail** for any instances that do not meet the requirement.
>  b. Otherwise, select **Pass**. Or, after you have marked all failures, select **Pass unmarked instances**.

## Are you willing to submit a PR?

No.

ferBonnin:
verified in Canary :)

PR #3399, opened 2020-09-24
fix: add corrected text for color contrast tests

shanisebarona:
#### Description of changes
Updated the text for the assessment's adaptable content contrast test and fastpass to address issue #3337.

image

image

#### Pull request checklist

- [x] Addresses an existing issue: #3337
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/components/cards/how-to-check-text.tsx
+++ b/src/common/components/cards/how-to-check-text.tsx
@@ -31,7 +31,7 @@ export const HowToCheckText = NamedFC<HowToCheckTextProps>('HowToCheckText', pro
         }
         case 'color-contrast': {
             checkText = (
-                <div classname={styles.combinationLists}>
+                <div className={styles.combinationLists}>
                     <ul className={styles.multiLineTextYesBullet}>
                         <li list-style-type="disc">
                             If the text is intended to be invisible, it passes.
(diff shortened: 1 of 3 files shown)


src/common/components/cards/how-to-check-text.tsx
```diff
@@ -54,7 +55,7 @@ export const HowToCheckText = NamedFC<HowToCheckTextProps>('HowToCheckText', pro
                         <li>
                             For detailed test instructions, see{' '}
                             <b>
-                                Assessment {'>'} Text legibility {'>'} Contrast
+                                Assessment {'>'} Adaptable content {'>'} Contrast
                             </b>
```
shanisebarona, line 59: I noticed this was previously in bold, so I left as is, but it seems that most of the bold styling is wrapped in <Markup.Term></Markup.Term>. Should this be changed to <Markup.Term> instead?

LiLoDavis (approved):
The text changes look good to me.

src/common/components/cards/how-to-check-text.tsx
dbjorge: Good catch! I agree that it would be an improvement to be consistent about using the semantically-named components; I think reusing `Term` would be better than using `<b>` here.

dbjorge (changes requested):
This looks good to me once the `<b>` -> `<Term>` thing is updated!

dbjorge (approved):

Issue #1796, opened 2019-12-04
Card from report doesn't respond to NVDA / Narrator table commands when reached via 'tab'


karanbirsingh:
## Describe the bug

This issue was created from this [comment](https://github.com/microsoft/accessibility-insights-web/issues/1494#issuecomment-548473227). When a user tabs to a card in the report, the row contents are not read by NVDA or Narrator.

**To Reproduce**

- Run a fastpass on [this website](https://www.washington.edu/accesscomputing/AU/before.html)
- Export a report
- Open NVDA on the report
- Pressing "tab" to get to the card leads to focus on the card. Pressing any arrow key or shortcut key from this point does not lead to the content of the card being read (only the first card row, which gets read by default when first focused on the card. The CTRL+ALT+arrow combinations also do not lead to any cell
- Before tabbing to the card, NVDA does recognize a "T" to get to the table. At this point arrow keys work properly to navigate through each table
- Pressing "T" while focused on the card (after getting there from tabbing) does not go to the table
- If focused on something other than the card, pressing down and up arrows reads through all the content in the card as well (similar to browse mode)

It seems like being focused on the card disallows arrow key & 'T' input from NVDA.

## Expected behavior

Are there benefits to this table being in the tab order - am I missing a common/canonical keystroke to read the contents?

## Context (please complete the following information)

- OS Name & Version: Windows 1903
- AI-Web Version & Environment: 2.10.3 production, 2.11.0 insider
- Browser Version: Version 78.0.3904.108 (Official Build) (64-bit)

## Did you search for similar existing issues?

This issue was separated from https://github.com/microsoft/accessibility-insights-web/issues/1494, which was auto-closed; we may want to add more details to that issue for closure.

ferBonnin:
@iamrafan could you please take a look at this?

ferBonnin:
Per conversation with Peter, moving to active and assigning to @waabid since you are working on it too

peterdur:
Moving back to Ready for Work since this is not actively being worked on.

CorgiDev:
It seems to have to do with the use of `<div>` tags that are then given the role of table instead of semantic tables. The functionality of the table role greatly depends on the screen reader / browser combination. For example, I was able to navigate the tables in the card using the NVDA keyboard shortcuts when using NVDA with Firefox, but not using NVDA with Chrome or the Chromium based Edge. Looking to see if there are any fixes that can be used for this short of it being changed to a semantic table.

[...]

PR #3401, opened 2020-09-25
Fix 1796 nvda table commands not work

CorgiDev:
#### Description of changes
Fix for issue #1796 where NVDA table commands were not able to properly access the cards in an exported report. This appears to be due in part to the inconsistent support for table related role attributes used on the divs containing the data tables. 

This appears to particularly be the case with the `gridcell` role, as mentioned in the MDN document on the [ARIA: gridcell role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Gridcell_role). Removing them made NVDA able to properly read the table without impacting the functionality of NVDA when interacting with the tables. 

I also added `scope="row"` to each row header for additional data relationship clarity. This was done because the first header in the first row was being incorrectly identified as a possible column and row header by the [ANDI](https://www.ssa.gov/accessibility/andi/help/install.html) accessibility testing tool used in Section 508 compliance testing.

I tested the functionality with NVDA 2020 and JAWS 2020. No visible UI changes. 

I attached a zip file containing 2 copies of the exported automated check report below. One prefixed with "BEFORE_" and the other "AFTER_" to indicate whether they occurred before or after I made my changes in case anyone wants to double-check the screen reader impact of my changes.

[Exported_Automated_Check_Files.zip](https://github.com/microsoft/accessibility-insights-web/files/5280035/Exported_Automated_Check_Files.zip)

#### Pull request checklist

- [x] Addresses an existing issue: #1796
- [x] Ran `yarn fastpass`


CorgiDev:
It looks like this had a fail because of me removing the roles from the divs surrounding the semantic table that contains the information. However, putting the roles back in causes the issue with NVDA because those roles are poorly supported, particularly when using certain combinations of screen reader and browser.

CorgiDev:
After thinking on it, I realized that the roles on the outer divs around the semantic table are probably meant to help with when there is more than one result so the user knows how many are there. However, leaving them there results in the incompatibility between NVDA and the cards in the exported report. I'm going to take a further look at it to see if I can figure it out.

karanbirsingh:
Hi @CorgiDev, thanks for the PR and info! It's helpful to know that the outer roles affect the table command behavior in this way. I think you're right there may be side effects with removing them though - this [prior PR](https://github.com/microsoft/accessibility-insights-web/pull/1532#discussion_r338170189) mentions the roles help with aria-selected announcements on the interactive cards. We will take a closer look at the issue to see what approach works best for both scenarios; I'll add any updates to #1796.

As a side note, initial failed tests on a UI PR like this are likely OK. We use [jest snapshot testing](https://jestjs.io/docs/en/snapshot-testing), which maintains a reference snapshot for each UI component. From their description: 

> "The test will fail if the two snapshots do not match: either the change is unexpected, or the reference snapshot needs to be updated to the new version of the UI component."

When a UI change is proposed, one can update the accompanying reference snapshots with `yarn test -u`. This adds the updated reference snapshots to the diff and would fix the tests.

dbjorge:
We had some discussion among ourselves last week trying to think of options for resolving the issue without causing other regressions, but didn't come to a conclusion. We've scheduled some more time to meet and think about options on Thursday this week; hopefully we'll have an update for you after that. Thanks for your patience!

CorgiDev:
> We had some discussion among ourselves last week trying to think of options for resolving the issue without causing other regressions, but didn't come to a conclusion. We've scheduled some more time to meet amongst ourselves and think about options on Thursday this week; hopefully we'll have an update for you after that. Thanks for your patience!

I've been looking up some articles to see if I can identify some solutions. This article might provide some ideas for discussion.
https://sarahmhigley.com/writing/grids-part1/

RobGallo:
Hi @CorgiDev ,

We have discussed a fix and I have added a description of it in issue #1796 . The fix will include the change you made here, which seems to have been the right thing to do (in the report UI).

Thank you so much for contributing to Accessibility Insights! You helped move the resolution of this issue forward.

PR #3415, opened 2020-09-29
feat(android-left-nav): Add left nav actions

RobGallo:
#### Description of changes

- Create `LeftNavActionsz class
- Create `LeftNavActionsCreator` class
- Remove the `defaultKey` from the `LeftNavStore` constructor because it can be set in the `initialize` function
- implement `LeftNavStore.addActionListeners`
- Move `LeftNavStore` and `LeftNavStoreData` alongside similar store and data classes under the flux directory

Next change will be to create the left nav store in render-initializer.


src/electron/flux/store/left-nav-store.ts
```diff
@@ -0,0 +1,30 @@
...
+import { LeftNavItemKey } from 'electron/types/left-nav-item-key';
+
+export class LeftNavStore extends BaseStoreImpl<LeftNavStoreData> {
+    constructor(private readonly actions: LeftNavActions) {
+        super(StoreNames.LeftNavStore);
+    }
+
+    public getDefaultState(): LeftNavStoreData {
+        return {
+            selectedKey: 'overview',
```
waabid, line 18: We should change this to automated checks since that is the default for unified.

waabid (approved):
Minor change. Otherwise LGTM

src/tests/unit/tests/electron/flux/store/__snapshots__/left-nav-store.test.ts.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`LeftNavStore returns default state 1`] = `undefined`;
```
lisli1, line 3: I'm curious why this snapshot has `undefined` as the value. Is this expected?

PR #3418, opened 2020-09-30
feat(android-left-nav): Set 'automated-checks' as default page for the left nav store

RobGallo:
#### Description of changes

This change resulted from a comment in a previous PR. Also fixed unit tests which were not properly initializing the snapshot.


lisli1 (approved):
LGTM!

PR #3423, opened 2020-09-30
fix: wrap urls in summary report

pownkel:
#### Description of changes

Currently, when we have long URLs in the summary report, they stretch out the results table and can make it longer than everything else on the page:

summary report overflow

This PR will cause long URLs to wrap even if they can't be broken up "by word":

summary report overflow fixed

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/report-sections/url-errors-table.tsx
+++ b/src/reports/components/report-sections/url-errors-table.tsx
@@ -4,7 +4,10 @@ import { NamedFC } from 'common/react/named-fc';
 import * as React from 'react';
 import { SummaryScanError } from '../../package/accessibilityInsightsReport';
 import { NewTabLink } from 'common/components/new-tab-link';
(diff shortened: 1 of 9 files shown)


src/reports/components/report-sections/url-errors-table.tsx
```diff
@@ -23,5 +23,14 @@ export const UrlErrorsTable = NamedFC<UrlErrorsTableProps>('UrlErrorsTable', pro
         return [errorType, urlLink, errorLogLink];
     });
 
-    return <SummaryResultsTable columnHeaders={headers} rows={rows} id={props.id} />;
+    const columnIsUrl = [false, true, false];
```
waabid, line 26: We should combine this with headers into a single object.

waabid (changes requested):
Largely LGTM. Would modify it such that we use an object instead of two loosely coupled arrays instead however.

waabid (approved):

PR #3427, opened 2020-10-01
feat(android-left-nav): Add content page info

RobGallo:
#### Description of changes

- Add a type, `ContentPagesInfo`, which is a dictionary of `ContentPageInfo` objects referenced by `LeftNavItemKey`
- Create `ContentPanelDeps`, which contains a `ContentPagesInfo` instance
- Add the `description` property to the `TestConfig` object, which is a `JSX.Element` so that each page can have whatever style of content is required for its description.

`ContentPageInfo` contains the static information needed to render the content page displayed when a given left nav item is selected.

There are now two separate factories which create left nav dependency information based on the `TestConfig` object. The `TestConfig` object is designed so that differencial information about each test can be stored in one place. But we have elected not to simply pass the `TestConfig` objects around because it tends to bloat dependencies, and it makes determining what a component requires from the `TestConfig` less obvious.


src/electron/common/content-page-info-factory.ts
```diff
@@ -0,0 +1,21 @@
+import { ContentPageInfo, ContentPagesInfo } from 'electron/types/content-page-info';
+import { LeftNavItemKey } from 'electron/types/left-nav-item-key';
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { TestConfig } from 'electron/types/test-config';
+
+export const createContentPagesInfo = (configs: TestConfig[]): ContentPagesInfo => {
+    const retVal = {} as ContentPagesInfo;
```
waabid, line 9: Since it's an empty object, it doesn't need the "as ContentPagesInfo. You should be able to use normal definition like:

`retVal: ContentPagesInfo`

src/electron/platform/android/test-configs/automated-checks/test-config.tsx
```diff
@@ -1,9 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+import * as React from 'react';
 import { TestConfig } from '../../../../types/test-config';
 
+const description = <>Here's the description for automated checks</>;
```
waabid, line 7: I would update this to use the description from the `HeaderSection` component (note: there are multiple header section components, I'm referring to the one under electron/views/automated-checks)

waabid (changes requested):
Minor changes, otherwise LGTM

src/electron/common/content-page-info-factory.ts
RobGallo: If this isn't there, we get an eslint error saying that other standard object properties can't be found.

waabid: Hm very odd.

waabid (approved):

Issue #3435, opened 2020-10-05
pass/fail radiobuttons in assessment don't show selected state (windows high contrast + edge)


karanbirsingh:
## Describe the bug

In Edge + Windows HC, the radio buttons in assessments are always white even if selected. It's therefore unclear what the state is. This issue only shows up when Windows high contrast mode is on. Toggling high-contrast mode from the extension itself (in both chrome and edge) produces the right behavior.

**To Reproduce**

1. Add Accessibility Insights for Web to Edge
2. Turn on Windows High Contrast mode
3. Open an assessment whose results have pass/fail radio buttons (e.g. links)
4. Notice the radio buttons don't show their state

## Screenshots

![screenshot of radio buttons in assessment](https://user-images.githubusercontent.com/7775527/95116505-fdd3b480-06fb-11eb-9718-b6837483a30b.png)

## Context (please complete the following information)

- OS Name & Version: Windows OS 19041.508
- AI-Web Version & Environment: 2.21.1 Production
- Browser Version: Microsoft Edge 86.0.622.31 (Official build) beta (64-bit)

dbjorge:
Potentially related fluentui issues/PRs:

* most similar issue, wontfixed: https://github.com/microsoft/fluentui/issues/13768
* similar issue, fixed in office-ui-fabric-react@v7.121.7: https://github.com/microsoft/fluentui/issues/13767
* PR that fixed 13767: https://github.com/microsoft/fluentui/pull/13846

dbjorge:
In the latest Fluent UI (7.144.2 as of writing), this issue is sort-of resolved; the choicegroup buttons do use our override colors in that version, but we're providing override colors that don't meet contrast requirements, so it's a different problem rather than a complete fix:

![image](https://user-images.githubusercontent.com/376284/95121777-8747ae80-071d-11eb-8ccb-f00248b09845.png)

(there are also a variety of other, unrelated high-contrast bugs introduced by upgrading from our current fabric version to fluentUI 7.144.2)

ferBonnin:
@iamrafan can we please take a look at the fluent UI update to see if it introduces other HC issues? @dbjorge was this pushed to playground?

dbjorge:
[...]

PR #3468, opened 2020-10-14
fix: use distinguishable colors for test status choice groups in OS high-contrast modes

dbjorge:
#### Description of changes

This PR addresses #3435, where in Windows high contrast mode, pass/fail radio buttons that were selected vs not were indistinguishable.

Per discussion with @ferBonnin, this intentionally causes the buttons to use the system-provided colors for selected buttons, which means they do not use different colors for pass vs fail like they do in our app-specific HC mode and in the default theme. Respecting the system theme is more important.

This makes use of some relatively new Edge features around forced-color mode (specifically, the `forced-color-adjust: none;` property). See discussion in #3435 for references.

Screenshots showing before on left and after on right in assorted HC setting modes:

**App HC off, Windows HC off (intentionally no change):**
image

**App HC on, Windows HC off (intentionally no change):**
image

**App HC off, Windows HC theme 1:**
image

**App HC off, Windows HC theme 2:**
image

**App HC off, Windows HC theme Black (motivating case):**
image

**App HC off, Windows HC theme White:**
image

#### Pull request checklist

- [x] Addresses an existing issue: #3435
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/test-status-choice-group.scss
+++ b/src/DetailsView/components/test-status-choice-group.scss
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-@import '../../common/styles/colors.scss';
+@import '../../common/styles/common.scss';
 
 .radio-button-group {
     float: left;
@@ -18,8 +18,8 @@
             margin-top: 0px !important;
         }
         .ms-ChoiceField-wrapper.is-inFocus .ms-ChoiceField-field {
-            @media screen and (-ms-high-contrast: active) {
-                outline: 3px highlighttext solid !important;
+            @include ms-high-contrast-override {
+                outline: 3px HighlightText solid !important;
             }
         }
         .ms-ChoiceField:nth-of-type(1) {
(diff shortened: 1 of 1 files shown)


src/DetailsView/components/test-status-choice-group.scss
```diff
@@ -18,8 +18,8 @@
             margin-top: 0px !important;
         }
         .ms-ChoiceField-wrapper.is-inFocus .ms-ChoiceField-field {
-            @media screen and (-ms-high-contrast: active) {
-                outline: 3px highlighttext solid !important;
+            @include ms-high-contrast-override {
```
dbjorge, line 21: This is just to make it more consistent; the updated `Highlight`/`HighlightText` capitalization is to match the CSS spec

waabid, line 21: Seems like we can replace this with the mixin in a few other places as well, maybe not in this PR.

waabid (approved):
LGTM

src/DetailsView/components/test-status-choice-group.scss
dbjorge: Yep, I have a different branch already set up with that

PR #3487, opened 2020-10-19
feat(android-nav-reflow): Plumb in narrow mode thresholds as dependencies

RobGallo:
#### Description of changes

This change is necessary so that web and unified can have differing narrow mode thresholds.

Kept a single set of thresholds for web instead of one per each view to make it easier to adjust globally and because it hasn't seemed to be a problem so far having one set.


src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
```diff
@@ -11,6 +11,10 @@ describe('Details View -> Assessment -> Reflow', () => {
     let browser: Browser;
     let detailsViewPage: DetailsViewPage;
     const height = 400;
+    const narrowModeThresholds: NarrowModeThresholds = {
```
waabid, line 14: Probably a good idea to link this with the source code, no?

src/tests/unit/tests/DetailsView/components/narrow-mode-detector.test.tsx
```diff
@@ -6,19 +6,29 @@ import {
...
     return <h1>Test component</h1>;
 });
 
+const narrowModeThresholds = {
+    collapseHeaderAndNavThreshold: 600,
+    collapseCommandBarThreshold: 960,
+};
+
+const getNarrowModeThresholds = (): NarrowModeThresholds => {
+    return narrowModeThresholds;
```
waabid, line 24: While this works out, we usually do use Mocks here. If we want to keep it this way, I would suggest naming them as mocks appropriately.

src/DetailsView/components/narrow-mode-thresholds.ts
```diff
@@ -0,0 +1,20 @@
...
+};
+
+export const getNarrowModeThresholdsForWeb = (): NarrowModeThresholds => {
+    return {
+        collapseHeaderAndNavThreshold: 600,
+        collapseCommandBarThreshold: 960,
+    };
+};
+
+export const getNarrowModeThresholdsForUnified = (): NarrowModeThresholds => {
```
waabid, line 15: nit: Might not be a bad idea to have this in separate file specifically inside the electron/unified file structure rather than the DetailsView.

waabid (changes requested):
Largely LGTM, some minor comments.

waabid (approved):

PR #3494, opened 2020-10-20
test(combined-report-ui): establish monorepo setup with new empty report-e2e-tests package

dbjorge:
#### Description of changes

This PR adds a new packages at `/packages/report-e2e-tests`. We'll be using this as part of the combined-report-ui feature to test report generation as it would work in real usage (ie, across package boundaries and with proper style embedding), but this also has the secondary benefit of setting up a lot of the infrastructure we'll be using for migrating the rest of the repo to the lerna/yarn workspaces monorepo format. Specifically, for generalized changes, it:

* Adds a new devDependency on lerna
* Establishes a `lerna.json` and a root level package.json `workspaces` property to establish the root of the repo as a yarn worktree/lerna monorepo (similar to the setup that service uses)
* Updates a variety of ignore/lint tool rules to ignore build output directories in subdirectories, not just the root
* Splits out the parts of `tsconfig.json` that will be shared among packages into `tsconfig.base.json` (the same layout that service uses)

Specific to the new `report-e2e-tests` package (empty with a placeholder test for now), it also:
* Creates the empty package, set up with typescript/jest/etc and a placeholder test
* Adds a root-level `yarn test:report:e2e` that runs the tests
* Adds a step to run the tests in CI/PR builds

Using yarn workspaces means that `yarn install` from the root like usual will cover all the requirements for installing the new package's dependencies. I used `lerna` to implement `yarn test:report:e2e` mostly as a POC (you could technically invoke `jest --projects packages/report-e2e-tests` and it would work equally well)

This PR **does not migrate /src to /packages/web/src** - instead, it leaves package.json to act as *both* the worktree root (the aggregator of workspaces) and *also* an individual workspace root representing web/unified/etc. This is supported by yarn, but not what we want to use long-term, since it's a bit confusing; this is an incremental step on the way towards using a layout closer to what service uses.

Out of scope for this PR:
* Migrating existing web code to `/packages/web`
* Migrating existing report package code to `/packages/report`
* Migrating the existing report integration tests to `/packages/report-e2e-tests`
* Extracting common Jest configuration to a `jest.config.base.js` like service does (this required a lot of changes, so I left it for a separate PR)

#### Pull request checklist

- [x] Addresses an existing issue: 1783601
- [x] Ran `yarn fastpass`


karanbirsingh (approved):

packages/report-e2e-tests/jest.config.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+const baseConfig = require('../../src/tests/unit/jest.config');
```
pownkel, line 3: Will we eventually have a separate jest.base.config, like tsconfig.base.json?

dbjorge: Yes! I actually tried to do that in this PR, but it required changing almost a dozen more files, so I noted it as out of scope and have a separate branch ready to go with that :)

pownkel (approved):

PR #3499, opened 2020-10-20
feat(combined-report-ui): Combined report html generator

pownkel:
#### Description of changes

Add CombinedResultsReport, CombinedReportHtmlGenerator, and CombinedReportSectionFactory. With this PR, the package will be able to generate a combined report containing the components that already exist and don't depend on the summary results data format (report header, footer, and details section):

minimal combined report

#### Pull request checklist

- [x] Addresses an existing issue: #1783608
- [x] Ran `yarn fastpass`


src/reports/package/reporter-factory.ts
```diff
@@ -130,9 +128,38 @@ const summaryResultsReportGenerator = (parameters: SummaryReportParameters) => {
     return new SummaryResultsReport(deps, parameters, toolData);
 };
 
+const combinedResultsReportGenerator = (parameters: CombinedReportParameters) => {
+    const { serviceName, axeVersion, userAgent } = parameters;
+
+    const toolData = createToolData(
+        serviceName,
+        '',
```
dbjorge, line 136: This feels a little off; can we pass a serviceVersion param along with the serviceName, rather than leaving this blank? Or if we do want to omit it, can we make it more clear in the typings (ie, allow `createToolData` to accept `null` explicitly, rather than using an empty string)?

(I do see that you're just matching what the other summary report is already doing, so feel free to defer this comment from the PR if you'd rather consider it a separate change)

src/reports/package/combined-results-report.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        const targetAppInfo = {
+            name: scanDetails.basePageTitle,
+            url: scanDetails.baseUrl,
+        };
+
+        const scanMetadata: ScanMetadata = {
+            targetAppInfo: targetAppInfo,
+            toolData: this.toolInfo,
+            timestamp: null,
```
dbjorge, line 30: It feels strange/redundant to set a `null` timestamp and then pass a second similar `timespan` parameter instead. Probably not a change for this PR, but I feel like it would be nice to combine those two (maybe replace `scanMetadata.timestamp` by putting the `timespan` in there instead and update the current `timestamp` usages to use `scanComplete` instead)

dbjorge (approved):
This looks good to me! Both of my comments are things that would be reasonable to defer to other PRs, so approving and leaving to your discretion whether to defer them or add them here.

PR #3500, opened 2020-10-20
feat(combined-report-ui): Add URL card row for future use

jalkire:
#### Description of changes
The new combined report UI needs a card row that displays one or more URLs. This PR introduces, but does not put to use, such a card row. 

Example of the new URL row in use (with dummy data; will not appear in fastpass like this):
image

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/components/cards/urls-card-row.tsx
+++ b/src/common/components/cards/urls-card-row.tsx
@@ -6,6 +6,7 @@ import * as React from 'react';
 import { CardRowProps } from 'common/configs/unified-result-property-configurations';
 import { NamedFC } from 'common/react/named-fc';
 import { SimpleCardRow } from './simple-card-row';
+import * as styles from './urls-card-row.scss';
 
 export interface UrlsCardRowProps extends CardRowProps {
     propertyData: string[];
@@ -14,13 +15,13 @@ export interface UrlsCardRowProps extends CardRowProps {
 export const UrlsCardRow = NamedFC<UrlsCardRowProps>('UrlsCardRow', ({ deps, ...props }) => {
     const renderUrlContent = () => {
         return (
-            <div>
+            <ul class={styles.urlsRowContent}>
                 {props.propertyData.map((url, index) => (
-                    <div key={`urls-${index}`}>
+                    <li key={`urls-${index}`}>
                         <NewTabLink href={url}>{url}</NewTabLink>
-                    </div>
+                    </li>
                 ))}
-            </div>
+            </ul>
         );
     };
 
--- a/src/common/components/cards/urls-card-row.tsx
+++ b/src/common/components/cards/urls-card-row.tsx
@@ -1,14 +1,19 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { NewTabLink } from 'common/components/new-tab-link';
 import * as React from 'react';
 
-import { CardRowProps } from 'common/configs/unified-result-property-configurations';
+import { CardRowDeps, CardRowProps } from 'common/configs/unified-result-property-configurations';
 import { NamedFC } from 'common/react/named-fc';
+import { LinkComponentType } from 'common/types/link-component-type';
 import { SimpleCardRow } from './simple-card-row';
 import * as styles from './urls-card-row.scss';
 
+export type UrlsCardRowDeps = CardRowDeps & {
+    LinkComponent: LinkComponentType;
+};
+
 export interface UrlsCardRowProps extends CardRowProps {
+    deps: UrlsCardRowDeps;
     propertyData: string[];
 }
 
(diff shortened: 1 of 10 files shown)


src/common/components/cards/urls-card-row.tsx
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { NewTabLink } from 'common/components/new-tab-link';
```
dbjorge, line 3: I think we'll need to use a `NewTabLinkWithConfirmationDialog` instead for the reports. Would probably be reasonable to have this file be agnostic with a `LinkComponent` dep.

src/common/components/cards/urls-card-row.tsx
```diff
@@ -0,0 +1,34 @@
...
+export interface UrlsCardRowProps extends CardRowProps {
+    propertyData: string[];
+}
+
+export const UrlsCardRow = NamedFC<UrlsCardRowProps>('UrlsCardRow', ({ deps, ...props }) => {
+    const renderUrlContent = () => {
+        return (
+            <div>
+                {props.propertyData.map((url, index) => (
+                    <div key={`urls-${index}`}>
```
dbjorge, line 19: I wonder if it would be better to use a `<ul>` of URLs; even if we aren't styling it with bullet points, it feels semantically like a list to me. @ferBonnin, thoughts?

src/common/components/cards/urls-card-row.tsx
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License.
+import { NewTabLink } from 'common/components/new-tab-link';
+import * as React from 'react';
+
+import { CardRowProps } from 'common/configs/unified-result-property-configurations';
+import { NamedFC } from 'common/react/named-fc';
+import { SimpleCardRow } from './simple-card-row';
+
+export interface UrlsCardRowProps extends CardRowProps {
+    propertyData: string[];
```
dbjorge, line 11: I know you can't change the name because it's set by the more general `CardRowProps` type, but I wonder if it might be helpful to clarify here exactly what this is supposed to contain (either with a comment or something like `propertyData: { urls: string[] }`

src/common/components/cards/urls-card-row.tsx
ferBonnin: I like that suggestion, it would make AT navigation and overall readability easier.

dbjorge (changes requested):
Looks good overall, a few minor comments + a question for Fer

src/common/components/cards/urls-card-row.tsx
jalkire: A <ul> works for me, I didn't use one because bullet points were not part of the design. To clarify @ferBonnin, do you still want this visually the same (no bullets)?

ferBonnin: lets add visuall bullet points as well, please; I believe it would help for readability

src/common/components/cards/urls-card-row.tsx
jalkire: I punted on creating a more descriptive propertyData Type in this PR in case the future work to integrate this lends itself better to nesting the array of URLs or not. Based on your comment, I'd prefer to just create the new type now vs adding a comment.

src/common/components/cards/urls-card-row.tsx
jalkire: Now uses `LinkComponent` dep.

src/common/components/cards/urls-card-row.tsx
jalkire: Now a bulleted `<ul>`! See screenshot.

src/common/components/cards/urls-card-row.tsx
jalkire: Introduced `UrlsPropertyData` to improve clarity.

src/common/components/cards/urls-card-row.tsx
```diff
@@ -1,26 +1,36 @@
...
 
-import { CardRowProps } from 'common/configs/unified-result-property-configurations';
+import { CardRowDeps, CardRowProps } from 'common/configs/unified-result-property-configurations';
 import { NamedFC } from 'common/react/named-fc';
+import { LinkComponentType } from 'common/types/link-component-type';
 import { SimpleCardRow } from './simple-card-row';
+import * as styles from './urls-card-row.scss';
+
+export type UrlsCardRowDeps = CardRowDeps & {
+    LinkComponent: LinkComponentType;
```
dbjorge, line 12: I think this is actually going to need to go into `CardRowDeps`, since that's what the parent container components are going to be composing with; the individualized rows contain their own type identifier and propertyData, so those can be coupled here, but the deps come from the parent container independent of knowing the row types

dbjorge (changes requested):
1 minor change in updated code, otherwise looks good!

dbjorge (approved):

PR #3515, opened 2020-10-22
feat(android-nav-reflow): Use ReflowCommandBar when narrow mode is enabled

RobGallo:
#### Description of changes

This change makes it so that when the window width is less than the narrow mode threshold for the command bar, the ReflowCommandBar component will be rendered instead of the standard CommandBar component.

The change only applies of course in the case where the feature flag is active.

### DIFF
--- a/src/electron/views/automated-checks/automated-checks-view.tsx
+++ b/src/electron/views/automated-checks/automated-checks-view.tsx
@@ -20,6 +20,7 @@ import {
 } from 'DetailsView/components/details-view-overlay/settings-panel/settings-panel';
 import { NarrowModeStatus } from 'DetailsView/components/narrow-mode-detector';
 import { UnifiedFeatureFlags } from 'electron/common/unified-feature-flags';
+import { LeftNavActionCreator } from 'electron/flux/action-creator/left-nav-action-creator';
 import { ScanActionCreator } from 'electron/flux/action-creator/scan-action-creator';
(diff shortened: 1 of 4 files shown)


src/electron/views/automated-checks/automated-checks-view.tsx
```diff
@@ -216,6 +219,34 @@ export class AutomatedChecksView extends React.Component<AutomatedChecksViewProp
...
+                setSideNavOpen={this.props.deps.leftNavActionCreator.setLeftNavVisible}
+            />
+        );
+    }
+
+    private renderCommandBar(
+        cardsViewData: CardsViewModel,
+        scanMetadata: ScanMetadata,
+    ): JSX.Element {
+        return this.props.narrowModeStatus.isCommandBarCollapsed
```
waabid, line 245: You shouldn't need this here since the reflow-command-bar will handle it.

src/electron/views/automated-checks/automated-checks-view.tsx
```diff
@@ -216,6 +219,34 @@ export class AutomatedChecksView extends React.Component<AutomatedChecksViewProp
         );
     }
 
+    private renderReflowCommandBar(
```
waabid, line 222: This method should be called on line 258 where we are showing the component that is enabled when the feature flag is enabled.

waabid (approved):

PR #3544, opened 2020-10-27
feat(android-nav-reflow): Add e2e tests

RobGallo:
#### Description of changes

Added tests to check the following scenarios

- When the window width is greater than the narrow mode threshold for unified, the hamburger button is not shown in the command bar
- When the window width is less than the narrow mode threshold for unified,
	- the hamburger button is shown in the command bar
	- the left nav bar is not initially visible
	- when the hamburger button is clicked, the left nav pane becomes visible
	- if the hamburger button is clicked again, the left nav pane disappears
	- if a link in the left nav pane is clicked, the left nav pane disappears

### DIFF
--- a/src/tests/electron/tests/automated-checks-view.test.ts
+++ b/src/tests/electron/tests/automated-checks-view.test.ts
@@ -234,41 +234,40 @@ describe('AutomatedChecksView', () => {
         expect(result.value).toBeNull();
     });
 
-    it('hamburger button click opens and closes left nav', async () => {
-        await setupWindowForCommandBarReflowTest(2);
+    const waitForFluentLeftNavToDisappear = async (): Promise<void> => {
+        return automatedChecksView.waitForSelectorToDisappear(
+            AutomatedChecksViewSelectors.fluentLeftNav,
+        );
+    };
 
+    const expectFluentLeftNavNotToBeRendered = async (): Promise<void> => {
         const result = await automatedChecksView.client.$(
             AutomatedChecksViewSelectors.fluentLeftNav,
         );
         expect(result.state).toBe('failure');
+    };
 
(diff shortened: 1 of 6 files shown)


src/tests/electron/tests/automated-checks-view.test.ts
```diff
@@ -207,4 +208,67 @@ describe('AutomatedChecksView', () => {
...
+            AutomatedChecksViewSelectors.fluentLeftNav,
+        );
+        expect(result.state).toBe('failure');
+
+        await automatedChecksView.client.click(AutomatedChecksViewSelectors.leftNavHamburgerButton);
+
+        // if the nav bar doesn't appear, the timeout has expired and the test fails.
+        await automatedChecksView.waitForSelector(AutomatedChecksViewSelectors.fluentLeftNav);
+
+        // The fluent left nav bar appears over the command bar
```
waabid, line 250: I would write this as:

`Specifically clicks the hamburger button inside the fluent left nav (as there is one within the command bar as well)`

src/tests/electron/tests/automated-checks-view.test.ts
```diff
@@ -207,4 +208,67 @@ describe('AutomatedChecksView', () => {
...
+        // so it requires a different selector
+        const selector = `${AutomatedChecksViewSelectors.fluentLeftNav} ${AutomatedChecksViewSelectors.leftNavHamburgerButton}`;
+        await automatedChecksView.client.click(selector);
+
+        // if the nav bar doesn't disappear, the timeout has expired and the test fails.
```
waabid, line 255: nit: I think the comment here feels a bit unnecessary (and below). Using an explicit expect statement instead or wrapping within a private method would be a better approach in my opinion.

waabid (approved):

PR #3545, opened 2020-10-27
feat(combined-report-ui): Introduce CollapsibleFailedResultsSection to contain Combined Results

jalkire:
#### Description of changes
Currently, the `CombinedReportSectionFactory` does not provide a value for `FailedInstancesSection`. This PR introduces `CollapsibleFailedResultsSection` to act as the combined report's `FailedInstancesSection`. This is distinct from the existing `FailedInstancesSection` because it is collapsible and from the `CollapsibleResultSection` because it displays full result cards. A future PR will update the header to match the spec.

Expanded:
image
Collapsed:
image

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/reports/package/combined-results-to-cards-model-converter.ts
```diff
@@ -77,10 +77,10 @@ export class CombinedResultsToCardsModelConverter {
             status: 'fail',
             ruleId: rule.ruleId,
             identifiers: {
+                urls: { urls: failureData.urls },
```
jalkire, line 80: I moved urls because the order of the properties here determines the order they appear in card rows, and urls should be first

pownkel (approved):
I notice a few things off from the figma (the HowToFix row is missing, and the failure count for each rule shows the count of cards, not the count of URLs), but those might take more significant refactoring so it might be preferred to address them in separate PRs instead. Other than that, LGTM.

jalkire:
> I notice a few things off from the figma (the HowToFix row is missing, and the failure count for each rule shows the count of cards, not the count of URLs), but those might take more significant refactoring so it might be preferred to address them in separate PRs instead. Other than that, LGTM.

Good catch, will have a followup PR for those!

PR #3549, opened 2020-10-28
feat(combined-report-ui): add "How to fix" card row

pownkel:
#### Description of changes

Add missing "how to fix" card row to combined report. This involves changing the input interface to accept the resolution data that we get from axe-core.

howToFixCard

#### Pull request checklist

- [x] Addresses an existing issue: #1772544
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/package/accessibilityInsightsReport.d.ts
+++ b/src/reports/package/accessibilityInsightsReport.d.ts
@@ -61,10 +61,16 @@ declare namespace AccessibilityInsightsReport {
         ruleUrl: string
     }
 
+    export interface FormattedCheckResult {
+        id: string;
+        message: string;
+        data: any;
+    }
+
     export type HowToFixData = {
-        any: axe.CheckResult[],
-        all: axe.CheckResult[],
-        none: axe.CheckResult[],
+        any: FormattedCheckResult[],
+        all: FormattedCheckResult[],
+        none: FormattedCheckResult[],
         failureSummary?: string,
     }
 
--- a/src/reports/package/accessibilityInsightsReport.d.ts
+++ b/src/reports/package/accessibilityInsightsReport.d.ts
@@ -62,9 +62,9 @@ declare namespace AccessibilityInsightsReport {
     }
 
     export type HowToFixData = {
-        any: FormattedCheckResult[],
-        all: FormattedCheckResult[],
-        none: FormattedCheckResult[],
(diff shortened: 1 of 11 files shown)


JGibson2019:
Figma clarifications:

- Should there be an unordered list in the URL row?

- Noticed that there are grey lines underneath each row; should those be removed to match the Figma? 
image

dbjorge:
> Should there be an unordered list in the URL row?
This is intentional, per earlier discussion with @ferBonnin 

> Noticed that there are grey lines underneath each row; should those be removed to match the Figma?
The gray separators are consistent with all other card views; I think we should leave them for consistency, but good to double check with @ferBonnin

JGibson2019 (commented):
LGTM, but wanted to check on the styling things before fully approving

JGibson2019:
> > Should there be an unordered list in the URL row?
> 
> This is intentional, per earlier discussion with @ferBonnin
> 
> > Noticed that there are grey lines underneath each row; should those be removed to match the Figma?
> 
> The gray separators are consistent with all other card views; I think we should leave them for consistency, but good to double check with @ferBonnin

Re 1 - Thank you! 
Re 2 - I'm also a fan of keeping them for consistency's sake

ferBonnin:
> > Noticed that there are grey lines underneath each row; should those be removed to match the Figma?
> 
> The gray separators are consistent with all other card views; I think we should leave them for consistency, but good to double check with @ferBonnin


yes, we should leave the grey separators for consistency.

JGibson2019 (approved):

src/reports/package/accessibilityInsightsReport.d.ts
```diff
@@ -61,11 +61,18 @@ declare namespace AccessibilityInsightsReport {
         ruleUrl: string
     }
 
+    export type HowToFixData = {
+        any: FormattedCheckResult[],
```
dbjorge, line 65: `FormattedCheckResult` doesn't seem to be defined or explicitly imported from this file (I think it's coming implicitly from `scanner.d.ts`, so I'm not sure it's going to be usable for a user of the package. It's a pretty short declaration so I'd lean towards just repeating it here.

dbjorge (changes requested):
One suggestion, looks good overall!

dbjorge (approved):

PR #3600, opened 2020-11-02
chore(lint): part 1 of reenabling no-floating-promises

dbjorge:
#### Description of changes

When we transitioned from tslint to eslint, we disabled the no-floating-promises rule due to high false positives. This PR is part 1 of working towards re-enabling it; it fixes about half the issues it catches.

It mostly does this by adding `.catch(logger.error)` behavior to floating cases; that essentially matches the behavior in most browsers from leaving them floating, so it shouldn't have a behavioral impact.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/common/fail-test-on-error-logger.ts
+++ b/src/tests/unit/common/fail-test-on-error-logger.ts
@@ -8,6 +8,8 @@ export const failTestOnErrorLogger: Logger = {
         // intentionally ignored
     },
     error: (message?: string, ...optionalParams: any[]) => {
-        fail(`FailTestOnErrorLogger.error invoked with "${message}" ${inspect(optionalParams)}`);
+        expect(
+            `FailTestOnErrorLogger.error invoked with "${message}" ${inspect(optionalParams)}`,
+        ).toBe('FailTestOnErrorLogger.error never invoked');
     },
 };
--- a/src/tests/unit/common/fail-test-on-error-logger.ts
+++ b/src/tests/unit/common/fail-test-on-error-logger.ts
@@ -1,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { inspect } from 'util';
+import { Logger } from 'common/logging/logger';
+
+export const failTestOnErrorLogger: Logger = {
+    log: () => {
+        // intentionally ignored
+    },
+    error: (message?: string, ...optionalParams: any[]) => {
+        fail(`FailTestOnErrorLogger.error invoked with "${message}" ${inspect(optionalParams)}`);
+    },
+};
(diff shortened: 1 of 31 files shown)


src/tests/unit/common/fail-test-on-error-logger.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { inspect } from 'util';
```
karanbirsingh, line 3: Can we remove the inspect import?

src/tests/unit/common/fail-test-on-error-logger.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Logger } from 'common/logging/logger';
+
+export const failTestOnErrorLogger: Logger = {
+    log: () => {
+        // intentionally ignored
+    },
+    error: (message?: string, ...optionalParams: any[]) => {
+        fail(`FailTestOnErrorLogger.error invoked with "${message}" ${inspect(optionalParams)}`);
```
karanbirsingh, line 11: I looked for this and found https://github.com/facebook/jest/issues/2129 mentioning it's a feature of jasmine, but I'm not sure if I'm missing something related to jest-circus that will help us use it. I added `fail` locally to a test and it failed because fail() was not defined. Is there an API I'm missing, or should we do this a different way?

src/tests/unit/common/fail-test-on-error-logger.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { inspect } from 'util';
+import { Logger } from 'common/logging/logger';
+
+export const failTestOnErrorLogger: Logger = {
```
karanbirsingh, line 6: You might have already handled it in part 2, I think https://github.com/microsoft/accessibility-insights-web/blob/ded30591606aeed2357977f33f38d28f7ce9f75a/src/tests/unit/tests/electron/platform/android/setup/state-machine/state-machine.test.ts#L210 will need to be updated to work with this

karanbirsingh (changes requested):
Looks good overall, I commented a couple questions

src/tests/unit/common/fail-test-on-error-logger.ts
dbjorge: Huh; I had used this based on this from `@types/jest/index.d.ts`:

```ts
/**
 * Fails a test when called within one.
 */
declare function fail(error?: any): never;
```

Based on that thread, I'll switch it to something based on `expect()`. Great catch!

src/tests/unit/common/fail-test-on-error-logger.ts
dbjorge: Yes, I think you'd want to add a `noOpLogger` instead if you wanted something that didn't fail the test on errors for tests that intentionally introduce errors, or probably better, a mockLogger or a recordingLogger or something to validate what errors get logged

karanbirsingh (approved):

PR #3626, opened 2020-11-06
feat(android-needs-review): Add initial implementation of needs review view e2e tests

RobGallo:
#### Description of changes

This is a small implementation. There's only one test currently. But this should serve as the beginning of the e2e tests once we know what else needs to be tested.


src/tests/electron/common/view-controllers/left-nav-controller.ts
```diff
@@ -0,0 +1,27 @@
...
+        await this.clickLeftNavItem(needsReviewTestConfig);
+    }
+
+    private async clickLeftNavItem(config: TestConfig): Promise<void> {
+        const selector = this.getSelectorForLeftNavItemLink(config);
+        await this.client.click(selector);
+    }
+
+    private getSelectorForLeftNavItemLink(config: TestConfig): string {
+        return `${this.leftNavSelector} li [name="${config.contentPageInfo.title}"] a`;
```
dbjorge, line 25: Like you mentioned in the teams chat, I think it would be preferable to be able to use a data automation selector here for the link itself. I didn't see an obvious way to add an automation id to the `<a>` tag, but it looks like we're already adding them to immediately children of the link (see `TestViewLeftNavLink` along with usage in `AutomatedChecksViewSelectors.nthTestInLeftNav`), which does what we want when clicked; I'd recommend using that (probably updating `TestViewLeftNavLink` to generate `data-automation-id` based on `link.key`, rather than using the same one for each nav link)

dbjorge (changes requested):
One suggested change, otherwise LGTM!

src/tests/electron/common/view-controllers/left-nav-controller.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    constructor(private readonly client: SpectronAsyncClient) {}
+
+    public async clickNeedsReview(): Promise<void> {
+        await this.clickLeftNavItem(needsReviewTestConfig);
+    }
+
+    private async clickLeftNavItem(config: TestConfig): Promise<void> {
+        const selector = this.getSelectorForLeftNavItemLink(config);
+        await this.client.click(selector);
```
dbjorge, line 21: It looks like there's some flakiness happening here in the e2e tests at this point, where the tests are trying to click on a selector that doesn't exist yet. I think it would probably help to include a `this.waitForSelector(selector)` before the click.

dbjorge, line 21: This should go through `this.click` rather than `this.client.click` (so we get screenshots when the click fails)

src/tests/electron/common/view-controllers/automated-checks-view-controller.ts
```diff
@@ -64,4 +65,14 @@ export class AutomatedChecksViewController extends ViewController {
 
         await this.waitForSelector(toggleInStateSelector);
     }
+
+    public async clickLeftNavItem(key: LeftNavItemKey): Promise<void> {
+        const selector = this.getSelectorForLeftNavItemLink(key);
+        await this.waitForSelector(selector);
+        await this.client.click(selector);
```
dbjorge, line 72: Would be preferable to use `this.click` here instead of `this.client.click`; that way it'll get wrapped in `screenshotOnError` for you

dbjorge (approved):

PR #3638, opened 2020-11-10
feat:(android-needs-review): Don't show the export report button when needs review is selected

RobGallo:
#### Description of changes

- Add `allowsExportReport` property to `ContentPageInfo` so whether or not the export report button is shown can be specified as part of a test config
- Modify `ReflowCommandBar` to respect the `allowsExportReport`  setting
- Add e2e tests to confirm the export report button is shown for automated checks, but not for needs review


dbjorge (approved):
LGTM!

PR #3633, opened 2020-11-10
feat(android-needs-review): Filter GetCardSelectionViewData based on TestConfig

jalkire:
#### Description of changes
Currently, `GetCardSelectionViewData` assumes that all results are applicable. This works in the web extension because we do not simultaneously hold failed and needs review results in the store. Now that we are adding needs review results to unified, that assumption is no longer valid, as we do not want to require the user to rescan their device when changing between Automated Checks and Needs Review. 

To accommodate this scenario, this PR introduces the concept/type of a `ResultsFilter` and uses it to filter the selection data returned by `GetCardSelectionViewData`. This is done by adding a `ResultsFilter` to the appropriate unified `TestConfig` files. No filtering is required on the web side at this time.

The immediate user impact of this PR should be minimal--only the unified Needs Review tab's experience is impacted (Automated Check instances' highlights no longer show up).

Out of scope: 
- In the future, we may want to consider using the same `ResultsFilter` to filter  `CardsViewData` as well
- The unified Needs Review tab will still not display color contrast results

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #3639, opened 2020-11-10
feat(android-needs-review): Include Android ColorContrast results in unified results as status unknown

jalkire:
#### Description of changes
For the new Needs Review display in unified, we want the following behavior:
- ColorContrast scan results with high confidence display in Needs Review
- No change to what displays in Automated Checks

Currently `RuleInformationProvider` and its usage results in the following:
- ColorContrast scan results are excluded from unified results
- Other scan results are included in unified results with their original status

This PR achieves the desired behavior by updating `RuleInformationProvider` and its usage as follows:
- ColorContrast scan results with a high confidence and status fail are included in unified results with status unknown
- Other ColorContrast scan results are included in unified results with status pass
- Other scan results with status fail are included in unified results with status fail
- Other scan results are included in unified results with status pass

With these changes, the status of unified results matches how we display results to users: Failing results are displayed in Automated Checks, unknown results are displayed in Needs Review, and passing results are not displayed.

Note that I noticed some issues with the highlighting behavior from #3633, which a future PR will fix.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/platform/android/rule-information-provider.ts
```diff
@@ -117,13 +120,25 @@ export class RuleInformationProvider {
...
     };
 
-    private includeAllResults = (ruleResultsData: RuleResultsData): boolean => {
-        return true;
+    private getStandardResultStatus = (ruleResultsData: RuleResultsData): InstanceResultStatus => {
+        switch (ruleResultsData.status) {
+            case 'FAIL':
+                return 'fail';
+            default:
+                return 'pass';
```
dbjorge, line 140: The previous behavior from `scan-results-to-unified-results` treated non-PASS non-FAIL as unknown, rather than pass, which sounds more reasonable to me; why the behavior change?

src/electron/platform/android/rule-information-provider.ts
```diff
@@ -117,13 +120,25 @@ export class RuleInformationProvider {
...
+    private getColorContrastResultStatus = (
+        ruleResultsData: RuleResultsData,
+    ): InstanceResultStatus => {
+        if (
+            ruleResultsData.status == 'FAIL' &&
+            ruleResultsData.props['Confidence in Color Detection'] === 'High'
+        )
+            return 'unknown';
+
+        return 'pass';
```
dbjorge, line 132: I feel like this creates a somewhat misleading data model; I would feel more inclined to come up with a way to omit the low-confidence cases, instead (maybe allow these functions to return `InstanceResultStatus | 'omit'`, say)

src/tests/unit/tests/electron/platform/android/scan-results-helpers.ts
```diff
@@ -129,12 +128,12 @@ export function buildRuleInformation(
...
         },
-        includeThisResultDelegate: r => {
+        getResultStatusDelegate: r => {
             expect('includeThisResultDelegate').toBe('This code should never execute');
             return null;
         },
-        includeThisResult: r => {
-            return includeResults;
+        getResultStatus: r => {
+            return r.status;
```
dbjorge, line 136: This created snapshots which made me concerned there was an implementation bug, since r.status isn't the expected return type of getResultStatus; can this maybe do something like `r.status === 'PASS' ? 'pass' : 'unknown'` instead?

dbjorge (changes requested):
Looks good overall, a few suggestions inline!

src/electron/platform/android/rule-information-provider.ts
jalkire: Unknown results are displayed in the Needs Review tab, and we only want to display Color Contrast results in the Needs Review tab. To accomplish the desired behavior, we have two options:
1. Only mark the appropriate CC results as unknown
2. Update the Needs Review display behavior to display results based on something other than status unknown

I can see the value of option 2, but choosing it raises the question in my mind: what is the practical meaning of status unknown if not to indicate that a result needs review? I couldn't come up with a good answer to that question, so I went with option 1. My understanding of the values of `InstanceResultStatus` is:
- `fail`: we believe the result indicates an accessibility issue, so we display it in Automated Checks
- `unknown`: we believe the result is likely to indicate an accessibility issue, so we display it in Needs Review
- `pass`: we do not believe the result indicates an accessibility issue, so we do not display it

I am certainly open to another approach if I'm missing something about what unknown should indicate!

src/electron/platform/android/rule-information-provider.ts
jalkire: I think my other comment largely addresses my thinking here, as well. I could see the benefit of an omit option here, but the need for it makes me think I may be thinking of the fail/unknown/pass distinction incorrectly

src/electron/platform/android/rule-information-provider.ts
dbjorge: That's fair enough; assuming PM confirms that we do want to suppress unknown cases for other rules, I'd be okay with sticking with this. It might be nice to add some comments on the definition of the `InstanceResultStatus` type to clarify the intended meaning (particularly, that "pass" may include results which are very low-confidence failures, in addition to high-confidence non-failures)

src/tests/unit/tests/electron/platform/android/scan-results-helpers.ts
jalkire: Updated! I used a slightly different conditional that is closer to what happens in prod code

src/electron/platform/android/rule-information-provider.ts
jalkire: Added comment as suggested!

dbjorge (approved):

PR #3646, opened 2020-11-12
feat(android-needs-review): Fix Android Needs Review screenshot highlighting behavior

jalkire:
#### Description of changes
PR #3633 introduced the concept of a `ResultsFilter` and used it to ensure that the proper elements are highlighted in Automated Checks and Needs Review when all cards are collapsed. Unfortunately, it failed to address the scenario where some cards are expanded or selected. This PR addresses those scenarios by ensuring that we always limit our `resultsHighlightStatus` data to reflect the currently selected tab's results.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/get-card-selection-view-data.ts
+++ b/src/common/get-card-selection-view-data.ts
@@ -46,10 +46,9 @@ export const getCardSelectionViewData: GetCardSelectionViewData = (
     viewData.visualHelperEnabled = cardSelectionStoreData.visualHelperEnabled || false;
     viewData.expandedRuleIds = getRuleIdsOfExpandedRules(cardSelectionStoreData.rules);
 
-    const allResultUids = getAllResultUids(
-        unifiedScanResultStoreData.results.filter(resultsFilter),
-        cardSelectionStoreData.rules,
-    );
+    const filteredResults = unifiedScanResultStoreData.results.filter(resultsFilter);
+    const filteredResultUids = filteredResults.map(res => res.uid);
+    const allResultUids = getAllResultUids(filteredResultUids, cardSelectionStoreData.rules);
 
     const unavailableResultUids = getResultsWithUnavailableHighlightStatus(
         unifiedScanResultStoreData,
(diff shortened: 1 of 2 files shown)


src/common/get-card-selection-view-data.ts
```diff
@@ -46,10 +46,9 @@ export const getCardSelectionViewData: GetCardSelectionViewData = (
     viewData.visualHelperEnabled = cardSelectionStoreData.visualHelperEnabled || false;
     viewData.expandedRuleIds = getRuleIdsOfExpandedRules(cardSelectionStoreData.rules);
 
-    const allResultUids = getAllResultUids(
-        unifiedScanResultStoreData.results.filter(resultsFilter),
-        cardSelectionStoreData.rules,
-    );
+    const filteredResults = unifiedScanResultStoreData.results.filter(resultsFilter);
+    const filteredResultUids = filteredResults.map(res => res.uid);
+    const allResultUids = getAllResultUids(filteredResultUids, cardSelectionStoreData.rules);
```
dbjorge, line 51: I got confused by these names; my expectation would be that `filteredResultUids` is a subset of `allResultUids`, not the other way around.

dbjorge: I'm not sure I'd expect there to be any reason for the rest of the implementation to actually use `filteredResults` instead of the version also filtered by the selectionStore rules, so maybe just combine these lines into something like `candidateResults` or `allCandidateResults`?

dbjorge (changes requested):
Besides the two suggestions inline, it looks like `getResultsWithUnavailableHighlightStatus` is still using the store results directly without filtering; it might be more clear/less error-prone to have it work in terms of the pre-filtered candidate results

src/common/get-card-selection-view-data.ts
jalkire: Turns out I needed what was `filteredResults` to do your suggested change to `getResultsWithUnavailableHighlightStatus`, so I kept that. I like the candidate name idea instead of filtered, so I changed to that!

src/common/get-card-selection-view-data.ts
```diff
@@ -120,23 +121,24 @@ function getRuleIdsOfExpandedRules(ruleDictionary: RuleExpandCollapseDataDiction
 }
 
 function getAllResultUids(
```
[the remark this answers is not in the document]
jalkire: Updated name to `getAllFilteredUids`. Now that I'm using `candidateResultUids`, this seemed like a more accurate place to describe the results as "filtered"

dbjorge (approved):

PR #3672, opened 2020-11-18
feat(android-needs-review): Remove LeftNavBar feature flag

RobGallo:
#### Description of changes

This feature represents the last of three features where we have added the left nav bar and needs review content. So we are removing the feature flag so that we can release the change.


src/electron/views/results/components/command-bar.tsx
```diff
@@ -1,116 +0,0 @@
...
-} from 'DetailsView/components/report-export-component';
-import { UnifiedFeatureFlags } from 'electron/common/unified-feature-flags';
-import { ScanActionCreator } from 'electron/flux/action-creator/scan-action-creator';
-import { ScanStatus } from 'electron/flux/types/scan-status';
-import { ScanStoreData } from 'electron/flux/types/scan-store-data';
-import * as React from 'react';
-import { ReportGenerator } from 'reports/report-generator';
-
-import { InsightsCommandButton } from 'common/components/controls/insights-command-button';
-import * as styles from './command-bar.scss';
```
dbjorge, line 21: Can `command-bar.scss` be removed as well?

dbjorge: On second look, it's reused by `reflow-command-bar.tsx`, so probably better to rename it to match the remaining file than to delete it

dbjorge (approved):
Looks good to me besides 1 minor rename suggestion!

Issue #653, opened 2019-05-06
Provide users with ability to import/export an assessment as data (e.g. json)?


mmafe:
Currently I can only work on a single assessment at a time. If I want to start a new assessment, I'm forced to lose my previous work. I am also unable to edit a previously exported assessment once I've started a new one.

# My Request
In addition, I would love to be able to:
- export an assessment as data (e.g. json file)
- re-import the data of that assessment

This would allow me to:
- simultaneously work on multiple assessments
- edit a previously exported assessment without having to redo the whole thing from scratch

nschonni:
This may be applicable if you start to look at output formats https://w3c.github.io/wcag-act/act-rules-format.html#appendix-data-example

scottsb:
Here is a more detailed explanation of the same approach @nschonni mentions. It would be amazing if this tool could generate interchangeable data files with the W3C's own (sadly poorly maintained) tool:
https://github.com/w3c/wcag-em-report-tool/blob/master/docs/EARL+JSON-LD.md

ferBonnin:
Hey @mmafe, thanks for your interest in Accessibility Insights for Web!

This is a request that requires more discussion and investigation on our side, we will update here when we know more and we will keep checking for upvotes/reactions from the community on this request

adbajaj:
This would be very helpful for my team as well! Would be interested to know ball park timings on when this is coming.
Thanks,
Aditya

mgifford:
Just wanting to check in on the status of this. Love the idea of being able to take this data and just plug it into W3C's WCAG-EM Report Tool /c @scottsb @nschonni

nitedog:
@ferBonnin I'd be happy to engage in such a discussion on interchangeable formats.

mgifford:
@WilcoFiers let me know that you should be able to do this in axe with https://github.com/dequelabs/axe-reporter-earl

In which case, it should be possible with Accessibility Insights.

ferBonnin:
Hi Everyone, thanks for the additional inputs on this feature request! This last comments seem to be focused around the ability to export the assessment as data.

To help us understand better the requirements, can you share a little bit about:
-  the scenarios where having the assessment as data would be helpful / how would you use this data? 
- tell us a bit more about what formats would be helpful and why?

mmafe:
@ferBonnin Thanks for looking into this!

I'd find an export of the assessment as data helpful if I need to:
- update the report (e.g. if I find a new issue)
[...]

PR #3676, opened 2020-11-18
feat(save-and-load-assessment): add versioning 

shanisebarona:
#### Description of changes
This PR adds versioning, formatting, and a11y extension to the assessment data to be downloaded.

#### Pull request checklist

- [x] Addresses an existing issue: Feature request #653
- [x] Ran `yarn fastpass`


src/DetailsView/components/save-assessment-factory.tsx
```diff
@@ -20,9 +22,15 @@ export function getSaveButtonForAssessment(props: SaveAssessmentFactoryProps): J
     const assessmentData = props.deps.assessmentDataFormatter.formatAssessmentData(
         props.assessmentStoreData.assessments,
     );
+
+    const currentDate = new Date();
+    const fileDate = props.deps.fileNameBuilder.getDateSegment(currentDate);
+    const targetPageTitle = props.assessmentStoreData.persistedTabInfo.title;
+    const fileTitle = props.deps.fileNameBuilder.getTitleSegment(targetPageTitle);
+    const fileName = `SavedAssessment_<${fileDate}>_${fileTitle}.a11yextension`;
```
dbjorge, line 30: `a11yextension` matches the spec, but as I think about this, I think we might want to update the spec here; I think it would makes sense for us to use something that's a little more specific to our product, since we don't intend to support this format for use with other a11y tools besides AI4Web. @ferBonnin, thoughts on using something a little more specific? Maybe `.aiwebassessment`?

src/tests/unit/tests/DetailsView/components/__snapshots__/save-assessment-factory.test.tsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`SaveAssessmentFactory getSaveButtonForAssessment renders save assessment button 1`] = `
 <SaveAssessmentButton
-  download="filename"
+  download="SavedAssessment_<undefined>_title.a11yextension"
```
dbjorge, line 5: The `<undefined>` here looks a little off; it suggests that either the test is missing setting something up, or the real code is misusing the file builder.

src/DetailsView/components/save-assessment-factory.tsx
ferBonnin: agree, per the spec in the feature the file extension should be` *.a11yassessment`

src/tests/unit/tests/DetailsView/components/save-assessment-factory.test.tsx
```diff
@@ -68,10 +74,15 @@ describe('SaveAssessmentFactory', () => {
                 .returns(() => 'fileURL')
                 .verifiable(Times.once());
 
+            fileNameBuilderMock.setup(f => f.getDateSegment(currentDate)).returns(() => 'date');
```
dbjorge, line 77: using `currentDate` here is probably not quite going to do what you want, since the actual real-world time returned by `new Date()` is going to be slightly different between you invoking it during the test on line 64 vs the real code invoking it (the name `Date` is misleading; it's actually a millisecond-precise timestamp). This is probably why you see an `<undefined>` for the date part of the filename in the snapshot; you're setting up the mock to return `'date'` if the code asks for "the current millisecond as of when line 64 ran", but the code actually asks for "the current millisecond a moment later when the code being tested ran", which your mock isn't expecting, so it doesn't have any return value defined.

To make this testable, what we normally do is inject a dependency to provide the current date, so the test can then mock that dependency into providing a fake current date to the thing being tested.

Doing this for a "current date" provider is really common, so there's actually already a dep for that set up; you can follow `DetailsViewCommandBarDeps`'s example and re-use the `getCurrentDate` dep that's already being set up in the detailsViewInitializer.

dbjorge (changes requested):
This looks really good overall! Just one question for Fer and one issue with the tests

src/DetailsView/components/save-assessment-factory.tsx
dbjorge: @ferBonnin and I discussed further offline, let's go with `.a11ywebassessment`

shanisebarona: oops. that was my bad, I accidentally transposed the second half of the file extension. thank you both!

dbjorge (approved):
LGTM!

Issue #3703, opened 2020-11-30
Update WCAG mapping for a few rules


iamrafan:
Update WCAG mapping for the following rules to match the mappings in axe-core

Rule name | Current WCAG mapping | Expected  WCAG mapping
-- | -- | --
aria-allowed-attr | WCAG 4.1.1, WCAG 4.1.2 | WCAG 4.1.2
aria-required-attr | WCAG 4.1.1, WCAG 4.1.2 | WCAG 4.1.2
aria-roles | WCAG 1.3.1, WCAG 4.1.1, WCAG 4.1.2 | WCAG 4.1.2
aria-valid-attr | WCAG 4.1.1 | WCAG 4.1.2
aria-valid-attr-value | WCAG 4.1.1, WCAG 4.1.2 | WCAG 4.1.2
frame-title | WCAG 2.4.1 | WCAG 2.4.1, WCAG 4.1.2

## Context

- AI-Web Version & Environment: Prod 2.22.0

ferBonnin:
Per conversation with Peter, let's try to tackle this as part of the axe-core update feature, if possible. cc @lisli1

ferBonnin:
@dbjorge can this be marked as resolved?

dbjorge:
Yep! This should be resolved as of prod 2.23.0

ferBonnin:
@iamrafan could you please verify and close?

iamrafan:
Validate in prod 2.23.0

PR #3709, opened 2020-12-02
fix: use axe-core tags to determine which WCAG links to show

dbjorge:
#### Description of changes

This PR changes the way we map from axe-core results/rules to WCAG guidance links.

Previously, we used a static mapping from rule ID to WCAG link. This mapping didn't have any guaranteed means of keeping in sync with axe-core's own WCAG tagging, so it drifts out of sync over time; #3703 notes several of the cases where we became inconsistent.

With this PR, we stop maintaining a rule->WCAG mapping and instead trust axe-core's tags to provide the rule->WCAG mapping; instead, we only maintain an axe tag -> WCAG link mapping.

The most meaningful implementation change is in converting `rule-to-links-mappings.ts` to `map-axe-tags-to-guidance-links.ts`; most the of the other changes are just updating the rest of the code to accomodate the change. Of note:

* `get-rule-inclusions` now looks for axe tags directly to determine if a rule should be excluded for being best-practice/experimental. There was one case (`form-field-multiple-labels`) where we were previously excluding a rule as "best practice" because its rule -> WCAG mapping said "best practice" even though axe-core didn't say so. I added an `explicitRuleOverrides` entry to make it more explicit why we exclude that case.
* `unique-landmark.ts` gained a `tags` entry, replacing the ad-hoc entry in the rules mapping
* The snapshot updates in the reporter E2Es are a good summary of the practical impact of the changes; mostly, it means that the WCAG links we present in the "passed"/"na" sections of reports change a bit.

Draft while we're going over the resulting changes in more detail

#### Pull request checklist

- [x] Addresses an existing issue: #3703
- [x] Ran `yarn fastpass`


src/tests/unit/tests/scanner/get-rule-inclusions.test.ts
```diff
@@ -56,44 +50,44 @@ describe('getRuleInclusions', () => {
...
-        const configWithBestPractice: DictionaryStringTo<HyperlinkDefinition[]> = {
-            'axe-enabled': [BestPractice],
-            'axe-disabled': [{} as HyperlinkDefinition],
-        };
-        const inclusions = getRuleInclusions(fakeRules, configWithBestPractice, {});
-        expect(inclusions).toMatchObject({
-            'axe-enabled': {
-                status: 'excluded',
-                reason: 'rule maps to BestPractice',
+    it('excludes rules mapped to best-practice tag and populates reason', () => {
```
lisli1, line 71: Should this say 'mapped to experimental tag' (rather than best-practice)?

src/tests/unit/tests/scanner/map-axe-tags-to-guidance-links.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    it('should sort the output', () => {
+        expect(mapAxeTagsToGuidanceLinks(['wcag111', 'wcag1411', 'best-practice'])).toEqual([
+            BestPractice,
+            link.WCAG_1_1_1,
+            link.WCAG_1_4_11,
+        ]);
+    });
+
+    const axe = Axe as any;
```
lisli1, line 68: Why do we do this instead of importing `* as axe from 'axe-core'`?

dbjorge: Good catch! I copied this from the obsolete rule mapping tests verbatim, but your suggestion is both more concise and more type safe. Will update

lisli1 (approved):
LGTM

PR #3786, opened 2020-12-28
feat(improve-manual-tests): UX improvements for Semantics > Header attributes

pownkel:
#### Description of changes

Changes the UX for the Semantics > Header attributes step of Assessment to match the hackathon project spec (slide 13). This adds a visual helper to this test step that highlights table cells with "headers" attributes, and the corresponding header element they reference
simple table with highlighted cells screenshot

When visibility is only turned on for one failure instance, the header referenced by that cell is also highlighted
simple table with one cell and its header highlighted

In more complex tables, all headers for a given cell/instance are highlighted
complex table with two levels of column headers and a row header. One cell and the relevant headers are all highlighted

Elements coded as table headers that have a "headers" attribute will also appear as instances, and will have both their "headers" attribute and id displayed when they are highlighted
complex table header highlighted

The instances list displays the HTML snippet for the given instance, as in the spec
instances list with HTML snippet

And the "How to test" text refers to the visual helper instead of the Web Developer extension:

>The visual helper for this requirement highlights `<td>` and `<th>` elements with a headers attribute set. Any element whose id is referenced by a headers attribute is also highlighted.
>
>1. Use the visual helper to reveal any headers attributes on the page.
>2. If a table has headers attributes, verify that they are coded correctly:
>    1. Each header cell (`<th>` element) must have an id attribute.
>    2. Each data cell (`<td>` element) must have a headers attribute.
>    3. Each data cell's headers attribute must reference all cells that function as headers for that data cell.
>Note: If a headers attribute references an element that is missing or invalid, it will fail an automated check.
>3. Record your results:
>    1. Select Fail for any instances that do not meet the requirement.
>    2. Otherwise, select Pass. Or, after you have marked all failures, select Pass unmarked instances.

#### Pull request checklist

- [x] Ran `yarn fastpass`


dbjorge:
Would it be possible to add the test page(s) you created for the screenshots to `/src/tests/end-to-end/test-resources/`?

src/injected/visualization/get-cell-and-header-elements.ts
```diff
@@ -0,0 +1,24 @@
...
+    result: AxeResultsWithFrameLevel,
+    dom: Document,
+) => {
+    const elements = Array.from(dom.querySelectorAll(result.target[result.targetIndex]));
+    const allElements = [...elements];
+    elements.forEach(element => {
+        const headers = element.getAttribute('headers')?.split(' ');
+        if (headers) {
+            headers.forEach(headerId => {
+                const headerElement = dom.querySelector(`#${headerId}`);
```
dbjorge, line 15: Maybe a little more expressive to use `dom.getElementById`?

src/injected/visualization/table-headers-formatter.ts
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { DialogRenderer } from '../dialog-renderer';
+import { AssessmentVisualizationInstance } from '../frameCommunicators/html-element-axe-results-helper';
+import { FailureInstanceFormatter } from './failure-instance-formatter';
+import { DrawerConfiguration } from './formatter';
+
+export interface HeadingStyleConfiguration {
```
dbjorge, line 9: `HeadingStyleConfiguration` and `StyleComputer` don't appear to be used; are these copy-paste leftovers?

src/injected/visualization/table-headers-formatter.ts
```diff
@@ -0,0 +1,87 @@
...
+            outlineStyle: 'solid',
+            showVisualization: true,
+            textAlign: 'right',
+            failureBoxConfig: this.getFailureBoxConfig(data),
+        };
+    }
+
+    private isHeader(element: HTMLElement): boolean {
+        if (element.matches('th')) {
+            return true;
```
dbjorge, line 59: Do we need to worry about elements like `<th role="somenonheader">`?

dbjorge: (if not, would be nice to leave a comment explaining why)

src/injected/visualization/table-headers-formatter.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+        return false;
+    }
+
+    private getTextForHeader(element: HTMLElement): string {
+        const idText = `id="${this.getAttribute(element, 'id')}"`;
+        const headersAttr = this.getAttribute(element, 'headers');
+        const headersText = headersAttr === null ? null : `headers="${headersAttr}"`;
+
+        return ['th', idText, headersText].join(' ');
```
dbjorge, line 73: This will show `th id="foo"` for `<div role="columnheader" id="foo">`; is that intended?

dbjorge, line 73: This will include an extra trailing space in the no-headers case; that won't matter in practice because it'll be stripped when rendered as html, but maybe would be a little less coupled to usage if it didn't depend on that

src/injected/visualization/table-headers-formatter.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private getTextForCell(element: HTMLElement): string {
+        const headersAttr = this.getAttribute(element, 'headers');
+        const headersText = headersAttr === null ? null : `headers="${headersAttr}"`;
+
+        return ['td', headersText].join(' ');
```
dbjorge, line 80: Same 2 comments as L73

src/scanner/custom-rules/headers-attribute-rule.ts
```diff
@@ -0,0 +1,27 @@
...
+export const headersAttributeRuleConfiguration: RuleConfiguration = {
+    checks: [
+        {
+            id: headersAttributeCheckId,
+            evaluate: () => true,
+        },
+    ],
+    rule: {
+        id: 'headers-attribute',
+        selector: 'th,td,[role=cell],[role=columnheader],[role=rowheader]',
```
dbjorge, line 16: Is it actually correct for us to include the `role=...` cases here? As far as I can tell from the [HTML 5.2 spec](https://www.w3.org/TR/html52/tabular-data.html#the-td-element), `headers` attributes are only supported for actual `tr`/`td` elements. Do we know what screen reader support is like here?

dbjorge: (if we do decide to allow `role="cell"` we should also consider whether to check `role="gridcell"` cases)

dbjorge (changes requested):
Structure of the changes looks great! A few individual comments (some will require LL/guru/fer review)

src/tests/end-to-end/test-resources/semantics/header-attributes.html
```diff
@@ -0,0 +1,66 @@
+<!--
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+-->
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <title>CodePen - Complex table example</title>
```
dbjorge, line 9: I think the `<title>` and the `./style.css` link are probably copy-paste leftovers

src/scanner/custom-rules/headers-attribute-rule.ts
pownkel: @LiLoDavis Do you have any input on whether we should be including non-td and non-th elements with table-relevant roles for this requirement?

src/injected/visualization/table-headers-formatter.ts
pownkel: My reasoning for this was that the text on the visual helper is already a bit cramped, and displaying "columnheader" on top of the table cells would take up more space and make it harder to read if the labels overlap. @LiLoDavis do you have any suggestions for whether we should be displaying something different for elements with roles (maybe something like the different capitalization we use for the headings visual helper)?

src/scanner/custom-rules/headers-attribute-rule.ts
LiLoDavis: > @LiLoDavis Do you have any input on whether we should be including non-td and non-th elements with table-relevant roles for this requirement?

@pownkel I agree with Dan - the `headers` attribute applies only to `<td>` and `<th>` elements, and there's no corresponding ARIA attribute.

src/injected/visualization/table-headers-formatter.ts
pownkel: Since we won't be including elements with columnheader/rowheader/gridcell roles (as per the thread below), this question is no longer relevant

src/injected/visualization/table-headers-formatter.ts
pownkel: According to the text for Semantics > Table headers, and the W3 pages that we link from that requirement, using a `<th>` element is sufficient to code that element as a header. I can't find any mention of roles that would cause a problem for table headers. This check only affects the visual helper, which will show that the element is coded as a header regardless of its role.

src/tests/unit/tests/injected/visualization/__snapshots__/table-headers-formatter.test.ts.snap
```diff
@@ -0,0 +1,61 @@
...
+Object {
+  "borderColor": "#0066CC",
+  "failureBoxConfig": null,
+  "outlineStyle": "solid",
+  "showVisualization": true,
+  "textAlign": "right",
+  "textBoxConfig": Object {
+    "background": "#0066CC",
+    "fontColor": "#FFFFFF",
+    "text": "th id=\\"id\\" ",
```
dbjorge, line 58: Extra trailing space in the text

dbjorge (approved):

PR #3812, opened 2021-01-11
Chore: Add eslint rule to sort import statements

RobGallo:
#### Description of changes

We discovered we had some out-of-order import statements from a CodeQL  pass on our code. This rule should address that.

Using the eslint-plugin-import package as it appears to be more configurable than the built-in 'sort-imports' rule.


peterdur:
I see that eslint-plugin-import groups imports by category; do we also want them alphabetically sorted within each group? It looks like the default is to ignore alphabetizing.

RobGallo:
@peterdur  Good catch! Updated the PR.

dbjorge (approved):

PR #3820, opened 2021-01-13
Chore: Add eslint rule react/no-access-state-in-setstate

RobGallo:
#### Description of changes

Ensures the current (soon to be previous) state of a react component is accessed from within the `setState` callback in order to protect against the possibility of the state being set by out-of-date data when calls to setState are batched.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -42,6 +42,8 @@ module.exports = {
             },
         ],
 
+        'react/no-access-state-in-setstate': 'error',
+
         // Disabled due to high existing-positive count during initial tslint -> eslint migration
         '@typescript-eslint/no-explicit-any': 'off',
         '@typescript-eslint/explicit-module-boundary-types': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -42,8 +42,6 @@ module.exports = {
             },
         ],
 
-        'react/no-access-state-in-setstate': 'error',
-
         // Disabled due to high existing-positive count during initial tslint -> eslint migration
         '@typescript-eslint/no-explicit-any': 'off',
         '@typescript-eslint/explicit-module-boundary-types': 'off',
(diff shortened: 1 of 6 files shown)


.eslintrc.js
```diff
@@ -42,6 +42,8 @@ module.exports = {
             },
         ],
 
+        'react/no-access-state-in-setstate': 'error',
```
dbjorge, line 45: nit: would be nice to group this with the other `react/*` rules down at L72-L78

src/DetailsView/components/export-dialog-with-local-state.tsx
```diff
@@ -49,11 +49,10 @@ export class ExportDialogWithLocalState extends React.Component<
     };
 
     private generateHtml = () => {
-        const exportData = this.props.htmlGenerator(this.state.exportDescription);
-        this.setState({
-            exportData,
+        this.setState(prevState => ({
+            exportData: this.props.htmlGenerator(prevState.exportDescription),
```
dbjorge, line 53: My understanding is that the same logic behind wanting to avoid `this.state` in `setState` input also applies to `this.props`. It's benign here because the `htmlGenerator` happens to be constant, but I think the better pattern would probably be to use the two-argument form of `setState`'s parameter, eg:

```ts
this.setState((prevState, prevProps) => ({
    exportData: prevProps.htmlGenerator(prevState.exportDescription),
```

src/DetailsView/components/report-export-component.tsx
```diff
@@ -58,8 +58,10 @@ export class ReportExportComponent extends React.Component<
 
     private generateHtml = () => {
         const { htmlGenerator } = this.props;
-        const exportData = htmlGenerator(this.state.exportDescription);
-        this.setState({ exportDescription: '', exportData });
+        this.setState(prevState => ({
+            exportDescription: '',
+            exportData: htmlGenerator(prevState.exportDescription),
```
dbjorge, line 63: same comment on `this.props` here (hidden access via L60)

dbjorge (changes requested):
Awesome, thanks for following up on this! A few suggestions inline

dbjorge (approved):

Issue #3322, opened 2020-09-04
Adaptable content > Text spacing - New visual helper


LiLoDavis:
## Is your feature request related to a problem? Please describe.
For this requirement, users must apply certain styling properties to verify that text adopts them with no clipping or overlap. At present, users must apply the styling manually to any text in an iframe. 

## Describe the desired outcome

- Turning on the visual helper should apply the required style properties to the root html document and to any html documents embedded in iframes. 
- Turning off the visual helper should restore the original styling by deleting the <style> element we created.

Here's the required styling:

```
<style>
* { line-height: 1.5 !important; letter-spacing: 0.12em !important; word-spacing: 0.16em !important; } 
p { margin-bottom: 2em !important; } 
iframe { border: thick dashed blue !important; background-color: yellow !important; }');
</style>
```

Here's JavaScript that creates <style> element with the above properties in the selected JavaScript context:

`var style = document.createElement('style'), styleContent = document.createTextNode('* { line-height: 1.5 !important; letter-spacing: 0.12em !important; word-spacing: 0.16em !important; } p { margin-bottom: 2em !important; } iframe { border: thick dashed blue !important; background-color: yellow !important; }'); style.appendChild(styleContent ); document.getElementsByTagName('head')[0].appendChild(style);`

## Describe alternatives you've considered
We tried using bookmarklets, but they can't apply the styling to text in iframes.

## Additional context
This feature was created in response to bug #3096.

joppekroon:
We unfortunately use a lot of nested iframes, so we are heavily impacted by Accessibility Insights referring to a bookmarklet that does not actually work for us.

Although it is possible to add this to [improve the code of the bookmarklet](https://github.com/joppekroon/bookmarklets/blob/master/text-adaptation.js), I believe using the same style injection system that Accessibility Insights already uses for visual helpers would be superior. 

P.S. I'm not sure there is a need for the border on the iframes as suggested on top.

LiLoDavis:
Looks good on most pages, 
However, it won't work on this target page: [Microsoft License Terms](https://www.microsoft.com/en-us/useterms). Visual helper is disabled, and it says **No matching/failing instances were found**.  There's text on the page, so it shouldn't say no instances.

ferBonnin:
@LiLoDavis I cannot repro this, for me it takes a second or so and then the toggle is made available. This also happens if you go to requirements in Headings or Landmarks

madalynrose:
[...]

PR #3836, opened 2021-01-21
feat(text-spacing-visualization): Adaptable content > Text spacing - New visual helper

madalynrose:
#### Description of changes

This PR addresses #3322 and implements a new visual helper for `Adaptable content > Text spacing` that applies CSS to text on the target page instead of directing users to use a third-party bookmarklet. This improves UX by centralizing functionality inside of Accessibility Insights itself and also allows users to test target page iframes, which the bookmarklet cannot reach. 

image

This visual helper injects a class into the target page and applies injected CSS, similar to the `Sensory > Color as Meaning` test. However, the existing Drawer and Formatter that is used to inject the CSS class for that test explicitly targets a single HTML element and therefore won't apply to iframes. ~~This PR abstracts the existing `SingleTargetDrawer ` and `SingleTargetFormatter` (changed to `InjectedClassDrawer` and `InjectedClassFormatter`) to accept a parameter that specifies whether its target is `Single` or `Multi`. This will be useful for future visual helpers that apply CSS to the target page but aren't drawing boxes around specific instances the way that the visual helpers for headers and landmarks do.~~

#### Pull request checklist

- [x] Addresses an existing issue: #3322 
- [x] Ran `yarn fastpass`


src/injected/styles/injected.scss
```diff
@@ -11,6 +11,17 @@
     filter: grayscale(100%) !important;
 }
 
+.insights-formatted-text-spacing-container {
+    * {
```
dbjorge, line 15: This is going to miss text that is a direct child of the `<body>` (eg,

```html
<body>
<h1>this works</h1>
this will be missed
</body>
```

I think you could just drop the `*` and make the 3 styles directly apply to the `.insights-formatted-text-spacing-container`. But if the `*` is important for the sake of precedence rule cases I haven't thought of, you could alternately do `&, * { ... }`

src/assessments/adaptable-content/test-steps/text-spacing.tsx
```diff
@@ -16,43 +18,23 @@ const textSpacingDescription: JSX.Element = (
...
-        <ol>
+        The visual helper for this requirement adjusts the target page's text styling as follows:
+        <ul>
             <li>
-                Open{' '}
-                <NewTabLink href="https://www.html5accessibility.com/tests/tsbookmarklet.html">
-                    Text spacing
-                </NewTabLink>{' '}
-                in a new browser window.
+                <Markup.Term>Letter spacing</Markup.Term> (tracking) at 0.12 times the font size{' '}
```
dbjorge, line 24: nitpick: The `{' '}`s aren't necessary here. We use those when we need to preserve spaces in between lines that JSX processing would normally remove whitespace for. For example, in something like

```jsx
<>
    some
    <em>emphasized</em>
    text
</>
```

...JSX's whitespace-collapsing behavior would normally result in there being no spaces between the three words in the rendered HTML. `{' '}` is the usual trick to avoid that for inline elements where you want to preserve the space, but it's generally fine/desirable to omit the space at the ends of `display: block` elements (like the `</li>` end tags in this code).

Some StackOverflow threads with more thorough descriptions:
* [How to avoid JSX component from condensing when React.js rendering it?](https://stackoverflow.com/q/32686728/2825668)
* [Best practice when adding whitespace in JSX](https://stackoverflow.com/q/40264084/2825668)

dbjorge (approved):
Looks great, awesome feature! One minor suggestion inline

src/assessments/adaptable-content/test-steps/text-spacing.tsx
madalynrose: oh awesome! good to know!

dbjorge (approved):

PR #3869, opened 2021-02-04
Chore: Add more files to eslint coverage

RobGallo:
#### Description of changes

Previously, only the files under ./src were checked by eslint. Now, all our non-production code should be covered as well.

Fixed the few eslint errors and warnings that popped up.


JGibson2019 (approved):
LGTM

JGibson2019 (commented):
LGTM

tsconfig.json
```diff
@@ -11,5 +11,15 @@
...
         "noUnusedLocals": true
     },
-    "include": ["src/**/*.ts", "src/**/*.tsx", ".eslintrc.js", "src/**/*.js"]
+    "include": [
+        ".eslintrc.js",
+        "src/**/*.ts",
+        "src/**/*.tsx",
+        "src/**/*.js",
+        "deploy/**/*",
+        "packages/**/*",
```
dbjorge, line 20: Stuff under `packages/` probably shouldn't be included here; tsc won't be able to build it correctly if they have separate `package.json`s with different dependencies (that's where the current PR build failures are mostly coming from).

Instead, this should probably follow `accessibility-insights-service`'s example, where `/package/*/package.json` specifies a `lint:check` script in terms of `../../.eslintrc.js` (like report-e2e-tests does already), and then use `lerna run lint:check` to run in all packages (maybe via a `lint:check:all` command at the root)

tsconfig.json
```diff
@@ -11,5 +11,15 @@
         "strictNullChecks": false,
         "noUnusedLocals": true
     },
-    "include": ["src/**/*.ts", "src/**/*.tsx", ".eslintrc.js", "src/**/*.js"]
+    "include": [
+        ".eslintrc.js",
```
dbjorge, line 15: This is redundant with `"./*.js"`, no?

.eslintignore
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+**/node_modules/**/*
+**/*bundle.js
+**/dist/*
```
dbjorge, line 6: `dist` isn't flat; should this be `**/dist/` instead?

.eslintignore
```diff
@@ -0,0 +1,9 @@
...
+**/drop/*
+**/extension/*
```
dbjorge, line 8: It feels like it would be easy for someone to accidentally create a directory like `/src/unified/platforms/extension/` in the future and miss that it wasn't being linted for a long time. Should most of these `**/` cases be more specific to the root directories?

.eslintignore
dbjorge: (same for most of these)

dbjorge (changes requested):
Thanks, some suggestions inline!

tsconfig.json
RobGallo: It is. But if I don't include it, I get an error parsing .eslintrc.js. My theory is the beginning '.' messes up the glob logic.

.eslintignore
RobGallo: I think perhaps "extension" could specifically be called out at the root, but since we have packages, "node_modules", "drop", and "dist" could occur elsewhere. There is a drop folder under packages/report for example.

dbjorge (approved):

PR #3873, opened 2021-02-05
feat(focus-mvp-client): Add code to send commands to device

DaveTryon:
#### Details

Adds code to communicate with the Android client for the focus order feature. It uses ADB to send keystrokes to the device, and calls new service endpoints (which will be implemented in the service in the next feature) to control the service state. 

The intended use is to add the `DeviceFocusControllerFactory` to the deps that get passed into the TabStops page. When loading, the TabStop page will need to call the factory with the port and the ADB location (both in the store, and both determined during device connection, which is why the TabStops page needs to use the factory to generate a `DeviceFocusController` object). It then uses the `DeviceFocusController` object to send commands to the device. Errors are bubbled up to the caller for error handling.

##### Motivation

Part of the Android focus order feature

##### Context

Telemetry will be added as a separate PR.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/platform/android/device-focus-command-sender.ts
+++ b/src/electron/platform/android/device-focus-command-sender.ts
@@ -3,10 +3,11 @@
 import axios from 'axios';
 
 export enum DeviceFocusCommand {
-    'Enable',
-    'Disable',
-    'Reset',
+    Enable = 'Enable',
+    Disable = 'Disable',
+    Reset = 'Reset',
 }
+
 export type DeviceFocusCommandSender = (port: number, command: DeviceFocusCommand) => Promise<void>;
 
 export type HttpGet = typeof axios.get;
--- a/src/electron/platform/android/device-focus-command-sender.ts
+++ b/src/electron/platform/android/device-focus-command-sender.ts
@@ -7,11 +7,11 @@ export enum DeviceFocusCommand {
     'Pause',
     'End',
 }
-export type FocusCommandSender = (port: number, command: DeviceFocusCommand) => Promise<void>;
+export type DeviceFocusCommandSender = (port: number, command: DeviceFocusCommand) => Promise<void>;
 
 export type HttpGet = typeof axios.get;
 
-export const createFocusCommandSender = (httpGet: HttpGet): FocusCommandSender => {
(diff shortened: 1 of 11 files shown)


src/electron/platform/android/device-focus-command-sender.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import axios from 'axios';
+
+export enum DeviceFocusCommand {
```
waabid, line 5: I think this looks okay for now but I do wonder whether this should be named Visualization/VisualHelper commands, since I figure they'd be applicable to other scenarios.

DaveTryon: I hadn't thought of that, but it would seem a little odd since we hardcode "/FocusTracking/" into the URI that this class produces. I guess I assumed that other scenarios would have a different URI path. I wanted to avoid having the client send 2 commands (for example, one to hide visualizations and one to disable focus tracking), since 2 commands would complicate the state management if either one failed.

It's an easy refactor if/when we have additional scenarios to take into account. We already have 2 other classes that build a similar URI, so perhaps we'd pull all of those into a single location (or we might just decide that we ought to have a shared formatting string that gets used to build all URI's)

src/electron/views/renderer-initializer.ts
```diff
@@ -410,6 +413,14 @@ getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch)
 
         scanController.initialize();
 
+        const deviceFocusControllerFactory = new DeviceFocusControllerFactory(
+            appiumAdbWrapperFactory,
+            createDeviceFocusCommandSender(axios.get),
+        );
+
+        // Placeholder--remove once we include deviceFocusControllerFactory in the deps
+        deviceFocusControllerFactory.initialize();
```
waabid, line 422: You can actually just pass this into the deps as it is and can remove this and the initialize method. I don't think TS/lint would complain if there is an additional property on the deps object that is not needed.

waabid (approved):
LGTM, just a minor comment on removing `initialize` and related code.

PR #3880, opened 2021-02-05
feat(focus-mvp-client): Add shell input keyevent commands to mock-adb

DaveTryon:
#### Details

Teach mock-adb how to handle ADB's `shell input keyevent` commands for keyevent codes of 19, 20, 21, 22, 61, and 66 (these are the codes that we intend to use for this feature, and are added in #3873).

##### Motivation

We're using these commands as part of the android focus feature. In order to validate these in E2E tests, we need to make sure that mock-adb is able to handle these commands.

##### Context

The spec calls for Shift+Tab, but we believe that to be infeasible at the moment, so we're doing only the 4 arrow keys (codes 19-22), Tab (code 61), and Enter (code 66). It would have been nice to pull these from the TypeScript enum, but I couldn't find a clean way to do that with the configurator being in straight JavaScript.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/miscellaneous/mock-adb/common-adb-configs.js
+++ b/src/tests/miscellaneous/mock-adb/common-adb-configs.js
@@ -15,6 +15,7 @@ const serviceInfoCommandMatch =
 const serviceIsRunningCommandMatch = 'shell dumpsys accessibility';
 const portForwardingCommandMatch = 'forward tcp:';
 const sdkVersionCommandMatch = 'shell getprop ro.build.version.sdk';
+const inputKeyeventCommandMatch = 'shell input keyevent';
 
 function addDeviceEnumerationCommands(id, output) {
     output[`-s ${id} ${devicesCommandMatch}`] = cloneDeep(output.devices);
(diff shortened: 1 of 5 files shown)


src/tests/miscellaneous/mock-adb/common-adb-configs.js
```diff
@@ -93,6 +94,16 @@ function addPortForwardingCommands(id, output, port) {
     };
 }
 
+function addInputKeyeventCommands(id, output) {
+    const keyeventCodes = [19, 20, 21, 22, 61, 66];
```
dbjorge, line 98: I don't mind this being repeated here but it would be nice for this to make it more clear how the magic numbers were arrived at, something like:

```ts
// These are the cases required for the virtual keyboard in unified tab stops
// See https://android.documentation/keycodes
const keyeventCodeMap = {
    'left-arrow': 19,
    'right-arrow': 20,
    // ... etc
};

for (const keyeventCode of Object.values(keyeventCodeMap)) {
    // ...
}
```

dbjorge (approved):
One readability suggestion, otherwise LGTM

Issue #3263, opened 2020-08-19
Android: on mac in dark mode, device setup dialog title bar has weird blank space if not focused


dbjorge:
## Describe the bug

On a Mac, when the device setup dialog isn't the currently focused window, its title bar has a weird looking blank space when the window isn't focused. I'm not sure what the "normal" mac behavior is here (is the window title text supposed to be centered?), but it probably isn't what we see in this screenshot:

![image](https://user-images.githubusercontent.com/376284/90687069-46154100-e23a-11ea-8d81-497cca1ac900.png)

**To Reproduce**

1. Open Accessibility Insights for Android on a mac
2. Click to some other window
3. Observe title bar

## Expected behavior

Styling should looks similar to other apps on the platform

## Context (please complete the following information)

- OS Name & Version: Mac 10.15.6 (I think? I took screenshot from Dave's demo presentation today)
- AI-Android Version & Environment: Accessibility Insights for Android v2020.702.3

## Are you willing to submit a PR?

yes

## Did you search for similar existing issues?

yes

DaveTryon:
For reference, here's what happens with a system app (the image viewer):
1. The title is always centered
2. Everything is greyed out (but still visible) when the focus is lost

Image with focus:
<img width="606" alt="Reference (with focus)" src="https://user-images.githubusercontent.com/45672944/90692022-39d4b800-e229-11ea-83bc-c3922ec4ac73.png">

Image without focus:
<img width="609" alt="Reference (without focus)" src="https://user-images.githubusercontent.com/45672944/90692033-3fca9900-e229-11ea-9491-e0e5826d5c8e.png">

dbjorge:
@ferBonnin Yes, this is technically feasible (iirc, the title bar already has custom behavior/styling between the 2 platforms)

DaveTryon:
@ferBonnin, what do you want to do about the disappearing icons that originally prompted this issue? Can this issue really be "Ready for Work" without that answer?

ferBonnin:
@DaveTryon can you clarify?

DaveTryon:
@ferBonnin: When @dbjorge opened this issue, the primary concern was about the blank space in the top left corner--that blank space is where the close/minimize/fullscreen buttons live. These buttons disappear when we lose the focus. The secondary concern was that the text wasn't centered, like most Mac apps.

[...]

PR #3891, opened 2021-02-09
fix: Use default title bar for unified client running on mac

DaveTryon:
#### Details

The unified app currently uses a custom title bar that works OK on Windows and Linux, but it has some funny behaviors on Mac when we lose the focus, as pointed out in #3263. We did some investigation here and found that this is a side effect of Electron on Mac in dark mode. We could potentially create a full dark mode option to address this, but it's a lot simpler to simply use the default title bar on Mac. We may revert this change if/when we create a dark mode for the unified app.

##### Motivation

Addresses issue #3263

##### Context

I tested the traffic light buttons, including saving the app size and state, and didn't detect any regressions. The one thing I noticed was that the system menu says "Electon" instead of "Accessibility Insights for Android", but the same thing also occurs when running a local build from the master branch. As it turns out, the string shown in the system menu comes from the app package, and isn't explicitly set by our code.

The titles are being rendered by the OS, not by our code, so we inherit system behavior automatically.

Screenshots (shadows omitted so they're all the same size, although in some cases the screenshots look like there's no border at all, which isn't really the case)

Description | Image
--- | ---
Dark mode, unfocused | Dark-Unfocused
Dark mode, focused | Dark-Focused
Light mode, unfocused | Light-Unfocused
Light mode, focused | Light-Focused
High Contrast, unfocused | High Contrast-Unfocused
High Contrast, focused | High Contrast-Focused

#### Pull request checklist

- [x] Addresses an existing issue: #3263 
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -39,7 +39,7 @@ const createWindow = () => {
         show: false,
         // enableRemoteModule required for spectron (https://github.com/electron-userland/spectron/issues/693#issuecomment-696957538)
         webPreferences: { nodeIntegration: true, enableRemoteModule: true },
-        titleBarStyle: 'hidden',
+        titleBarStyle: platformInfo.isMac() ? 'default' : 'hidden',
         width: mainWindowConfig.defaultWidth,
         height: mainWindowConfig.defaultHeight,
         frame: os === OSType.Mac,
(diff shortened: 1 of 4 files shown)


DaveTryon:
I missed the caption changes--will close for now and reopen once that's fixed

PR #3906, opened 2021-02-11
feat(focus-mvp-client): Add mock adb + android service command logging for future E2E tests

jalkire:
#### Details
This PR updates mock adb and the mock service for Android to log their received commands. These logs can then be referenced in E2E tests to confirm that the proper adb and service commands occurred. This PR has no other impact on current E2E tests and does not impact preexisting logging behavior.

##### Motivation
Currently, our E2E tests implicitly validate that adb and service commands occurred because the UI updates based on the results of the commands. The new Tab stops UI will allow users to send adb and service commands without displaying UI feedback for those commands. This creates a new testing requirement--we need to explicitly validate that the proper commands were sent.

##### Context
This PR does not add or modify any E2E tests. Instead, it adds the backing infrastructure to support future E2E tests. New tests will be able to validate mock adb and service commands in the following manner:
```ts
it('does service for Android stuff', async () => {
    // start with an empty log
    logController.resetServerLog(logsContext);

    // do some test stuff related to the mock server

    // verify that mock server's log is as expected
    const serverLog = logController.getServerLog(logsContext);
    expect(serverLog).toMatchSnapshot();
});

it('does adb stuff', async () => {
    // start with an empty log
    logController.resetAdbLog(logsContext);

    // do some test stuff related to mock adb

    // verify that mock ADB's log is as expected
    const adbLog = logController.getAdbLog(logsContext);
    expect(adbLog).toMatchSnapshot();
});
```

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/electron/common/view-controllers/app-controller.ts
+++ b/src/tests/electron/common/view-controllers/app-controller.ts
@@ -1,7 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import * as fs from 'fs';
-import * as path from 'path';
 import { AndroidSetupStepId } from 'electron/platform/android/setup/android-setup-step-id';
 import { Application } from 'spectron';
 import { AndroidSetupViewController } from 'tests/electron/common/view-controllers/android-setup-view-controller';
@@ -125,38 +123,6 @@ export class AppController {
                 timeout: DEFAULT_WAIT_FOR_ELEMENT_TO_BE_VISIBLE_TIMEOUT_MS,
                 timeoutMsg: `was expecting window.${propertyName} to be defined`,
             },
-        );
-    }
-
-    public getServerLog(logName: string, extraLogNames: string): string {
-        return fs
-            .readFileSync(
-                path.normalize(
-                    `drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/server.log`,
-                ),
-            )
-            .toString();
-    }
-
-    public getAdbLog(logName: string, extraLogNames: string): string {
-        return fs
-            .readFileSync(
-                path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/adb.log`),
-            )
-            .toString();
-    }
-
-    public resetServerLog(logName: string, extraLogNames: string): void {
-        fs.rmdirSync(
-            path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/server.log`),
-            { recursive: true },
-        );
-    }
-
-    public resetAdbLog(logName: string, extraLogNames: string): void {
-        fs.rmdirSync(
-            path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/adb.log`),
-            { recursive: true },
         );
     }
 }
--- a/src/tests/electron/common/view-controllers/app-controller.ts
+++ b/src/tests/electron/common/view-controllers/app-controller.ts
@@ -145,4 +145,18 @@ export class AppController {
             )
             .toString();
     }
+
+    public resetServerLog(logName: string, extraLogNames: string): void {
+        fs.rmdirSync(
+            path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/server.log`),
+            { recursive: true },
+        );
+    }
+
+    public resetAdbLog(logName: string, extraLogNames: string): void {
+        fs.rmdirSync(
+            path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/adb.log`),
+            { recursive: true },
+        );
+    }
 }
--- a/src/tests/electron/common/view-controllers/app-controller.ts
+++ b/src/tests/electron/common/view-controllers/app-controller.ts
@@ -1,5 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import * as fs from 'fs';
+import * as path from 'path';
 import { AndroidSetupStepId } from 'electron/platform/android/setup/android-setup-step-id';
 import { Application } from 'spectron';
 import { AndroidSetupViewController } from 'tests/electron/common/view-controllers/android-setup-view-controller';
@@ -124,5 +126,23 @@ export class AppController {
                 timeoutMsg: `was expecting window.${propertyName} to be defined`,
             },
         );
+    }
+
+    public getServerLog(logName: string, extraLogNames: string): string {
+        return fs
+            .readFileSync(
+                path.normalize(
+                    `drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/server.log`,
+                ),
+            )
+            .toString();
+    }
+
+    public getAdbLog(logName: string, extraLogNames: string): string {
+        return fs
+            .readFileSync(
+                path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/adb.log`),
+            )
+            .toString();
     }
 }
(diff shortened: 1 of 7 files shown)


src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -125,4 +127,22 @@ export class AppController {
             },
         );
     }
+
+    public getServerLog(logName: string, extraLogNames: string): string {
+        return fs
+            .readFileSync(
+                path.normalize(
+                    `drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/server.log`,
```
DaveTryon, line 135: It feels odd to repurpose the mock-adb folder for server logs. Would it logical/practical instead to create a per-test log folder (independent of the mock-adb folder) and have a way to tell dump both the mock-adb and server logs into that folder?

src/tests/miscellaneous/mock-adb/app/bin.js
```diff
@@ -58,6 +58,10 @@ async function main() {
     const outputLogsDir = path.join(path.dirname(process.execPath), 'logs', currentContext);
     fs.mkdirSync(outputLogsDir, { recursive: true });
 
+    const testLogsDir = path.join(outputLogsDir, 'testLogs');
+    fs.rmdirSync(testLogsDir, { recursive: true });
+    fs.mkdirSync(testLogsDir);
```
DaveTryon, line 63: Mock-adb gets called several times during an E2E test. If I'm reading this correctly, each call will delete all previous log files. That might make sense in specific scenarios, but I'm not sure it does in the general scenario. Would it make sense to have this behavior controlled by the config file? The default behavior might be to retain the logs, but E2E tests could specify an option that would delete on each call

src/tests/miscellaneous/mock-adb/app/bin.js
```diff
@@ -92,6 +96,10 @@ async function main() {
 
     result.input = process.argv;
     result.inputCommand = inputCommand;
+
+    fs.writeFileSync(path.join(testLogsDir, 'adb.log'), `ADB ${inputCommand}\n`, {
```
DaveTryon, line 100: Would be it practical to make the test logging behavior controlled by a flag in the mock-adb-config.json file? It's probably not a huge deal to log in 2 places, but it feels cleaner to only log it when we need it

DaveTryon, line 100: Might consider EOL instead of '\n'
```suggestion
    fs.writeFileSync(path.join(testLogsDir, 'adb.log'), `ADB ${inputCommand}` + EOL, {
```

src/tests/electron/common/view-controllers/app-controller.ts
jalkire: Per offline conversation, leaving this as is because the server is being started and given this logging location by mock adb

src/tests/miscellaneous/mock-adb/app/bin.js
jalkire: Good catch! Updated bin.js to no longer clear the directory. Instead, added some helper methods to `AppController` to handle this.

src/tests/miscellaneous/mock-adb/app/bin.js
jalkire: Per offline conversation, leaving this as is because we're already doing a lot of logging that can't be disabled and the new logs live alongside the old logs.

jalkire: Since we expect this file to be compared to a snapshot in our E2E tests, we don't want different EOL characters on different OSs.

src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -125,4 +127,36 @@ export class AppController {
...
+    public getAdbLog(logName: string, extraLogNames: string): string {
+        return fs
+            .readFileSync(
+                path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/adb.log`),
+            )
+            .toString();
+    }
+
+    public resetServerLog(logName: string, extraLogNames: string): void {
+        fs.rmdirSync(
```
DaveTryon, line 150: I *think* that server.log is a file, but we're use `fs.rmdirSync` to remove it. Should we be using `fs.unlink`, instead?

src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -125,4 +127,36 @@ export class AppController {
...
+            path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/server.log`),
+            { recursive: true },
+        );
+    }
+
+    public resetAdbLog(logName: string, extraLogNames: string): void {
+        fs.rmdirSync(
```
DaveTryon, line 157: I *think* that adb.log is a file, but we're use `fs.rmdirSync` to remove it. Should we be using `fs.unlink`, instead?

src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -125,4 +127,36 @@ export class AppController {
...
+        return fs
+            .readFileSync(
+                path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/adb.log`),
+            )
+            .toString();
+    }
+
+    public resetServerLog(logName: string, extraLogNames: string): void {
+        fs.rmdirSync(
+            path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/server.log`),
```
DaveTryon, line 151: We build this path in 4 different places (3 in this file, 1 in the mock server). Would it improve the cohesion if we pulled that construction into a single method that we referenced? For the mock server, maybe we could pass in the full file name instead of just the path, so that we still build the name in just 1 place?

src/tests/electron/common/view-controllers/app-controller.ts
```diff
@@ -125,4 +127,36 @@ export class AppController {
...
+                    `drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/server.log`,
+                ),
+            )
+            .toString();
+    }
+
+    public getAdbLog(logName: string, extraLogNames: string): string {
+        return fs
+            .readFileSync(
+                path.normalize(`drop/mock-adb/logs/${logName}/${extraLogNames}/testLogs/adb.log`),
```
DaveTryon, line 144: We build the path to adb.log in 3 different places. Would it improve the cohesion if we pulled that construction into a single method that we referenced from the other locations?

src/tests/electron/common/view-controllers/app-controller.ts
jalkire: I'm only seeing two server.log path creations in this file (same for adb.log). Am I just missing the third ones? I'm taking a look at a good way to separate this out.

src/tests/electron/common/view-controllers/app-controller.ts
jalkire: Yes! Now using `fs.unlinkSync`

src/tests/electron/common/view-controllers/app-controller.ts
jalkire: Refactored and introduced `generate-log-paths.js` to handle log path generation on a central location.

src/tests/electron/common/view-controllers/app-controller.ts
jalkire: Yes! I also moved this logic out of `AppController` and into a new `LogController` class

src/tests/electron/common/view-controllers/app-controller.ts
jalkire: Yes! Now using `fs.unlinkSync`

DaveTryon (approved):

PR #3921, opened 2021-02-17
feat(focus-mvp-client): Remove Instances component from tab stops test and update styling

brocktaylor7:
#### Details

- Removes instances component from tab stops test
- Brings tab stops styling in line with web's tab stops test
- refactors layout to fix elements being nested within a paragraph element

image

##### Motivation

Unifying tab stops test with the web version

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/views/results/test-view.scss
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+.test-view {
+    .tab-stops-test {
+        .how-to-test-list {
```
waabid, line 6: Is there a reason we need to nest these styles? As in, can we not just use the "how-to-test-list" style by itself?

waabid: Regardless, we try to isolate styles on a per-component basis (unless generic/common styling). Currently, this style file seems to be for `test-view` but is being used by both `test-view` and the tab stops description component, even though it is only really applying styling to one component (the latter).

I think it would be better if we created a specific component tsx file for the description (instead of doing so inline) and then an appropriately named scss file instead.

src/electron/platform/android/test-configs/tab-stops/test-config.tsx
```diff
@@ -1,19 +1,23 @@
...
 
 export const tabStopsTestConfig: TestConfig = {
     key: 'tab-stops',
     contentPageInfo: {
         title: 'Tab stops',
         description: (
-            <>
-                <strong>How to test:</strong>
-                <ol>
+            <div className={styles.tabStopsTest}>
```
waabid, line 16: I'm not sure I see the reason for this to be a div?

src/electron/views/results/test-view.scss
brocktaylor7: Yeah, this makes sense.  I wasn't 100% sure of the way things were done in the rest of the repo.  I've updated this now. I moved the file location and removed the nesting.

src/electron/platform/android/test-configs/tab-stops/test-config.tsx
brocktaylor7: I changed this back to not being a div, I had made it a div to handle the nested css that I've now removed.

waabid (approved):

src/electron/views/results/test-view.tsx
```diff
@@ -34,21 +34,25 @@ export const TestView = NamedFC<TestViewProps>('TestView', props => {
     const { title, description } = contentPageInfo;
     const headerSection = <HeaderSection title={title} description={description} />;
 
+    if (contentPageInfo.instancesSectionComponent == null) {
```
waabid, line 37: An alternative might be to use the null pattern and remove this case.

Basically, just pass a null component for the instances section and eliminates the optional parameter and logic from the UI.

waabid: We specifically do have one called `NullComponent`.

waabid: I think that might be better (and then in the future moving the spinner inwards to the instances section/something specific to where it applies).

PR #3939, opened 2021-02-19
feat(focus-mvp-client): add toggle for tab stops

waabid:
#### Details

Renames virtual keyboard folder to tab stops and that is most of the changes (**please use the commit view to see changes**). Otherwise, moves the how to test instructions for tab stops to the instances property (since it seems the most appropriate considering what we're doing). That property will be renamed in the future.

##### Motivation

Needed for tab stops test.

##### Context

Property rename not included on content page interface.

Considered adding a separate property to the content page interface. Decided against it because effectively we already have a pattern for including the visualization toggle within the instances section and this would be consistent.

image

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/views/results/test-view.tsx
```diff
@@ -60,6 +60,7 @@ export const TestView = NamedFC<TestViewProps>('TestView', props => {
                 scanMetadata={scanMetadata}
                 shouldAlertFailuresCount={true}
                 cardsViewData={cardsViewData}
+                showTabStops={true}
```
waabid, line 63: Decided to leave as is here (since the property is not used/won't impact users). In a future PR, this will be passed from store data.

waabid: Actually on second thought, I'll leave this out.

DaveTryon (approved):
I'm guessing that connecting the toggle actions will be in a separate PR?

PR #3945, opened 2021-02-20
feat(focus-mvp-client): Refactor Start over button

DaveTryon:
#### Details

The start over button needs to be customizable for different tabs. We do this by determining the control for the `onClick` and the `disabled` properties into the `TestConfig` class, then retrieving them at initialization time. Following the existing patter, the `TestConfig` class just collects up existing components, and populating the actual properties is farmed out to a couple of helpers--one that is specific to TabStops, and one that is shared between AutomatedChecks and NeedsReview.

##### Motivation

Needed for tabstops feature

##### Context

This currently leaves the Start over button disabled in the TabStops tab. We'll enable it once we have the correct actions ready to wire in.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/platform/android/test-configs/automated-checks/test-config.tsx
```diff
@@ -22,5 +25,11 @@ export const automatedChecksTestConfig: TestConfig = {
         resultsFilter: automatedChecksResultsFilter,
         allowsExportReport: true,
         visualHelperSection: ScreenshotView,
+        startOverButtonTabSettings(props: ReflowCommandBarProps): StartOverButtonTabSettings {
```
jalkire, line 28: Currently, the `resultsFilter` values for automated checks and needs review live separately from their respective test configs and have their own test files. That way, the test configs contain no logic, just keys and values, and don't need tests. Do we want to follow a similar pattern for these `startOverButtonTabSettings`?

DaveTryon: Good call. I created a function that gets shared in AutomatedChecks and NeedsReview, and another function (just a placeholder until we have the connecting code) for TabStops. These functions have unit tests to replace the ones that were removed from the ReflowCommandBar class

jalkire (approved):
🚢

PR #3951, opened 2021-02-23
feat(focus-mvp-client): add tab stops store and update tab stops action creator

waabid:
#### Details

Adds a store for tracking tab stops state. Also updates the action creator to use the device focus controller (and adds functionality).

##### Motivation

Feature work.

##### Context

The alternative approach to the action creator changes: we could instead listen to the tab stops store and then act accordingly. However, that would require a lot of state management and make the action creator complicated. What that would require:
- We would need to listen to the store update in device focus controller, compare past state, and then choose accordingly to either enable/disable focus tracking.
- Either inconsistency with how we send keyboard commands (direct device focus controller invocation) vs. enabling/disabling focus tracking (on store update) OR do store related infra work that will contain the virtual keyboard commands to send.

Due to this, I figured direct invocation of methods on the controller was the better approach to take.

Finally: wiring up these changes to the UI will be in a future PR.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/flux/action/tab-stops-action-creator.ts
+++ b/src/electron/flux/action/tab-stops-action-creator.ts
@@ -2,19 +2,50 @@
 // Licensed under the MIT License.
 
 import { TabStopsActions } from 'electron/flux/action/tab-stops-actions';
+import { DeviceFocusController } from 'electron/platform/android/device-focus-controller';
 
 export class TabStopsActionCreator {
-    constructor(private readonly tabStopsActions: TabStopsActions) {}
+    constructor(
+        private readonly tabStopsActions: TabStopsActions,
+        private readonly deviceFocusController: DeviceFocusController,
+    ) {}
 
     public enableTabStops = () => {
         this.tabStopsActions.enableFocusTracking.invoke();
+        this.deviceFocusController.enableFocusTracking();
     };
 
     public disableTabStops = () => {
         this.tabStopsActions.disableFocusTracking.invoke();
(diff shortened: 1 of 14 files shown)


src/electron/flux/action/tab-stops-action-creator.ts
```diff
@@ -2,19 +2,50 @@
...
+import { DeviceFocusController } from 'electron/platform/android/device-focus-controller';
 
 export class TabStopsActionCreator {
-    constructor(private readonly tabStopsActions: TabStopsActions) {}
+    constructor(
+        private readonly tabStopsActions: TabStopsActions,
+        private readonly deviceFocusController: DeviceFocusController,
+    ) {}
 
     public enableTabStops = () => {
```
DaveTryon, line 13: Consider the following scenario:

1. User clicks the toggle to start focus tracking
2. `tabStopsActions.enableFocusTracking` is invoked and the store gets updated
3. `deviceFocusController.enableFocusTracking` fails because the device is disconnected

Once the DeviceDisconnectedPopup goes away, `TabStopsStore.focusTracking` thinks tab stops are enabled on the device but the device never received the command.

What if, instead of having the TabStopsActionCreator call the DeviceFocusController, the DeviceFocusController called the TabStopsActionCreator? That would cause the TabStopsStore to change only when the device was changed, and would allow us to remove the functions that are just one-line wrappers around the DeviceFocusController

DaveTryon (approved):
Looks good once the couple of things we discussed are in. Signing off

PR #3966, opened 2021-02-24
feat(focus-mvp-client): turn off tab stops visualization on browser close

brocktaylor7:
#### Details

Disables visualizations on device when the client is closed.

##### Motivation

Improved UX, focus tracking visualizations will be turned off when the client is closed.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/electron/tests/__snapshots__/tab-stops-view.test.ts.snap
+++ b/src/tests/electron/tests/__snapshots__/tab-stops-view.test.ts.snap
@@ -5,6 +5,11 @@ exports[`TabStopsView clicking start over sends corresponding service command 1`
 "
 `;
 
+exports[`TabStopsView leaving tab stops sends reset service command 1`] = `
+"GET /AccessibilityInsights/FocusTracking/Reset
+"
+`;
+
 exports[`TabStopsView toggling show tab stops sends corresponding service commands 1`] = `
 "GET /AccessibilityInsights/FocusTracking/Enable
 GET /AccessibilityInsights/FocusTracking/Disable
--- a/src/tests/electron/tests/tab-stops-view.test.ts
+++ b/src/tests/electron/tests/tab-stops-view.test.ts
@@ -90,6 +90,15 @@ describe('TabStopsView', () => {
         expect(serverLog).toMatchSnapshot();
     });
 
+    it('leaving tab stops sends reset service command', async () => {
+        logController.resetServerLog();
+
+        await resultsViewController.clickLeftNavItem('automated-checks');
+
+        const serverLog = await logController.getServerLog();
+        expect(serverLog).toMatchSnapshot();
+    });
+
     it('should pass accessibility validation in all contrast modes', async () => {
         await scanForAccessibilityIssuesInAllModes(app);
     });


src/tests/unit/tests/electron/ipc/async-action.test.ts
```diff
@@ -17,4 +17,16 @@ describe('AsyncAction', () => {
...
+    it("calls listeners in order they're created", async () => {
+        const testSubject = new AsyncAction();
+        const listenerMocks = [];
+        for (let i = 0; i <= 3; i++) {
+            listenerMocks.push(Mock.ofInstance(async () => {}));
+            testSubject.addAsyncListener(listenerMocks[i].object);
+        }
+        await testSubject.invokeAsync();
+
+        //any way to verify the order in which the mocks were called?
```
brocktaylor7, line 30: Does anyone know of a way to verify that functions were called in a particular order? I'd also like to add that to the android-browser-close-cleanup-initializer.test.ts

dbjorge: Rather than using `Mock.ofInstance(async () => {})`, you can use stub listeners that update a list (`async () => { listenerCallHistory.push(i) }`) and then write an expectaion against the list at the end.

If you want to verify not just that they're called in order but that each one is `await`ed in sequence before starting the next, you might instead do ```listenerCallHistory.push(`${i} start`); await tick(); listenerCallHistory.push(`${i} end`);```

brocktaylor7: thanks @dbjorge this worked great!

src/electron/platform/android/setup/android-browser-close-cleanup-initializer.ts
```diff
@@ -0,0 +1,16 @@
...
+import { AndroidTabStopsVisualizationCleaner } from 'electron/platform/android/setup/android-tab-stops-visualization-cleaner';
+
+export class AndroidBrowserCloseCleanupInitializer {
+    constructor(
+        private readonly androidTabStopsVisualizationCleaner: AndroidTabStopsVisualizationCleaner,
+        private readonly androidPortCleaner: AndroidPortCleaner,
+    ) {}
+
+    public initialize(): void {
+        this.androidTabStopsVisualizationCleaner.initialize();
```
DaveTryon, line 13: I get that this class exists to enforce the ordering, but it somehow feels odd to bury the initialize call inside another class. I'd love to hear from @waabid on this one.

Assuming that we keep this pattern, there's still a risk that someone would re-add the call to initialize the androidPortCleaner, which would silently undo what this class does. We can buy some cheap "insurance" here if we make AndroidPortCleaner count its calls to initialize, then throw an error if it's ever initialized more than once.

brocktaylor7: Yeah, I deliberated for a while about how to approach this. I'm certainly not married to this pattern if anyone has suggestions on how to improve it.

Additional calls to initialize on AndroidPortCleaner.initialize() would be fine, as long as they happen after initialize is called on AndroidTabStopsVisualizationCleaner the first time.  If it happened before, it would indeed silently undo the tab stops visualization cleaner.  If anyone has ideas of improvements on how to prevent that from happening in the future I'd appreciate it. I couldn't figure out a good way to test for that scenario.

brocktaylor7: I was able to add a test that guarantees the order in this class, if we decide to keep this approach.

Issue #3431, opened 2020-10-02
Details dialog not rendering Fix the following on some pages


jalkire:
## Describe the bug

**To Reproduce**

1. Go to https://flow.microsoft.com/en-us/
2. Run Automated Checks
3. Click on a failure instance on the target page
4. Observe that fix the following renders an empty bullet list

## Expected behavior
Fix the following displays text

## Screenshots

![image](https://user-images.githubusercontent.com/4615491/94960211-58280780-04a7-11eb-990f-9554f62d1425.png)

## Context (please complete the following information)

- OS Name & Version: Windows 10 version 2004 19041.508
- AI-Web Version & Environment: 2.21.1 prod
- Browser Version: Edge 85.0.564.68
- Target Page: 

## Did you search for similar existing issues?
Yes

## Additional context
Removing `aria-hidden="true"` from the `<span>` containing the how to fix text causes it to display.

ferBonnin:
Per conversation with Peter, looks like this particular page has a CSS rule that if it has aria-hidden = true, then it doesn't display the element. The color boxes are hidden on purpose so the screen reader doesn't read them (since we already add the HEX as part of the text).

Given that this particular page is doing something odd, we are marking this as out of scope.

PR #3969, opened 2021-02-25
feat(dialog-iframe): implement and wire up InjectedDialogStore

dbjorge:
#### Details

This PR adds a new `InjectedDialogStore` and associated action/action-message/etc wiring. This will be used to track whether the details dialog in the target page is currently open and if so, against which element.

This is part 1 of a series of PRs that will eventually move the rendering of our `DetailsDialog` component (the dialog that renders to show details of issues in the target page when using the ad-hoc issues tool) into a separate iframe. This PR itself doesn't change any functionality; it sets up the ability to replace the behavior of clicking a highlight box in the target page from "invoke dialog rendering behavior directly" to "update store state so that a different script context can know to do dialog rendering".

##### Motivation

Moving dialog rendering into our own iframe resolves several issues and bits of technical debt; it enables us to use more typical office fabric and react components for rendering that we are currently unable to do because of the high likelihood of conflicts with code/styles/etc in the target page, and it prevents bugs like #3431 where target page logic interferes with our rendering process.

##### Context

This PR will be followed by separate PRs which:

* Introduce a new webpack entry point/HTML file for the dialog iframe
* Moves `DialogRenderer` logic and associated react components into the iframe
* Add some sort of iframe controller to main-window-initializer which renders the dialog iframe as a high-z-order overlay iff `InjectedDialogStore` shows `isOpen: true`
* Update `highlight-box-drawer` and friends to send a `InjectedDialog.Open` message instead of invoking `DialogRenderer` directly

#### Pull request checklist

- [x] Addresses an existing issue: 1818340
- [x] Ran `yarn fastpass`


src/background/stores/injected-dialog-store.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    protected addActionListeners(): void {
+        this.injectedDialogActions.getCurrentState.addListener(this.onGetCurrentState);
+        this.injectedDialogActions.openDialog.addListener(this.onOpenDialog);
+        this.injectedDialogActions.closeDialog.addListener(this.onCloseDialog);
+
+        this.visualizationActions.enableVisualization.addListener(this.onCloseDialog);
+        this.visualizationActions.disableVisualization.addListener(this.onCloseDialog);
+        this.visualizationActions.updateSelectedPivot.addListener(this.onCloseDialog);
+        this.visualizationActions.updateSelectedPivotChild.addListener(this.onCloseDialog);
```
lisli1, line 26: To clarify, these are for the other visualizations on the page (specifically not the injected dialog), and we want to ensure that the dialog doesn't stay open when these actions happen?

src/background/actions/injected-dialog-action-creator.ts
```diff
@@ -0,0 +1,41 @@
...
+            this.onCloseDialog,
+        );
+        this.interpreter.registerTypeToPayloadCallback(
+            getStoreStateMessage(StoreNames.InjectedDialogStore),
+            () => this.injectedDialogActions.getCurrentState.invoke(null),
+        );
+    }
+
+    private onOpenDialog = (payload: InjectedDialogOpenPayload): void => {
+        this.telemetryEventHandler.publishTelemetry(TelemetryEvents.ISSUES_DIALOG_OPENED, payload);
```
lisli1, line 34: Should we update the name for this telemetry event to `INJECTED_DIALOG_OPENED` for consistency? I also see the argument that we don't want to rename it because we already have a history of telemetry for `ISSUES_DIALOG_OPENED`, and we want them to be grouped together.

src/tests/unit/tests/background/actions/injected-dialog-action-creator.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+        const testSubject = new InjectedDialogActionCreator(
+            interpreterMock.object,
+            actionsMock.object,
+            telemetryEventHandlerMock.object,
+        );
+
+        testSubject.registerCallbacks();
+
+        openDialogMock.verifyAll();
```
lisli1, line 41: Does this need a `.verifiable()` call before `.verifyAll()`?

src/background/actions/injected-dialog-action-creator.ts
```diff
@@ -0,0 +1,41 @@
...
+            Messages.InjectedDialog.Open,
+            this.onOpenDialog,
+        );
+        this.interpreter.registerTypeToPayloadCallback(
+            Messages.InjectedDialog.Close,
+            this.onCloseDialog,
+        );
+        this.interpreter.registerTypeToPayloadCallback(
+            getStoreStateMessage(StoreNames.InjectedDialogStore),
+            () => this.injectedDialogActions.getCurrentState.invoke(null),
```
peterdur, line 29: Consider adding this and the other new files to the strict null checks list? (When I add this file, an error is reported at line 29.)

lisli1 (approved):
Left 2 questions but overall LGTM

src/background/stores/injected-dialog-store.ts
dbjorge: Yep! Similar to what `TabStore` does

src/background/actions/injected-dialog-action-creator.ts
dbjorge: Yep, that argument is why I left it as-is.

src/background/actions/injected-dialog-action-creator.ts
dbjorge: Good idea!

dbjorge:
Closing this for now until we're more decided on continuing the feature.

PR #4004, opened 2021-03-04
chore(null): null fixes for 11 /src/electron files

dbjorge:
#### Details

Addresses a few layers of `/src/electron` in `yarn null:find`

Before: **62%** complete (**885**/1426 non-test files)
After: **63%** complete (**896**/1426 non-test files)

##### Motivation

See #2869

##### Context

I debated how to best address the `unified-result-builder`-related cases; personally I think it would probably be better to fail a little faster when we get half-populated results from the scanner, rather than trying to barge onwards with potentially half-filled-in-results, but I decided that was too big a change to want to embed as part of the null fixes.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/electron/platform/android/unified-result-builder.ts
```diff
@@ -34,14 +34,14 @@ export const createBuilder = (
     const payload: UnifiedScanCompletedPayload = {
         scanResult: getUnifiedResults(scanResults, ruleInformationProvider, uuidGenerator),
         rules: getUnifiedRules(scanResults, ruleInformationProvider, uuidGenerator),
-        platformInfo: getPlatformData(scanResults),
+        platformInfo: getPlatformData(scanResults) ?? undefined,
```
RobGallo, line 37: What's the reason we don't want to possibly return undefined from getPlatformData?

src/electron/main/main.ts
```diff
@@ -81,7 +81,7 @@ const createWindow = () => {
     mainWindow.on('closed', () => {
         // Drop all references to the window object, to force garbage collection
         ipcMessageDispatcher.unregisterMessageSink(mainWindowMessageSink);
-        mainWindow = null;
+        mainWindow = null!;
```
DaveTryon, line 84: Doesn't the ! postfix operator throw an assertion? That seems like an odd choice for something that is part of the typical shutdown...

dbjorge: No; the `!` is called the "non-null assertion operator", but I think you're misunderstanding what the word "assertion" implies in the context of a "type assertion". The [docs for the operator](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-) are a good starting point; the usage here is essentially telling typescript "I know this variable's type suggests it can't be null, but please allow me to pass it a possibly-null (here: definitely null) value anyway". I'm doing this here because it's specifically in the context of the on-closed event, where we know nothing will be referencing it afterwards; the alternative would be to make the type of `mainWindow` nullable, but that felt worse because it forced unnecessary checks everywhere the variable is used.

src/electron/platform/android/unified-result-builder.ts
dbjorge: In general, my preference is to use `null` over `undefined` to represent "we explicitly decided to return a non-value here", which is what `getPlatformData`'s `null` case represents. I decided to keep its current implementation (which already returned `null` for this case) based on a combination of that preference and minimizing the change size.

The obvious followup question is "if `null` makes more sense here why not update `UnifiedScanCompletedPayload` to use that instead?" I debated how far to go here with myself for quite some time. Ultimately, there's not a lot of practical difference for this case (or most of our cases) in using `null` vs `undefined` (we typically check for it using `!= null`, which treats the cases identically), and all of the `unified-data-interface` types use `property?: Type` instead of `property: Type | null` throughout, so the options would have been either "do it for just the `platformInfo` property, but then it's inconsistent", "do it for all the properties, but then it's a huge PR for negligible value", or "just do the conversion here". I opted for the latter as the best of bad options.

RobGallo (approved):

PR #4020, opened 2021-03-10
feat(focus-mvp-android-device) Grant overlay permission

RobGallo:
#### Details

Add a step to the Android setup state machine which automatically grants the overlay permission required to display tabstops indicators on the Android device screen.

##### Motivation

Part of the tabstops feature

##### Context

The reason for automatically granting the permission is that in Android SDK version 30, the overlay permission is silently granted by the system when the media projection permission (`SYSTEM_ALERT_WINDOW`) is granted. If the user has a device running SDK version 29 or lower, since the user must grant the media projection permission, we believe it is acceptable to automatically grant the overlay permission as Google has chosen to do in SDK 30. If there is any objection to granting the overlay permission silently, we can of course add UX later to inform the user the permission is being granted on their behalf.

The granting of the permission is not currently behind the tabstops feature flag because 

- There is a lot of plumbing that would need to be done to add it, and then almost immediately undone when the feature is released soon.
- Granting the permission causes no ill effects even when an older version of the Accessibility Insights service, one which does not require the `SYSTEM_ALERT_WINDOW` permission, is installed on the device.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/electron/platform/android/appium-adb-wrapper.ts
+++ b/src/electron/platform/android/appium-adb-wrapper.ts
@@ -89,18 +89,16 @@ export class AppiumAdbWrapper implements AdbWrapper {
         await this.adb.shell(['input', 'keyevent', keyEventCode]);
     };
 
-    public grantOverlayPermission = async (deviceId: string): Promise<void> => {
+    public grantOverlayPermission = async (
+        deviceId: string,
+        packageName: string,
+    ): Promise<void> => {
         this.adb.setDeviceId(deviceId);
-        await this.adb.shell([
-            'cmd',
-            'appops',
-            'reset',
-            'com.microsoft.accessibilityinsightsforandroidservice',
-        ]);
+        await this.adb.shell(['cmd', 'appops', 'reset', packageName]);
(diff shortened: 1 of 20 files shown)


src/electron/platform/android/appium-adb-wrapper.ts
```diff
@@ -88,4 +88,20 @@ export class AppiumAdbWrapper implements AdbWrapper {
...
         await this.adb.shell(['input', 'keyevent', keyEventCode]);
     };
+
+    public grantOverlayPermission = async (deviceId: string): Promise<void> => {
+        this.adb.setDeviceId(deviceId);
+        await this.adb.shell([
+            'cmd',
+            'appops',
+            'reset',
+            'com.microsoft.accessibilityinsightsforandroidservice',
```
waabid, line 98: The android service configurator has a property for the package name that it uses to pass to other methods on the wrapper. I think we should likely take that same pattern here.

waabid (approved):

Issue #3995, opened 2021-03-02
Date stamp of report is always 19700101


Bram-diederik:
If you export a report it should have the date added to the file name. but with the current version: 2.23.0 
the date in the filename is always 19700101

jalkire:
I am able to repro this with the following steps:
1. Navigate to a new page
2. Launch assessment directly from the launchpad
3. Click start new on the dialog
4. Export result

If I first launch fastpass and then navigate to assessment, the problem goes away until I follow those steps again.

ferBonnin:
Per conversation with Peter, lets fix this. The time we should show on the report should be either the time the user selects the " export result" button or the time when the user did their first requirement. The first option is probably the more straightforward one.

ferBonnin:
Hi @Bram-diederik this has been addressed and its now available in our [Canary extension](https://chrome.google.com/webstore/detail/accessibility-insights-fo/hbcplehnakffdldhldncjlnbpfgogbem) please let us know what you think!

PR #4080, opened 2021-04-01
fix: update scanDate to get current date

shanisebarona:
#### Details

Update scanDate to return current date for the date stamp populated as part of the exported results file title.

##### Motivation

Addresses the bug in issue  #3995.

#### Pull request checklist

- [x] Addresses an existing issue:  #3995
- [x] Ran `yarn fastpass`
- [/] (UI changes only) Added screenshots/GIFs to description above
- [/] (UI changes only) Verified usability with NVDA/JAWS

### DIFF
--- a/src/DetailsView/components/report-export-dialog-factory.tsx
+++ b/src/DetailsView/components/report-export-dialog-factory.tsx
@@ -32,7 +32,7 @@ export function getReportExportDialogForAssessment(
         deps: deps,
         reportExportFormat: 'Assessment',
         pageTitle: scanMetadata.targetAppInfo.name,
-        scanDate: scanMetadata.timespan.scanComplete,
+        scanDate: props.deps.getCurrentDate(),
         htmlGenerator: description =>
             reportGenerator.generateAssessmentReport(
                 assessmentStoreData,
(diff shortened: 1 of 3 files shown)


dbjorge (approved):

dbjorge (changes requested):
It's surprising to me that this doesn't require a unit test update; there should probably be a test pinning the behavior here

shanisebarona:
> It's surprising to me that this doesn't require a unit test update; there should probably be a test pinning the behavior here

@dbjorge I thought so too! In `report-export-dialog-factory.test.tsx`, it's seems to already be mocking the date function as desired. 

`const theDate = new Date(Date.UTC(2019, 2, 12, 9, 0));`
 and `getCurrentDate: () => theDate` in the `deps`.

unless this itself should be updated?

dbjorge:
@shanisebarona Ideally we want the unit test to be verifying that the date used by the component comes from the place we expect it to; here, the test isn't verifying which of `scanCompleted` or `getCurrentDate` is being used. The way the test is wiring up the mock `getCurrentDate` implementation would be fine in isolation; the only issue is that the same date stub (`theDate`) is being re-used in several locations. Ideally, we'd want to pass in different dates for `scanCompleted` vs `getCurrentDate`, so the test can verify which one the actual component is using.

shanisebarona:
Thank you @dbjorge, that was helpful, I've updated the test :)

dbjorge (approved):

PR #4088, opened 2021-04-06
feat: Show friendlier device name in unified client

DaveTryon:
#### Details

Currently, the unified client identifies the device only by the value that gets reported from Android's [Build.OS.MODEL](https://developer.android.com/reference/android/os/Build#MODEL) field. On emulators, this reports something like "AOP on IA Emulator". On physical devices, instead of reporting a user-friendly value, it reports a model code--on a Samsung Galaxy Tab S4, for example, it reports SM-T830. That's not very useful, I'm afraid...

After discussion with @juchampa, this PR uses the [android-device-list](https://www.npmjs.com/package/android-device-list) package to convert the codes for physical devices into a string that includes both the code and the market name of the device. This is closer to the original intent of the friendly name in the unified client. The friendly name gets inserted immediately after we fetch the devices from ADB, so it flows seamlessly through all setup-related workflows. It isn't yet integrated into the exported results, but that would be a simple extension once this PR is merged.

##### Motivation

Improve the user experience with physical devices (this has no impact on emulators since emulators don't have a corresponding entry in the list of known devices).

##### Context

The [android-device-list](https://www.npmjs.com/package/android-device-list) library has 2 components--a runtime component and an optional build-time script. The runtime component adds 2 json files (about 3.5MB total) to the local install and uses the local json files to convert from the model code to a string that users would recognize. The list remains in memory throughout the lifetime of the process, with no current way to free the memory (not a huge deal, but just calling it out for completeness).

By default, the device list gets updated when the package updates (typically every couple of months), so it's possible to find ourselves with an outdated device list if the package updates were to stop occurring (at the moment, only 1 person is a contributor to the project, and previous PR's received no reply from the maintainer). That said, the optional build-time script grabs the data directly from [Google's documented location](https://storage.googleapis.com/play_public/supported_devices.csv) and creates updated versions of the json files, meaning that we have the option to always have the most recent device lists if we eventually choose to incorporate it into our build pipeline. 

##### Screenshots:
Before the change, using a Samsumg Galaxy Tab S4 (the model code is highlighted)
image

After the change, using the same device (the friendly name is highlighted)
image

The [android-device-list](https://www.npmjs.com/package/android-device-list) library is MIT license, so we could always make our own version and completely remove the external dependency, but this seems counter to open-source principles. It seems best to use it (or some other package that might better suit our needs) until something actually prevents us from using it.

#### Pull request checklist

- [x] Ran `yarn fastpass`


DaveTryon:
Moving from draft since @juchampa has approved this user experience

jalkire (approved):
🚢

PR #4095, opened 2021-04-08
fix(performance): avoid unnecessary drawer updates

dbjorge:
#### Details

`TargetPageVisualizationUpdater` is responsible for tracking the current state of visualizers in a specific target page and responding to changes to the visualization store by checking whether a new store state mismatches a tracked visualizer state, triggering the appropriate `drawerInitiator` methods if any drawer needs to be updated to a new state.

The `previousVisualizationSelectorMapStates` it maintains is a mapping from `VisualizationType` to a `SelectorMap`, where ~a `SelectorMap` is a mapping from a particular step/requirement's `configId` to an `AssessmentVisualizationInstance` with information about which selector(s) are relevant to that requirement.~ edit: see comment below

The point where this map is *read* (L84, where it's passed it to `visualizationNeedsUpdate`) correctly uses this state by indexing into it using `visualizationType`s, but the point where it is *written* (L58) was incorrectly indexing into it using `configId`s. The compiler doesn't catch this because our compiler settings are not strict enough; it does detect this as an error with `"noImplicitAny": true` added to `tsconfig.base.ts`, but unfortunately it also catches 938 other errors, many of which are false positives.

The actual error is mostly-benign because of the way `visualization-needs-update` uses this field; there is no overlap between `configId`s (strings) and `visualizationType`s (int enums), so `visualizationNeedsUpdate` always ends up thinking that the previous selector map for the visualization it's considering was `undefined`. It compares it to the output of `SelectorMapHelper.getSelectorMap`, which is not very consistent (it uses empty objects for some empty cases and `null` for others) but does consistently avoid returning `undefined`), and that comparison always shows as "unequal". The practical effect is that it will say that *every* visualization type which has ever been enabled previously on that target page will always show as needing an update.

The three cases where these updates aren't necessary are:
1. A previously-enabled visualization is still enabled, and selector map state hasn't changed.
2. A previously-disabled visualization is still disabled, and selector map state hasn't changed.
3. A previously-disabled visualization is still disabled, even if selector map state has changed.

The bug fix in `target-page-visualization-updater` fixes cases 1 and 2 to not result in updates. The fix in `visualization-needs-update` is for case 3.

##### Motivation

In pages with iframes, each drawer intitiation involves a lot of messaging traffic between the frames, which slightly delays the legitimate messages for drawer updates that are actually required. The result is that after this PR, toggling visualizations will feel a little snappier in pages with lots of iframes (especially nested iframes).

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/target-page-visualization-updater.ts
+++ b/src/injected/target-page-visualization-updater.ts
@@ -15,7 +15,7 @@ import {
 } from './visualization-needs-update';
 
 type TestStepVisualizationStateMap = {
-    [visualizationType in VisualizationType]: {
+    [visualizationType: number]: {
         [testStepConfigId: string]: TestStepVisualizationState;
     };
 };
@@ -26,7 +26,7 @@ export type UpdateVisualization = (
     storeData: TargetPageStoreData,
 ) => void;
 export class TargetPageVisualizationUpdater {
-    private previousVisualizationStates: TestStepVisualizationStateMap;
+    private previousVisualizationStates: TestStepVisualizationStateMap = {};
 
     constructor(
         private visualizationConfigurationFactory: VisualizationConfigurationFactory,
--- a/src/injected/target-page-visualization-updater.ts
+++ b/src/injected/target-page-visualization-updater.ts
@@ -55,7 +55,7 @@ export class TargetPageVisualizationUpdater {
             configuration,
             configId,
         );
-        this.previousVisualizationSelectorMapStates[configId] = selectorMap;
+        this.previousVisualizationSelectorMapStates[visualizationType] = selectorMap;
     };
 
     private executeUpdate = (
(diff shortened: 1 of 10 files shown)


RobGallo (dismissed):

dbjorge:
@waabid pointed out in an offline code review that I had misunderstood `SelectorMap`, and convinced me that it was more correct to continue to store the map state on a per-`configId` basis rather than a per-`visualizationType` basis. I've substantially rewritten the two classes to make this more obvious!

src/injected/target-page-visualization-updater.ts
```diff
@@ -105,4 +76,39 @@ export class TargetPageVisualizationUpdater {
...
+
+    private updatePreviousVisualizationState(
+        visualizationType: VisualizationType,
+        configId: string,
+        newState: TestStepVisualizationState,
+    ) {
+        if (this.previousVisualizationStates[visualizationType] === undefined) {
+            this.previousVisualizationStates[visualizationType] = {};
+        }
+        this.previousVisualizationStates[visualizationType][configId] = newState;
```
waabid, line 88: nit?: could also write this as:

```
this.previousVisualizationStates[visualizationType] = {
            ...this.previousVisualizationStates[visualizationType],
            [configId]: newState,
};
```

waabid: Either work, but depends on your style preference, of course.

waabid (approved):
LGTM

PR #4115, opened 2021-04-14
fix: fix telemetry for tab stops in android

waabid:
#### Details

Passes through telemetry for publishing unlike before.

##### Motivation

Need telemetry on tab stop usage.

##### Context

The main "fix" is around the fact that the payload for publish telemetry cannot be null. If it is, nothing is published. This is why key event telemetry already works.

Passed through events from UI so I can use the telemetry data factory. Seemed appropriate considering that's how we usually do things.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/electron/views/results/components/reflow-command-bar.test.tsx
+++ b/src/tests/unit/tests/electron/views/results/components/reflow-command-bar.test.tsx
@@ -15,6 +15,7 @@ import {
     ReflowCommandBarProps,
 } from 'electron/views/results/components/reflow-command-bar';
 import { mount, shallow } from 'enzyme';
+import { isMatch } from 'lodash';
 import { IButton } from 'office-ui-fabric-react';
 import * as React from 'react';
 import { ReportGenerator } from 'reports/report-generator';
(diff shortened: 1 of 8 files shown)


src/tests/unit/tests/electron/views/results/components/reflow-command-bar.test.tsx
```diff
@@ -72,7 +72,7 @@ describe('ReflowCommandBar', () => {
                 allowsExportReport: true,
                 startOverButtonSettings: _ => {
                     return {
-                        onClick: () => {},
+                        onClick: _ => {},
```
dbjorge, line 75: GitHub won't let me leave a comment far away from a change (grumble), but it'd be nice if the interaction test down on L134 could verify that the expected event stub gets passed through

dbjorge (changes requested):
One minor test suggestion, otherwise LGTM

src/tests/unit/tests/electron/views/results/components/reflow-command-bar.test.tsx
waabid: Updated!

dbjorge (approved):

Issue #4102, opened 2021-04-08
Mark "duplicate-id" rule as a best practice rule 


iamrafan:
Based on feedback from users we tested to check the user impact of duplicate ID failures on static elements. We found no user impact on the experience with any of the ATs including Narrator, JAWS, NVDA, Dragon, Windows Speech Recognition.

After further discussions with the Tooling Advisory Board, we decided to make this a best practice rule

## Context
- AI-Web Version : Prod 2.24.0

tohmjudson-ms:
Happy to pick this fix up

ferBonnin:
thanks @tohmjudson-ms! assigning to you and marking this as active

dbjorge:
Resolved by @tohmjudson-ms in #4127

ferBonnin:
this is now in production, @iamrafan can you please verify and close?

iamrafan:
Validated in AI Web Prod 2.26.0

PR #4127, opened 2021-04-21
feat(duplicate-id): exclude duplicate-id rule from automated checks

tohmjudson-ms:
#### Details

Adds a custom rule to tag 'duplicate-id' errors as best practice

##### Motivation

Addresses issue #4102

##### Context

This PR only addresses `duplicate-id`, which is marked as **Impact**: Minor by axe-core. It does not change and not `duplicate-id-active` (**Impact**: Serious) or `duplicate-id-aria` (**Impact**: Critical).

#### Pull request checklist

- [x] Addresses an existing issue: #4102 
- [x] Ran `yarn fastpass`


tohmjudson-ms:
Note: this most likely needs a test as well. I could definitely use some help on that to meet standards

dbjorge:
Thanks for the contribution, @tohmjudson-ms! Our usual mechanism for suppressing rules that axe-core includes by default is the `explicitRuleOverrides` variable in `/src/scanner/get-rule-inclusions.ts`; I think that would probably be the preferred place to update, rather than using a custom rule config. You can use the description from #4102 in the `reason` field, something like this:

```ts
    'duplicate-id': {
        status: 'excluded',
        reason: `
            Based on feedback from users we tested to check the user impact of duplicate ID failures
            on static elements. We found no user impact on the experience with any of the ATs
            including Narrator, JAWS, NVDA, Dragon, Windows Speech Recognition. After further
            discussions with the Tooling Advisory Board, we decided to make this a best practice
            rule. See #4102.`,
    },
```

To update the corresponding unit test, you'd need to update a jest snapshot (`yarn test -- -- -u -- get-rule-inclusions.test`)

It looks like there are also a few end to end tests that involve pinning scan results which are impacted by this change; the PR build results should point you at which test cases need updates. It looks like there will be one case from the web e2es you'll need to change a number for and that you'll need to regenerate snapshots for the report e2es (`yarn test:report:e2e -u`)

tohmjudson-ms:
Awesome @dbjorge! That makes things so much more clear to me, thank you! I updated as you requested. I still need to update the test cases, which I hope to do very shortly

dbjorge (approved):
Thanks for the contribution, LGTM!

Issue #4140, opened 2021-04-22
Unified client's controls are not visible when running on Linux in Parallels VM


DaveTryon:
## Describe the bug

The combination of Electron, Linux, and  the latest Parallels Tools (installed if you're running on a Mac using Parallels VM) have a known problem. If hardware acceleration is enabled (which is the default state), then the window's background color gets displayed, but the window contents don't render on screen. The controls are still there and accessible via keyboard or mouse, but they're not visible.

VSCode reports this on [ubuntu](https://github.com/microsoft/vscode/issues/111290) and [debian](https://github.com/microsoft/vscode/issues/118710), and has a [documented workaround](https://code.visualstudio.com/updates/v1_40#_disable-gpu-acceleration) of launching the app with hardware acceleration disabled (either by command line or by a config file).

**To Reproduce**

1. Open a linux VM (I used ubuntu 20.04) under Parallels
2. Start Accessibility Insights for Android
3. The window contents are all white (see the screenshot)

## Expected behavior

Window contents should display correctly

## Screenshots

This is captured from the machine in question (the "frame" is just the default desktop):
![image](https://user-images.githubusercontent.com/45672944/115792013-5c9e0a00-a37e-11eb-8c23-70da1f479312.png)

## Context (please complete the following information)

- OS Name & Version: Ubuntu 20.04 LTS
- AI-Android Version & Environment: 2121.414.9 Production

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?
Yes

## Additional context
I've reported this to the Parallels support team, but there's no way to know how quickly they might be able to update their software.

DaveTryon:
This has been fixed.

PR #4141, opened 2021-04-22
fix: Add display workaround for linux in Parallels VM's

DaveTryon:
#### Details

Disable hardware acceleration when running on Linux as a workaround for an issue in the Parallels VM software.

##### Motivation

Address issue #4140  

##### Context

This is an unfortunately broad change, impacting all linux installations, including those not running in a VM. I'd love to scope it down to apply only when running in a Parallels VM (or even only in a VM), but found no javascript-friendly to make that determination. It might be possible to exec **systemd-detect-virt**, but that felt "heavy". Our graphics are very limited (display a screenshot with highlights), so disabling hardware acceleration doesn't seem to have a discernible impact on the product.

#### Screenshots
These are taken on the same machine running Ubuntu 20.04 in a Parallels VM:
 Before | After
--- | ---
image | image

#### Pull request checklist

- [x] Addresses an existing issue: #4140
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -32,6 +32,7 @@ log.transports.file.level = 'info';
 autoUpdater.logger = log;
 
 if (platformInfo.isLinux()) {
+    // Avoid a conflict between Parallels VM and Electron. See https://github.com/microsoft/accessibility-insights-web/issues/4140
     app.disableHardwareAcceleration();
 }
 
--- a/src/electron/main/main.ts
+++ b/src/electron/main/main.ts
@@ -31,6 +31,10 @@ const nativeHighContrastModeListener = new NativeHighContrastModeListener(
 log.transports.file.level = 'info';
 autoUpdater.logger = log;
 
+if (platformInfo.isLinux()) {
+    app.disableHardwareAcceleration();
+}
+
 let recurringUpdateCheck;
 const electronAutoUpdateCheck = new AutoUpdaterClient(autoUpdater);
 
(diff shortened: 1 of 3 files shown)


src/electron/main/main.ts
```diff
@@ -31,6 +31,10 @@ const nativeHighContrastModeListener = new NativeHighContrastModeListener(
 log.transports.file.level = 'info';
 autoUpdater.logger = log;
 
+if (platformInfo.isLinux()) {
```
DaveTryon, line 34: @dbjorge, I'd love to get your thoughts on how we might be able to make this more selective

dbjorge: I spent 15 minutes researching and came to the same conclusion you did. It looks like most utilities for detecting "parallels" vs not boil down to using [DMI](https://en.wikipedia.org/wiki/Desktop_Management_Interface) to check for a Parallels-specific BIOS entry, but the tools available to do that query seem to vary per distro (we probably shouldn't assume the presence of systemd, in particular, but the most common alternatives require more privileges than we're going to have).

Given that there isn't a substantial performance difference, I agree with just disabling it on linux generally. We have a very, very low proportion of linux users already, so I feel it's hard to argue a strong value prop for doing anything complicated here.

It would be nice to leave a comment here explaining why we're disabling it on linux with a pointer to the issue, though!

DaveTryon: Comment added

dbjorge (approved):

Issue #3938, opened 2021-02-18
Android: No graceful recovery from "Disconnected device" scenario


DaveTryon:
## Describe the bug

If the user disconnects their Android device, we have a device disconnected dialog and associated recovery path that attempt to get the user back into a usable state. Due to port forwarding and ADB's behavior, though, the recovery doesn't work and the user's only viable option seems to be to exit the app and restart.

The root problem is that ADB knows when a device is disconnected and removes any port forwarding for that device.

**To Reproduce**
1. Open a command prompt where you can run adb
2. Before starting the unified client, run **adb forward --list** and you'll get back an empty string
3. Start the unified client and run an automated scan
4. From your command prompt, run **adb forward --list**. You'll get back the device ID, the local port, and the device port, something like `ce11182b4587381d0b7e tcp:62442 tcp:62442`
5. Disconnect the device. If you run **adb forward --list**, you'll get back an empty string because ADB automatically removes the port forwarding when the device is disconnected
6. Click "Start over". You'll get the device disconnected dialog. 
    1. If you choose the rescan option, you'll just get the same device disconnected dialog
    2. If you choose to connect a new device, you'll be taken back to the "Connected and ready to go!" dialog (independent of if you've actually connected the device or not, which is a related bug). Assuming you've attached the device, port forwarding is not configured, so you can't talk to the service.
    3. If you choose to go all the way back to the page where we scan for devices, you'll eventually reach a point where you receive the "Port forwarding failed" dialog, and you're stuck. I've confirmed that occurs on Mac and Windows, and am assuming that it will also happen on Linux.

## Expected behavior

We should be able to gracefully recover and get the user connected to the same or a different device. The flow should follow the same pattern as connecting a device when the app is first launched.

Some implementation ideas:
- Remove the "Rescan" button from the dialog if the port forwarding no longer exists (detect using `adb forward --list`).
- When the user clicks on "Reconnect device", start with the first setup step instead of the last, so the user always goes through all setup steps. Most of these will transition automatically if the service is already running on the device, so the user impact is negligible in the simple case.
- Explicitly clear the the forwarded port every time we enter the port forwarding step. If this seems too heavy-handed, we could choose to keep the port previously mapped to this device if `adb forward --list` shows that an appropriate port mapping already exists.

## Context (please complete the following information)

- OS Name & Version: Repro'ed on Windows and Mac, didn't check Linux
- AI-Web Version & Environment: Current unified builds

## Are you willing to submit a PR?

[...]

PR #4149, opened 2021-04-27
fix: Improve disconnected device experience

DaveTryon:
#### Details

Improve our disconnected device scenario. As called out in #3938, when the device connection is removed, ADB automatically removes port forwarding, which breaks assumptions in the reconnect code (which pre-dates the auto-configuration of port forwarding).

The PR basically consists of 3 pieces:
1. We already try to remove previously configured port forwarding, but we don't account for the error that gets thrown if we're removing a non-existent port. Fix this.
2. The notion of the "Start over" button is doomed to fail if we've really lost the connection, because we've lost our port forwarding and can only restore it by going through the device setup steps. I've updated the disconnected dialog to remove the "Start over" button.
3. When the user clicks the "Connect a new device" button, we return to the device setup view. Its state is still on the "start testing" page, which is doomed to fail because it's after the port forwarding step. I've changed the code to reset the state to the "connect-device" step when the users presses "Start testing", so there's always a consistent experience for reconnecting. To better reflect the workflow, I changed the "Connect a new device" button (in the disconnected dialog) to "Re-detect device", then we let the user connect an appropriate device before proceeding. @juchampa has reviewed and approved the user experience.

##### Motivation

Address #3938.

##### Screenshots

Connection dialog 
Before | After
--- | ---
image | image

Destination dialog after clicking through (these dialogs are unchanged, but shown purely to convey the user experience--note that the old experience landed in the "start testing" step with all previous information intact, while the new experience lands in the "connect your device" step):
Before | After
--- | ---
image | image

##### Context

This assumes that we're really disconnected when we display the device disconnected dialog. If the service were to throw an error (as opposed to being disconnected), then it's overkill to reconnect the device. I'm not aware of any user feedback to suggest that we're getting service errors, but I'd argue for a separate error handler dialog if/when we feel the need to make the distinction between the 2 error conditions.

I experimented with auto-detecting the newly connected device when the user clicked from the disconnected device dialog, but it added some unfortunate coupling between the results view (where the disconnected dialog is hosted) and the setup action creator. Always returning to the "Connect a device" step avoided this. Users may need to click 1 more button, but it lets us leverage an existing esperience.

#### Pull request checklist

- [x] Addresses an existing issue: #3938
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/electron/platform/android/setup/steps/configuring-port-forwarding.ts
+++ b/src/electron/platform/android/setup/steps/configuring-port-forwarding.ts
@@ -3,17 +3,25 @@
 
 import { AndroidSetupStepConfig } from 'electron/platform/android/setup/android-setup-steps-configs';
 
+const removeOldForwardedPort = async (deps, store) => {
+    const existingPort = store.getScanPort();
+    if (existingPort != null) {
+        try {
+            deps.logger.log(`removing old tcp:${existingPort} forwarding`);
+            await deps.removeTcpForwarding(existingPort);
+            store.setScanPort(null);
+        } catch (e) {
+            deps.logger.log(`Ignoring error : ${e}`);
+        }
+    }
+};
+
 export const configuringPortForwarding: AndroidSetupStepConfig = (stepTransition, deps, store) => ({
     actions: {},
     onEnter: async () => {
         try {
-            const existingPort = store.getScanPort();
-            if (existingPort != null) {
-                deps.logger.log(`removing old tcp:${existingPort} forwarding`);
-                await deps.removeTcpForwarding(existingPort);
-                store.setScanPort(null);
-                store.setApplicationName(null);
-            }
+            await removeOldForwardedPort(deps, store);
+            store.setApplicationName(null);
 
             const hostPort = await deps.setupTcpForwarding();
             deps.logger.log(`configured forwarding to tcp:${hostPort}`);
--- a/src/electron/platform/android/setup/steps/configuring-port-forwarding.ts
+++ b/src/electron/platform/android/setup/steps/configuring-port-forwarding.ts
@@ -1,11 +1,16 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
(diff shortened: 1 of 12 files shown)


src/electron/platform/android/setup/steps/configuring-port-forwarding.ts
```diff
@@ -1,19 +1,36 @@
...
+const removeOldForwardedPort = async (
+    deps: AndroidSetupDeps,
+    store: AndroidSetupStoreCallbacks,
+) => {
+    const existingPort = store.getScanPort();
+    if (existingPort) {
+        try {
+            deps.logger.log(`removing old tcp:${existingPort} forwarding`);
+            await deps.removeTcpForwarding(existingPort);
+            store.setScanPort(null);
```
RobGallo, line 17: Don't we want to set the scan port to null regardless of whether or not an exception was thrown?

src/tests/unit/tests/electron/views/device-disconnected-popup/device-disconnected-popup.test.tsx
```diff
@@ -18,32 +18,18 @@ describe('DeviceDisconnectedPopup', () => {
     });
 
     describe('user interaction', () => {
-        it('handles on connect a new device', () => {
-            const onConnectNewDeviceMock: () => void = jest.fn(() => {});
+        it('handles on redetect device', () => {
+            const onRedecteDeviceMock: () => void = jest.fn(() => {});
```
RobGallo, line 22: Spelling: "onRedect"

src/electron/platform/android/setup/steps/configuring-port-forwarding.ts
DaveTryon: Yup. Moved after the try/catch block

src/tests/unit/tests/electron/views/device-disconnected-popup/device-disconnected-popup.test.tsx
DaveTryon: Fixed

RobGallo (approved):

PR #4177, opened 2021-04-30
feat(complete-save-load-assessment): Refactor Change Assessment Dialog and Wire Up Load Assessment Dialog

brocktaylor7:
#### Details

image

 - refactors target change dialog to use new change assessment dialog
 - creates and wires up new load assessment dialog
 - updated tests

##### Motivation

Wiring up of the feature

##### Context

We decided that it would be best to have a single component responsible for changing assessments whether it's a target change or a load. Much of this PR is refactoring to make that work properly.  Previously there was a LoadAssessmentHelper that handled all the loading logic that is now in the LoadAssessmentButton class, but due to some additional technical complications it would introduce, we decided to consolidate the LoadAssessmentHelper into the LoadAssessmentButton class.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/load-assessment-button.tsx
```diff
@@ -3,31 +3,125 @@
...
+        }
+
+        return {
+            id: this.props.tabStoreData.id,
+            url: this.props.tabStoreData.url,
+            title: this.props.tabStoreData.title,
+        };
+    }
+
+    public getAssessmentForLoad() {
```
waabid, line 73: It might make sense to move this back to the separate method/class and instead add parameters for what this requires from the button (state and function that toggles load dialog).

src/DetailsView/components/load-assessment-button.tsx
```diff
@@ -3,31 +3,125 @@
...
+                </InsightsCommandButton>
+
+                <LoadAssessmentDialog
+                    {...this.props}
+                    tabId={this.props.tabStoreData.id}
+                    loadedAssessmentData={this.state.loadedAssessmentData}
+                    prevTab={this.state.prevTab}
+                    newTab={this.state.newTab}
+                    show={this.state.show}
+                    onClose={this.toggleLoadDialog.bind(this)}
```
waabid, line 122: FYI: this is not necessary. We can just make toggleLoadDialog an arrow method and pass in appropriately.

src/DetailsView/components/load-assessment-button.tsx
```diff
@@ -3,31 +3,125 @@
...
-                iconProps={{ iconName: 'FabricOpenFolderHorizontal' }}
-                onClick={() => this.props.deps.loadAssessmentHelper.getAssessmentForLoad()}
-            >
-                Load assessment
-            </InsightsCommandButton>
+            <>
+                <InsightsCommandButton
+                    data-automation-id={loadAssessmentButtonAutomationId}
+                    iconProps={{ iconName: 'FabricOpenFolderHorizontal' }}
+                    onClick={() => this.getAssessmentForLoad()}
```
waabid, line 110: Just a minor note here that doesn't ultimately matter (since this is the format we'll end up having with the above suggested changes) but in this case, if we were to not pass any local parameters to method, we could simply pass "this.getAssessmentForLoad" to the onClick prop.

src/DetailsView/components/load-assessment-dialog.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const dialogProps: ChangeAssessmentDialogProps = {
+        deps: props.deps,
+        prevTab: props.prevTab,
+        newTab: props.newTab,
+        dialogContentTitle: 'Assessment in progress',
+        subtitleAriaId: 'load-assessment-dialog-description',
+        divId: 'load-assessment-dialog-description',
+        leftButtonText: 'Continue previous',
+        leftButtonOnClick: continuePreviousAssessment,
+        rightButtonText: 'Load Assessment',
```
waabid, line 55: Should be: "Load **a**ssessment"

src/DetailsView/components/load-assessment-button.tsx
```diff
@@ -3,31 +3,125 @@
...
+    fileReader: FileReader;
+    document: Document;
 };
 export interface LoadAssessmentButtonProps {
     deps: LoadAssessmentButtonDeps;
+    tabStoreData: TabStoreData;
+    assessmentStoreData: AssessmentStoreData;
+}
+export interface LoadAssessmentButtonState {
+    newTab: Tab;
```
waabid, line 31: nit: I think we need to move away from the 'Tab' moniker in this kind of use-case. The idea of a tab doesn't have much to do with this and while I don't like using `Tab` the interface due to previous naming, I definitely want to avoid props with that naming convention.

Ideas:
- `fromData` and `toData`
- `previousTargetPageData' and `newTargetPageData`

src/DetailsView/components/load-assessment-button.tsx
```diff
@@ -3,31 +3,125 @@
...
+    prevTab: PersistedTabInfo;
```
waabid, line 32: Here and ChangeAssessmentDialog: we should not use "PersistedTabInfo" unless we need to.

waabid (changes requested):
Left some comments.

src/DetailsView/components/load-assessment-dialog.tsx
brocktaylor7: good catch here.

src/DetailsView/components/load-assessment-button.tsx
brocktaylor7: happy with this suggestion, I used previousTargetPageData and newTargetPageData

src/DetailsView/components/load-assessment-helper.ts
```diff
@@ -1,34 +1,40 @@
...
 
-export class LoadAssessmentHelper {
-    constructor(
-        private readonly assessmentDataParser: AssessmentDataParser,
-        private readonly detailsViewActionMessageCreator: DetailsViewActionMessageCreator,
-        private readonly fileReader: FileReader,
-        private readonly document: Document,
-        private readonly tabId: number,
-    ) {}
+export function getAssessmentForLoad(
```
waabid, line 7: I think my suggestion here was more so to _go back_ to what we previously had with load-assessment-helper. Specifically you can create the class with the non-dynamic parameters and then the latter can be passed when the method is called.

This would also avoid adding a static dependency as this currently does to LoadAssessmentButton.

src/DetailsView/components/load-assessment-button.tsx
waabid: As far as I can tell, this has not been fixed.

We should use the Tab interface in these places if that suffices. The PersistedTabInfo interface links with our store data/app knowledge and we should limit components that have it, unless necessary.

brocktaylor7: Don't know how I missed that the first time around, sorry about that! I've implemented this change now.

waabid:
Noticed this discrepancy between prod and this branch. Is this intentional (left is prod, right is branch; ignore high contrast):

image

src/DetailsView/components/load-assessment-button.tsx
```diff
@@ -3,31 +3,104 @@
...
+            prevTargetPageData: props.assessmentStoreData.persistedTabInfo,
+            show: false,
+        };
+    }
+
+    private toggleLoadDialog = () => {
+        this.setState(prevState => ({ show: !prevState.show }));
+    };
+
+    private getNewTab(): Tab {
```
waabid, line 54: Not sure if I follow the intention behind this code: this is only called in the constructor (and that's only called on creation of the component). As a result, it will always return the tab store data and we never enter the if-block. Is there something I'm missing here?

src/DetailsView/components/change-assessment-dialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { DefaultButton, DialogFooter, DialogType, TooltipHost } from 'office-ui-fabric-react';
+import * as React from 'react';
+export interface ChangeAssessmentDialogProps {
+    prevTab: Tab;
+    newTab: Tab;
+    dialogContentTitle: string;
+    subtitleAriaId: string;
+    divId: string;
+    leftButtonText: string;
+    leftButtonOnClick;
```
waabid, line 20: Missing types on these props.

src/DetailsView/components/load-assessment-helper.ts
```diff
@@ -9,18 +10,31 @@ export class LoadAssessmentHelper {
         private readonly detailsViewActionMessageCreator: DetailsViewActionMessageCreator,
         private readonly fileReader: FileReader,
         private readonly document: Document,
-        private readonly tabId: number,
     ) {}
 
-    public getAssessmentForLoad() {
+    public getAssessmentForLoad(
+        setAssessmentState: Function,
```
waabid, line 16: We should use appropriate types for these functions instead of the `Function` type.

src/DetailsView/components/change-assessment-dialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { NewTabLink } from 'common/components/new-tab-link';
+import { Tab } from 'common/itab';
+import { NamedFC } from 'common/react/named-fc';
+import * as commonDialogStyles from 'DetailsView/components/common-dialog-styles.scss';
+import * as styles from 'DetailsView/components/target-change-dialog.scss';
+import { DefaultButton, DialogFooter, DialogType, TooltipHost } from 'office-ui-fabric-react';
+import * as React from 'react';
+export interface ChangeAssessmentDialogProps {
+    prevTab: Tab;
+    newTab: Tab;
```
waabid, line 15: This newTab is not required since it's not being used in the component anywhere.

src/DetailsView/components/load-assessment-button.tsx
```diff
@@ -3,31 +3,104 @@
...
 };
 export interface LoadAssessmentButtonProps {
     deps: LoadAssessmentButtonDeps;
+    tabStoreData: TabStoreData;
+    assessmentStoreData: AssessmentStoreData;
+}
+export interface LoadAssessmentButtonState {
+    newTargetPageData: Tab;
+    prevTargetPageData: Tab;
+    loadedAssessmentData?: VersionedAssessmentData;
```
waabid, line 30: Any reason for this to be optional? Since we set it as null in constructor, I imagine it will always exist (even if it is null).

src/DetailsView/components/load-assessment-dialog.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    const continuePreviousAssessment = (event: React.MouseEvent<any, MouseEvent>) => {
+        props.deps.detailsViewActionMessageCreator.continuePreviousAssessment(event);
+        props.onClose();
+    };
+
+    const dialogProps: ChangeAssessmentDialogProps = {
+        deps: props.deps,
+        prevTab: props.prevTab,
+        newTab: props.newTab,
```
waabid, line 48: Since this is just being passed through and is not used later, we can also remove this.

src/DetailsView/components/load-assessment-button.tsx
```diff
@@ -3,31 +3,104 @@
...
-export class LoadAssessmentButton extends React.Component<LoadAssessmentButtonProps> {
+export class LoadAssessmentButton extends React.Component<
+    LoadAssessmentButtonProps,
+    LoadAssessmentButtonState
+> {
+    public constructor(props) {
+        super(props);
+        this.state = {
+            loadedAssessmentData: null,
+            newTargetPageData: this.getNewTab(),
```
waabid, line 44: Since we don't need this here or the child components, we can actually remove this entirely. The only place we would use it, line 87 in this file, can be replaced by using `this.props[...]`.

src/DetailsView/components/load-assessment-helper.ts
brocktaylor7: addressed this.

src/DetailsView/components/load-assessment-button.tsx
brocktaylor7: there wasn't a reason, I removed the optional

src/DetailsView/components/load-assessment-dialog.tsx
brocktaylor7: this has been removed.

src/DetailsView/components/change-assessment-dialog.tsx
brocktaylor7: addressed this.

src/DetailsView/components/load-assessment-button.tsx
brocktaylor7: this was needed but I was missing where it should've been called, I added the logic in that uses this as intended.

src/DetailsView/components/load-assessment-button.tsx
brocktaylor7: added in logic that uses this. It should've been there in the first place, it was an oversight on my part.

src/DetailsView/components/change-assessment-dialog.scss
```diff
@@ -0,0 +1,41 @@
...
+            margin-top: 4px;
+        }
+        .action-cancel-button-col + .action-cancel-button-col {
+            margin-left: 8px;
+        }
+        button {
+            padding: 0;
+            height: 32px;
+        }
+        .continue-button {
```
waabid, line 35: This style is probably not necessary since you specify load/restart separately.

waabid: Talked offline: misunderstood this one and resolving.

src/tests/end-to-end/common/element-identifiers/details-view-selectors.ts
```diff
@@ -76,6 +76,8 @@ export const overviewSelectors = {
     outcomeSummaryBar: getAutomationIdSelector(outcomeSummaryBarAutomationId),
     testOutcomeChips: (testName: string) =>
         getAutomationIdSelector(testSummaryStatusAutomationId(testName)) + ' .outcome-chip',
+    loadAssessmentDialog: '.ms-Dialog-inner',
```
waabid, line 79: We generally prefer not to use CSS styles for this. In this case specifically, these are also styles that we do not own. I would find a way to use data-automation-id here.

brocktaylor7: I was able to put a data-automation-id on the button, this has been updated!

waabid (approved):
LGTM

PR #4193, opened 2021-05-06
feat(complete-save-load-assessment): rewire load assessment dialog to fix reflow issue

brocktaylor7:
#### Details

This rewires the load assessment dialog to render it from the details view command bar rather than from the load assessment button.  This was necessary because when the button would reflow the dialog would disappear.

##### Motivation

To fix the reflow issue

##### Context

I debated whether to put the state in the details view command bar or on the load assessment dialog itself.  I decided to handle it similar to how things are handled for other rendered dialogs in the details view command bar and kept the stateful information in that component.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/details-view-command-bar.tsx
+++ b/src/DetailsView/components/details-view-command-bar.tsx
@@ -10,6 +10,7 @@ import {
     UnifiedScanResultStoreData,
 } from 'common/types/store-data/unified-data-interface';
 import { VisualizationStoreData } from 'common/types/store-data/visualization-store-data';
+import { VersionedAssessmentData } from 'common/types/versioned-assessment-data';
 import { VisualizationType } from 'common/types/visualization-type';
 import { DetailsViewActionMessageCreator } from 'DetailsView/actions/details-view-action-message-creator';
 import { CommandBarButtonsMenu } from 'DetailsView/components/command-bar-buttons-menu';
@@ -20,6 +21,10 @@ import {
     LoadAssessmentButton,
     LoadAssessmentButtonDeps,
 } from 'DetailsView/components/load-assessment-button';
+import {
+    LoadAssessmentDialogFactoryDeps,
+    LoadAssessmentDialogFactoryProps,
+} from 'DetailsView/components/load-assessment-dialog-factory';
 import { NarrowModeStatus } from 'DetailsView/components/narrow-mode-detector';
 import { ReportExportButton } from 'DetailsView/components/report-export-button';
 import { ReportExportDialogFactoryProps } from 'DetailsView/components/report-export-dialog-factory';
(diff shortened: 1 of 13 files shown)


src/DetailsView/components/details-view-command-bar.tsx
```diff
@@ -229,11 +242,49 @@ export class DetailsViewCommandBar extends React.Component<
 
     private renderLoadAssessmentButton = (): JSX.Element | null => {
         if (this.props.featureFlagStoreData.saveAndLoadAssessment) {
```
waabid, line 244: I think we can use FlaggedComponent here instead of using the feature flag like this.

src/DetailsView/components/details-view-command-bar.tsx
```diff
@@ -229,11 +242,49 @@ export class DetailsViewCommandBar extends React.Component<
...
+            ...this.props,
+            isOpen: this.state.isLoadAssessmentDialogOpen,
+            prevTab: this.props.assessmentStoreData.persistedTabInfo,
+            loadedAssessmentData: this.state.loadedAssessmentData,
+            tabId: this.props.tabStoreData.id,
+            onClose: this.toggleLoadAssessmentDialog,
+        });
+    };
+
+    private toggleLoadAssessmentDialog = () => {
```
waabid, line 267: This and the two methods below do not have tests validating them.

brocktaylor7: I've added tests to validate these.

src/DetailsView/components/details-view-command-bar.tsx
```diff
@@ -232,25 +234,25 @@ export class DetailsViewCommandBar extends React.Component<
     }
 
     private renderSaveAssessmentButton = (): JSX.Element | null => {
-        if (this.props.featureFlagStoreData.saveAndLoadAssessment) {
-            return this.props.switcherNavConfiguration.SaveAssessmentFactory({
-                ...this.props,
-            });
-        }
-        return null;
+        return this.props.switcherNavConfiguration.SaveAssessmentFactory({
```
waabid, line 237: I think we can probably do the same flagged component bit here instead of moving it down a layer to the save assessment factory.

Additionally, after that we can change the type for this method and below to remove the `| null`.

brocktaylor7: This has been addressed.

src/tests/unit/tests/DetailsView/components/load-assessment-button.test.tsx
```diff
@@ -11,33 +12,63 @@ import {
...
+            id: 5,
+        } as TabStoreData;
+
+        assessmentStoreData = {} as AssessmentStoreData;
+    });
 
     it('should render per the snapshot', () => {
         const rendered = shallow(<LoadAssessmentButton {...props} />);
-        expect(rendered.getElement()).toMatchSnapshot();
+        expect(rendered).toMatchSnapshot();
```
waabid, line 64: Think you're missing `.getElement()` here because the snapshot is empty.

waabid (approved):
LGTM, with the snapshot updates.

src/tests/unit/tests/DetailsView/components/load-assessment-button.test.tsx
brocktaylor7: good catch!

src/tests/unit/tests/DetailsView/components/load-assessment-button-factory.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+    getLoadButtonForFastPass,
+} from 'DetailsView/components/load-assessment-button-factory';
+
+describe('LoadAssessmentFactory', () => {
+    let props: LoadAssessmentButtonProps;
+
+    describe('getLoadButtonForAssessment', () => {
+        test('renders load assessment button', () => {
+            const rendered = getLoadButtonForAssessment(props);
+            expect(rendered).toMatchSnapshot();
```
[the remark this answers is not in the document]
brocktaylor7: This one the method returns the element and the snapshot is not empty.

PR #4282, opened 2021-05-25
chore: Tweak HC icons and scoping

DaveTryon:
#### Details

The changes in #4253 weren't quite correct. They omitted a bit of scoping for the styles, and they introduced some inconsistency in how icons (especially the file bug icon) were represented in HC White mode. This corrects those errors.

##### Screenshots

Note how the fill in the "Before" column of HC White is inconsistent with other HC modes, making the icons kind of "splotchy". In the "After" column, the fill is the same color as the button, so the icon appearance is cleaner and consistent across all HC modes.

HC Mode | Before | After
--- | --- | ---
None | image | image
HC 1 | image | image
HC 2 | image | image
HC Black | image | image
HC White | image | image

##### Motivation

Better user experience

##### Context

I tried unsuccessfully to get the file bug icon to have the same fill in HC that it has in non-HC. I think that would require reauthoring the SVG.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/styles/injected.scss
+++ b/src/injected/styles/injected.scss
@@ -131,13 +131,14 @@
         }
     }
 
-    .file-issue-button,
-    .copy-issue-details-button,
-    .insights-dialog-button-inspect {
+    .insights-dialog-main-override .file-issue-button,
+    .insights-dialog-main-override .copy-issue-details-button,
+    .insights-dialog-main-override .insights-dialog-button-inspect {
         svg path {
             @media screen and (forced-colors) {
                 forced-color-adjust: none !important;
                 stroke: ButtonText !important;
+                fill: ButtonFace !important;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/injected/styles/injected.scss
```diff
@@ -131,13 +131,14 @@
         }
     }
 
-    .file-issue-button,
-    .copy-issue-details-button,
-    .insights-dialog-button-inspect {
+    .insights-dialog-main-override .file-issue-button,
```
DaveTryon, line 134: The scoping here was required to get the fill override to work. It's specified everywhere else in this file and should have been included in #4253

dbjorge (approved):

src/injected/styles/injected.scss
```diff
@@ -266,14 +267,14 @@
         text-align: center !important;
     }
 
-    .insights-highlight-box {
+    .insights-highlight-container .insights-highlight-box {
```
DaveTryon, line 270: Scoping on lines 270 and 277 weren't required but I added them to keep the styling consistent. I confirmed that this didn't create a regression of the styles that #4253 addressed.

PR #4320, opened 2021-06-01
chore(deps-dev): migrate from Jest 26 to Jest 27

dbjorge:
#### Details

This PR updates all web packages to Jest 27.

* Removes explicit jest-circus settings (these are now Jest's defaults)
* Removes explicit jest-circus dependency (this is now bundled with Jest by default)
* Regenerates lockfile (avoids many double-dependencies on large deps like babel)
* ~Changes web e2e tests to use `testEnvironment: node` - this is better for performance and is also required because Jest 27 dropped `setImmediate`, which `playwright` uses, from the `jsdom` environment (because most browsers don't implement it)~ I attempted this and it turned out to be very disruptive, since hundreds of e2e snapshots rely on jsdom as part of how our element snapshot formatter works. As a more immediate scoped fix, added a `setImmediate` polyfill to the web e2e's Jest `globals`

There still isn't an updated `@types/jest` package, so we're still pulling in two versions of a few jest support libs, but I wanted to pull in the update now because it's still a net decrease in total dependency tree size.

##### Motivation

Keep deps up to date

##### Context

Obsoletes #4313, #4314

#### Pull request checklist

- [x] Ran `yarn fastpass`


brocktaylor7 (approved):
LGTM

Issue #3959, opened 2021-02-24
Add needs review rule: label-content-name-mismatch


LiLoDavis:
## Is your feature request related to a problem? Please describe.
Increase automated check coverage

## Describe the desired outcome
The axe experimental rule [label-content-name-mismatch](https://dequeuniversity.com/rules/axe/4.1/label-content-name-mismatch?application=RuleDescription) appears in **FastPass** > **Needs review**

## Describe alternatives you've considered
N/A

## Additional context
[How to fix - label-content-name-mismatch.docx](https://github.com/microsoft/accessibility-insights-web/files/6038228/How.to.fix.-.label-content-name-mismatch.docx)

ferBonnin:
@ThanyaLeif 
for the **how to check text**: Inspect the element using the Accessibility pane in the browser Developer tools and verify that the element’s accessible name contains its visible text.

ferBonnin:
verified this is now available in Canary!

PR #4339, opened 2021-06-05
Enabled label-content-name mismatch rule for needs-review

ThanyaLeif:
#### Details

The axe experimental rule label-content-name-mismatch was added in FastPass

image
image

##### Motivation

Addresses issue #3959

#### Pull request checklist

- [x] Addresses an existing issue: #3959
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/analyzers/filter-results.ts
+++ b/src/injected/analyzers/filter-results.ts
@@ -10,7 +10,8 @@ export const filterNeedsReviewResults = (results: ScanResults): ScanResults => {
         v =>
             v.id !== 'aria-input-field-name' &&
             v.id !== 'color-contrast' &&
-            v.id !== 'th-has-data-cells',
+            v.id !== 'th-has-data-cells' &&
+            v.id !== 'label-content-name-mismatch',
     );
     results.incomplete = results.incomplete.filter(i => i.id !== 'link-in-text-block');
 
--- a/src/scanner/get-rule-inclusions.ts
+++ b/src/scanner/get-rule-inclusions.ts
@@ -24,6 +24,10 @@ export const explicitRuleOverrides: DictionaryStringTo<RuleIncluded> = {
         status: 'excluded',
         reason: 'Based on feedback from users we tested to check the user impact of duplicate ID failures on static elements. We found no user impact on the experience with any of the ATs including Narrator, JAWS, NVDA, Dragon, and Windows Speech Recognition. After further discussions with the Tooling Advisory Board, we decided to make this a best practice rule. See #4102.',
     },
+    'label-content-name-mismatch': {
+        status: 'included',
+        reason: 'for parity with video-caption, which axe-core includes by default',
+    },
 };
 
 export const getRuleInclusions = (
--- a/src/ad-hoc-visualizations/needs-review/visualization.tsx
+++ b/src/ad-hoc-visualizations/needs-review/visualization.tsx
@@ -17,7 +17,13 @@ const needsReviewTestKey = AdHocTestkeys.NeedsReview;
 const issuesTestKey = AdHocTestkeys.Issues;
 
 const needsReviewRuleAnalyzerConfiguration: RuleAnalyzerConfiguration = {
-    rules: ['aria-input-field-name', 'color-contrast', 'th-has-data-cells', 'link-in-text-block'],
+    rules: [
+        'aria-input-field-name',
+        'color-contrast',
+        'th-has-data-cells',
+        'link-in-text-block',
+        'label-content-name-mismatch',
+    ],
     resultProcessor: (scanner: ScannerUtils) => scanner.getFailingInstances,
     telemetryProcessor: (telemetryFactory: TelemetryDataFactory) =>
         telemetryFactory.forNeedsReviewAnalyzerScan,
--- a/src/tests/unit/tests/injected/analyzers/filter-results.test.ts
+++ b/src/tests/unit/tests/injected/analyzers/filter-results.test.ts
@@ -10,6 +10,7 @@ describe('filterNeedsReviewResults', () => {
         const inapplicableOnlyRule0 = 'color-contrast';
         const inapplicableOnlyRule1 = 'aria-input-field-name';
         const inapplicableOnlyRule2 = 'th-has-data-cells';
+        const inapplicableOnlyRule3 = 'label-content-name-mismatch';
         const testRule = 'test rule';
         const given: ScanResults = {
             violations: [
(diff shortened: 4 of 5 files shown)


src/scanner/get-rule-inclusions.ts
```diff
@@ -24,6 +24,10 @@ export const explicitRuleOverrides: DictionaryStringTo<RuleIncluded> = {
         status: 'excluded',
         reason: 'Based on feedback from users we tested to check the user impact of duplicate ID failures on static elements. We found no user impact on the experience with any of the ATs including Narrator, JAWS, NVDA, Dragon, and Windows Speech Recognition. After further discussions with the Tooling Advisory Board, we decided to make this a best practice rule. See #4102.',
     },
+    'label-content-name-mismatch': {
+        status: 'included',
+        reason: 'for parity with video-caption, which axe-core includes by default',
```
ferBonnin, line 29: nit: reason to include this rule should be different (since the rule is not related with video-caption). Could be  something like " to include in Needs Review since its an experimental rule"

src/injected/analyzers/filter-results.ts
```diff
@@ -10,7 +10,8 @@ export const filterNeedsReviewResults = (results: ScanResults): ScanResults => {
         v =>
             v.id !== 'aria-input-field-name' &&
             v.id !== 'color-contrast' &&
-            v.id !== 'th-has-data-cells',
+            v.id !== 'th-has-data-cells' &&
+            v.id !== 'label-content-name-mismatch',
```
RobGallo, line 14: It looks like the list of rules for needs review here is the same as the list provided in needsReviewRuleAnalyzerConfiguration.rules in src/ad-hoc-visualizations/needs-review/visualization.tsx. If these two sets of data are always intended to be the same, could we have them in one place to eliminate the possibility of either getting out of sync?

dbjorge: It's not intended to be the same, though I don't think that fact is very clear from the naming of the function here.

The purpose of this function (`filterNeedsReviewResults`) is *not* to take a full set of axe-core results and to either filter out *or* to filter-keeping just needs-review results. Its purpose is to take the set of axe-core results *produced by the needs review visualizer's analyzer configuration*, which enables just those rules that can produce needs review results, and to *filter out results that are already expressed as automated checks failures*.

For example, the `color-contrast` rule needs to be enabled as part of the "needs review" analyzer configuration because we want any "incomplete" results it produces to be present in the "needs review" UI. However, we *don't* want the "violations" it produces to be present in the "needs review" UI, even though we need the rule enabled, since color-contrast violations are full automated checks failures. That's why `filterNeedsReviewResults` filters out color contrast violations.

The `label-content-name-mismatch` rule doesn't work the same way; it is an experimental rule that produces violations, not incompletes, when it detects problems. That means:

* We *don't* want it to run at all in the automated checks ruleset; there is nothing the rule can report that we would ever want to show up in an automated check. Therefore, it shouldn't be marked as included in "get-rules".
* We *do* want to see any violations that the rule reports as part of needs review. Therefore, we shouldn't filter out its violations in `filterNeedsReviewResults`.
* We *need PM input* on whether we want users to see anything that the rule reports as "incomplete" as part of *needs review* output.
    * Based on a quick review of [the rule's implementation](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/label/label-content-name-mismatch-evaluate.js), it looks like that will occur if-and-only-if either the visible text or the aria-label of the element being tested is determined to be "not human interpretable", which axe-core's purposes here essentially means "is something like an emoji, punctuation mark, or a single character that might be being used similarly, like "x").
    * I would recommend that we probably *do* want such cases to be marked as needs review. If that's the case, `filterNeedsReviewResults` should *not* filter out incomplete results for this id.
    * If PM decides that we *do not* want such cases to be marked as needs review (ie, we want to ignore them silently), then `filterNeedsReviewResults` should filter out incomplete results for this rule similar to what it already does for `link-in-text-block`.

---

I don't think this function's naming makes it very clear what the intended behavior is; I think it would be a very good idea to improve the name and/or add a comment on the function that explains what the intended purpose of the function is.

An even better option (which would also addressing Rob's point about duplication) might be to separate out the notions of "what behavior do we want configured for each needs review rule" vs "how do we apply that configuration to filter axe result objects". I could imagine there being some sort of `NeedsReviewRuleConfiguration` that we initialize for each "needs review" rule that includes properties for "should violations be treated as need-review? should incompletes be treated as needs-review?", for this filtering logic to use those properties as input, and then for the visualization configuration to use the same configuration objects to identify which needs review rules exist and reuse that for its "analyzer rules" configuration.

I think that would be a much bigger change though; if you'd like to pursue that refactoring, I would probably lean towards doing it in a separate PR.

src/scanner/get-rule-inclusions.ts
```diff
@@ -24,6 +24,10 @@ export const explicitRuleOverrides: DictionaryStringTo<RuleIncluded> = {
         status: 'excluded',
         reason: 'Based on feedback from users we tested to check the user impact of duplicate ID failures on static elements. We found no user impact on the experience with any of the ATs including Narrator, JAWS, NVDA, Dragon, and Windows Speech Recognition. After further discussions with the Tooling Advisory Board, we decided to make this a best practice rule. See #4102.',
     },
+    'label-content-name-mismatch': {
```
dbjorge, line 27: Per the big comment I left in filter-results.ts, I don't think this should actually be included at all here, but it's easy to see how you'd think it should be; this file doesn't make itself clear that it only applies to automated checks. It would be a good idea to improve the naming and/or leave a comment to make it more clear (and to fix the very misleading comments on `empty-table-header` and `form-field-multiple-labels`; they should probably stop at "only reports needs-review results", it's misleading to imply that we'd want to include them once we implement needs-review content for them).

dbjorge (changes requested):
It looks like the new rule is showing up under "automated checks" rather than "needs review" right now. The big comment I left in `filter-results.ts` explains why I think this is happening:

* it shows up in automated checks because of adding it as included in `get-rules`; I think that should be left out
* it doesn't show up in needs review because `filter-results` is filtering it out

Once you fix those, I think you're going to encounter an issue where the "How to check" text in the "Needs review" UI for these issues won't be filled in with any text. You'll want to update `src\common\components\cards\rich-resolution-content.tsx` to include the how to check content for this type of issue (you can find the content to use in the document linked in the "Additional context" section of #3959)

ThanyaLeif:
> It looks like the new rule is showing up under "automated checks" rather than "needs review" right now. The big comment I left in `filter-results.ts` explains why I think this is happening:
> 
> * it shows up in automated checks because of adding it as included in `get-rules`; I think that should be left out
> * it doesn't show up in needs review because `filter-results` is filtering it out
> 
> Once you fix those, I think you're going to encounter an issue where the "How to check" text in the "Needs review" UI for these issues won't be filled in with any text. You'll want to update `src\common\components\cards\rich-resolution-content.tsx` to include the how to check content for this type of issue (you can find the content to use in the document linked in the "Additional context" section of #3959)

I added the additional context as you suggested in my last commits

PR #4360, opened 2021-06-14
feat(recommend-color): Able to recommend a color or two for insufficient color contrast ratio

banne001:
#### Details

This new feature will enable FastPass users to receive a color recommendation when running FastPass on webpages. The color recommendation will be displayed on the webpage when clicking a point of interest as well as in the FastPass output. 
When there is a color conflict, the current build informs users of this issue but does not recommend a way to fix the issue. With the new recommend color feature, a foreground and/or background is provided that will achieve success when testing for color contrast accessibility.

##### Motivation

MinT Project through Green River College

##### Context

Screen Shot 2021-06-14 at 12 29 18 PM
Screen Shot 2021-06-14 at 12 29 28 PM

#### Pull request checklist

- [n/a] Addresses an existing issue: #0
- [x] Ran `yarn fastpass`


jalkire:
To get the PR build passing, you can run:
`yarn test -u`
and
`yarn build:package:report`  
`yarn test:report:e2e -- -- -u`
to update the snapshots. The web e2e failures are on our end, so you don't need to worry about them 🙂

jalkire:
Closing/reopening to rerun PR build

src/common/components/fix-instruction-processor.tsx
```diff
@@ -89,6 +171,99 @@ export class FixInstructionProcessor {
...
+                            endSubstring,
+                        )}
+                    </span>,
+                );
+            } else {
+                colorEndIndex = sortedMatches[i].splitIndex + this.recommendEndSentence.length;
+                endSubstring = fixInstruction.substring(insertionIndex, colorEndIndex);
+                result.push(
+                    <span key={`instruction-split-${keyIndex++}`}>
+                        <ul key={`instruction-split-${keyIndex++}`}>
```
jalkire, line 228: I gave the changes a quick pass in NVDA, and it looks like there is a slight accessibility issue with the new rows. Specifically, the way the list is now updated, NVDA always reads it as having one item, even when it now has more. NVDA read the following to me as I navigated through the list: "...Fix the following: list  with 1 items  •   Element has insufficient color contrast of 4.07 (foreground color: #0074d9, background color: #e6f5d0, font size: 10.5pt (14px), font weight: normal). Expected contrast ratio of 4.5:1 Use foreground color: #0068c3 and the original background color: #e6f5d0 to meet a contrast ratio of 4.87:1. Or use background color: #fdffe4 and the original foreground color: #0074d9 to meet a contrast ratio of 4.58:1...". Instead, I would expect NVDA to announce that I am entering a new list with two items before reading the suggested fixes.

Part of the issue is that each new `<li>` is being added in its own new `<ul>`--it would be more correct to create a new `<ul>` and then add the one or two suggested color `<li>` elements to that list. 

The other part of the issue relates to lines 168-9 of this file where we add separate screen reader text. I think we might be best served by removing that text, so I am going to approve this PR so we can deal with this in a follow-up PR if you would prefer (I could also do the follow up, given the timing of things). Please let me know if you would like to try fixing it before this PR is merged or not 🙂

jalkire (approved):
Looks good to me! I did notice a bit of not quite optimal behavior from NVDA around this, we'll want to get that fixed either here or later 🙂

PR #3606, opened 2020-11-03
chore(null): null safety fixes for all null:find results with <5 refs

dbjorge:
#### Description of changes

This fixes most of the "leaf" instances reported by `null:find`. New progress as of this PR:

```markdown
## Web strict-null progress

**49%** complete (**675**/1382 non-test files)

*Contribute at [#2869](https://github.com/microsoft/accessibility-insights-web/issues/2869). Last update: Tue Nov 03 2020*
```

#### Pull request checklist

- [x] Addresses an existing issue: #2869
- [x] Ran `yarn fastpass`
- [n/a] Added/updated relevant unit test(s) (and ran `yarn test`)
- [n/a] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #4377, opened 2021-06-16
chore: Make AxeRule.description required again

DaveTryon:
#### Details

While working on strict null checks, I noticed that `AxeRule.description` is optional, which seemed counterintuitive since we'd presumably always want to have a description for a rule. My curiosity piqued, I checked the history and learned that this was changed in #3606, which was early in the strict null check work. As an experiment, I reverted 2 changes related to the description and found only 1 place where it was still flagged as a problem. I added a placeholder string for the description in a place where we used to have no description at all.

##### Motivation

Undo what was apparently intended to be a temporary change, hopefully simply future strict null check efforts

##### Context

`IRuleConfiguration.description` should arguably be non-optional, but that goes back to well before the strict null check work. I tried the change locally and it broke multiple custom rules and multiple tests, so that feels like a separate conversation than restoring the intent before #3606.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/scanner.d.ts
+++ b/src/injected/scanner.d.ts
@@ -8,7 +8,7 @@ declare interface AxeResult {
 declare interface AxeRule {
     id: string;
     nodes: AxeNodeResult[];
-    description?: string;
+    description: string;
     help?: string;
 }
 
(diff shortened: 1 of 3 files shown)


DaveTryon:
After conversation with @dbjorge, the change to optional description is intended to be a long-term change. Closing without merging.

PR #4378, opened 2021-06-16
chore(web-compliance-21h1): enable self-checking of FastPass needs review rules

dbjorge:
#### Details

This PR renames `window.A11YSelfValidator.validate()` to `window.selfFastPass.automatedChecks()` and adds a new `window.selfFastPass.needsReview()` which runs our FastPass Needs Review checks.

To do this, it rewrites the self validator to go through the `IssuesAdHocVisualization` and `NeedsReviewAdHocVisualization` config objects that contain the definitions of which rules should run for each case. This isn't totally accurate for Needs Review because it doesn't include the filtering performed by `UnifiedResultSender.filterNeedsReviewResults`; I decided to leave that out of scope for this PR, but it would be nice to do some sort of refactoring like we discussed in #4339 to enable reuse of that filtering logic. In the meantime, the main implication of not doing the filtering here is that some automated check failures would additionally show up as needs review failures, which isn't a big impact since we're generally automated-checks-clean.

##### Motivation

Wanted to run needs review checks as part of web compliance feature.

##### Context

I was initially concerned that the new dependencies on the visualizer configs would explode the size of the popup/insights bundles, but that doesn't seem to be the case in practice; it does increase them, but only by 1-2%.

#### Pull request checklist

- [x] Addresses an existing issue: 1849910
- [x] Ran `yarn fastpass`


brocktaylor7 (approved):
LGTM

Issue #4361, opened 2021-06-14
Text spacing violation (clipping) in some assessment left nav items


dbjorge:
## Describe the bug

In the web details view, the left nav menu for assessments has several items which get clipped when you apply the styles suggested by our own Text Spacing test:

![screenshot of issue with the clipped nav items circled](https://user-images.githubusercontent.com/376284/121971822-a46e5b80-cd47-11eb-933d-7591f631f3a8.png)

Additionally, the `scrollable-region-focusable` automated check gets clipped slightly even with default styling:

![screenshot showing scrollable-region-focusable clipping](https://user-images.githubusercontent.com/376284/121971908-dd0e3500-cd47-11eb-97f7-1aec7dde9d36.png)

**To Reproduce**

1. Open a web assessment against any page
2. Right click on one of the menu items and select "Inspect"
3. In the F12 developer tools panel that opens, apply the following styles:
```css
    line-height: 1.5 !important;
    letter-spacing: 0.12em !important;
    word-spacing: 0.16em !important;
```
4. Observe the results from the above screenshot

## CodePen repro example

n/a

## Expected behavior

Per our own Text Spacing requirement (emphasis mine):

"Verify that all of the following are true:
* All text responds to each change in styling.
* ***All text remains visible (no clipping)***.
* There is no overlapping text."

## Screenshots

see inline

## Context (please complete the following information)

- OS Name & Version: Windows 21H1 OS Build 19043.1052
- AI-Web Version & Environment: Production 2.27.0 and dev cb03f4912
- Browser Version: Microsoft Edge Version 91.0.864.41 (Official build) (64-bit)
- Target Page: n/a

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

Yes

## Additional context

Discovered during biannual compliance pass when the scrollable-region-focusable issue with default styling tripped a need-review check run as part of a self-assessment

dbjorge:
@juchampa / @ferBonnin , we'll need design input for what you'd like to do about these cases. I'm imagining something like a combination of "enable the text in the menu items to reflow for the sake of users that want to use custom text styling, and also expand the width slightly so `scrollable-region-focusable` doesn't reflow under default styles" - does that sound reasonable?

ferBonnin:
[...]

PR #4393, opened 2021-06-22
chore: remove obsolete pre-reflow code

dbjorge:
#### Details

While I was working on #4361, I noticed that the existing left nav styles had a few obsolete leftovers nearby from back when we implemented the reflow feature in v2.19. This PR chases down the rabbit hole of unused code and removes net ~800 lines.

I've tried to keep individual commits to one type of refactor at a time, so it might be easier to go through commit-by-commit.

Notable updates:
* Renamed a bunch of stuff named "ReflowThing" to either just "Thing" or "DetailsViewThing"
* Removed several styles/style vars that were always-true (eg, `.reflow-ui` stuff)
* Removed an unnecessary layer of `ReflowAssessmentView` into its parent `AssesssmentTestView` (the parent used to dispatch between the old and new versions based on a feature flag, but this responsibility no longer exists since the old version had already been ripped out)
    * As part of this, moved some logic (eg, `nextRequirement` calculation) that was specific to the child `RequirementView` into the `RequirmentView`, rather than calculating it in the parent
* The "order"  and "requirementOrder" properties of requirements/assessments were no longer a necessary abstraction, removed them and a bunch of comparator helpers
    * This also enabled removing an `AssessmentTestResult` supporting class whose only practical purpose was to sort requirements by requirement order

##### Motivation

Remove obsolete code, simplify #4361 fix

##### Context

See web@2.19.3 for details/context on the reflow-ui features' release

No UI changes as part of this; manually verified old/new assessment behavior side by side (would be good for reviewer to do the same)

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/assessment-test-view.tsx
```diff
@@ -124,7 +75,22 @@ export const AssessmentTestView = NamedFC<AssessmentTestViewProps>(
                     test={assessmentNavState.selectedTestType}
                 />
                 <TargetChangeDialog deps={deps} prevTab={prevTarget} newTab={currentTarget} />
-                {mainContent}
+                {isGettingStartedSelected ? (
```
waabid, line 78: Usually don't like having conditionals in the component html itself but this reads quite nicely.

waabid (approved):
LGTM

src/DetailsView/components/left-nav/common-left-nav-link.scss
```diff
@@ -11,7 +11,7 @@
 .left-nav-link-container {
     padding-left: 5%;
     padding-right: 8%;
-    width: $detailsViewLeftNavWidth;
+    width: 200px;
```
dbjorge, line 14: This is maybe the most notable change in terms of "things that are probably a bug that the refactoring uncovered"; this was the only point where the pre-reflow `detailsViewLeftNavWidth` was still being used, and it's *not* consistent with the actual left nav width today. This PR is just maintaining the current behavior by inlining the old value in this one spot where it was still used, but this line will probably change to something less questionable in a followup PR to fix #4361

PR #4406, opened 2021-06-24
chore(google-rules-mvp): Add converter for ATFA results

DaveTryon:
#### Details

Add the ability to convert ATFA results into Unified rules and Unified results. It includes some refactors to the converters:
- Add ATFA types in `atfa-data-types.cs` and expose their results in `android-scan-results`. ATFA uses dots in its key names, which forces us to access fields using the `foo['key']` instead of just using `foo.key`. This is unfortunate but it should only be needed in the ATFA converter classes and their associated tests.
- `unified-result-buider.ts` is the focal point for including converters for results from both Axe and ATFA
- The Axe-specific code that was in `scan-results-to-unified-rules.ts` and `scan-results-to-unified-results.ts` now lives in `axe-scan-results-to-unified-rules.ts` and `axe-scan-results-to-unified-results.ts`
- The ATFA-specific code has been added to `atfa-scan-results-to-unified-rules.ts` and `atfa-scan-results-to-unified-results.ts`
- `scan-results-to-unified-rules.ts` and `scan-results-to-unified-results.ts` now just iterate through a list of converter delegates, then return the aggregated results

##### Context

This PR excludes adding the ATFA rules to the `RuleInformationProvider`. This PR is already large enough (even if most of that is from refactoring) that I wanted to get it out. It's an easy thing to update the rule information in a separate PR.

#### Pull request checklist

- [x] Addresses an existing issue: [1853274](https://mseng.visualstudio.com/1ES/_workitems/edit/1853274)
- [x] Ran `yarn fastpass`


src/electron/platform/android/android-scan-results.ts
```diff
@@ -73,4 +74,8 @@ export class AndroidScanResults {
     private getAxeResults(): any {
         return this.rawData?.AxeResults ?? this.rawData;
     }
+
+    public get atfaResults(): AccessibilityHierarchyCheckResult[] {
+        return this.rawData.ATFAResults || [];
```
lisli1, line 79:
```suggestion
        return this.rawData?.ATFAResults || [];
```

src/electron/platform/android/atfa-data-types.ts
```diff
@@ -0,0 +1,33 @@
...
+    'ViewHierarchyElement.contentDescription'?: SpannableString;
+    'ViewHierarchyElement.text'?: SpannableString;
+}
+
+// Source: https://github.com/google/Accessibility-Test-Framework-for-Android/blob/master/src/main/java/com/google/android/apps/common/testing/accessibility/framework/AccessibilityHierarchyCheckResult.java
+export interface AccessibilityHierarchyCheckResult {
+    'AccessibilityHierarchyCheckResult.element': ViewHierarchyElement;
+    'AccessibilityHierarchyCheckResult.resultId': number;
+    'AccessibilityHierarchyCheckResult.checkClass': string;
+    'AccessibilityHierarchyCheckResult.type': string; // TODO: make an enum?
```
lisli1, line 31: +1 on making this an enum since there is only a handful of result types but it doesn't have to be in this PR

src/electron/platform/android/atfa-scan-results-to-unified-results.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function getRawString(spannableString?: SpannableString): string | null {
+    if (spannableString) {
+        return spannableString['SpannableString.rawString'] ?? null;
+    }
+    return null;
+}
+
+function includeBasedOnResult(atfaResult: AccessibilityHierarchyCheckResult): boolean {
```
lisli1, line 100:
```suggestion
function includeBasedOnResultType(atfaResult: AccessibilityHierarchyCheckResult): boolean {
```
Naming this `includeBasedOnResultType` seems to better fit what the function does

DaveTryon: Done

src/electron/platform/android/atfa-data-types.ts
DaveTryon: Later PR seems reasonable

src/electron/platform/android/atfa-scan-results-to-unified-rules.ts
```diff
@@ -0,0 +1,40 @@
...
+    uuidGenerator: UUIDGenerator,
+): UnifiedRule[] {
+    if (!scanResults) {
+        return [];
+    }
+
+    const unifiedRules: UnifiedRule[] = [];
+    const ruleIds: Set<string> = new Set();
+
+    for (const atfaResult of scanResults.atfaResults) {
```
lisli1, line 27: Since ATFA rules are returned as part of the scan results, could we add a method in android-scan-results.ts to get the rules and then iterate through those when converting to unified rules? That way, we wouldn't have to loop through all the results to find their corresponding rules, and we could pretty much guarantee that this converts all the possible ATFA rules into unified rules (elements that pass ATFA checks are not included in ATFA results, so it's possible that an ATFA rule would not show up in the results).

lisli1: I know this is intended to be very similar to the axe-scan-results-to-unified-results but the way the axe results are formatted (rules embedded into the results) requires us to have to search the results for the rules, which is different from how ATFA results are formatted.

src/electron/platform/android/atfa-scan-results-to-unified-results.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function createUnifiedResult(
+    ruleInformation: RuleInformation,
+    atfaResult: AccessibilityHierarchyCheckResult,
+    viewElement: ViewHierarchyElement,
+    uuidGenerator: UUIDGenerator,
+): UnifiedResult {
+    const ruleResult: RuleResultsData = {
+        axeViewId: `atfa-${viewElement['ViewHierarchyElement.id']}`,
```
lisli1, line 66: It feels strange to have "axe" in the property name. Maybe in a future PR, we could look into changing this to just `viewId`, since RuleResultsData can refer to either axe or ATFA rule result.

src/electron/platform/android/atfa-scan-results-to-unified-rules.ts
lisli1: If we do use the ATFA rules from the ATFARules data (returned from Android Service), the package name would have to be removed to get the ruleId because the format for that is something like `com.google.android.apps.common.testing.accessibility.framework.checks.LinkPurposeUnclearCheck` (or alternatively, it's possible to update the returned data on the Android Service side to not include the prepended package name).

src/electron/platform/android/atfa-scan-results-to-unified-results.ts
DaveTryon: `axeViewId` was inherited from the Axe-Android data structure. I like the idea of renaming this in a future PR

src/electron/platform/android/atfa-scan-results-to-unified-rules.ts
DaveTryon: @lisli1, what you describe was my first inclination, as well. The difference in the names was the primary reason that I didn't follow that approach, because I have a strong aversion to heuristic-based code unless it's related to gaming or AI.

If it's important to ensure that we always have at least 1 result for each defined rule, then just iterating through the returned set of rules will be insufficient. We only display results where the type is `ERROR` or `WARNING`, and many of the results in the test file I was working with were type of `INFO` or `NOT_RUN`. The code currently ignores the type, but that's an easy enough thing to add if we need it.

Eventually, I suspect that some rules might need to map to more than one card, based on the resultID. At that point, this code will probably get more complex and need to assess both the rules and the results. That's out of scope for the MVP, though

lisli1 (approved):

PR #4448, opened 2021-07-12
fix: suppress failures on browser close for E2E tests

waabid:
#### Details

This wraps the `close` method in the browser class with catch statement which will suppress failures.

##### Motivation

Seeing transient issues in our E2E due to recent playwright updates with this specific function.

##### Context

I chose this approach because it touches the least amount of files; I could create a new method called `closeSafely` but all signatures would need to change. Figured, renaming to unsafeClose for the function details with a comment would be a better approach.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/tests/end-to-end/common/browser.ts
+++ b/src/tests/end-to-end/common/browser.ts
@@ -26,6 +26,16 @@ export class Browser {
     }
 
     public async close(): Promise<void> {
+        try {
+            await this.unsafeClose();
+        } catch (e) {
+            console.log('Suppressing failure on browser close: ', e);
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


waabid:
Example of failure can be seen here: https://dev.azure.com/accessibility-insights/accessibility-insights-web/_build/results?buildId=23055&view=logs&j=04f28fcc-3e74-5513-5910-075f7a6ec2dd&t=419e1bc8-be06-5c99-ae3d-64fc84a94342

dbjorge (approved):

dbjorge:
LGTM as a workaround, but it would be great if we could get a Playwright issue opened to track this (and ideally reference it in the comments you've added here)

PR #4463, opened 2021-07-13
feat(google-rules-mvp): Support spanning text

DaveTryon:
#### Details

The rule conversion code was not handling the case where ATFA returns an object as `SpannableString.rawString`, which led to some unusual behavior at rendering time. This PR adds support for this case, tidies up some of the associated unit tests, and removes support for the view and result ID, which never get consumed by the converters.

##### Motivation

Fix case that was missed previously.

##### Screenshot of result with fix:
image

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/unit/tests/electron/platform/android/scan-results-helpers.ts
```diff
@@ -208,27 +207,67 @@ function buildAtfaSpannableString(rawString: string): SpannableString {
...
+    }
 
-    if (boundsInScreen) {
-        e['ViewHierarchyElement.boundsInScreen'] = buildAtfaRectangle(boundsInScreen);
+    if (elementParameters.contentDescription && elementParameters.contentDescriptionViaSpan) {
+        throw Error("contentDescription can't be both direct and indirect");
     }
+    if (elementParameters.text && elementParameters.textViaSpan) {
+        throw Error("text can't be both direct and indirect");
+    }
```
lisli1, line 252: Are these lines needed if `verifyAndBuildSpannableString` throws the same error message? It looks like `verifyAndBuildSpannableString` is called for both the `contentDescription` and `text` below.

lisli1 (approved):
Minor comment inline, otherwise LGTM

src/tests/unit/tests/electron/platform/android/scan-results-helpers.ts
DaveTryon: Oops! I refactored and forgot to remove these. Will fix

Issue #4262, opened 2021-05-21
Failure instances has clipped border in all Windows HC modes


DaveTryon:
## Describe the bug

In Windows HC modes, the border of the "Failed instances" counter gets clipped. It doesn't happen if non-HC mode

**To Reproduce**

1. Enable High Contrast
2. Run automated checks on a page with errors (I used https://www.washington.edu/accesscomputing/AU/before.html)
3. Look at the Failed instances marker
4. The top of the border is clipped

## Expected behavior

The border should not be clipped

## Screenshots
HC Mode | Screenshot
--- | ---
None | ![image](https://user-images.githubusercontent.com/45672944/119206090-0987a780-ba4f-11eb-9c43-bc114688a873.png)
HC 1 | ![image](https://user-images.githubusercontent.com/45672944/119205956-ad248800-ba4e-11eb-95db-1282ce68b955.png)
HC 2 | ![image](https://user-images.githubusercontent.com/45672944/119206004-c594a280-ba4e-11eb-8ad1-99dfea355fbf.png)
HC Black | ![image](https://user-images.githubusercontent.com/45672944/119206034-df35ea00-ba4e-11eb-8c52-430923c17ccf.png)
HC White | ![image](https://user-images.githubusercontent.com/45672944/119206071-f70d6e00-ba4e-11eb-949e-a7f9bc9780d4.png)

## Context (please complete the following information)

- OS Name & Version: Windows 10 21He
- AI-Web Version & Environment: 2.26.0 Production
- Browser Version: Chrome Version 90.0.4430.212 (Official Build) (64-bit)
- Target Page: https://www.washington.edu/accesscomputing/AU/before.html

DaveTryon:
Confirmed fixed in Canary

PR #4500, opened 2021-07-19
fix: Stop clipping chip count borders in HC modes

DaveTryon:
#### Details

As called out in #4262, borders around chip counters get clipped in HC modes. @Peterdur pointed out this [Edge Blog post](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/) that describes how HC rendering changes rendering of the Text Backplate, and since the Text Backplate overlaps the border, the border gets clipped. The fix is to explicitly disable the `forced-color-adjust` attribute and control foreground and background colors ourselves.

##### Motivation

Address #4262 

##### Screenshots
HC Mode | 100% Before | 100% After | 400% Before | 400% After
--- | --- | --- | --- | ---
None | image | image | image | image
HC 1 | image | image | image | image
HC 2 | image | image | image | image
HC Black | image | image | image | image
HC White | image | image | image | image

#### Pull request checklist

- [x] Addresses an existing issue: #4262
- [x] Ran `yarn fastpass`
- [n/a] (UI changes only) Verified usability with NVDA/JAWS (css only)


src/reports/components/outcome.scss
```diff
@@ -60,6 +60,11 @@ $outcome-count-border-size: 1.5px;
         font-weight: 700;
         margin-left: -8px;
         height: $outcome-chip-icon-size - (2 * $outcome-count-border-size);
+
+        @media screen and (forced-colors: active) {
+            padding-top: 2px;
+            padding-bottom: 1px;
```
peterdur, line 66: @DaveTryon Something about the bigger border bugged me, so I banged my head against this for a while and I think I have something that will remove the clipping but not require extra padding:

```suggestion
            forced-color-adjust: none;
            border-color: CanvasText !important;
            color: CanvasText;
```

In looking at this, I thought that the problem was the background color was being overridden from transparent to the canvas color, so I tried disabling forced colors here and then specifying the [high contrast colors](https://www.w3.org/TR/css-color-4/#css-system-colors) explicitly. Per [an Edge blog post](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/#:~:text=New%2C%20standard%20system%20colors%20replace%20the%20old%20values%3A,Highlight%20..%20...%20%202%20more%20rows%20) I found later, this is not quite the full story; high contrast mode turns on text backplates (see heading about 60% down the page) and setting `forced-color-adjust: none` turns these off.

I think the `!important` that sets the border color may be avoidable by applying it directly to the `.outcome-chip.outcome-chip-fail .count` and other flavors when `forced-colors` is `active`.

DaveTryon: Thanks! I greatly prefer this solution. I left the !important on the .count, since it seems to bring the changes together in a more cohesive way that spreading it around.

peterdur (approved):
Ship it!

ferBonnin:
@DaveTryon talked with the design team today as we agreed. They mentioned that this option is the most visually correct but there are way too many colors in the small pill. It would be great if we can make the red circle and the white X change colors with the HC mode so the UI looks less busy

PR #4514, opened 2021-07-22
chore: (Android) Remove remnants of v1 results

DaveTryon:
#### Details

To facilitate with migration, the feature to add ATFA rules shipped with a feature flag that allowed v1 or v2 results to be selected. Now that the feature has shipped, and before starting work on changing the service to a non-HTTP protocol, the feature flag and the ability to produce/consume v1 results needs to be removed.

##### Motivation

Preparation for the protocol change feature.

##### Context

The commits took a "peel the onion" approach from the outside in. It might be simpler to review on a per-commit basis. Note that the `ColorContrast` rule was kept around for V1 compatibility, even though it was excluded from V2 results. It goes away completely in this PR

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/electron/platform/android/android-scan-results.ts
```diff
@@ -75,7 +75,7 @@ export class AndroidScanResults {
     }
 
     private getAxeResults(): any {
-        return this.rawData?.AxeResults ?? this.rawData;
```
DaveTryon, line 78: V2 moved the AxeResults from rawData to rawData.AxeResults. This line was here to provide automatic handling of either version

src/tests/unit/tests/electron/platform/android/scan-controller.test.ts
```diff
@@ -208,92 +205,6 @@ describe('ScanController', () => {
         telemetryEventHandlerMock.verifyAll();
     });
 
-    it('scans and handle successful v2 response', async () => {
```
DaveTryon, line 211: Even though this test name includes "v2 response', it was actually processing a v1 response. The name was wrong.

dbjorge (approved):
This is really helpful, thanks very much Dave!

PR #4515, opened 2021-07-22
refactor: accessibility-insights-ui's package.json

lisli1:
#### Details

Similar to the refactor from #3519, this PR moves the contents from `/src/packages/accessibility-insights-ui/root` to `/packages/ui` and updates relevant `package.json` files. Also updates `/packages/README.md` to reflect current package layout.

When running `yarn build:package:ui` locally from this branch, `/packages/ui/drop/` is created and contains `index.js` and `ui.css`

##### Motivation

The change to `workingDir` in `pipeline/build-package.template.yaml` caused the npm publish step to stop working for the accessibility-insights-ui package, because it could no longer find the `package.json`. With this refactor, `package.json` will be in the specified directory, `/packages/ui`.

##### Context

Considered adding a `"style"` field to `package.json` for the `ui.css` file, similar to the `typings` field added in #3519 for the report package, but it appears to not be standardized and could cause breaks in the future

#### Pull request checklist

- [x] Ran `yarn fastpass`


packages/README.md
```diff
@@ -41,8 +41,8 @@ this or not, but in the interest of keeping it easy, we want to lean towards usi
 decisions between the two of them for "directory layout", "which tools run repo-wide vs
 per-package", etc.
 
-## Current (Oct 2020) package layout
+## Current (July 2021) package layout
 
-* `/src/packages/accessibility-insights-ui/root`: current home of future `/packages/ui`
-* `/src/reports/package/root`: current home of future `/packages/report`
+* `/packages/ui`: current home of former `/src/packages/accessibility-insights-ui/root`
+* `/packages/report`: current home of former `/src/reports/package/root`
```
lisli1, line 47: Since the contents of these `src/**/**/root` directories have been already made, should we remove these notes? I left it in and slightly modified the text because it is titled "Current package layout" instead of something like "Changes needed to be made"

dbjorge: I think removing the notes makes sense!

dbjorge: On second reading, maybe keeping the section makes sense (since it also talks about where to find the web/unified packages outside the directory, which isn't obvious). I'm not sure we need to note the "of former" locations, but it might make sense to note a one-sentence summary of what each folder represents.

dbjorge:
It looks like there are a few dependencies in the ui package that have missed updates, so running a `yarn install` against this PR branch results in a lockfile update. It would be good to update the relevant ui/package.json entries to match the version numbers in the root package.json to avoid this (`uuid`, `lodash`, `luxon`, and `classnames` appear to be the impacted cases)

dbjorge (changes requested):
Looks good overall, the outdated packages thing is the only significant comment!

packages/README.md
lisli1: There seems to already be a description for each of these folders under "Short-term (~Dec 2020) planned package layout" in the same README. Would it make sense to have a similar summary here as well?

dbjorge (approved):
Oops, you already fixed the dependencies! LGTM; up to you whether to change the readme comment

PR #4544, opened 2021-08-03
chore(null): LiveAndroidSetupDeps

waabid:
#### Details

Adds LiveAndroidSetupDeps to null checks.

##### Motivation

#2869

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/electron/platform/android/setup/live-android-setup-deps.ts
```diff
@@ -27,6 +27,10 @@ export class LiveAndroidSetupDeps implements AndroidSetupDeps {
     public hasAdbPath = async (): Promise<boolean> => {
         try {
             const adbLocation = this.configStore.getState().adbLocation;
+            if (adbLocation == null) {
```
dbjorge, line 30: I don't think this is correct; `createValidatedAdbWrapper` has specific logic in place to handle the undefined case by passing `appium-adb` an undefined parameter blob, which `appium-adb` handles by falling back to looking based on the `ANDROID_HOME` and `ANDROID_SDK_HOME` environment variables.

I think it would be more correct to update the parameter type in `createValidatedAdbWrapper` to accept an undefined `sdkRoot`.

dbjorge (changes requested):
See inline

dbjorge (approved):

PR #4580, opened 2021-08-11
chore(strict-null): fixes in tab store and e2e test

codeofdusk:
#### Details
This PR includes partial fixes for tab store (did not add to file as there are still errors I'm unsure how to fix) and fixes for a file in the e2e tests.

##### Motivation
Part of strict null checks.

#### Pull request checklist

- [n/a] Ran `yarn fastpass`
- [n/a] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.

### DIFF
--- a/src/background/stores/tab-store.ts
+++ b/src/background/stores/tab-store.ts
@@ -54,7 +54,7 @@ export class TabStore extends BaseStoreImpl<TabStoreData> {
     };
 
     private onNewTabCreated = (payload: Tab): void => {
-        if (payload.id !== null && payload.id !== undefined) { 
+        if (payload.id !== null && payload.id !== undefined) {
             this.state.id = payload.id;
         }
         if (payload.title) {
@@ -92,7 +92,7 @@ export class TabStore extends BaseStoreImpl<TabStoreData> {
     };
 
     private originsMatch = (url1: string | null, url2: string | null): boolean => {
-        if (url1 === null && url2 === null) {
+        if (url1 == null && url2 == null) {
             return true;
         }
         if (url1 !== null && url2 !== null) {
--- a/src/background/stores/tab-store.ts
+++ b/src/background/stores/tab-store.ts
@@ -54,15 +54,9 @@ export class TabStore extends BaseStoreImpl<TabStoreData> {
     };
 
     private onNewTabCreated = (payload: Tab): void => {
-        if (payload.id !== null && payload.id !== undefined) {
-            this.state.id = payload.id;
-        }
-        if (payload.title) {
-            this.state.title = payload.title;
-        }
-        if (payload.url) {
-            this.state.url = payload.url;
-        }
+        this.state.id = payload.id;
+        this.state.title = payload.title;
+        this.state.url = payload.url;
         this.state.isClosed = false;
         this.state.isChanged = false;
         this.state.isOriginChanged = false;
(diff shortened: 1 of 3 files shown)


tsconfig.strictNullChecks.json
```diff
@@ -4,6 +4,7 @@
         "strictNullChecks": true
     },
     "files": [
+        "././src/tests/end-to-end/common/page-controllers/nested-iframe-target-page.ts",
```
dbjorge, line 7: Extra `./` at the beginning of the new line (it's benign but breaks sorting of the list)

src/background/stores/tab-store.ts
```diff
@@ -85,11 +91,11 @@ export class TabStore extends BaseStoreImpl<TabStoreData> {
         }
     };
 
-    private originsMatch = (url1: string, url2: string): boolean => {
-        if (url1 == null && url2 == null) {
+    private originsMatch = (url1: string | null, url2: string | null): boolean => {
+        if (url1 === null && url2 === null) {
```
dbjorge, line 95: In most cases we do prefer to use `===` over `==`, but comparisons against `null` are usually intentional exceptions; when you're comparing against `null` specifically, the difference is that `==` will cause `undefined` to be treated the same as `null` where `===` will cause it to be treated differently. It's very rare that we want to treat those two differently in practice.

src/background/stores/tab-store.ts
```diff
@@ -54,9 +54,15 @@ export class TabStore extends BaseStoreImpl<TabStoreData> {
     };
 
     private onNewTabCreated = (payload: Tab): void => {
-        this.state.id = payload.id;
-        this.state.title = payload.title;
-        this.state.url = payload.url;
+        if (payload.id !== null && payload.id !== undefined) {
```
dbjorge: I would lean against using individual conditionals for each of the individual `Tab` properties like this. With the individual conditionals, if a `payload` were to ever contain a mix of null and non-null properties, it would result in `state` being updated into a torn state where some state information is from the new tab and some is from the previous tab. I think that's probably not behavior we want; I'd expect that we probably want `state` to be self-consistent, so I think it probably makes more sense to pick one of:

a) if we get an incomplete new tab notification, assume it's either invalid or irrelevant and that it's safer for us to keep the old state as-is. Log an error and don't mutate the state.
b) if we get an incomplete new tab notification, assume that the old information has been invalidated, whether or not the new information is complete. Put incomplete information in the state and make sure that consumers of the state are able to handle incomplete information.

I think we'd need to research/test whether there are any conditions in which the browser will actually report new tab events with incomplete state, and think about how we want to handle those conditions. If we think there are realistic situations where the browser can report new tab events with incomplete information, we probably need to use approach b. If we are convinced that the browser won't report such events, approach a probably makes more sense.

dbjorge (changes requested):

dbjorge:
You've marked all the items in the PR checklist as n/a, but I think several of them are applicable:

* This is related to existing issue #2869 
* `yarn fastpass` is a good idea; it would have caught the indentation formatting issue
* Adding/updating relevant unit tests would be good for any cases where you end up making observable behavioral changes (eg, if we did keep the new conditionals in the new tab handler, we'd want unit tests verifying the new behavior)
* Verifying code coverage is a useful tool to check whether there are any new paths not covered by unit tests
* The PR title format is pretty much always relevant (you've formatted it correctly, but it's not n/a)

dbjorge (approved):

PR #4673, opened 2021-09-10
feat(electron-playwright): Upgrade Electron and use Playwright for end-to-end tests

madalynrose:
#### Details

* Upgrades Electron to 14.0.0
* Migrates tests from using Spectron to using Playwright
* Includes updates to run Electron e2e tests locally in docker

##### Motivation

As Spectron is no longer being actively maintained, we'd like to migrate to a better-supported end-to-end testing framework.

##### Implementation 

Upgrading Electron to 14.0.0 means:
* updating package.json and yarn.lock to include the new version number
* updating the electron build id

Migrating tests from using Spectron to using Playwright involves:
* adapting our tests from using our custom `SpectronAsyncClient` to using Playwright's `electron` client
  *  [src/tests/electron/common/create-application.ts](src/tests/electron/common/create-application.ts) launches an electron client and passes it to AppController, which is used in all of the tests. Additional changes were made to ensure the client has [access to environment variables](https://github.com/microsoft/accessibility-insights-web/pull/4673/files#diff-bca738149badeee3f1f88acbc46e9d7e0510df1fa62b274abe55a78e68a6db4cR34) and [starts with no persisted data](https://github.com/microsoft/accessibility-insights-web/pull/4673/files#diff-bca738149badeee3f1f88acbc46e9d7e0510df1fa62b274abe55a78e68a6db4cR33). The retry around launching is no longer needed because we are no longer using chromedriver.
  * [src/tests/electron/common/view-controllers/](src/tests/electron/common/view-controllers/) mostly consist of changes to the function calls to match the playwright `Page` api. 
    * One of the major changes involves the difference between Spectron's `waitUntil` and Page's [`waitForFunction`](https://playwright.dev/docs/next/api/class-page/#page-wait-for-function), which pass arguments differently and have scoping differences (waitForFunction does not grant access to any libraries or variables to the function you pass in.) This presented a challenge in the [`LogController`](src/tests/electron/common/view-controllers/log-controller.ts), which watches adb logs and confirms that the correct commands are sent when the UI is interacted with, so it uses a [custom `waitUntil` function](https://github.com/microsoft/accessibility-insights-web/pull/4673/files#diff-87784374bee324a2304fa0187fe96928be73ed273d80a8f8c256dcc41c8ed9e6R65). 
    * Another change is that many of the functions (e.g. `click()`, `isEnabled()` ) already wait for the element to exist, so many `waitForSelector` calls could be deleted because they are redundant.
 * Some changes needed to be made to the app itself. These include:
    * including [`webPreferences` for testing in a linux environment](https://stackoverflow.com/questions/51141037/unable-to-render-webgl-in-electron-with-xvfb)
    * adding an option to [launch electron without persisted data](https://github.com/microsoft/accessibility-insights-web/pull/4673/files#diff-fae1c8f97fb226ed4143bd1ea0c4f7969d846f46d45b44f679dc2344048ef00dR189) (this is specifically to get around electron saving the adb location and not using the mock adb location provided)
    ~~* catching errors around tests attempting to re-register the handler for opening a window and closing windows that have already been closed~~
* Updating configuration to allow our tests to run on Linux:
  *  [Dockerfile](Dockerfile) now includes multistep processes and our calls target specific stages (`web`, `unified`)
  * [pipeline/unified/unified-e2e-test-linux.yaml](pipeline/unified/unified-e2e-test-linux.yaml) includes `DOCKER_BUILDKIT`
  *  [unified-entrypoint.sh](unified-entrypoint.sh) ensures xvfb can be configured since this needs to happen inside the actual docker container when it's running, not during  build.

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/background/get-persisted-data.ts
```diff
@@ -24,8 +24,12 @@ const keyToPersistedDataMapping = {
 export function getPersistedData(
     indexedDBInstance: IndexedDBAPI,
     dataKeysToFetch: string[],
+    options?: { noPersistedData: boolean }, // this option is for tests to ensure they can use mock-adb
 ): Promise<PersistedData> {
     const persistedData = {} as PersistedData;
+    if (options?.noPersistedData) {
```
waabid, line 30: nit: maybe `ignorePersistedData` might be better?

src/electron/main/main.ts
```diff
@@ -36,14 +36,24 @@ if (platformInfo.isLinux()) {
     app.disableHardwareAcceleration();
 }
 
+const webPreferences =
+    process.env.ACCESSIBILITY_INSIGHTS_ELECTRON_LINUX_TESTS === 'true'
+        ? {
```
waabid, line 41: I think it might be better to name these objects.

src/tests/electron/common/create-application.ts
```diff
@@ -1,68 +1,61 @@
...
-    DEFAULT_CHROMEDRIVER_START_RETRIES,
-    DEFAULT_CHROMEDRIVER_START_TIMEOUT_MS,
-} from 'tests/electron/setup/timeouts';
+import { _electron as electron } from 'playwright';
 import { AppController } from './view-controllers/app-controller';
-
-export interface AppOptions extends Partial<AppConstructorOptions> {
+export interface AppOptions {
     suppressFirstTimeDialog: boolean;
+    env?: {
```
waabid, line 8: Should this be optional? In the code below, we're assuming it exists. If we can safely make that assumption then we should change this or otherwise check in other locations.

madalynrose: Yes we rarely pass `env` when calling createApplication (basically only used to set the adb folder to something fake to force a prompt)

waabid (approved):
LGTM

src/tests/electron/common/create-application.ts
waabid: Sorry, I actually misread it. We are checking to validate the options object.

PR #4719, opened 2021-09-23
chore: fix embed-styles escaping

dbjorge:
#### Details

This PR fixes an issue with our custom Grunt `embed-styles` task. The task currently includes logic to escape `\n` and `"` characters as `\"` and `\\n`, but it doesn't escape `\` characters, which means that if one of the CSS files being embedded contains a `\` (or worse, a `\\n` or `\"`), the "escaped" characters will end up unescaped in practice and cause very confusing errors at runtime.

As of current main, this is benign, since no `\` characters appear in any of the css files output by our chain of CSS webpack loaders. However, as of #4681, `mini-css-extract-plugin` now embeds comment banners in dev mode signifying which CSS came from which source file, and those comment banners happen to include lines that end in `\` characters. This causes a mis-escaping problem that results in dev builds of web not being able to successfully load the details view. It's very challenging to debug this issue because it gives Chrome so much trouble parsing the javascript bundles that it can't associate source maps to anything.

This is the "fast and cheap, but not good" tradeoff option to fix this. The much better fix would be to rip out the custom `embed-style` task completely and replace it with a more robust solution for embedding styles in our generated HTML reports (probably something based on creating a template HTML file for the reports to use based on `css-loader` with `exportType: string`, rather than assembling the template ourselves with `<style dangerouslySetInnerHTML={{ __html: detailsViewBundledCSS.styleSheet }} />` like we do today). This might require some substantial changes to how the report package works, though, and I'd rather take the small fix now to unblock the `mini-css-extract-plugin` update and track the better fix separately. #4720 tracks the longer-term solution.

##### Motivation

Unblocks #4681

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn fastpass`


peterdur (approved):
Reviewed and approved.
MS internal folks can look for "tricky dependabot" in OneNote for a walkthrough of the investigation and fix.

Issue #4615, opened 2021-08-25
web: color visualizer does not affect `body` element's background


dbjorge:
## Describe the bug

Accessibility Insights for Web's Color ad-hoc visualizer, which is intended to turn a page grayscale, does not work against background colors/images applied to a page's body element.

**To Reproduce**

1. Go to any page that uses a `background-color` or `background-image` style on its `body` element. For example, https://dequeuniversity.com/demo/mars/ or a local page using the repro in the next section of this issue
2. Active the Color ad-hoc tool
3. Observe that the page's background color does not change

## Repro example

```html
<html>
<head>
<style>
body {
    /* Either one of these repros */
    background-color: #ffaaaa;
    background-image: linear-gradient(#ffaaaa, #aaffaa)
}
</style>
</head>
<body>
<p>some text</p>
</body>
</html>
```

## Expected behavior

The entire page, including the body background, should be filtered to grayscale

## Screenshots

![screenshot showing repro page with colorful background despite active color ad-hoc tool](https://user-images.githubusercontent.com/376284/130868967-b9f04a83-728e-4b58-b79a-0f21540858c2.png)

## Context (please complete the following information)

- OS Name & Version: Windows 21H1 (OS Build 19043.1165)
- AI-Web Version & Environment: Repros on both 2.27.0 (prod) and 2.28.0 (insider)
- Browser Version: Microsoft Edge Version 92.0.902.78 (Official build) (64-bit)
- Target Page: see repro example

## Are you willing to submit a PR?

Yes. I believe the fix is to apply our grayscale filter style to the HTML element, rather than the body element.

## Did you search for similar existing issues?

yes

## Additional context

n/a

peterdur:
This bug feels very familiar, but I've had no luck finding an older instance with a reason that this couldn't be fixed.

Validated that applying the filter on the html element rather than on the body element appears to fix this.

ferBonnin:
reviewed with Peter. To consider that this is also used in Assessment

PR #4728, opened 2021-09-27
fix: color visualizer does not affect body element's background

ThanyaLeif:
#### Details

This PR address #4615 
Before this PR, Accessibility Insights for Web's Color ad-hoc visualizer, does not work against background colors / images applied to a page's body element.

I changed the viewer selector so that the properties are applied to the entire document. This also fixes the bug related to assessment 16.1
Below is the behavior of the ad-hoc color visualizer before and after the changes

The first image shows a page with a gradient background, which is not affected by the visualizer.
image

The second image shows the entire page in grayscale, after the changes were applied.
image

The third image shows how the color rule is also applied from assessment 16.1
image

##### Motivation

Address #4615 

##### Context
For some reason it seems that the 'body' selector only affects its children, not the body itself.

#### Pull request checklist

- [x] Addresses an existing issue: #4615 
- [x] Ran `yarn fastpass`


src/injected/analyzers/filter-results.ts
```diff
@@ -10,7 +10,8 @@ export const filterNeedsReviewResults = (results: ScanResults): ScanResults => {
         v =>
             v.id !== 'aria-input-field-name' &&
             v.id !== 'color-contrast' &&
-            v.id !== 'th-has-data-cells',
+            v.id !== 'th-has-data-cells' &&
+            v.id !== 'label-content-name-mismatch',
```
dbjorge, line 14: Did you intend to include the `filter-results` change (and its tests) here? I suspect they're intended to be for #4339, not this PR

src/scanner/custom-rules/color-rule.ts
```diff
@@ -13,7 +13,7 @@ export const colorConfiguration: RuleConfiguration = {
     ],
     rule: {
         id: 'select-body',
-        selector: 'body',
+        selector: '*',
```
dbjorge, line 16: I think the `select-body` id is misleading now that this is selecting more than just the body element. Could you update the ID to something that reflects the new selector?

src/injected/analyzers/filter-results.ts
ThanyaLeif: Right, this belongs to #4339
I undid changes to this file and `src/tests/unit/tests/scanner/custom-rules/color-rule.test.ts` in my last commit.

src/scanner/custom-rules/color-rule.ts
ThanyaLeif: Done
I renamed this and other related ids in my last commit.

dbjorge: Thanks! However, in this context, the term "document" has a specific meaning which is still different from what's being selected here. Maybe `select-each-element` or `select-all-elements`?

waabid: I would also extract the name out to a variable so you can reference that in `use-of-color.tsx` and `visualization.tsx`.

src/scanner/custom-rules/color-rule.ts
```diff
@@ -12,8 +12,8 @@ export const colorConfiguration: RuleConfiguration = {
         },
     ],
     rule: {
-        id: 'select-body',
-        selector: 'body',
+        id: 'select-document',
```
waabid, line 15: While I think this solution works, I think a simpler solution may suffice. With regards to both uses of this rule, we only add the css styles to a single/first element in the `SingleTargetDrawer` (line 47 onwards `single-target-drawer.ts`).

In the webpage examples given, I believe the first element returned is actually the `html` element, so if we change the selector to `html` we should get the same result (and can also rename to `select-html`, per Dan's suggestion).

waabid (changes requested):
Left a comment.

src/scanner/custom-rules/color-rule.ts
ThanyaLeif: Makes sense to me
Is there a specific file in the project where we put those common variables @waabid ?

waabid: I think the `color-rule.tsx` file should be fine for this purpose in this case.

dbjorge (approved):

waabid (approved):
🚢

PR #4774, opened 2021-10-05
feat(download-report-json): Modify export telemetry to include export service

jalkire:
#### Details
This PR updates the telemetry logged on report export click in two ways:
1. `exportResultsService` is added to the event with potential values of `html`, `json`, and `codepen`
2. `exportResultsData` is removed from the event (read more in Context)

##### Motivation
Telemetry updates for the json export feature

##### Context
In theory, `exportResultsData` was recording the length of the downloaded html string. In reality, it had a bug which caused it to record the length of the _previously downloaded_ html string from that session, which was usually 0. This PR fixes that bug by removing the value entirely (per discussion with @ferBonnin, we don't need that data).

#### Pull request checklist

- [x] Ran `yarn fastpass`


brocktaylor7 (approved):
LGTM

PR #4836, opened 2021-10-27
feat(tab-stop-details-view): Add Tab stop requirement actions to ActionCreator

madalynrose:
#### Details

Create action listeners and messages for the new Tab Stops actions from #4826 

This includes:

###### Actions/Listeners
* moving tab stop requirement actions into their own file as `TabStopRequirementActions`
* including those actions in the `ActionHub`, where `VisualizationScanResultActions` are also surfaced
* registering callbacks for those actions in the `TabStopRequirementActionCreator`
* passing these actions to the `VisualizationScanResultStore` in the `TabContextStoreHub`
* adding listeners to update `VisualizationScanResultStore` if those actions are triggered
* adding messages to call these background actions from other parts of the extension
* creating a new action message creator `TabStopRequirementActionMessageCreator` inside `DetailsView/actions` to send information to the background using those messages

###### Telemetry
* adding telemetry events for each of these actions
* creating a new TelemetryData method to generate telemetry for a tab stop requirement

##### Motivation

Feature Work

#### Pull request checklist
- [x] Ran `yarn fastpass`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`).

### DIFF
--- a/src/DetailsView/actions/tab-stop-requirement-action-message-creator.ts
+++ b/src/DetailsView/actions/tab-stop-requirement-action-message-creator.ts
@@ -14,11 +14,14 @@ import { TabStopRequirementId } from 'types/tab-stop-requirement-info';
 import { TabStopRequirementStatus } from '../../common/types/store-data/visualization-scan-result-data';
 const messages = Messages.Visualizations.TabStops;
 
-export class DetailsViewActionMessageCreator extends DevToolActionMessageCreator {
+export class TabStopRequirementActionMessageCreator extends DevToolActionMessageCreator {
     public addTabStopInstance(requirementId: TabStopRequirementId, description: string): void {
+        const telemetry = this.telemetryFactory.forTabStopRequirement(requirementId);
+
         const payload: AddTabStopInstancePayload = {
             requirementId,
             description,
+            telemetry,
         };
 
         this.dispatcher.dispatchMessage({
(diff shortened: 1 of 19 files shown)


src/background/actions/action-creator.ts
```diff
@@ -99,6 +106,26 @@ export class ActionCreator {
             this.getScanResultsCurrentState,
         );
 
+        this.interpreter.registerTypeToPayloadCallback(
```
waabid, line 109: I would move the tab stop actions to a separate action-creator; better to have smaller action creators and this seems like a good scope for one.

src/DetailsView/actions/tab-stop-requirement-action-message-creator.ts
```diff
@@ -0,0 +1,84 @@
...
+            messageType: messages.RemoveTabStopInstance,
+            payload,
+        });
+    }
+
+    public updateTabStopRequirementStatus(
+        requirementId: TabStopRequirementId,
+        status: TabStopRequirementStatus,
+    ): void {
+        const telemetry = this.telemetryFactory.forTabStopRequirement(requirementId);
```
waabid, line 71: We usually send telemetry for how the action was initiated/triggered (keyboard or mouse for example); maybe we should do that here (or check with PM team on whether we still need to do that).

madalynrose: I based it on the instances actions from DetailsView but would like to know what is preferred

src/background/actions/action-creator.ts
madalynrose: done!

src/DetailsView/actions/tab-stop-requirement-action-message-creator.ts
waabid: Yeah, on a look-through, it's fairly inconsistent. Let's check with the PM team on what they would like.

waabid: Just checked with PM team on this: this is good as is!

waabid (approved):

PR #4865, opened 2021-11-03
feat(tab-stops-details-view): add action for undo in requirements table

madalynrose:
#### Details

This sets up the action for resetting a requirement status in the store and allows it to be called from the requirements table.

I chose to name the methods using `reset` and not `undo` in the code for clarity around behavior (e.g. If a user selects pass (`pass` status) and then selects fail (`fail` status) and then clicks the undo arrow icon, the status will be set to `unknown` and not back to `pass`)

#### Pull request checklist

- [x] Ran `yarn fastpass`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`).


src/common/types/store-data/visualization-scan-result-data.ts
```diff
@@ -17,7 +17,13 @@ export interface TabbedElementData extends TabStopEvent {
     propertyBag?: TabOrderPropertyBag;
 }
 
-export type TabStopRequirementStatus = 'pass' | 'fail' | 'unknown';
+export enum TabStopRequirementStatuses {
```
madalynrose, line 20: I thought it might be good for us to be able to reference the statuses directly instead of hardcoding their strings places

brocktaylor7 (approved):
I like the change to the enum, LGTM!

Issue #4811, opened 2021-10-20
[Bug] Next requirement button doesn't expose "next" or ">" behavior to ATs 


karanbirsingh:
## Describe the bug

This bug was made from https://github.com/microsoft/accessibility-insights-web/issues/3688, which had a couple issues bundled together.

Here are comments from Rob about the next requirement button used in assessment:

> Note: The button gives no accessible indication it is meant to move to the next requirement. My understanding is there is a visual next indicator (the presence of a rightward pointing arrow). This could be resolved by
> 
> Labeling the button simply as "next" or "next requirement" for everyone
> Adding a character in the accessible label which connotes next such as '>' or '→'.
> Adding a brief text indicator to the accessible label before or after the text of the next requirement, e.g., "next: On focus"
> 

**To Reproduce**

1. Start an assessment
2. Select any requirement that has a 'next requirement' button on the bottom right

## Screenshots

![relevant next requirement button circled, with its devtools accessibility properties highlighted](https://user-images.githubusercontent.com/7775527/138159737-6f2fcdb3-40c3-4e0d-a2d6-0c9fffb64125.png)

## Context (please complete the following information)

- AI-Web Version & Environment: Version 2.29.0

PR #4866, opened 2021-11-03
fix: Add aria label to chevron on next requirement button, so next is read aloud

jlperkins:
#### Details

This PR adds an aria label to the chevron icon on the next requirement button.  This adds the word "next" to what screen readers say aloud when describing the next requirement button. For example, screen readers now say "xyz next button" instead of "xyz button." This allows users who use a screen reader to become apprised of the same information as visual users; the word "next" provides the same indication as the chevron visually does. 
 
##### Motivation

Addresses issue #4811 

##### Context

Adding the aria label to the chevron was preferable over adding an aria label to the button itself because of varying NVDA behavior.  If the aria label was on the button, upon clicking the button, NVDA would then read aloud the aria label of the new next requirement button that was loaded on the following page.  JAWS, however, would not.  Before this change, neither screen reader would read the newly loaded button's label aloud.  Putting the label on the chevron made it so that neither NVDA nor JAWS read the newly loaded button aloud, which is consistent with current expectations, and still fixed the bug as intended. 

#### Pull request checklist

- [x] Addresses an existing issue: #4811
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/next-requirement-button.tsx
+++ b/src/DetailsView/components/next-requirement-button.tsx
@@ -40,7 +40,7 @@ export const NextRequirementButton = NamedFC<NextRequirementButtonProps>(
                 onClick={selectNextRequirement}
             >
                 <span>
-                    <Icon iconName="ChevronRight" />
+                    <Icon iconName="ChevronRight" ariaLabel={'next'} />
                 </span>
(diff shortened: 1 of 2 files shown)


shanisebarona (approved):
lgtm!

PR #4868, opened 2021-11-04
feat(tab-stops-details-view): Add tab stops requirement table to DOM tree

madalynrose:
#### Details

This PR adds the new `TabStopsRequirementTableView` to the `AdhocTabStopsTestView`. It also adds `ContentLink` (guidance) and `VisualizationToggle`.

##### Motivation
screenshot showing guidance, toggle, and requirements table in UI

Feature Work

##### Context

Right now `addFailureInstanceForRequirement` remains unimplemented as it needs to hook up with work in another PR.

#### Pull request checklist
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [x] (UI changes only) Added screenshots/GIFs to description above
- [ ] (UI changes only) Verified usability with NVDA/JAWS

PR #4876, opened 2021-11-10
feat(tab-stops-details-view): Add tab stops requirement table to DOM tree

shanisebarona:
#### Details

This PR merges #4868 authored by @madalynrose. 

----------------------------------------------------

This PR adds the new `TabStopsRequirementTableView` to the `AdhocTabStopsTestView`. It also adds `ContentLink` (guidance) and `VisualizationToggle`.

##### Motivation
screenshot showing guidance, toggle, and requirements table in UI

Feature Work

##### Context

Right now `addFailureInstanceForRequirement` remains unimplemented as it needs to hook up with work in another PR.

#### Pull request checklist
- [x] Ran `yarn fastpass`


src/DetailsView/components/adhoc-tab-stops-test-view.tsx
```diff
@@ -1,29 +1,49 @@
...
 import * as React from 'react';
+import { ContentLink, ContentLinkDeps } from 'views/content/content-link';
+import { ContentReference } from 'views/content/content-page';
 import * as Markup from '../../assessments/markup';
 
-export type AdhocTabStopsTestViewDeps = TabStopsFailedInstanceSectionDeps;
+export type AdhocTabStopsTestViewDeps = TabStopsRequirementsTableDeps &
+    TabStopsFailedInstanceSectionDeps &
+    ResultSectionDeps &
+    StaticContentDetailsViewDeps &
```
waabid, line 34: Neither ResultSection or StaticContentDetailsView are child components here so we should remove the deps for them. If we need something in this component we should be adding it here explicitly.

src/DetailsView/components/adhoc-tab-stops-test-view.tsx
```diff
@@ -1,29 +1,49 @@
...
+    ContentLinkDeps;
 
 export interface AdhocTabStopsTestViewProps {
     deps: AdhocTabStopsTestViewDeps;
     configuration: VisualizationConfiguration;
+    tabStoreData: Pick<TabStoreData, 'isChanged'>;
```
waabid, line 40: I don't see where this prop is being used.

src/DetailsView/components/adhoc-tab-stops-test-view.tsx
```diff
@@ -1,29 +1,49 @@
...
     featureFlagStoreData: FeatureFlagStoreData;
+    visualizationStoreData: VisualizationStoreData;
     visualizationScanResultData: VisualizationScanResultData;
     selectedTest: VisualizationType;
+    clickHandlerFactory: DetailsViewToggleClickHandlerFactory;
+    guidance?: ContentReference;
```
waabid, line 46: Don't think guidance should be optional here.

src/tests/unit/tests/DetailsView/components/adhoc-tab-stops-test-view.test.tsx
```diff
@@ -2,30 +2,118 @@
...
+            .returns(() => scanDataStub)
+            .verifiable();
+
+        clickHandlerFactoryMock
+            .setup(chfm => chfm.createClickHandler(selectedTest, !scanDataStub.enabled))
+            .returns(() => clickHandlerStub)
+            .verifiable();
+    });
+
+    it('should return target page changed view as tab is changed', () => {
```
waabid, line 81: This uses the tabStoreData; we should either remove this test case here or not remove tabStoreData and get this test case to be correct since currently it does not return the target page changed view in the snap shot.

src/tests/unit/tests/DetailsView/components/adhoc-tab-stops-test-view.test.tsx
```diff
@@ -2,30 +2,118 @@
...
+        props.tabStoreData = {
+            isChanged: true,
+        };
+
+        const actual = shallow(<AdhocTabStopsTestView {...props} />);
+        expect(actual.debug()).toMatchSnapshot();
```
waabid, line 87: We should use getElement instead of debug (and below as well).

src/DetailsView/components/adhoc-tab-stops-test-view.tsx
shanisebarona: Looks like `guidance` is optional elsewhere, in src\common\configs\visualization-configuration.ts, and removing it fails the e2e test.

waabid (approved):

PR #4885, opened 2021-11-12
feat(tab-stops-details-view): create tab stops view store, actions, and controller

waabid:
#### Details

Creates a store to be used in details view to manage state of the tab stops view.

##### Motivation

feature work.

##### Context

This is a new approach to a kind of problem where we usually have stateful components that pass state-management functions down to children. This would be complicated and messy in this case (although possible) since there are multiple places that the failed instance panel could open from (not included in this PR).

In this case, we create a store that only exists in details view and thus our components can be simpler and not do any state management themselves.

The controller is a more palatable approach than interacting with actions directly (don't want to create confusion with our usage of action-message-creators).

The usage of this in UI components will be in a future PR.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/tab-stops/tab-stops-test-view-controller.ts
+++ b/src/DetailsView/components/tab-stops/tab-stops-test-view-controller.ts
@@ -1,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import {
+    EditExistingFailureInstancePayload,
+    TabStopsViewActions,
+} from 'DetailsView/components/tab-stops/tab-stops-view-actions';
+
+export class TabStopsTestViewController {
+    constructor(private actions: TabStopsViewActions) {}
+
+    public createNewFailureInstancePanel: (payload: string) => void = payload =>
+        this.actions.createNewFailureInstancePanel.invoke(payload);
+
+    public editExistingFailureInstance: (payload: EditExistingFailureInstancePayload) => void =
+        payload => this.actions.editExistingFailureInstance.invoke(payload);
+
+    public dismissPanel = () => this.actions.dismissPanel.invoke();
+
+    public updateDescription: (payload: string) => void = payload =>
+        this.actions.updateDescription.invoke(payload);
+}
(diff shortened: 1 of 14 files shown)


src/common/types/captured-instance-action-type.tsx
```diff
@@ -0,0 +1,4 @@
+export enum CapturedInstanceActionType {
```
brocktaylor7, line 1: Missing copyright/license.

src/DetailsView/components/tab-stops/tab-stops-test-view-controller.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import {
+    EditExistingFailureInstancePayload,
+    TabStopsViewActions,
+} from 'DetailsView/components/tab-stops/tab-stops-view-actions';
+
+export class TabStopsTestViewController {
```
brocktaylor7, line 9: I think the controller is a great way to handle what we chatted about yesterday. I like this and it feels like it will definitely be less confusing and more palatable!

src/tests/unit/tests/DetailsView/components/tab-stops/tab-stops-view-store.test.ts
```diff
@@ -0,0 +1,108 @@
...
+            requirementId,
+        };
+        createStoreForTabStopsViewActions('editExistingFailureInstance')
+            .withActionParam(payload)
+            .testListenerToBeCalledOnce(initialState, finalState);
+    });
+
+    test('onCreateNewFailureInstancePanel', () => {});
+
+    test('onCreateNewFailureInstancePanel', () => {});
```
brocktaylor7, line 95: These should probably be removed, already tested above.

brocktaylor7 (approved):
LGTM, a couple minor comments.

src/tests/unit/tests/DetailsView/components/tab-stops/tab-stops-view-store.test.ts
brocktaylor7, line 95: These should probably be removed, already tested above

PR #4888, opened 2021-11-12
fix: Adjust styling so text is visible at high zoom level

jlperkins:
#### Details

This PR adds the white-space: normal style to information displayed in the details view in tables during an assessment.  This makes it so that at a high zoom level, instead of being truncated with "..." like before, information in a table cell that extends past one line will now wrap and continue in the same cell on the next line.  This makes that content visible to users who are using a high zoom level. 

##### Motivation

Addresses ADO Trusted Tester bug #1883866.

##### Context

Assessment test steps that render instances in a table either do so through generating an AssessmentInstanceDetailsColumn or by using the PropertyBagColumnRendererFactory.  So, styling here needed to be updated in two spots to account for the two different places we generate content. 

image
Pic above: The issue mentioned in the trusted tester bug, with a long header instance being truncated and no way to reveal the entire sentence

image
Pic above: The issue mentioned in the trusted tester bug, now resolved with word wrap, so the instance in the table is taller and the entire sentence is visible. 

image
Pic above: The same issue, but in a table for the images test, whose instances are rendered via a different method than for headings. Instead of being truncated with ellipses, the text is simply truncated at the edge of the table cell

image
Pic above: The same issue, but in a table for the images test, now resolved in the same manner (word wrap)

image
Pic above: The issue affecting manually added instance failures

image
Pic above: The issue, now resolved, no longer affecting manually added instance failures. 

#### Pull request checklist

- [x] Addresses an existing issue: Trusted Tester bug
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/Styles/detailsview.scss
+++ b/src/DetailsView/Styles/detailsview.scss
@@ -39,6 +39,12 @@ body {
     flex-direction: column;
     height: 100%;
 
+    .ms-DetailsRow {
+        .property-bag-div {
+            white-space: normal;
+        }
+    }
+
     .ms-DetailsRow,
     .ms-GroupHeader {
         user-select: text;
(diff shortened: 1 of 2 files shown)


shanisebarona (approved):
lgtm!

PR #4901, opened 2021-11-18
fix: Modify screen-reader-only class to improve Orca's reading behavior

jalkire:
#### Details
Currently, we use a screen-reader-only class that makes elements offscreen and 1px x 1px. Unfortunately, shortening the element's width to 1px causes the Orca screen reader to behave in an undesirable manner wherein it reads one word at a time and re-reads the element type with each word. To fix this, this PR removes width: 1px from the screen-reader-only class. Instead, clip-path is used to ensure the element is not visible.

##### Motivation
Addresses accessibility issue in AI-Android (and AI-Web) on Linux.

##### Context
I considered simply eliminating width since the class already sets left:-10000px, but decided to add the clip-path as a redundancy. 

This should have no visible impact and no impact to other screen readers' behavior.

#### Pull request checklist

- [x] Addresses an existing issue: ADO 1886155
- [x] Ran `yarn fastpass`

### DIFF
--- a/packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
+++ b/packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
@@ -400,9 +400,9 @@
         position: absolute;
         left: -10000px;
         top: auto;
-        width: 1px;
         height: 1px;
         overflow: hidden;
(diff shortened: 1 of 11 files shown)


lisli1 (approved):
LGTM

PR #4904, opened 2021-11-19
fix: Use aria-label on NavLinkButton to improve Orca reading behavior

jalkire:
#### Details
Currently, we set aria-current="page" on the left nav bar button for the currently selected page. This is right and proper. Unfortunately, the Orca screen reader ignores this setting in both the Web and Android products. To improve the user's experience, this PR replaces the use of aria-current with an aria-label to simulate the proper reading behavior. 

##### Motivation
Addresses accessibility bug in Linux.

##### Context
I considered leaving the aria-current value as-is, but that results in a double reading of current page from non-Orca screen readers.

#### Pull request checklist

- [x] Addresses an existing issue: ADO 1886192
- [x] Ran `yarn fastpass`


ThanyaLeif (approved):
Looks good to go.

lisli1 (approved):
LGTM

PR #4918, opened 2021-11-23
fix: Replace issue filing URL placeholders with separate label to improve Orca behavior

jalkire:
#### Details
Currently, we set the placeholder value of our azure boards and github issue filing TextFields to be an example URL. Unfortunately, the Orca screen reader ignores these placeholder values. To improve the experience for Linux screen reader users, this PR removes the placeholder values and replaces them with "example: \<url>" text beneath the TextFields. 

GitHub in the issue filing popup:
GitHub in the issue filing popup screenshot
Azure Boards in the settings panel:
Azure Boards in the settings panel screenshot

##### Motivation
Addresses accessibility issue.

##### Context

- I considered leaving the placeholder and just adding on the URL as an aria-description. This worked alright with Orca, but then caused other platforms' screen readers to read both the placeholder and the aria-description.
- I considered not setting the TextField's aria-describedby to be the new span, and honestly I'm still on the fence about it. On the plus side, now AT will read the example URL by default like the placeholder. On the negative side, AT will still read the example URL even after the user has entered their own URL.
- I considered pursuing a method to provide different accessibility information on different platforms, but agreed with @ferBonnin that this should be avoided

#### Pull request checklist

- [x] Addresses an existing issue: ADO 1886158
- [x] Ran `yarn fastpass`

### DIFF
--- a/packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
+++ b/packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
@@ -782,34 +782,34 @@
         margin-bottom: unset;
       }</style
     ><style>
-      .report-congrats-message--BoHie {
+      .report-congrats-message--UNXUb {
         word-break: break-word;
         overflow-wrap: break-word;
       }
-      .report-congrats-head--_Lf9p {
+      .report-congrats-head--_IdFb {
         font-size: 16px;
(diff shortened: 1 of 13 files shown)


ThanyaLeif (approved):
LGTM

lisli1 (approved):

Issue #4492, opened 2021-07-15
Android: export dialog disappears if window is resized


waabid:
## Describe the bug

The export result dialog will disappear if the window is resized to where the export button moves to the "..." menu.

**To Reproduce**

1. Use the export result button.
2. The dialog will show up.
3. Resize window such that the export result dialog moves from the vertical "..." menu to outside of it or vice versa.

## Expected behavior

It should still show the dialog

## Context (please complete the following information)

- OS Name & Version: n/a
- AI-Web Version & Environment: Android

## Are you willing to submit a PR?

Yes

ferBonnin:
reviewed with @peterdur. we should address this

PR #4947, opened 2021-12-01
Fix: Android export dialog disappears on reflow

brocktaylor7:
#### Details

This fixes a reflow issue that causes the export dialog to disappear when it shouldn't. This was because of the button and export dialog being rendered in a single component, so when that component would move to the collapsed menu due to reflow, the export dialog would disappear.

Before with reflow making dialog disappear:
original_unified_resize

After the fix:
new_unified_resize

##### Motivation

Addresses issue #4492 

##### Context

The diff is a little confusing and it may make sense to look at the first two commits separately. This is due to the fact that we had two components doing the same thing in web and unified. One was called "ReportExportComponent" and one was called "ExportDialogWithLocalState".  I much prefer the name "ReportExportComponent" but the code in ExportDialogWithLocalState was much closer to what I needed to unify the two. So I made the changes in ExportDialogWithLocalState in the first commit, then renamed it back to ReportExportComponent in the second commit. (Reach out if there is any confusion :))

I completely removed the entity formerly known as ReportExportComponent (which was only used in Unified).

#### Pull request checklist

- [x] Addresses an existing issue: #4492
- [x] Ran `yarn fastpass`


ThanyaLeif (approved):
Looks good to me, nicely done!

PR #4967, opened 2021-12-08
feat(tabstops-report) add NeedsReviewScanResultStore and NeedsReviewCardSelectionStore

brocktaylor7:
#### Details

This adds two stores, NeedsReviewScanResultStore and NeedsReviewCardSelectionStore. These will be used to independently store information for Needs Review, rather than reusing the UnifiedScanResultStore and CardSelectionStore that is being shared with Automated Checks.

##### Motivation

Feature work, we need this data persistence to be separate to enable the report to always be able to export Automated Checks data, without it being overwritten by Needs Review data.

##### Context

There is a lot of context that went into this decision. We considered going the route of turning the existing UnifiedScanResultStore into a dictionary with keys for each test type. We also considered having multiple instances of the UnifiedScanResultStore. Due to a number of technical and time constraints, we decided ultimately that our best option was the break these stores into separate stores for automated checks and needs review.  The UnifiedScanResultStore will likely eventually be renamed the AutomatedChecksScanResultStore but I'm leaving it for now.

These stores are partially wired up but will be fully wired up in a later PR.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/types/store-data/needs-review-card-selection-store-data.ts
+++ b/src/common/types/store-data/needs-review-card-selection-store-data.ts
@@ -1,19 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export type NeedsReviewCardSelectionData = { [resultInstanceUid: string]: boolean };
+import { CardSelectionStoreData } from 'common/types/store-data/card-selection-store-data';
 
-export interface RuleExpandCollapseData {
-    isExpanded: boolean;
-    cards: NeedsReviewCardSelectionData;
-}
-
-export interface RuleExpandCollapseDataDictionary {
-    [ruleId: string]: RuleExpandCollapseData;
-}
-
-export interface NeedsReviewCardSelectionStoreData {
-    rules: RuleExpandCollapseDataDictionary;
-    visualHelperEnabled: boolean;
-    focusedResultUid: string | null;
-}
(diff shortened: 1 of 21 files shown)


src/common/types/store-data/needs-review-card-selection-store-data.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface RuleExpandCollapseData {
+    isExpanded: boolean;
+    cards: NeedsReviewCardSelectionData;
+}
+
+export interface RuleExpandCollapseDataDictionary {
+    [ruleId: string]: RuleExpandCollapseData;
+}
+
+export interface NeedsReviewCardSelectionStoreData {
```
jalkire, line 15: What would your thoughts be on sharing some of these types/interfaces with the other CardSelectionStore? `RuleExpandCollapseDataDictionary` in particular doesn't feel like it needs to be duplicated, but I could see an argument for sharing the rest as a standard set of card selection store related types.

brocktaylor7: This is a good point, I've changed it to use a type that simply points to the existing data types. There's no reason we can't re-use them currently, and this makes it fairly easy to decouple down the road if we ever decide we need to! Good catch!

jalkire (approved):
🚢

PR #4974, opened 2021-12-10
feat(tabstops-report): Separate Needs Review and Automated Checks into relevant stores

brocktaylor7:
#### Details

This wires up the Needs Review section to the new NeedsReviewScanResultStore and NeedsReviewCardSelectionStore. 

##### Motivation

Feature work. We needed to be able to store both needs review and automated checks at the same time so that if you navigated to needs review, scan, and navigate back to tab stops to export, the data for automated checks will still exist.

##### Context

This doesn't change the current workflow for when you must scan/rescan a test. You still have to click the button to scan needs review, and if you return to automated checks you must rescan. This will be updated in a later PR to improve the user experience.

I considered various approaches to try to avoid the mass propagating of CardSelectionMessageCreators, including using the visualization configuration to hold the relevant CardSelectionMessageCreator. This would be potentially problematic on the android side, as we don't use visualization configurations. After trying various approaches, the solution that was the most straight forward was creating a generic interface (CardSelectionMessageCreator) that the AutomatedChecksCardSelectionMessageCreator and the NeedsReviewCardSelectionMessageCreator both implement. This allows us to pass the relevant one via props. This approach allowed me to reduce the amount of code duplication that would've been necessary to separate various shared props between the two.

Sorry for the very large PR. Once I started separating out Needs Review and Automated Checks I didn't see a straight-forward way of breaking it into smaller PRs without leaving the application in a broken intermediate state.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/adhoc-issues-test-view.tsx
+++ b/src/DetailsView/components/adhoc-issues-test-view.tsx
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 import { CommonInstancesSectionProps } from 'common/components/cards/common-instances-section-props';
 import { VisualizationConfiguration } from 'common/configs/visualization-configuration';
+import { CardSelectionMessageCreator } from 'common/message-creators/card-selection-message-creator';
 import { ScanIncompleteWarningId } from 'common/types/scan-incomplete-warnings';
 import { CardsViewModel } from 'common/types/store-data/card-view-model';
 import { FeatureFlagStoreData } from 'common/types/store-data/feature-flag-store-data';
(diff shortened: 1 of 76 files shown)


jalkire:
I'm seeing an issue where selecting needs review instance cards does not cause the target page to scroll to the corresponding element. I think the cause might be that the `FocusChangeHandler` (which is responsible for scrolling the target page) [only pays attention to the old unified results/card selection stores](https://github.com/brocktaylor7/accessibility-insights-web/blob/00b09800a29f920fc609f2ec4d72d0551d7b0097/src/injected/focus-change-handler.ts#L55)

brocktaylor7:
> I'm seeing an issue where selecting needs review instance cards does not cause the target page to scroll to the corresponding element. I think the cause might be that the `FocusChangeHandler` (which is responsible for scrolling the target page) [only pays attention to the old unified results/card selection stores](https://github.com/brocktaylor7/accessibility-insights-web/blob/00b09800a29f920fc609f2ec4d72d0551d7b0097/src/injected/focus-change-handler.ts#L55)

ooh good catch, I'm taking a look into a fix for this now!

jalkire (approved):
👍

PR #5004, opened 2022-01-03
feat(tabstops-report): Initial FastPass Report E2E setup

brocktaylor7:
#### Details

This adds E2E testing for the fastpass report. It allows us to download the report and use browser automation on the downloaded html file. For this initial PR, the E2E test checks whether the report has rendered and does an accessibility scan.

##### Motivation

Feature work.

##### Context

Currently there is an existing accessibility issue in the report. Due to this, I've made the accessibility check save as a snapshot rather than having it check the length of the accessibility check results array. Once the accessibility check has been resolved we can remove the snapshot and change back to checking the results array.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/end-to-end/tests/details-view/fast-pass-report.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        reportDownload = await detailsViewPage.downloadExportReport(
+            detailsViewSelectors.singleExportToHtmlButton,
+            reportSaveAsFilePath,
+        );
+        reportPage = await browser.newHtmlReportPage(reportDownloadUrl);
+    });
+
+    afterAll(async () => {
+        // delete GUID file auto-generated by the download
+        await detailsViewPage.deleteDownloadedFile(reportDownload);
```
dbjorge, line 54: It would be good to make sure that the `browser?.close()` step still runs if either/both of the `delete` steps fail. Particularly, `afterAll` still runs if the `beforeAll` fails; if any setup step fails before the download happens, we want the browser to still get cleaned up

src/tests/end-to-end/common/page-controllers/html-report-page.ts
```diff
@@ -0,0 +1,15 @@
...
+import * as Playwright from 'playwright';
+import { CommonSelectors } from 'tests/end-to-end/common/element-identifiers/common-selectors';
+import { Page, PageOptions } from './page';
+
+export class HtmlReportPage extends Page {
+    constructor(underlyingPage: Playwright.Page, options?: PageOptions) {
+        super(underlyingPage, options);
+    }
+
+    public async ensureNoModals(): Promise<void> {
```
dbjorge, line 12: Is this a copy/paste leftover? I don't see this being used anywhere

src/tests/end-to-end/tests/details-view/fast-pass-report.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const reportDownloadUrl = getTestResourceUrl('fast-pass-report.html');
+
+describe('Details View -> FastPass -> Report', () => {
+    let browser: Browser;
+    let targetPage: TargetPage;
+    let detailsViewPage: DetailsViewPage;
+    let backgroundPage: BackgroundPage;
+    let reportPage: HtmlReportPage;
+    const reportSaveAsFilePath = 'src/tests/end-to-end/test-resources/fast-pass-report.html';
```
dbjorge, line 27: I think it would be preferable to avoid having a test mutate a file under `src` at runtime (unless it's through an explicit snapshotting mechanism). Could we instead just omit the `saveAs` call entirely and use `await reportDownload.path()` as the `reportDownloadUrl`?

src/tests/end-to-end/common/page-controllers/details-view-page.ts
```diff
@@ -122,14 +123,26 @@ export class DetailsViewPage extends Page {
...
             this.clickSelector(selector),
         ]);
-        const suggestedName = download.suggestedFilename();
+        if (saveAsFileName !== undefined) {
+            await download.saveAs(saveAsFileName);
+        }
+        return download;
+    }
+
+    public async getDownloadFileName(download: Download) {
```
dbjorge, line 140: I think the only test that uses this would probably read a bit more clearly if we just inlined this helper - when I first read `getDownloadFileName` in the context of `overview.test.ts`, I wondered if this meant "the filename our extension suggested" vs "the generated filename Playwright actually used". I think it's more obvious that the test is testing the right thing to have it read `suggestedFilename`.

src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -138,6 +138,15 @@ export class Page {
         });
     }
 
+    public async waitForAndClickSelector(selector: string): Promise<void> {
```
dbjorge, line 141: Did you find that this was necessary for the test to work? Playwright documents that `Page.click` already begins by waiting for the selector in question, so I would expect that this is actually just doubling the timeout for the operation in question without making it more reliable.

src/tests/end-to-end/common/page-controllers/page.ts
```diff
@@ -138,6 +138,15 @@ export class Page {
...
+        await this.waitForSelector(selector);
+        await this.clickSelector(selector);
+    }
+
+    public deleteFile(filePath: string): void {
```
dbjorge, line 146: This can be dropped if you accept the other comment suggesting removing the `saveAs` step

dbjorge (changes requested):
See inline

src/tests/end-to-end/tests/details-view/fast-pass-report.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    await detailsViewPage.clickSelector(tabStopsSelectors.primaryAddFailedInstanceButton);
+
+    //check for failed instances section
+    await detailsViewPage.waitForSelector(tabStopsSelectors.failedInstancesSection);
+
+    // expand collapsible content to reveal failed instance
+    await detailsViewPage.clickSelector(tabStopsSelectors.collapsibleComponentExpandToggleButton);
+
+    //await text of failed instance
+    await detailsViewPage.waitForSelector(tabStopsSelectors.instanceTableTextContent);
```
dbjorge, line 102: It might be a good smoke test to include a check here that the actual text content matches `initialFailureInstanceText`

src/tests/end-to-end/tests/details-view/fast-pass-report.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    it('renders', async () => {
+        await reportPage.waitForSelector(fastPassReportSelectors.reportHeaderSection);
+    });
+});
+
+async function openTabStopsPage(detailsViewPage: DetailsViewPage): Promise<void> {
+    await detailsViewPage.waitForAndClickSelector(tabStopsSelectors.navDataAutomationId);
+}
+
+async function addTabStopsFailure(detailsViewPage: DetailsViewPage): Promise<void> {
```
dbjorge, line 77: It's surprising to me that an earlier feature didn't add some of this infrastructure around e2e testing for the tab stops view. Do we know if leaving out e2e tests from the earlier tab stops UI feature was an intentional bit of tech debt? If not, we should probably note for a future retro that it would be good to circle back on how to avoid forgetting these in future features

src/tests/end-to-end/common/page-controllers/html-report-page.ts
brocktaylor7: Good catch!

src/tests/end-to-end/tests/details-view/fast-pass-report.test.ts
brocktaylor7: Per our conversation, I moved it to use the temp `test-results` directory.

src/tests/end-to-end/common/page-controllers/page.ts
brocktaylor7: This was a misunderstanding on my part. I wasn't aware that Page.click waited for the selector, I've removed it!

src/tests/end-to-end/common/page-controllers/page.ts
brocktaylor7: per our conversation this has been left in.

src/tests/end-to-end/tests/details-view/fast-pass-report.test.ts
brocktaylor7: This is handled now with the refactor of the tabstops e2e test. Part of the `addFailedTabStopsInstance` function is testing that the text content matches.

src/tests/end-to-end/tests/details-view/fast-pass-report.test.ts
brocktaylor7: There was a fair bit done, it wasn't in a very reusable format. I've refactored it to use the same code in both places.

dbjorge (approved):

PR #5030, opened 2022-01-06
feat(manifest-v3): Create dev-mv3 build as a sandbox for MV3 conversion

DaveTryon:
#### Details

This PR updates the build system to create both a `dev` and a `dev-mv3` build. The `dev-mv3` build will eventually replace the `dev` build, but only after we complete the work needed to migrate from manifest version 2 (MV2) to manifest version 3 (MV3). Since the Chrome team chose to change the schema between MV2 and MV3, some extra work is needed to create a version-specific manifest.json file. MV3 doesn't support `eval`, so we had to change the webpack's [devtool](https://webpack.js.org/configuration/devtool/) option from `eval-source-map` to an option that doesn't use `eval`. I experimented with both `inline-cheap-source-map` and `inline-source-map`, which is slightly faster. I opted for the slower option because the difference was only about 5 seconds per build.

The `dev-mv3` build doesn't scan--it's only action when you click the button is to log some data to the service worker's console. Future changes will work on migrating the current background worker to the service worker.

Finally, the MV3 manifest file intentionally minimal. We'll need to vet the settings as we port the current background page over, and this makes it easy to keep track of what still needs to be vetted.

##### Motivation

Feature to allow us to keep running after MV2 support is turned off.

#### Pull request checklist

- [x] Addresses an existing issue: [1906550](https://mseng.visualstudio.com/1ES/_workitems/edit/1906550) and [1906551](https://mseng.visualstudio.com/1ES/_workitems/edit/1906551)
- [x] Ran `yarn fastpass`

### DIFF
--- a/targets.config.js
+++ b/targets.config.js
@@ -39,8 +39,22 @@ module.exports = {
             options: {
                 ...commonExtensionOptions,
                 ...icons.dev,
+                manifestVersion: 2,
                 fullName: 'Accessibility Insights for Web - Dev',
                 telemetryBuildName: 'Dev',
+            },
+        },
+        bundleFolder: 'devBundle',
+        mustExistFile: 'background.bundle.js',
+    },
+    'dev-mv3': {
+        config: {
+            options: {
+                ...commonExtensionOptions,
+                ...icons.dev,
+                manifestVersion: 3,
+                fullName: 'Accessibility Insights for Web - Dev (Manifest 3)',
+                telemetryBuildName: 'DevMV3',
             },
         },
         bundleFolder: 'devBundle',
--- a/targets.config.js
+++ b/targets.config.js
@@ -25,6 +25,7 @@ const commonExtensionOptions = {
         'Accessibility Insights for Web helps developers quickly find and fix accessibility issues.',
     bundled: true,
     productCategory: 'extension',
+    manifestVersion: 2, // Will get overwritten for MV3 configs
 };
 
 const commonUnifiedOptions = {
@@ -39,7 +40,6 @@ module.exports = {
             options: {
                 ...commonExtensionOptions,
                 ...icons.dev,
-                manifestVersion: 2,
                 fullName: 'Accessibility Insights for Web - Dev',
                 telemetryBuildName: 'Dev',
             },
--- a/targets.config.js
+++ b/targets.config.js
@@ -57,8 +57,8 @@ module.exports = {
                 telemetryBuildName: 'DevMV3',
             },
         },
-        bundleFolder: 'devBundle',
-        mustExistFile: 'background.bundle.js',
+        bundleFolder: 'devMv3Bundle',
+        mustExistFile: 'serviceWorker.bundle.js',
     },
     playground: {
         release: true,
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -175,6 +175,25 @@ const devConfig = {
     },
 };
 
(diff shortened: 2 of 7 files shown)


targets.config.js
```diff
@@ -46,6 +47,19 @@ module.exports = {
...
+        config: {
+            options: {
+                ...commonExtensionOptions,
+                ...icons.dev,
+                manifestVersion: 3,
+                fullName: 'Accessibility Insights for Web - Dev (Manifest 3)',
+                telemetryBuildName: 'DevMV3',
+            },
+        },
+        bundleFolder: 'devBundle',
```
dbjorge, line 60:
```suggestion
        bundleFolder: 'devMv3Bundle',
```

targets.config.js
```diff
@@ -46,6 +47,19 @@ module.exports = {
...
+        mustExistFile: 'background.bundle.js',
```
dbjorge, line 61:
```suggestion
        mustExistFile: 'serviceWorker.bundle.js',
```

webpack.config.js
```diff
@@ -174,6 +175,25 @@ const devConfig = {
...
+const devMv3Config = {
+    ...commonConfig,
+    entry: {
+        ...commonEntryFiles,
+        detailsView: [...reactDevtoolsEntryFiles, ...commonEntryFiles.detailsView],
+        popup: [...reactDevtoolsEntryFiles, ...commonEntryFiles.popup],
+    },
+    name: 'dev-mv3',
+    mode: 'development',
+    devtool: 'inline-cheap-source-map',
```
dbjorge, line 187: > I experimented with both inline-cheap-source-map and inline-source-map, which is slightly faster. I opted for the slower option because the difference was only about 5 seconds per build.

I also tested this and my findings matched yours: On my machine, `inline-source-map` was faster than `inline-cheap-source-map` (in 3 trials with 1 warmup iteration each, `yarn build:dev:mv3` took `51±2s` for the former and `57±8s` for the latter on my Surface Laptop 3 running win11).

[Webpack's documentation](https://webpack.js.org/configuration/devtool/) suggests that the only functional difference between them should be that the `cheap` variant will have a slightly less accurate source map (it maps only line numbers, where the non-`cheap` variant maps both line and column numbers). Did you find some other reason to prefer the `cheap` option? It seems like the non-cheap one is both faster and more accurate, so I think it would make more sense to use that.

(I think we found similar timings in the past when we last benchmarked `eval-source-map` vs `eval-cheap-source-map`, and that's why the existing `dev` config uses the former)

dbjorge (changes requested):
This looks great overall, just a few suggestions! I appreciate that you built the new build command into `yarn build:all`.

Besides code review, I verified:

* the build artifact for the PR build includes the new drop
* `yarn build:dev:mv3` worked locally (after making the `bundleFolder` fix in the targets config).
* Adding the new drop as an unpacked extension loaded as expected and I was able to see the interval/counter messages in the console 
* Source mapping worked as expected with devtools (with both `cheap-` and non-cheap settings)

webpack.config.js
DaveTryon: Ironically, I intended to use `inline-source-map`, but then forgot to make the change--I should know better than to write PR's in a brain fog ☹️ 

Fixed to use `inline-source-map`

dbjorge (approved):
This looks great, thanks Dave!

PR #5031, opened 2022-01-06
feat(tabstops-report): Fix logic for when to show export button

brocktaylor7:
#### Details

This adjusts the logic for when the export report button shows. Here are the before and after scenarios for when it shows in each test:
1. Automated Checks
- Before: Shows whenever visualization is enabled.
- After: Shows whenever target page changed view is not visible, which equates to whenever the visualization is enabled with the new navigation flow.
2.  Tab Stops:
- Before: Always shows
- After: Shows whenever target page changed view is not visible.
3. Needs Review: 
- Before: incorrectly shows when test is enabled.
- After: Never shows.

##### Motivation

feature work

##### Context

This passes in `featureFlagStoreData` as a second parameter. The `featureFlagStoreData` parameter will be removed before release, at which point `shouldShowExportReport` in `tab-stops/visualization.tsx` will also need to be updated to exclude it.

I had initially used the existing scanData.isEnabled to make this change but realized that the tab stops test is not enabled by default. Rather than have the tab stops view always display the export button, I realized that the only time we don't want the export to show in Automated Checks and Tab Stops is when the target page changed view is visible and no test has been started over. Due to this, it made sense to pass in the tabStoreData instead, as it captures the desired result in both Automated Checks and Tab Stops.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/should-show-report-export-button.ts
+++ b/src/DetailsView/components/should-show-report-export-button.ts
@@ -23,5 +23,8 @@ export function shouldShowReportExportButtonForFastpass(
     props: ShouldShowReportExportButtonProps,
 ): boolean {
     const config = props.visualizationConfigurationFactory.getConfiguration(props.selectedTest);
-    return config.shouldShowExportReport(props.tabStoreData, props.featureFlagStoreData);
+    const isTargetPageChangedViewVisible = props.tabStoreData.isChanged;
+    return (
+        config.shouldShowExportReport(props.featureFlagStoreData) && !isTargetPageChangedViewVisible
+    );
 }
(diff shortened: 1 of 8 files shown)


src/DetailsView/components/should-show-report-export-button.ts
```diff
@@ -3,31 +3,25 @@
...
 export function shouldShowReportExportButtonForFastpass(
     props: ShouldShowReportExportButtonProps,
 ): boolean {
     const config = props.visualizationConfigurationFactory.getConfiguration(props.selectedTest);
-    const shouldShow = config.shouldShowExportReport(props.featureFlagStoreData);
-
-    const scanData = config.getStoreData(props.visualizationStoreData.tests);
-    const isEnabled = config.getTestStatus(scanData);
-    return shouldShow || isEnabled;
+    return config.shouldShowExportReport(props.tabStoreData, props.featureFlagStoreData);
```
dbjorge, line 26: Can we have the `!tabStoreData.isChanged` check happen here instead of repeating it in each visualization config? The "does this config support exporting to reports?" part feels reasonable for the individual config to understand, but I don't think it feels right for each individual visualization to need to understand of how the target page URL state ties into the report export UX flow.

brocktaylor7: This is a good call, I've updated to have the check happen here.

dbjorge (approved):

PR #5033, opened 2022-01-07
feature(tabstops-report): fix target page changed reset issue

brocktaylor7:
#### Details

This handles clearing automated checks and needs review data when the target page changed view happens on tab stops. There was logic implemented to handle this in the `AdhocIssuesTestView` that Automated Checks and Needs review uses, but it wasn't added to the `AdhocTabStopsTestView`. This caused the reset to not happen properly on target page changed when tab stops is the currently selected test.

Before:
tab-stops-reset-before

After:
tab-stops-reset-after

##### Motivation

feature work

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/unit/tests/DetailsView/components/adhoc-tab-stops-test-view.test.tsx
```diff
@@ -98,8 +102,18 @@ describe('AdhocTabStopsTestView', () => {
...
+            } as AdhocTabStopsTestViewDeps;
+
+            detailsViewActionMessageCreatorMock
+                .setup(m => m.targetPageChangedResetData())
+                .verifiable();
+
             const wrapper = shallow(<AdhocTabStopsTestView {...props} />);
             expect(wrapper.getElement()).toMatchSnapshot();
+
+            detailsViewActionMessageCreatorMock.verifyAll();
```
dbjorge, line 116: Could you put the action message verification in a separate test case from the rendering snapshot test, so we can avoid having to include a snapshot of a mock proxy object? Proxy snapshots tend to require extraneous snapshot updates when typemoq/swc/typescript/node versions update.

src/DetailsView/components/adhoc-tab-stops-test-view.tsx
```diff
@@ -103,6 +103,7 @@ export const AdhocTabStopsTestView = NamedFC<AdhocTabStopsTestViewProps>(
         );
 
         if (props.tabStoreData.isChanged) {
+            props.deps.detailsViewActionMessageCreator.targetPageChangedResetData();
```
dbjorge, line 106: I think this makes sense as a quick fix, but I feel like it might be preferable to have whatever is responding to this message instead hook directly into whatever message sets `tabStoreData.isChanged`; it feels off for the data clearing behavior to rely on whether particular bits of UI are rendered

brocktaylor7: This makes sense, it's tied into `Messages.Tab.ExistingTabUpdated` which is listened to from several stores. I made the change to hook directly into this message instead.

src/tests/unit/tests/DetailsView/components/adhoc-tab-stops-test-view.test.tsx
brocktaylor7: With my latest update this is no longer an issue, but I also don't fully follow how to do this? The mock verification doesn't change the snapshot as far as I know, but the mock did need to exist in the props for the view to render, because `props.deps.detailsViewActionMessageCreator.targetPageChangedResetData()` was being called. How do we get around having a proxy object in the snapshot in a scenario like that?

dbjorge: The two options I use most often are:

* If the snapshot test in question doesn't need to capture a "deep" copy of the props to be acting as a valuable test, you can snapshot `theEnzymeShallowInstance.debug()` instead of `theEnzymeShallowInstance.getElement()`; this produces a "shallower" snapshot that doesn't encode so much implementation detail of the mock proxy
* If the test is more valuable for verifying other "deep" subproperties of props, you can replace the use of `Mock.ofInstance`/etc with a less-smart stub object. Usually this route is much more realistic if you can separate out the tests of the rendering (which can use simple stub objects) vs tests of callback behavior (which are easier to write with full mock objects). For example, here, you might have the rendering test pass a `detailsViewActionMessageCreator` whose value was `{ targetPageChangedResetData: () => {} }`, rather than `Mock.ofType<DetailsViewActionMessageCreator>()`

src/tests/unit/tests/DetailsView/components/adhoc-tab-stops-test-view.test.tsx
```diff
@@ -12,6 +12,7 @@ import {
     VisualizationStoreData,
 } from 'common/types/store-data/visualization-store-data';
 import { VisualizationType } from 'common/types/visualization-type';
+import { DetailsViewActionMessageCreator } from 'DetailsView/actions/details-view-action-message-creator';
```
dbjorge, line 15: Should the changes to this test (+ its snapshot) be reverted now that the changes to the component were reverted?

brocktaylor7: Good catch!

src/tests/unit/tests/DetailsView/components/adhoc-tab-stops-test-view.test.tsx
brocktaylor7: This is great, thanks for taking the time to explain it!

dbjorge (approved):
Looks great, thanks Brock! Love fixes with negative diff size :)

PR #5034, opened 2022-01-07
feat(tabstops-report): enable fast pass report generation without automated checks

dbjorge:
#### Details

The new FastPass report allows a report export to happen when Automated Checks aren't complete (eg, by changing the target page URL while on the Tab Stops test and then exporting a report). However, today there are a few issues that cause the details view to just crash when attempting to do so:

* The report uses the `scanMetadata` from the automated checks scan to determine certain metadata about the report (eg, scan timestamp, axe version), and if those aren't populated by a scan, you get an undefined-dereference.
* The report's `PassedChecksSection` and `IncompleteChecksSection`s assume that their respective `cardsViewData` will be populated, and crash if the data doesn't exist

This PR fixes both issues (and plugs in the now-non-crashing sections into the report for testing):

* For the former, it updates the data flow into the FastPass report to avoid pulling that data from the Automated Checks' `UnifiedResultStore`. This both avoids crashes in the "incomplete automated checks" case and is also more accurate (previously, the "report data" was actually "the last automated checks scan date", which was less accurate than the new strategy of using the date at which the report was generated).
* For the latter, it updates the section components to allow for card data to be non-present and falls back to treating that case the same as the results being present as an empty list

View of a report generated without automated checks data (by opening a FastPass to "Needs Review", changing the target URL, then switching to "Tab Stops" to export a report:

Before: it crashes the details view before it can be generated
After: screenshot of report without any automated checks

##### Motivation

See 1906107

##### Context

This is part 1 of supporting incomplete automated checks; it doesn't complete the user story because it doesn't actually populate the list of incomplete checks. That will come in a separate PR.

#### Pull request checklist

- [x] Addresses an existing issue: 1906107
- [x] Ran `yarn fastpass`


jalkire (approved):
LGTM (though it appears to need a snapshot update)

PR #5036, opened 2022-01-08
feat(tabstops-auto-impl): moving frame-communication code outside of tab-stops-listener

karanbirsingh:
#### Details

This PR refactors our tab stop listener - functionality should be identical to before (It would be helpful if a reviewer can verify tab stops still works for them on this branch. I have tested on our Accessible University site and a local sample site with iframes.)

This PR replaces `tab-stops-listener.ts` with two files:
- `single-frame-tab-stop-listener.ts`: responsible for DOM event listeners, transforming tabbed HTML elements to TabStopEvent objects, and reporting results
- `all-frame-runner.ts`: responsible for running an `AllFrameRunnerTarget` like `single-frame-tab-stop-listener.ts` in all child frames; communicates data through frames so that child instances of `single-frame-tab-stop-listener` can report results to the main window (and thus to `background`)

Before this PR, `tab-stops-listener.ts` covered both frame communication and tab-stop-analysis. This is unwieldy to work with, especially as we look to increase the analysis done on tabbed HTML elements before transforming them to payloads for `background`.

##### Motivation

Make it easier to write tab-stop-listeners without dealing directly with iframe-related code

##### Context

We plan to add a new `AllFrameRunnerTarget` for upcoming "automated" tab stop analysis. We will run requirements defined in #5042 per frame before transforming the tabbed HTML element to some "result" object that we can forward to `background`. We can switch between the new and old frame runners with a feature flag to avoid regressing existing behavior.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/injected/all-frame-runner.ts
```diff
@@ -0,0 +1,151 @@
...
+            this.onResultFromChildFrame,
+        );
+
+        this.listener.setResultCallback(async payload => {
+            await this.reportResults(payload);
+            return null;
+        });
+    }
+
+    public start = async (): Promise<CommandMessageResponse | null> => {
```
karanbirsingh, line 61: start & stop are explicitly listed here and in AllFrameRunnerTarget - this feels OK to me, but I briefly considered an 'operations' type that maps from string IDs ("start" | "stop", for instance) to functions. This would remove the similar code between start/stop and enable arbitrary operations defined by AllFrameRunnerTarget. It felt premature so I left it out.

src/injected/all-frame-runner.ts
```diff
@@ -0,0 +1,151 @@
...
+} from 'injected/frameCommunicators/respondable-command-message-communicator';
+
+export interface AllFrameRunnerTarget<T> {
+    start: () => void;
+    teardown: () => void;
+    transformChildResultForParent: (fromChild: T, messageSourceFrame: HTMLIFrameElement) => T;
+    setResultCallback: (reportResults: (payload: T) => Promise<null>) => void;
+}
+
+/*
```
karanbirsingh, line 19: note for reviewer: most of the iframe code is moved over from tab-stops-listener

src/injected/all-frame-runner.ts
```diff
@@ -14,3 +16,124 @@ export interface AllFrameRunnerTarget<T> {
...
+    private sendCommandToFrames = async (
+        command: string,
+    ): Promise<CommandMessageResponse | null> => {
+        const iframes = this.getAllFrames();
+        for (let i = 0; i < iframes.length; i++) {
+            await this.frameMessenger.sendMessageToFrame(iframes[i], {
+                command,
+            });
+        }
+        return { payload: null };
```
sfoslund, line 93: nit: why bother returning anything if we're always returning null?

sfoslund (approved):

src/injected/all-frame-runner.ts
```diff
@@ -1,11 +1,13 @@
...
+import { HTMLElementUtils } from 'common/html-element-utils';
+import { WindowUtils } from 'common/window-utils';
+import { FrameMessenger } from 'injected/frameCommunicators/frame-messenger';
+import {
+    CommandMessage,
+    CommandMessageResponse,
+} from 'injected/frameCommunicators/respondable-command-message-communicator';
 
 export interface AllFrameRunnerTarget<T> {
     name: string;
```
karanbirsingh, line 13: Chatted with Sarah, will rename this in a follow-up PR to be more clear

src/tests/unit/tests/injected/all-frame-runner.test.ts
```diff
@@ -0,0 +1,314 @@
...
+        };
+        test('calls topWindowCallback if in main window', async () => {
+            const topWindowCallbackMock = Mock.ofInstance((_: UnitTestTargetType) => {});
+
+            const { commandFunc } = captureOnResultFromChildFrameCallback(
+                unitTestListenerMock,
+                topWindowCallbackMock.object,
+            );
+
+            htmlElementUtilsMock
```
waabid, line 174: Might not be a bad idea to refactor some of these setups to functions to reuse in multiple tests.

waabid (approved):
LGTM. Just a minor comment.

karanbirsingh:
Merged to unblock iterating on this in main; making follow-up PR to address:
- renaming "name" field
- return type of `{payload : null}`
- seeing if it makes sense to move more of the tests into helper functions

PR #5037, opened 2022-01-08
feat(tabstops-report): add results data to outcome summary bar

shanisebarona:
#### Details

This PR adds the results data to the `OutcomeSummaryBar` in `FastPassReportSummary`.

##### Motivation

Feature 1872889.

##### Context

Incomplete automated checks data will be added after the completion of 1906107.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/fast-pass-report-summary.tsx
+++ b/src/reports/components/fast-pass-report-summary.tsx
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { CardRuleResult } from 'common/types/store-data/card-view-model';
 import { requirements } from 'DetailsView/components/tab-stops/requirements';
 import { TabStopsFailedCounter } from 'DetailsView/tab-stops-failed-counter';
 import * as React from 'react';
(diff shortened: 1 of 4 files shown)


src/reports/components/fast-pass-report-summary.tsx
```diff
@@ -1,24 +1,86 @@
...
 
 export class FastPassReportSummary extends React.Component<FastPassReportSummaryProps> {
     public render(): JSX.Element {
+        const tabResults = [];
+
+        for (const [requirementId, data] of Object.entries(this.props.results.tabStops)) {
+            if (data.status !== 'fail') {
+                continue;
+            }
+
```
brocktaylor7, line 30: I think it might be more straightforward to just do this loop once. Inside the loop you can check the `data.status` and push to the correct array of results, rather than looping 3 times and checking for the correct status in each one.

src/reports/components/fast-pass-report-summary.tsx
```diff
@@ -1,24 +1,86 @@
...
+                description: requirements[requirementId].description,
+                instances: data.instances,
+                isExpanded: data.isExpanded,
+            });
+        }
+        const totalIncompleteTabCount: number = incompleteTabResults.length;
+
+        const failedAutomatedChecks = this.props.results.automatedChecks.cards.fail.length;
+        const passedAutomatedChecks = this.props.results.automatedChecks.cards.pass.length;
+        const incompleteAutomatedChecks = this.props.results.automatedChecks.cards.unknown.length;
```
brocktaylor7, line 78: These probably need to use a failed counter similar to the one used for tab stops. Currently the `fail`, `pass`, and `unknown` properties of `cards` returns an array for each rather than the number of actual failed instances. If there isn't a similar counter in the codebase we'd probably just need to make sure we're summing up all the failed instances of each rule.

src/tests/unit/tests/reports/components/__snapshots__/fast-pass-report-summary.test.tsx.snap
```diff
@@ -15,9 +15,9 @@ exports[`FastPassReportSummary renders per the snapshot 1`] = `
     }
     outcomeStats={
       Object {
-        "fail": 1,
+        "fail": NaN,
```
brocktaylor7, line 18: Is this supposed to be `NaN`? It seems like we'd either want a number >= 0 always for this, right?

brocktaylor7 (approved):

PR #5053, opened 2022-01-13
feat(tabstops-auto-impl): Create tab stops orchestrator

waabid:
#### Details

Creates tab stops orchestrator which should be used in all frames via all frame runner in a future PR.

##### Motivation

feature work.

##### Context

This is **definitely** making an assumption that when the orchestrator is stopped is equivalent to when the user is done testing. If that assumption ends up being not true, we'll have to expose a different API where we get the keyboard navigation results.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/injected/analyzers/tab-stops-orchestrator.ts
+++ b/src/injected/analyzers/tab-stops-orchestrator.ts
@@ -1,0 +1,104 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { WindowUtils } from 'common/window-utils';
+import { AllFrameRunnerTarget } from 'injected/all-frame-runner';
+import {
+    TabStopRequirementResult,
+    TabStopsRequirementEvaluator,
+} from 'injected/tab-stops-requirement-evaluator';
+import { TabbableElementGetter } from 'injected/tabbable-element-getter';
+import { FocusableElement } from 'tabbable';
+
+export class TabStopRequirementOrchestrator
+    implements AllFrameRunnerTarget<TabStopRequirementResult>
+{
+    public readonly name: string = 'TabStopRequirementOrchestrator';
+    private tabbableTabStops: FocusableElement[];
+    private actualTabStops: Set<HTMLElement> = new Set();
+    private latestVisitedTabStop: HTMLElement = null;
+    private reportResults: (payload: TabStopRequirementResult) => Promise<void>;
+
+    constructor(
+        private readonly dom: Document,
+        private readonly tabbableElementGetter: TabbableElementGetter,
+        private readonly windowUtils: WindowUtils,
+        private readonly tabStopsRequirementEvaluator: TabStopsRequirementEvaluator,
+        private readonly getUniqueSelector: (element: HTMLElement) => string,
+    ) {}
+
+    public start = () => {
+        this.dom.addEventListener('keydown', this.onKeydownForFocusTraps);
+        this.dom.addEventListener('focusin', this.addNewTabStop);
+        this.tabbableTabStops = this.tabbableElementGetter.getRawElements();
+        const tabbableFocusOrderResults =
+            this.tabStopsRequirementEvaluator.getTabbableFocusOrderResults(this.tabbableTabStops);
+        tabbableFocusOrderResults.forEach(result => {
+            this.reportResults(result);
+        });
+    };
+
+    public stop = () => {
+        this.dom.removeEventListener('keydown', this.onKeydownForFocusTraps);
+        this.dom.removeEventListener('focusin', this.addNewTabStop);
+        const keyboardNavigationResults =
+            this.tabStopsRequirementEvaluator.getKeyboardNavigationResults(
+                this.tabbableTabStops,
+                this.actualTabStops,
+            );
+        keyboardNavigationResults.forEach(this.reportResults);
+    };
+
+    public setResultCallback = (
+        reportResultCallback: (payload: TabStopRequirementResult) => Promise<void>,
+    ) => {
+        this.reportResults = reportResultCallback;
+    };
+
+    public transformChildResultForParent = (
+        result: TabStopRequirementResult,
+        messageSourceFrame: HTMLIFrameElement,
+    ): TabStopRequirementResult => {
+        const frameSelector = this.getUniqueSelector(messageSourceFrame);
+        result.selector = [frameSelector, ...result.selector];
+        return result;
+    };
+
+    private addNewTabStop = (focusEvent: FocusEvent) => {
+        const newTabStop = focusEvent.target as HTMLElement;
+        if (this.actualTabStops.size > 0 && this.actualTabStops.has(newTabStop)) {
+            return;
+        }
+
+        this.actualTabStops.add(newTabStop);
+        this.latestVisitedTabStop = newTabStop;
+
+        const result = this.tabStopsRequirementEvaluator.getFocusOrderResult(
+            this.latestVisitedTabStop,
+            newTabStop,
+        );
+
+        if (result == null) {
+            return;
+        }
+        this.reportResults(result);
+    };
+
+    private onKeydownForFocusTraps = (e: KeyboardEvent) => {
+        if (e.key !== 'Tab') {
+            return;
+        }
+
+        const oldElement = this.dom.activeElement;
+        this.windowUtils.setTimeout(() => {
+            const result = this.tabStopsRequirementEvaluator.getKeyboardTrapResults(
+                this.dom.activeElement,
+                oldElement,
+            );
+            if (result == null) {
+                return;
+            }
+            this.reportResults(result);
+        }, 500);
+    };
+}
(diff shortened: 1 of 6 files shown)


src/tests/unit/tests/injected/tabbable-element-getter.test.ts
```diff
@@ -64,4 +64,23 @@ describe('TabbableElementGetter', () => {
 
         expect(testSubject.get()).toEqual(expectedTabbleElementInfo);
     });
+
+    test('getRawElements', () => {
+        docMock.setup(m => m.documentElement).returns(() => documentElementStub);
+
+        const focusableElementsStub: FocusableElement[] = [
```
sfoslund, line 71: nit: can we factor this out into a class constant so it doesn't have to be declared multiple times in the multiple tests?

src/tests/unit/tests/injected/tabbable-element-getter.test.ts
```diff
@@ -64,4 +64,23 @@ describe('TabbableElementGetter', () => {
...
+            {
+                outerHTML: 'some outer html',
+            },
+            {
+                outerHTML: 'some other outer html',
+            },
+        ] as FocusableElement[];
+
+        getTabbableElementsMock
```
sfoslund, line 80: nit: this can probably be done in `beforeEach` and removed from both tests

src/injected/analyzers/tab-stops-orchestrator.ts
```diff
@@ -0,0 +1,104 @@
...
+    TabStopRequirementResult,
+    TabStopsRequirementEvaluator,
+} from 'injected/tab-stops-requirement-evaluator';
+import { TabbableElementGetter } from 'injected/tabbable-element-getter';
+import { FocusableElement } from 'tabbable';
+
+export class TabStopRequirementOrchestrator
+    implements AllFrameRunnerTarget<TabStopRequirementResult>
+{
+    public readonly name: string = 'TabStopRequirementOrchestrator';
```
sfoslund, line 16: What will this be used for?

src/injected/analyzers/tab-stops-orchestrator.ts
```diff
@@ -0,0 +1,104 @@
...
+    ) {}
+
+    public start = () => {
+        this.dom.addEventListener('keydown', this.onKeydownForFocusTraps);
+        this.dom.addEventListener('focusin', this.addNewTabStop);
+        this.tabbableTabStops = this.tabbableElementGetter.getRawElements();
+        const tabbableFocusOrderResults =
+            this.tabStopsRequirementEvaluator.getTabbableFocusOrderResults(this.tabbableTabStops);
+        tabbableFocusOrderResults.forEach(result => {
+            this.reportResults(result);
```
sfoslund, line 37: I think it's probably possible that we'll have duplicated results between `getTabbableFocusOrderResults` and `getFocusOrderResult`. Are we okay with reporting those duplicates or do we need to add logic to filter those out?

src/injected/analyzers/tab-stops-orchestrator.ts
```diff
@@ -0,0 +1,104 @@
...
+import { FocusableElement } from 'tabbable';
+
+export class TabStopRequirementOrchestrator
+    implements AllFrameRunnerTarget<TabStopRequirementResult>
+{
+    public readonly name: string = 'TabStopRequirementOrchestrator';
+    private tabbableTabStops: FocusableElement[];
+    private actualTabStops: Set<HTMLElement> = new Set();
+    private latestVisitedTabStop: HTMLElement = null;
+    private reportResults: (payload: TabStopRequirementResult) => Promise<void>;
```
sfoslund, line 20: It looks like we're not using `TabStopRequirementResults`, which contains the test id. Do we want to remove it and just add the id to TabStopRequirementResult?

src/injected/analyzers/tab-stops-orchestrator.ts
```diff
@@ -0,0 +1,104 @@
...
+        if (this.actualTabStops.size > 0 && this.actualTabStops.has(newTabStop)) {
+            return;
+        }
+
+        this.actualTabStops.add(newTabStop);
+        this.latestVisitedTabStop = newTabStop;
+
+        const result = this.tabStopsRequirementEvaluator.getFocusOrderResult(
+            this.latestVisitedTabStop,
+            newTabStop,
```
sfoslund, line 78: Since you just assigned ` this.latestVisitedTabStop` on line 74, this calls `getFocusOrderResult` on the same tab stop for both curr and last, right? Do we need to move this call up a few lines before that assignment?

sfoslund (commented):
Looks good overall, a few questions and nits :)

src/injected/analyzers/tab-stops-orchestrator.ts
karanbirsingh: my fault - usage of 'name' is in all-frame-runner from https://github.com/microsoft/accessibility-insights-web/pull/5036. We need the target to supply a suffix to frame-messenger commands since command strings should be globally unique. Discussed offline & will update this field in a follow-up PR.

src/injected/analyzers/tab-stops-orchestrator.ts
```diff
@@ -0,0 +1,104 @@
...
+    ): TabStopRequirementResult => {
+        const frameSelector = this.getUniqueSelector(messageSourceFrame);
+        result.selector = [frameSelector, ...result.selector];
+        return result;
+    };
+
+    private addNewTabStop = (focusEvent: FocusEvent) => {
+        const newTabStop = focusEvent.target as HTMLElement;
+        if (this.actualTabStops.size > 0 && this.actualTabStops.has(newTabStop)) {
+            return;
```
karanbirsingh, line 70: Do you envision the "done" signal happening here or elsewhere? No worries if it's in a future PR, just want to make sure we're tracking it

src/tests/unit/tests/injected/tabbable-element-getter.test.ts
waabid: I think I'm going to remove the 'get' API completely actually, since it's not being used.

waabid: I guess that will have to wait for Karan's PR since he's removing the usage of it.

src/injected/analyzers/tab-stops-orchestrator.ts
waabid: Discussed offline.

src/injected/analyzers/tab-stops-orchestrator.ts
waabid: Thought about that as well and figured it might make more sense to do later, in the process that is.

src/injected/analyzers/tab-stops-orchestrator.ts
waabid: I've been going back and forth on how exactly we should propagate the results back ... I went with this because it's a bit simpler as is, doesn't require the array of results, and we do not intend to send objects back, at least in the orchestrator, multiples at a time.

Okay, after writing this out, I'm more convinced to include id on the result object itself; we can determine what to do with these result objects at a higher level than the orchestrator (maybe analyzer or some new layer between that).

src/injected/analyzers/tab-stops-orchestrator.ts
waabid: Good catch. I was wondering myself but didn't go back to look into the notes myself..

src/injected/analyzers/tab-stops-orchestrator.ts
```diff
@@ -0,0 +1,111 @@
...
+        const newTabStop = focusEvent.target as HTMLElement;
+
+        if (this.latestVisitedTabStop == null) {
+            this.actualTabStops.add(newTabStop);
+            this.latestVisitedTabStop = newTabStop;
+            return;
+        }
+
+        if (this.actualTabStops.has(newTabStop)) {
+            this.latestVisitedTabStop = newTabStop;
```
sfoslund, line 77: Why don't we want to rungetFocusOrderResult when this tab element has been seen before? Are we just assuming that it will have been checked previously? Is it possible to get to the same newTabStop from two different lastVisitedTabStops?

sfoslund (approved):
Looks great! One more question but not blocking.

src/injected/analyzers/tab-stops-orchestrator.ts
waabid: Good question. My understanding/thinking here is that there's a few reasons for this to happen but none would require us to get a 'focus-order' result. Effectively, when this happens we're seeing an element that we've seen before which could be three scenarios:

- we're stuck on this specific tab-stop => keyboard-trap; detected elsewhere.
- we're stuck in a loop including this tab-stop => keyboard-trap or keyboard-navigation issue, detected elsewhere.
- we've completed the full tab stop loop => we are done tabbing; detected elsewhere ... possibly.

Now, whether all these are behaving the way we expect is a different story but I think that's the intent.

PR #5059, opened 2022-01-13
feat(tabstops-report): fix duplicate-id-aria accessibility issue

brocktaylor7:
#### Details

This fixes an accessibility issue in the fastpass report caused by the duplication of ReportCollapsibleComponent controls having the same id. It parameterizes a test key to differentiate the ids between the duplicated components.

This also reverts the E2E test that was set to be a snapshot of the accessibility issue to avoid blocking the feature. This restores functionality of checking that there are no accessibility issues on the report page.

##### Motivation

feature work

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/reports/components/fast-pass-report.tsx
```diff
@@ -40,48 +40,54 @@ export type FastPassReportProps = Omit<
...
+                            }}
+                            tabStopRequirementState={props.results.tabStops}
+                            alwaysRenderSection={true}
+                        />
+                        <TabStopsChecksSectionWrapper
+                            key={6}
+                            checksSection={IncompleteChecksSection}
+                            tabStops={props.results.tabStops}
+                            {...props}
+                            testKey={tabStopsTestKey}
```
jalkire, line 76: One thought: since `testKey` is only set to tab-stops here, it seems reasonable that `TabStopsChecksSectionWrapper` could take responsibility for setting it and `FastPassReport` could remain unchanged. Just an idea; totally optional!

jalkire (approved):
Approve with one idea

src/reports/components/fast-pass-report.tsx
brocktaylor7: This is a good call, I made the change! :)

PR #5070, opened 2022-01-15
chore(tabstops-auto-impl): small refactors to all-frame-runner

karanbirsingh:
#### Details

This PR applies some changes based on comments in #5036 - we decided to merge that in early and address follow-up comments in a separate PR:
- change `AllFrameRunnerTarget.name` to `AllFrameRunnerTarget.commandSuffix` to better illustrate its usage
- change `AllFrameRunner.start()` and `AllFrameRunner.stop()`  to `void` functions. While the frame-messenger has support for responses from child frames, we weren't making use of it for start/stop. The frame-messenger listener expects an explicit return type of `CommandMessageResponse | null`, but this didn't make sense for clients of `AllFrameRunner` who need to just fire-and-forget start and stop. When registering commands to the frame messenger, we wrap the public-facing void methods and explicitly return `null`. If we need response data later (maybe we want to "collect" data on stop), we can, but so far we've been using the `reportResults` callback to accomplish the same thing.
- refactor all-frame-runner.test.ts to improve readability. I don't like that `start` and `stop` share so much test code, but I'm wary of creating more helper methods that obfuscate the unit tests. Happy to adjust in either direction based on feedback.

##### Motivation

respond to feedback from original PR

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/injected/all-frame-runner.ts
+++ b/src/injected/all-frame-runner.ts
@@ -67,12 +67,12 @@ export class AllFrameRunner<T> {
 
     public start = async () => {
         this.listener.start();
-        this.sendCommandToFrames(this.startCommand);
+        await this.sendCommandToFrames(this.startCommand);
     };
 
     public stop = async () => {
         this.listener.stop();
-        this.sendCommandToFrames(this.stopCommand);
+        await this.sendCommandToFrames(this.stopCommand);
     };
 
     private reportResultsThroughFrames = async (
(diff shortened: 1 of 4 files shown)


src/injected/all-frame-runner.ts
```diff
@@ -58,14 +65,14 @@ export class AllFrameRunner<T> {
         });
     }
 
-    public start = async (): Promise<CommandMessageResponse | null> => {
+    public start = async () => {
         this.listener.start();
-        return this.sendCommandToFrames(this.startCommand);
+        this.sendCommandToFrames(this.startCommand);
```
sfoslund, line 70: Does this need to be awaited? (Same with stop)

sfoslund (approved):

PR #5087, opened 2022-01-19
feat(tabstops-auto-impl): Trigger tabbing completed based on details view focus

sfoslund:
#### Details

Adds `tabbingCompleted` field to store data to keep track of if the tab stop analyzer is listening for user tabs or not. Tabbing is marked as complete when focus shifts from the target page to the details page. 

##### Motivation

Feature work

##### Context

A future PR will add another criteria for tabbing 'done-ness'- the tabbingCompleted field will also be set to true if we reach the same element multiple times in our tabbing. 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/adhoc-tab-stops-test-view.tsx
+++ b/src/DetailsView/components/adhoc-tab-stops-test-view.tsx
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import { CollapsibleComponent } from 'common/components/collapsible-component';
+import { FocusComponent } from 'common/components/focus-component';
 import { VisualizationConfiguration } from 'common/configs/visualization-configuration';
 import { NamedFC } from 'common/react/named-fc';
 import { FeatureFlagStoreData } from 'common/types/store-data/feature-flag-store-data';
@@ -9,6 +10,7 @@ import { TabStoreData } from 'common/types/store-data/tab-store-data';
 import { VisualizationScanResultData } from 'common/types/store-data/visualization-scan-result-data';
 import { VisualizationStoreData } from 'common/types/store-data/visualization-store-data';
 import { VisualizationType } from 'common/types/visualization-type';
+import { WindowUtils } from 'common/window-utils';
 import { DetailsViewActionMessageCreator } from 'DetailsView/actions/details-view-action-message-creator';
 import * as styles from 'DetailsView/components/adhoc-tab-stops-test-view.scss';
 import * as requirementInstructionStyles from 'DetailsView/components/requirement-instructions.scss';
@@ -36,6 +38,7 @@ import * as Markup from '../../assessments/markup';
 
 export type AdhocTabStopsTestViewDeps = {
     detailsViewActionMessageCreator: DetailsViewActionMessageCreator;
+    windowUtils: WindowUtils;
 } & TabStopsRequirementsTableDeps &
     TabStopsFailedInstancePanelDeps &
     TabStopsFailedInstanceSectionDeps &
@@ -150,6 +153,14 @@ export const AdhocTabStopsTestView = NamedFC<AdhocTabStopsTestViewProps>(
                         deps={props.deps}
                         failureInstanceState={props.tabStopsViewStoreData.failureInstanceState}
                         requirementState={requirementState}
+                    />
+                    <FocusComponent
+                        windowUtils={props.deps.windowUtils}
+                        configuration={props.configuration}
+                        visualizationStoreData={props.visualizationStoreData}
+                        tabStopRequirementActionMessageCreator={
+                            props.deps.tabStopRequirementActionMessageCreator
+                        }
                     />
                 </div>
             </div>
--- a/src/DetailsView/components/adhoc-tab-stops-test-view.tsx
+++ b/src/DetailsView/components/adhoc-tab-stops-test-view.tsx
@@ -2,15 +2,16 @@
 // Licensed under the MIT License.
 
 import { CollapsibleComponent } from 'common/components/collapsible-component';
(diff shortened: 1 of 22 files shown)


src/DetailsView/components/adhoc-tab-stops-test-view.tsx
```diff
@@ -151,6 +154,14 @@ export const AdhocTabStopsTestView = NamedFC<AdhocTabStopsTestViewProps>(
                         failureInstanceState={props.tabStopsViewStoreData.failureInstanceState}
                         requirementState={requirementState}
                     />
+                    <FocusComponent
+                        windowUtils={props.deps.windowUtils}
```
karanbirsingh, line 158: It seems like other components in this file define their own `deps` to easily distinguish between dependencies leveraged in this component and dependencies required by child components. I wonder if we should do the same so that `AdhocTabStopsTestViewDeps` adds an additional `& FocusComponentDeps` constraint?

src/common/components/focus-component.tsx
```diff
@@ -0,0 +1,49 @@
...
+        this.props.windowUtils.removeEventListener(
+            this.props.windowUtils.getWindow(),
+            FocusComponent.focusEventName,
+            this.handleFocusEvent,
+            false,
+        );
+    }
+
+    private handleFocusEvent = () => {
+        const tabbing = this.props.configuration.getStoreData(
```
karanbirsingh, line 38: In `AdhocTabStopsTestView` we call `getStoreData` on line 99 before instantiating these components; is it possible to reuse the `scanData` variable from there?

https://github.com/microsoft/accessibility-insights-web/blob/18832555826b0666423c956529749ba6b05cd40b/src/DetailsView/components/adhoc-tab-stops-test-view.tsx#L99 

I'm curious about the `checked` value passed to `Toggle` on line 129 and whether this is similar:

https://github.com/microsoft/accessibility-insights-web/blob/18832555826b0666423c956529749ba6b05cd40b/src/DetailsView/components/adhoc-tab-stops-test-view.tsx#L129

src/common/components/focus-component.tsx
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { VisualizationConfiguration } from 'common/configs/visualization-configuration';
+import { VisualizationStoreData } from 'common/types/store-data/visualization-store-data';
+import { WindowUtils } from 'common/window-utils';
+import { TabStopRequirementActionMessageCreator } from 'DetailsView/actions/tab-stop-requirement-action-message-creator';
+import * as React from 'react';
+
+export interface FocusComponentProps {
```
karanbirsingh, line 9: Some of these props feel more like `deps`; here are some examples of [windowUtils](https://github.com/microsoft/accessibility-insights-web/search?l=TSX&q=windowutils) and [tabStopRequirementActionMessageCreator](https://github.com/microsoft/accessibility-insights-web/search?l=TSX&q=tabStopRequirementActionMessageCreator) usage. Functionally `deps` still are passed through the `props` chain, but we often separate them in our codebase. Usually `deps` includes methods/classes whereas other `props` include more statelike data. This allows us to instantiate `deps` once at the root level:

https://github.com/microsoft/accessibility-insights-web/blob/18832555826b0666423c956529749ba6b05cd40b/src/DetailsView/details-view-initializer.ts#L488

and only reference them in components that use them directly (example in `Toast.tsx`):

https://github.com/microsoft/accessibility-insights-web/blob/18832555826b0666423c956529749ba6b05cd40b/src/common/components/toast.tsx#L9

Then components in between who are just passing those `deps` along can reference child component `dep` types to construct their own (example in `copy-failure-issues-details-button`):

https://github.com/microsoft/accessibility-insights-web/blob/0cf0164cc07552761ca5010f2071e3cc891a1527/src/common/components/copy-issue-details-button.tsx#L13

Following the same pattern here would allow you to remove the explicit reference to `windowUtils` in `AdhocTabStopsTestView`.

sfoslund:
Just realized that I didn't use the feature flag here. Is it enough to include the `FocusComponent` in the page conditionally based on the flag, but leave the rest as is, since none of the messages/ logic will be triggered without the listeners in the focus component?

karanbirsingh:
> Just realized that I didn't use the feature flag here. Is it enough to include the `FocusComponent` in the page conditionally based on the flag, but leave the rest as is, since none of the messages/ logic will be triggered without the listeners in the focus component?

I think this seems reasonable

waabid:
> Just realized that I didn't use the feature flag here. Is it enough to include the `FocusComponent` in the page conditionally based on the flag, but leave the rest as is, since none of the messages/ logic will be triggered without the listeners in the focus component?

You should be able to use `FlaggedComponent` to make that a bit easier as well (avoids having to add an extra test case). You may want to look at other places we use the component for reference.

src/common/components/focus-component.tsx
sfoslund: Would reusing that value work? I'm requerying for the state in the focus handler because I figured that the handler will be called separately/ more often than the code for rendering this component, is that incorrect?

karanbirsingh: I agree the handler will be called separately/more often; I think that is OK. It feels like reusing the value would work because `getStoreData` is a pure function and [relies fully](https://github.com/microsoft/accessibility-insights-web/blob/e3995ceaf8f0161ebcc88c63f56775d09834eac3/src/ad-hoc-visualizations/tab-stops/visualization.tsx#L33) on `props.visualizationStoreData.tests` - and if `props.visualizationStoreData.tests` changes, I'm imagining AdhocTabStopsTestView will re-compute line 99 with its new value.

Issue #5099, opened 2022-01-22
Web FastPass Tab stops view and report fit and finish inconsistencies


dbjorge:
## Describe the bug

This issue tracks a collection of small fit and finish issues discovered during release validation for `feat(tabstops-report)` in Insider 2.30.0.

#### Summary bar

* [x] (#5109 The Summary bar is intended to be contained inside a white "card" rectangle
* [x] (#5109) The "x" and "check" icons are misaligned; they should be vertically aligned with the numbers next to them
* Note that the icons' colors are mismatched from the Figma (the check and x should be colored icons in a white circle, not the current white icons in a colored circle). However, they are mismatched the same way in the Assessment report; for consistency, they can stay as-is.

#### Heading levels
 
* [x] (#5121) The heading level for "Failed Instances"/"Passed checks"/"Incomplete checks" is h2, should be h3
* [x] (#5121) The heading level for failed instance cards is h3, should be h4
* Note: Be sure not to accidentally break heading levels in other reports, may need to parameterize levels
* Note: Heading weights and sizes can stay as-is, despite mismatching Figma slightly, since the current versions meet the Heading Level assessment requirement and the Figma version does not

#### Spacing between sections
 
* [x] (#5119) Spacing between comments in tab stops failure instance should be bigger; it should match the details view UI
* [x] (#5119) Spacing between "failed instances", "incomplete checks", "passed checks" sections should be smaller (see figma)
* [x] (#5119) Spacing between "Automated checks" and "Tab stops" sections should be bigger, see Figma (be careful to keep the "Automated checks haven't been run" case consistent)

### Details view FastPass Tab stops UI

* [x] (#5107) Changing a pass/fail option control from "unselected" to "selected" currently causes the containing row to change height slightly, such that all components underneath it shifts downward slightly. This shouldn't happen.
* [x] (#5111) Rows for individual comments should have an on-hover gray background effect similar to the rows in instance tables in assisted Assessment requirements
* [x] (#5110) The report and the details view should be consistent about whether there is a colon after the word "Comment"
* [x] (#5115) At window widths from 600px to 704px, a spurious horizontal scroll bar appears for the `DetailsList` Requirements table. This is not technically a WCAG violation since this qualifies as a data table, but it is still an unnecessary usability issue.
* [x] (#5115) At window widths from 600px to 678px, the above issue also occurs with the `DetailsList` for failed instances' comments.
[...]

PR #5111, opened 2022-01-25
fix(tabstops-report): enable hover/focus highlighting on tab stops requirement rows

dbjorge:
#### Details

This PR enables the Details View > Fast Pass > Tab Stops' new Requirements table to use Office Fabric's default on-hover/on-focus styling that darkens the row backgrounds, to be consistent with the similar behavior in the similar looking Assisted Instance tables in Assessment.

The implementation for this is a little more complicated than expected. The on-hover/on-focus behavior is actually the default fabric behavior; Assessment doesn't do anything special to enable it. The reason we weren't seeing it in FastPass was that we were previously using a very hacky bit of CSS to override the background of the Fabric `DetailsView` to match the slightly-darker-than-usual-gray background we use in FastPass pages (it's different from Assessment to support Cards-based views that want to use our "normal" white background color for Card backgrounds):

```scss
.requirement-table {
    // ...
    * {
        background-color: $neutral-2;
    }
}
```

This was used because the fabric `DetailsView` contains many implementation-detail subcomponents (for rows, cells, headers, etc) that each individually apply overlapping background-color styles; if you just apply background-color to our containing `requirement-table` class, it doesn't have any effect on its own. However, this also has the effect of overriding all of the `:hover` and `:focus` styling under the table.

I originally tried using various combinations of `:not(:focus):not(:hover)` styles, but it rapidly devolved into a big mess of very hacky CSS. Instead, this PR implements a solution based on Office Fabric Themes, which are Fabric's intended way to solve this problem but not the way we normally use in our codebase (we usually use hacky overrides of CSS styles that are Fabric implementation details instead).

The typical Fabric-recommended solution for this is to use their `<Customizer>` to temporarily override the globally-loaded theme; however, this has the disadvantage that it isn't cognizant of the way we swap between "default" and "high-contrast" theme palettes at runtime in our top-level `<Theme>` component. This PR implements a new `<ThemeFamilyCustomizer>` which wraps the fabric `<Customizer>` in a way which respects our app's high contrast mode setting.

##### Animated screenshots

Before:

animation showing no background color changes on hover/focus

After (default theme):

animation showing background color changing on hover/focus

After (app high contrast setting):

animation showing background color changing on hover/focus

After (Windows "Night Sky" HC theme):

animation showing background color changing on hover/focus

##### Motivation

* Make table hover/focus behavior consistent between Assessment and FastPass
* Set up infrastructure that can be used in other components to avoid some types of CSS hacks

##### Context

n/a

#### Pull request checklist

- [x] Addresses an existing issue: Part of #5099
- [x]  Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/adhoc-tab-stops-test-view.tsx
+++ b/src/DetailsView/components/adhoc-tab-stops-test-view.tsx
@@ -4,6 +4,7 @@
 import { CollapsibleComponent } from 'common/components/collapsible-component';
 import { FlaggedComponent } from 'common/components/flagged-component';
 import { FocusComponent, FocusComponentDeps } from 'common/components/focus-component';
+import { ThemeFamilyCustomizer } from 'common/components/theme-family-customizer';
 import { VisualizationConfiguration } from 'common/configs/visualization-configuration';
 import { FeatureFlags } from 'common/feature-flags';
 import { NamedFC } from 'common/react/named-fc';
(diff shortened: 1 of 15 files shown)


src/common/styles/fast-pass-theme-palette.ts
```diff
@@ -0,0 +1,22 @@
...
+const neutral6 = '#f2f2f2';
+
+// This is a copy of our DefaultThemePalette slightly adjusted to account for the page background
+// of FastPass-like content views using our "neutral2" instead of our "white" for a background color
+// (because they use Card-like elements that want to keep "white" background colors for themselves).
+export const FastPassThemePalette: IPartialTheme = {
+    ...DefaultThemePalette,
+    palette: {
+        ...DefaultThemePalette.palette,
+        white: neutral2,
```
dbjorge, line 18: `white` is chosen to match the FastPass background pages. `neutralLighterAlt` and `neutralLighter` are the colors from our default Palette which are slightly darker whites, but which would have ended up lighter than `white` if not updated; I picked values from our existing `neutral-N` color definitions that blended in the correct direction.

src/common/components/theme-family-customizer.tsx
```diff
@@ -0,0 +1,40 @@
...
+import * as React from 'react';
+
+export type ThemeFamily = 'default' | 'fast-pass';
+
+// createTheme is somewhat expensive; don't call it in render()
+const themeFamilyDefaultThemes: { [themeFamily in ThemeFamily]: ITheme } = {
+    default: createTheme(DefaultThemePalette),
+    'fast-pass': createTheme(FastPassThemePalette),
+};
+const highConstrastTheme = createTheme(HighContrastThemePalette);
```
karanbirsingh, line 18: small typo, `highContrastTheme`

src/common/components/theme-family-customizer.tsx
```diff
@@ -0,0 +1,40 @@
...
+// enabled our High Contrast Mode setting.
+export const ThemeFamilyCustomizer = NamedFC<ThemeFamilyCustomizerProps>(
+    'ThemeCustomizer',
+    props => {
+        const isHighContrastEnabled = props.userConfigurationStoreData?.enableHighContrast === true;
+        const defaultTheme = themeFamilyDefaultThemes[props.themeFamily];
+
+        const activeTheme = isHighContrastEnabled ? highConstrastTheme : defaultTheme;
+
+        return <Customizer settings={{ theme: activeTheme }}>{props.children}</Customizer>;
```
karanbirsingh, line 38: It looks like `Customizer` is being deprecated in Fluent 8 in favor of `ThemeProvider`: https://github.com/microsoft/fluentui/wiki/Version-8-migration-guide#theming-changes It seems like there was a preview of `ThemeProvider` for fluent 7, but I don't know how much of it was portable to 8. This PR seems like a great improvement so I don't have reservations about `Customizer` usage but thought I'd mention the changes in 8. We may want to make a note in the Fluent 8 investigation spec.

dbjorge: Thanks for the pointer! Noted the suggested upgrade path in a comment and let the Fluent UI 8 upgrade feature crew know about its existence

src/DetailsView/components/adhoc-tab-stops-test-view.tsx
```diff
@@ -118,50 +121,63 @@ export const AdhocTabStopsTestView = NamedFC<AdhocTabStopsTestViewProps>(
...
-                    {description}
+        const tabStopsTestViewContents = (
+            <>
+                <h1>
+                    {displayableData.title}
+                    {` ${stepsText} `}
+                    <ContentLink deps={props.deps} reference={props.guidance} iconName="info" />
+                </h1>
+                {description}
+                <ThemeFamilyCustomizer
```
dbjorge, line 132: The diff here is a little awkward to read. The semantic changes are that the whole test view is wrapped in a `<ThemeFamilyCustomizer themeFamily={'fast-pass`}>` so that all the `DetailsLists` get the correct colors by default, and then this specific toggle button is wrapped in an inner `<ThemeFamilyCustomizer themeFamily={'default'}>` so that it can keep the white background in the "off" state consistent with other pages (per PM feedback)

karanbirsingh (approved):
nice improvement & appreciate the detailed description!

PR #5112, opened 2022-01-26
feat(tabstops-auto-impl): remove duplicate results & reset state between consecutive tab-stop-analyzer invocations

karanbirsingh:
#### Details

This PR addresses two issues in the current automated tab-stops implementation:
- duplicate results are not ignored. The impact is that users can be presented with unnecessary results they must manually delete. Our focus-order test implementation will produce duplicate results often (by design), so we need to filter those out.
- orchestrator-held state (in the target page) is not reset between consecutive start()/stop() calls. The impact is that hitting "start over" will act as a continuation of the last test (for example, the test will ignore any tab stops it found the first time)

##### Motivation

Fix behavior, unblock correct e2e tests

##### Context

I expect this code to change again in the upcoming visualization feature. The current feature is mainly scoped to gathering automated tab stops instances in /background. We have an upcoming feature that:
- changes the visualization experience in the target page
- decouples the details view toggle behavior from starting/stopping the test
- (most relevant) defers issue reporting until the scan is done

Depending on where the visualization feature crew decides to store requirement results before reporting, this duplication logic might move elsewhere. Therefore duplication logic is simple for now and merely checks deep equality per result.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/injected/analyzers/tab-stops-analyzer.ts
+++ b/src/injected/analyzers/tab-stops-analyzer.ts
@@ -11,7 +11,7 @@ import { AllFrameRunner } from 'injected/all-frame-runner';
 import { BaseAnalyzer } from 'injected/analyzers/base-analyzer';
 import { ScanIncompleteWarningDetector } from 'injected/scan-incomplete-warning-detector';
 import { TabStopRequirementResult } from 'injected/tab-stops-requirement-evaluator';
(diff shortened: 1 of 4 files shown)


sfoslund (approved):
LGTM!

PR #5113, opened 2022-01-26
feat(tabstops-auto-impl): Fix results reporting loop bug

sfoslund:
#### Details

Fix bug introduced in latest tabstops-auto-impl feature work. If a user does not finish tabbing through the target page then the analyzer-controller enters a loop in which it sends automatic tabbing results to the store over and over, causing a crash. This PR fixes this bug and adds test coverage. 

##### Motivation

Bug fix for feature work.

##### Context

This fix adds an additional field to tab stops store data which allows us to track both if the user is finished tabbing (`tabbingCompleted`) as well as if results have not been reported and need to be updated (`needToCollectTabbingResults`). The second field is necessary because we only want to report results once, the very first time tabbing is completed. We considered fixing this bug by marking `tabbingCompleted = false` in `analyzer-controller`'s `onResultsChangedState`, but without the second `needToCollectTabbingResults` field, this meant that we would resend results every time we received a `tabbingCompleted` trigger- so we would resend results every time the user refocused on the details view. With 2 fields, we're able to only send results when both `tabbingCompleted=true` and `needToCollectTabbingResults=true`, which only happens once in the testing.

An alternative approach could be to only fire the `updateTabbingCompleted` message once, when we get a done trigger (by either focus shifting to the details view or reaching the same tabbed element twice) and we know the test is still active. The problem here is that we don't necessarily know when the test is active at the moment. We expect that a future refactor/ future feature work in this area may end up devising a better way to represent if the test is in progress or not, and if that is the case then using that heuristic to determine if we should fire off an `updateTabbingCompleted` message will likely be more readable than this current implementation. 

This fix runs up against another known bug with the tab stops visualization toggle, which will be fixed in the following feature: 
- Repro steps:
  - Open tab stops tab, toggle visual helper on, tab through target page as normal
  - Return focus to details view to see failures
  - Toggle visual helper on and immediately back off
  - BUG: new failure instances are added to the list
- Explanation
  - Current behavior of the visual helper is that it resets the tabbed elements in the data store and restarts the tab-stops-listener, which can be viewed as a sort of 'soft start over' of the tabbing test
  - The tab stops analyzer tries to re-run the automated checks, and comes up with new results, which are returned and displayed in addition to the results that were already present in the details view
  - The issue seems to be that the behavior of the toggle is unclear- at the moment it is acting similarly to start over, when the desired behavior is that it controls only the visualization on the target page and does not effect the tab stops analyzer at all
- Way forward
  - The way forward is likely to change the behavior of the toggle so it only turns on and off the existing visualization but does not make changes to the tabbed elements or the tab stop analyzer

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/actions/tab-stop-requirement-action-message-creator.ts
+++ b/src/DetailsView/actions/tab-stop-requirement-action-message-creator.ts
@@ -10,6 +10,7 @@ import {
     ResetTabStopRequirementStatusPayload,
     ToggleTabStopRequirementExpandPayload,
     UpdateTabbingCompletedPayload,
+    UpdateNeedToCollectTabbingResultsPayload,
 } from 'background/actions/action-payloads';
 import { DevToolActionMessageCreator } from 'common/message-creators/dev-tool-action-message-creator';
 import { Messages } from 'common/messages';
@@ -120,6 +121,17 @@ export class TabStopRequirementActionMessageCreator extends DevToolActionMessage
 
         this.dispatcher.dispatchMessage({
             messageType: Messages.Visualizations.TabStops.TabbingCompleted,
+            payload,
+        });
+    };
+
+    public updateNeedToCollectTabbingResults = (needToCollectTabbingResults: boolean) => {
+        const payload: UpdateNeedToCollectTabbingResultsPayload = {
+            needToCollectTabbingResults,
+        };
+
+        this.dispatcher.dispatchMessage({
+            messageType: Messages.Visualizations.TabStops.NeedToCollectTabbingResults,
             payload,
         });
     };
--- a/src/background/actions/tab-stop-requirement-action-creator.ts
+++ b/src/background/actions/tab-stop-requirement-action-creator.ts
@@ -9,6 +9,7 @@ import {
     RemoveTabStopInstancePayload,
     ResetTabStopRequirementStatusPayload,
     ToggleTabStopRequirementExpandPayload,
+    UpdateNeedToCollectTabbingResultsPayload,
     UpdateTabbingCompletedPayload,
     UpdateTabStopInstancePayload,
     UpdateTabStopRequirementStatusPayload,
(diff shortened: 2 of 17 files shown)


src/DetailsView/actions/tab-stop-requirement-action-message-creator.ts
```diff
@@ -123,4 +124,15 @@ export class TabStopRequirementActionMessageCreator extends DevToolActionMessage
             payload,
         });
     };
+
+    public updateNeedToCollectTabbingResults = (needToCollectTabbingResults: boolean) => {
+        const payload: UpdateNeedToCollectTabbingResultsPayload = {
```
karanbirsingh, line 129: This function is missing accompanying unit tests - jest code cov shows uncovered lines "129-133"

src/background/actions/tab-stop-requirement-action-creator.ts
```diff
@@ -112,4 +118,10 @@ export class TabStopRequirementActionCreator {
     private onTabbingCompleted = (payload: UpdateTabbingCompletedPayload): void => {
         this.tabStopRequirementActions.updateTabbingCompleted.invoke(payload);
     };
+
+    private onNeedToCollectTabbingResults = (
```
karanbirsingh, line 122: This callback does not have accompanying unit tests

src/background/stores/visualization-scan-result-store.ts
```diff
@@ -253,4 +258,11 @@ export class VisualizationScanResultStore extends BaseStoreImpl<VisualizationSca
         this.state.tabStops.tabbingCompleted = payload.tabbingCompleted;
         this.emitChanged();
     };
+
+    private onUpdateNeedToCollectTabbingResults = (
+        payload: UpdateNeedToCollectTabbingResultsPayload,
+    ): void => {
+        this.state.tabStops.needToCollectTabbingResults = payload.needToCollectTabbingResults;
```
karanbirsingh, line 265: This callback is missing unit tests

karanbirsingh (approved):

karanbirsingh:
Great context in the PR description 🙂

Issue #5120, opened 2022-01-28
Divs with role=heading should be replaced with native heading elements


jalkire:
## Describe the bug
In a few places in our codebase, we use div elements with role=heading instead of actual heading elements. According to [our guidance](https://github.com/microsoft/accessibility-insights-web/blob/5deba6151546207ecb802540184a0c7e46947827/src/content/adhoc/headings/guidance.tsx) and [other best practices](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#best_practices), we should use native heading elements when possible. This bug tracks the work to refactor those. 

## Are you willing to submit a PR?
Yes

## Did you search for similar existing issues?
Yes

majaokholm:
@jalkire I would like to give this one a go, if not already taken? It would be my first contribution here but looks pretty straight forward.

jalkire:
> @jalkire I would like to give this one a go, if not already taken? It would be my first contribution here but looks pretty straight forward.

@majaokholm Sure thing, thank you for checking! One hint is to take a look at how we're using [`HeadingElementForLevel`](https://github.com/microsoft/accessibility-insights-web/blob/main/src/common/components/heading-element-for-level.tsx) right now as a starting point 🙂

ferBonnin:
@majaokholm thanks! I have assigned this issue to you 😄

majaokholm:
@jalkire sorry for the delay, got fir by sick kids. Nevertheless, I made a few changes but excluding e2e tests and unit tests I only found the div element with role=heading being used three places. I have refactored these three instances but am wondering if I might be overlooking something? I have created a draft pull request for my changes

jalkire:
> @jalkire sorry for the delay, got fir by sick kids. Nevertheless, I made a few changes but excluding e2e tests and unit tests I only found the div element with role=heading being used three places. I have refactored these three instances but am wondering if I might be overlooking something? I have created a draft pull request for my changes

No problem at all, thank you for opening the PR! You found all the right div elements. I left a few pieces of feedback on the PR

ferBonnin:
@jalkire can this be closed now?

PR #5121, opened 2022-01-28
feat(tabstops-report): Set results section heading levels dynamically

jalkire:
#### Details
Because the new FastPass report has a different hierarchy of headings than other reports and views, we now need a way to set the heading level of results sections dynamically. To do this, `HeadingElementForLevel` is introduced to allow creating h1-h6 elements based on a heading level of 1-6. It is then used to replace existing h2 elements in results sections.

New heading hierarchy:
screenshot of fastpass report with headings visualized

##### Motivation
Addresses part of #5099

##### Context
At first, I considered using our existing pattern of divs with role=heading ([example](https://github.com/microsoft/accessibility-insights-web/blob/d86b5c3c9614c1d059a68a3d65fa68284ac40a1f/src/common/components/cards/collapsible-component-cards.tsx#L46)). However, this is [against best practices](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#best_practices) and [our own guidance](https://github.com/microsoft/accessibility-insights-web/blob/5deba6151546207ecb802540184a0c7e46947827/src/content/adhoc/headings/guidance.tsx). Given the size of this PR and some style impacts, I haven't included refactoring our existing role=heading elements in this PR. That work is tracked by #5120

#### Pull request checklist

- [x] Addresses an existing issue: #5099
- [x] Ran `yarn fastpass`


src/common/components/cards/failed-instances-section.tsx
```diff
@@ -40,7 +41,8 @@ export const FailedInstancesSection = NamedFC<CommonInstancesSectionProps>(
                 visualHelperEnabled={cardsViewData.visualHelperEnabled}
                 allCardsCollapsed={cardsViewData.allCardsCollapsed}
                 outcomeCounter={OutcomeCounter.countByCards}
-                headingLevel={3}
+                headingLevel={titleHeadingLevel + 1}
```
dbjorge, line 44: imo, it would make more sense for the `ResultSection` to be the component responsible for understanding that the "titleHeadingLevel" should always be one above the content heading level, since it's the component responsible for how those two are composed; maybe a single `headingLevel` prop for the section and have the section know that `titleHeadingLevel = sectionHeadingLevel` and `contentHeadingLevel = sectionHeadingLevel + 1`?

src/common/components/heading-element-for-level.tsx
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { NamedFC } from 'common/react/named-fc';
+import * as React from 'react';
+
+export type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6;
```
dbjorge, line 6: This vs aliasing `number` is an interesting choice. I was surprised when I saw this that it wasn't causing compilation failures everywhere the PR uses `childHeadingLevel={parentHeadingLevel + 1}`; it turns out, the only reason there aren't errors about that is because `tab-stops-instance-section-props-factory.tsx` and `result-section-content.tsx` both have `headingLevel: number` props still instead of `headingLevel: HeadingLevel`.

I think to enable the `headingLevel + 1` cases properly you need a runtime check somewhere. You could either do:

* Have `HeadingLevel` alias `number` instead of being an enum type, and make `HeadingElementForLevel` throw an error if it's given something that isn't an integer 1-6, or:
* Keep `HeadingLevel` and `HeadingElementForLevel` as is, but replace all the `headingLevel + 1` invocations with invocations of a new function along the lines of `nextHeadingLevel` that encapsulates the addition alongside a "throw if called with 6" check

I lean towards the second one personally but I think either would be ok.

dbjorge: if you take the second approach, you might also consider having `HeadingLevel` be non-numeric (eg, `'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'`) to make it harder for consumers to misuse it with "+ and cast"

src/common/components/cards/failed-instances-section.tsx
jalkire: Good idea! Moved this logic into `ResultSection`. I also prefer `sectionHeadingLevel` to `titleHeadingLevel`, so I renamed that

src/common/components/heading-element-for-level.tsx
jalkire: This makes sense. I considered the nextHeadingLevel helper idea originally but didn't go with it because it could easily be ignored. If `HeadingLevel` is made non-numeric, that would make the helper function more useful/harder to skip.

For this PR, I've implemented the first part of the suggested change (performing headingLevel + 1 only in a helper function). I think it would make the most sense to make the non-numeric change to the type in the follow up refactoring PR, since that will eliminate are usage of number-based heading levels entirely.

jalkire: This makes sense. I considered the nextHeadingLevel helper idea originally but didn't go with it because it could easily be ignored. If HeadingLevel is made non-numeric, that would make the helper function more useful/harder to skip.

For this PR, I've implemented the first part of the suggested change (performing headingLevel + 1 only in a helper function). I think it would make the most sense to make the non-numeric change to the type in the follow up refactoring PR, since that will eliminate our usage of number-based heading levels entirely.

dbjorge (approved):

PR #5137, opened 2022-02-03
feat(tabstops-auto-target-page-vis): Add telemetry for automated tab stops results

sfoslund:
#### Details

Add telemetry for automated tab stops results. Telemetry event includes the number of issues by requirement when results are found. When results are not found, no telemetry is sent.

##### Motivation

Feature work.

##### Context

Ideally we wouldn't have to add another message for adding telemetry. I considered reusing the `NeedToCollectTabbingResults` message for this but at the moment it only keeps track of if tabbing results need to be collected or not (a boolean) and doesn't know what the results actually are. Further, we only want to send the telemetry event once so we would have to add telemetry only when the payload is false (that is, we have finished collecting tabbing results, which should only happen once). We could update this message to take in the results as well and handle the telemetry, but I opted to create a new message instead to keep it logically separated. 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/common/telemetry-data-factory.test.ts
+++ b/src/tests/unit/tests/common/telemetry-data-factory.test.ts
@@ -770,7 +770,7 @@ describe('TelemetryDataFactoryTest', () => {
         expect(result).toEqual(expected);
     });
 
-    test('forAutomatedTabStopsResults returns null when no results', () => {
+    test('forAutomatedTabStopsResults returns undefined when no results', () => {
         const result = testObject.forAutomatedTabStopsResults([]);
         expect(result).toBeUndefined();
     });
--- a/src/tests/unit/tests/common/telemetry-data-factory.test.ts
+++ b/src/tests/unit/tests/common/telemetry-data-factory.test.ts
@@ -18,6 +18,7 @@ import {
     SetAllUrlsPermissionTelemetryData,
     SettingsOpenSourceItem,
     SettingsOpenTelemetryData,
+    TabStopsAutomatedResultsTelemetryData,
     TelemetryEventSource,
     ToggleTelemetryData,
     TriggeredByNotApplicable,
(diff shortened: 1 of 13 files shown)


lisli1 (approved):
LGTM

waabid (approved):

src/tests/unit/tests/injected/analyzers/tab-stops-analyzer.test.ts
```diff
@@ -249,6 +250,10 @@ describe('TabStopsAnalyzer', () => {
             testSubject.teardown();
             await flushSettledPromises();
             verifyAll();
+            tabStopRequirementActionMessageCreatorMock.verify(
+                m => m.automatedTabbingResultsCompleted(It.isAny()),
```
waabid, line 254: Feels like we should be explicitly checking what the requirement results object looks like here.

src/tests/unit/tests/common/telemetry-data-factory.test.ts
```diff
@@ -749,4 +751,27 @@ describe('TelemetryDataFactoryTest', () => {
...
+        const expected: TabStopsAutomatedResultsTelemetryData = {
+            tabStopAutomatedFailuresInstanceCount: { 'tab-order': 2, 'keyboard-traps': 1 },
+            source: TelemetryEventSource.DetailsView,
+            triggeredBy: 'N/A',
+        };
+
+        expect(result).toEqual(expected);
+    });
+
+    test('forAutomatedTabStopsResults returns null when no results', () => {
```
waabid, line 773: nit: should be "returns undefined"

waabid (commented):
LGTM, just a couple minor comments

src/tests/unit/tests/injected/analyzers/tab-stops-analyzer.test.ts
sfoslund: Good point, updated, thanks!

PR #5154, opened 2022-02-08
feat(tabstops-auto-target-page-vis): Add tab stops auto detected failures pop up dialog

sfoslund:
#### Details

Add pop up to tab stops page when automatic failures are detected.

AutoDetectedPopUpTabStops
(screenshot of tab stops details view page with popup in front. Popup text is: 'Possible failures auto-detected. Review each finding to be sure it's valid. Add any more failures you discovered manually')

##### Motivation

Feature work.

##### Context

Pop up is behind `tabStopsAutomation` feature flag and only displays when failures have been auto detected. Ideally we will add a `Don't show again` checkbox later in this feature, but this is a stretch goal.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/adhoc-tab-stops-test-view.tsx
+++ b/src/DetailsView/components/adhoc-tab-stops-test-view.tsx
@@ -16,6 +16,7 @@ import { VisualizationStoreData } from 'common/types/store-data/visualization-st
 import { VisualizationType } from 'common/types/visualization-type';
 import { DetailsViewActionMessageCreator } from 'DetailsView/actions/details-view-action-message-creator';
 import * as styles from 'DetailsView/components/adhoc-tab-stops-test-view.scss';
+import { AutoDetectedFailuresDialog } from 'DetailsView/components/auto-detected-failures-dialog';
 import * as requirementInstructionStyles from 'DetailsView/components/requirement-instructions.scss';
 import {
     TabStopsFailedInstanceSection,
@@ -192,6 +193,15 @@ export const AdhocTabStopsTestView = NamedFC<AdhocTabStopsTestViewProps>(
                     featureFlagStoreData={props.featureFlagStoreData}
                     enableJSXElement={
                         <FocusComponent deps={props.deps} tabbingEnabled={scanData.enabled} />
+                    }
+                />
+                <FlaggedComponent
+                    featureFlag={FeatureFlags.tabStopsAutomation}
+                    featureFlagStoreData={props.featureFlagStoreData}
+                    enableJSXElement={
+                        <AutoDetectedFailuresDialog
+                            visualizationScanResultData={props.visualizationScanResultData}
+                        />
                     }
                 />
             </>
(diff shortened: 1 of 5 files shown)


src/DetailsView/components/adhoc-tab-stops-test-view.tsx
```diff
@@ -61,9 +61,55 @@ export interface AdhocTabStopsTestViewProps {
...
 
-export const AdhocTabStopsTestView = NamedFC<AdhocTabStopsTestViewProps>(
-    'AdhocTabStopsTestView',
-    props => {
+export type AdhocTabStopsTestViewState = {
+    autoDetectedFailuresDialogEnabled: boolean;
+};
+
+export class AdhocTabStopsTestView extends React.Component<
+    AdhocTabStopsTestViewProps,
```
waabid, line 69: I think the AdHocTabStopsTestView component already has too many responsibilities and we should try to avoid adding more. In this case, can we not just keep the state in the dialog component?

Additionally, instead of having stateful components, this might be a good opportunity to shift more "view" information into the view store (specifically tab-stops-view-store.ts). This would mean the AdhocTabStopsTestView can remain a functional component and so can the new dialog.

waabid (changes requested):
Left comment.

src/tests/unit/tests/DetailsView/components/auto-detected-failures-dialog.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+        expect(wrapper.getElement()).toMatchSnapshot();
+    });
+
+    describe('with dialogEnabled = true', () => {
+        beforeEach(() => {
+            props.dialogEnabled = true;
+        });
+
+        it('is dismissed when got it is clicked', () => {
```
lisli1, line 36: It's not immediately obvious from this test file that the button is labeled `Got it`. For clarity, should we update this to say something like `is dismissed when "Got it" button is clicked`?

lisli1 (approved):
Left minor comment, otherwise LGTM

src/DetailsView/components/adhoc-tab-stops-test-view.tsx
sfoslund: @waabid Yeah, I can definitely shift the state to the dialog component. How would it work to have the state in the TabStopsViewStore though, since that's in the details view? If we keep the dialog state in a store, I would think we would want to update it based on `AutomatedTabbingResultsCompleted`, which is kept in the background.

sfoslund: For now I've switched over to keeping the state as part of the dialog component, let me know if you have any thoughts on moving the state to the store and I can look more into that as well :)

waabid: This is good as is.

I understand what you mean now; you need the state from the other store data. I guess you could still move the dialog state to the store but the componentDidUpdate and what not still would be required, so the benefit is minimal.

waabid (approved):

PR #5160, opened 2022-02-10
feat(tabstops-auto-target-page-vis): process both tabbed elements and requirement instances before data is visualized.

waabid:
#### Details

The data from the store needs to be normalized to minimize changes to the drawer layer. This PR does that.

##### Motivation

Feature work.

##### Context

Problems not solved in this PR:
- What about automated instances that share selectors? We don't have a way to represent that information currently.
- We need to include information about type of failure in these instances to visualize them differently.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/src/common/types/store-data/visualization-scan-result-data.ts
+++ b/src/common/types/store-data/visualization-scan-result-data.ts
@@ -33,12 +33,14 @@ export type TabStopRequirementInstance = {
     html?: string;
 };
 
+export type SingleTabStopRequirementState = {
+    status: TabStopRequirementStatus;
+    instances: TabStopRequirementInstance[];
+    isExpanded: boolean;
+};
+
 export type TabStopRequirementState = {
-    [requirementId: string]: {
-        status: TabStopRequirementStatus;
-        instances: TabStopRequirementInstance[];
-        isExpanded: boolean;
-    };
+    [requirementId: string]: SingleTabStopRequirementState;
 };
 
 export interface TabStopsScanResultData {
--- a/src/injected/visualization/get-visualization-instances-for-tab-stops.ts
+++ b/src/injected/visualization/get-visualization-instances-for-tab-stops.ts
@@ -1,0 +1,48 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { FeatureFlags } from 'common/feature-flags';
+import { FeatureFlagStoreData } from 'common/types/store-data/feature-flag-store-data';
+import { TabStopsScanResultData } from 'common/types/store-data/visualization-scan-result-data';
+import { AssessmentVisualizationInstance } from 'injected/frameCommunicators/html-element-axe-results-helper';
+import { SelectorToVisualizationMap } from 'injected/selector-to-visualization-map';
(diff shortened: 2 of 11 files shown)


src/common/types/store-data/visualization-scan-result-data.ts
```diff
@@ -33,12 +33,14 @@ export type TabStopRequirementInstance = {
     html?: string;
 };
 
+export type SingleTabStopRequirementState = {
```
sfoslund, line 36: nit: personally don't love this name, maybe TabStopRequirementStateInstance?

src/injected/visualization/get-visualization-instances-for-tab-stops.ts
```diff
@@ -0,0 +1,48 @@
...
+        forOwn(tabStopScanResultData.requirements, obj => {
+            obj.instances.forEach(instance => {
+                const newInstance = {
+                    isFailure: true,
+                    isVisualizationEnabled: true,
+                    ruleResults: null,
+                    target: instance.selector,
+                    propertyBag: {},
+                };
+                ret[newInstance.target.join(';')] = newInstance;
```
sfoslund, line 42: Is it intentional that this could be overwriting elements added in the first foreach loop?

src/tests/unit/tests/injected/visualization/get-visualization-for-tab-stops.test.ts
```diff
@@ -0,0 +1,113 @@
...
+                timestamp: tabbedElements[0].timestamp,
+            }),
+            'another;target': buildVisualizationInstance(tabbedElements[1].target, false, {
+                tabOrder: tabbedElements[1].tabOrder,
+                timestamp: tabbedElements[1].timestamp,
+            }),
+        } as SelectorToVisualizationMap;
+    });
+
+    test('GetVisualizationInstancesForTabStops: without requirement instances with feature flag off', () => {
```
sfoslund, line 44: nit: I found these test names confusing, as it seems to imply a testing matrix with 4 possibilities (requirement instances on/off, feature flag on/off). Can we either add those other tests (with requirement instances and feature flag off/ without requirement instances and feature flag on) or alter these test names slightly?

sfoslund (commented):
A couple nits and a clarifying question

src/common/types/store-data/visualization-scan-result-data.ts
waabid: Neither do I, to be honest.

But not sure how I feel about `StateInstance` either ... since the only instance-like thing about it is that it's an instanced object ... and could be confused with what we have generally referred to as instances before as well.

I was thinking re-naming the original type might be best but did not want to do that in this PR since it would get quite large.

How about:

`TabStopRequirementState` to `TabStopTestState` and using `TabStopRequirementState` for this situation?

src/injected/visualization/get-visualization-instances-for-tab-stops.ts
waabid: Discussed offline.

lisli1 (approved):

PR #5161, opened 2022-02-11
feat(tabstops-auto-target-page-vis): Fix FocusTrackingAnalyzer bug

sfoslund:
#### Details

Fixes bug in which automatic tab stops requirement checks are run for any test that uses the FocusTrackingAnalyzer (for example, Keyboard Navigation requirement under Assessment). As a result, extra failure instances will be collected, stored, and displayed when the user switches to the tabstops report. 

##### Motivation

Fix bug introduced by feature work.

##### Context

Added an extra method to `analyzer-provider` to avoid having to condition based on `VisualizationType` in the TabStopsAnalyzer. 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/tab-stops-requirement-evaluator.ts
+++ b/src/injected/tab-stops-requirement-evaluator.ts
@@ -44,7 +44,7 @@ export class DefaultTabStopsRequirementEvaluator implements TabStopsRequirementE
         actualTabStops: Set<HTMLElement>,
     ): AutomatedTabStopRequirementResult[] {
         const requirementResults: AutomatedTabStopRequirementResult[] = [];
-        (tabbableTabStops as HTMLElement[]).forEach(expectedTabStop => {
+        ((tabbableTabStops as HTMLElement[]) ?? []).forEach(expectedTabStop => {
             if (!actualTabStops.has(expectedTabStop)) {
                 const selector = this.generateSelector(expectedTabStop);
                 requirementResults.push({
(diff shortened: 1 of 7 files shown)


src/injected/tab-stops-requirement-evaluator.ts
```diff
@@ -44,7 +44,7 @@ export class DefaultTabStopsRequirementEvaluator implements TabStopsRequirementE
         actualTabStops: Set<HTMLElement>,
     ): AutomatedTabStopRequirementResult[] {
         const requirementResults: AutomatedTabStopRequirementResult[] = [];
-        (tabbableTabStops as HTMLElement[]).forEach(expectedTabStop => {
+        ((tabbableTabStops as HTMLElement[]) ?? []).forEach(expectedTabStop => {
```
waabid, line 47: I think it'd be better to use an if statement here since that's more readable.

waabid: Also is this change necessary due to something in this PR or something else?

sfoslund: With a different iteration of this fix I was noticing this method erroring when `tabbableTabStops` is null, so I figured I'd include the fix since it's so small. I don't think it's actually an issue anymore with this implementation of the fix

waabid (approved):

PR #5175, opened 2022-02-14
fix: Show full focus on target page link

### LINKED ISSUE #5067: Focus Indicator in FastPass and Assessment report not displaying correctly
## Describe the bug

In both the FastPass and Assessment reports the focus indicator for the target page link is not displaying correctly.

**To Reproduce**

1. Run FastPass or Assessment
2. Export the report as HTML
3. Tab until the Target Page link in the top-right has focus

## Expected behavior

The focus indicator should be a full box around the focused element.

## Screenshots

image

- OS Name & Version: Windows 11 Pro 21H2 Build 22000.376
- AI-Web Version & Environment: 2.29.0 Production
- Browser Version: Microsoft Edge Version 97.0.1072.55 (Official build) (64-bit)
- Target Page: [](http://www.washington.edu/accesscomputing/AU/before.html)

Yes

Yes

katydecorah:
#### Details

This PR fixes the focus on the target link in the FastPass and Assessment report. The focus was getting clipped due to the parent's `overflow: hidden`. 

- Replace `@include ellipsedText();` on parent element (`.target-page`) with `white-space: nowrap;`. This will keep the content on one line, but not clip the overflow. The CSS already sets the ellipsed text include on the anchor element and that's the only content we want to be truncated.
- Replace `display: inherit` with `display: flex`, while these values are the same, from a readability perspective, it may be helpful to be explicit with this value and help make the code modular. Using flex is important here so that only the link is truncated.
- Set `min-width: 0` on `.target-page`. I found this to be the best option to avoid using `overflow: hidden` on `.target-page` since hiding the overflow clips the focus. By setting `min-width` the element will contain the contents. This post on [flebox and truncated text](https://css-tricks.com/flexbox-truncated-text/#aa-the-solution-is-min-width-0-on-the-flex-child) explains more about using min-width on a flex child. An alternative would be to set `max-width: calc(100% -16px)` -- this accounts for the margin-left, but this option feels a little more brittle since it could break if `margin` changes.

Current | Proposed
---|---
screenshot of fast pass results, the target link focus does not complete enclose the link | screenshot of fast pass result, the target link focus, which is a black border, completely surrounds the link

On smaller width devices, the target link continues to truncate the text:
screenshot showing the target link is truncated

##### Motivation

Addresses #5067

##### Context

In my initial solution, I had made `.target-page` a block element (removed flexbox), increased the line-height, and remove the ellipsed text include from the anchor element, see commit [e664889](https://github.com/microsoft/accessibility-insights-web/pull/5175/commits/e664889c589688a048907f1dbaf96ed43f2fee6a). While this technically worked, on smaller devices the truncated ellipses use the text color, not the link color. Ultimately, we should truncate the link itself, not the parent element.

#### Pull request checklist

- [x] Addresses an existing issue: #5067
- [x] Ran `yarn fastpass`

### DIFF
--- a/packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
+++ b/packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
@@ -656,19 +656,13 @@
         overflow: hidden;
         text-overflow: ellipsis;
         margin: 0px 0px 0px 16px;
-        display: inherit;
         color: var(--primary-text);
         font-family: "Segoe UI Web (West European)", "Segoe UI", "-apple-system",
           BlinkMacSystemFont, Roboto, "Helvetica Neue", Helvetica, Ubuntu, Arial,
           sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
         font-size: 14px;
-        line-height: 20px;
+        line-height: 24px;
         font-weight: normal;
-      }
-      .report-header-command-bar .target-page a {
-        white-space: nowrap;
-        overflow: hidden;
-        text-overflow: ellipsis;
       }
       .report-congrats-message {
         word-break: break-word;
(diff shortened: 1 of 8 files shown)


katydecorah:
Switching this back to draft. I just noticed that the ellipsis are black and not the link color -- the ellipsis should match the link as shown in production. I think that's why the prent elements inherited `flex`. I'm going to look a little more closely at this.

JGibson2019 (approved):
LGTM!

PR #5176, opened 2022-02-14
feat(tabstops-auto-target-page-vis): introduce TabStopVisualizationInstance for multiple failures on one element

waabid:
#### Details

Allows us to have multiple failures attributed to one element.

##### Motivation
 
feature work

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/frameCommunicators/html-element-axe-results-helper.ts
+++ b/src/injected/frameCommunicators/html-element-axe-results-helper.ts
@@ -24,7 +24,7 @@ export interface AssessmentVisualizationInstance extends AxeResultsWithFrameLeve
 
 export type TabStopVisualizationRequirementResults = Partial<{
     [requirementId in TabStopRequirementId]: {
-        description: string;
+        instanceId: string;
     };
 }>;
 
--- a/src/injected/frameCommunicators/html-element-axe-results-helper.ts
+++ b/src/injected/frameCommunicators/html-element-axe-results-helper.ts
@@ -22,12 +22,14 @@ export interface AssessmentVisualizationInstance extends AxeResultsWithFrameLeve
     propertyBag?: any;
 }
 
-export interface TabStopVisualizationInstance extends AssessmentVisualizationInstance {
-    requirementResults: {
-        [requirementId in TabStopRequirementId]: {
-            description: string;
-        };
+export type TabStopVisualizationRequirementResults = Partial<{
+    [requirementId in TabStopRequirementId]: {
+        description: string;
     };
+}>;
+
+export interface TabStopVisualizationInstance extends AssessmentVisualizationInstance {
+    requirementResults: TabStopVisualizationRequirementResults;
 }
 
 export class HtmlElementAxeResultsHelper {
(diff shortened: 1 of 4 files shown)


lisli1 (approved):
LGTM

src/tests/unit/tests/injected/visualization/get-visualization-instances-for-tab-stops.test.ts
```diff
@@ -74,12 +90,60 @@ describe('GetVisualizationInstancesForTabStops', () => {
...
+
+        expectedResults = {};
+        expectedResults['some;requirement result selector'] = buildVisualizationInstance(
+            firstRequirementResults[0].selector,
+            true,
+            {},
+            {
+                'keyboard-navigation': { description: firstRequirementResults[0].description },
+                'keyboard-traps': { description: secondRequirementResults[0].description },
+            },
```
sfoslund, line 146: Other test cases we might want to add: 
- Requirement results are null (also tabbed elements, is there ever a situation in which this would be called when they are null?)
- There are multiple instances of the same requirement id (is it ever possible for a single element to have multiple instances of the same requirement id? Maybe if an element is out of order in relation to multiple other elements?)

sfoslund (approved):

src/tests/unit/tests/injected/visualization/get-visualization-instances-for-tab-stops.test.ts
waabid: > Requirement results are null (also tabbed elements, is there ever a situation in which this would be called when they are null?)

There should be no situations where requirement results are null (instantiated in the store as an empty array).

> There are multiple instances of the same requirement id (is it ever possible for a single element to have multiple instances of the same requirement id? Maybe if an element is out of order in relation to multiple other elements?)

Good question! I was thinking of this as well ... I figured, for the moment, this captures the most important cases for the consumer of this data (the visualizations/drawer) and if we want to surface more data in the target page pertaining to multiple requirement failures (which can't happen currently due to the dedupe logic in the analyzer AFAIK), we can do that in the future. The main scenario I was worried about, which this covers, was losing out on tab ordering when a failure was created.

src/injected/frameCommunicators/html-element-axe-results-helper.ts
```diff
@@ -22,12 +22,14 @@ export interface AssessmentVisualizationInstance extends AxeResultsWithFrameLeve
...
 }
 
-export interface TabStopVisualizationInstance extends AssessmentVisualizationInstance {
-    requirementResults: {
-        [requirementId in TabStopRequirementId]: {
-            description: string;
-        };
+export type TabStopVisualizationRequirementResults = Partial<{
+    [requirementId in TabStopRequirementId]: {
+        instanceId: string;
```
waabid, line 27: Changed this to use instanceId instead because I don't see a use case for description but instance id would be useful for if/when we implement the scroll-to-instance stretch goal.

PR #5189, opened 2022-02-18
refactor: replacing div with role=heading with native heading element issue #5120

majaokholm:
#### Details

Refactoring the codebase to use native heading elements instead of div elements with role = heading  as per issue #5120 

##### Motivation

addresses issue #5120 

##### Context

I replaced previously hardcoded levels with the corosponding h element directly and the previously dynamic levels with the element HeadingElementForLevel as this elemnt seems to serve the purpose of dynamically creating a native h-element of a given level.

I intentionally left out the tests (both e2e and unit) as I am not certain what the correct action is here. 

#### Pull request checklist

- [x] Addresses an existing issue: #5120 
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/components/report-sections/report-collapsible-container.tsx
+++ b/src/reports/components/report-sections/report-collapsible-container.tsx
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { css } from '@uifabric/utilities';
+import { HeadingElementForLevel, HeadingLevel } from 'common/components/heading-element-for-level';
 import { NamedFC } from 'common/react/named-fc';
 import * as React from 'react';
 
@@ -26,7 +27,10 @@ const ReportCollapsibleContainer = NamedFC<ReportCollapsibleContainerProps>(
 
         return (
             <div className={outerDivClassName}>
-                <div className="title-container" role="heading" aria-level={headingLevel}>
+                <HeadingElementForLevel
+                    headingLevel={headingLevel as HeadingLevel}
+                    className="title-container"
+                >
                     <button
                         className="collapsible-control"
                         aria-expanded="false"
@@ -35,7 +39,7 @@ const ReportCollapsibleContainer = NamedFC<ReportCollapsibleContainerProps>(
                     >
                         {header}
                     </button>
-                </div>
+                </HeadingElementForLevel>
                 <div id={contentId} className="collapsible-content" aria-hidden="true">
                     {content}
                 </div>
--- a/src/content/adhoc/tabstops/how-to-test.tsx
+++ b/src/content/adhoc/tabstops/how-to-test.tsx
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { HeadingElementForLevel, HeadingLevel } from 'common/components/heading-element-for-level';
 import { create, React } from '../../common';
 
 export const createHowToTest = (headingLevel: number) => {
@@ -10,9 +11,7 @@ export const createHowToTest = (headingLevel: number) => {
                     Note: this test requires you to use a keyboard and to visually identify interactive elements.
                 </Markup.Emphasis>
             </p>
-            <div role="heading" aria-level={headingLevel}>
-                How to test
-            </div>
+            <HeadingElementForLevel headingLevel={headingLevel as HeadingLevel}>How to test</HeadingElementForLevel>
             <ol>
                 <li>Refresh the target page to put it in its default state.</li>
                 <li>Turn on the Show tab stops toggle. An empty circle will highlight the element with focus.</li>
--- a/src/popup/components/launch-pad.tsx
+++ b/src/popup/components/launch-pad.tsx
@@ -59,13 +59,7 @@ export class LaunchPad extends React.Component<LaunchPadProps, undefined> {
         return (
             <div className="ms-Grid main-section">
                 <main>
-                    <div
-                        role="heading"
-                        aria-level={2}
-                        className="launch-pad-title ms-fontWeight-semibold"
-                    >
-                        Launch pad
-                    </div>
+                    <h2 className="launch-pad-title ms-fontWeight-semibold">Launch pad</h2>
                     <hr className="ms-fontColor-neutralTertiaryAlt launch-pad-hr" />
                     <div className="launch-pad-main-section">
                         {this.renderLaunchPadItemRows(this.props.rowConfigs)}
--- a/src/tests/unit/tests/popup/components/launch-pad.test.tsx
+++ b/src/tests/unit/tests/popup/components/launch-pad.test.tsx
@@ -56,13 +56,7 @@ describe('LaunchPad', () => {
         const expected = (
             <div className="ms-Grid main-section">
                 <main>
-                    <div
-                        role="heading"
-                        aria-level={2}
-                        className="launch-pad-title ms-fontWeight-semibold"
-                    >
-                        Launch pad
-                    </div>
+                    <h2 className="launch-pad-title ms-fontWeight-semibold">Launch pad</h2>
                     <hr className="ms-fontColor-neutralTertiaryAlt launch-pad-hr" />
                     <div className="launch-pad-main-section">
                         <div key="row-item-1">


src/reports/components/report-sections/report-collapsible-container.tsx
```diff
@@ -26,7 +27,10 @@ const ReportCollapsibleContainer = NamedFC<ReportCollapsibleContainerProps>(
 
         return (
             <div className={outerDivClassName}>
-                <div className="title-container" role="heading" aria-level={headingLevel}>
+                <HeadingElementForLevel
+                    headingLevel={headingLevel as HeadingLevel}
```
jalkire, line 31: Rather than using a type assertion here, it might be better to update the type of `headingLevel` in `ReportCollapsibleContainerProps` to be `HeadingLevel`. I believe there are a few other places in our props where `headingLevel: number` would make more sense as `headingLevel: HeadingLevel` with this change, as well 🙂

jalkire (changes requested):
Thanks for opening this PR! I believe these 4 headings cover the scope of the bug, and your changes look on the right track to me!

One thing to keep in mind: the switch to <h1-6> elements has some minor styling impacts around spacing and font. For example, here is Launch pad heading change before the h2 change:
screenshot showing the current launch pad popup heading
and after the change, now with some minor spacing differences:
screenshot showing launch pad with different spacing than before

We probably need to introduce a few css tweaks to make sure the new headings look like the old ones 🙂

src/reports/components/report-sections/report-collapsible-container.tsx
majaokholm: @jalkire thanks a lot for looking at this! 

You are right I overlooked the css change, sorry about that. COuld you guide me to the use of the report-collapsible-container? I am not 100% sure I am looking at the right element. For the issue you pointed out, I believe I found a solution for the issue, I have added a new commit. 

I can see that there is also an issue with the styling of the heading in how-to-test, it used to not be styled as a heading and now has the heading styling. I just want to verrify that this is really how it is supposed to be (since I am completly new here)?  
old:

old_how to test

new: 
new_how to test

majaokholm: @jalkire one more question, the snapshot test src/tests/unit/tests/reports/components/report-sections/report-collapsible-container.test.tsx:35:45 is failing as it is expecing a <div> element but getting a </HeadingElementForLevel> now, but I can't seem to find the source of the error. With my very limited knowledge of snapshot testing I assume a snapshot needs to be updated somewhere but can't find the correct one. Can you guide me? Thanks again!

jalkire: Great, thank you for those updates! To your questions:

- For the use of report-collapsible-container, we use that in the html reports that the extension exports. The easiest way to see it is by clicking "Export result" from a FastPass in the extension--the parts of the downloaded html report that expand/collapse on click are in the `ReportCollapsibleContainer`. 
- For the snapshots, if you add `-u` after the test command that will update the snapshots. Then you can confirm that the changes make sense 🙂 
- For the tab stops how to test heading, I would default to leaving it as-is, but tagging @ferBonnin in case the heading style should be changed.

ferBonnin: thanks for this contributions @majaokholm! For the Tab stops _How to test_ Heading, I recommend we have it as an **H2**, so its the same as the _why it matters_ headings 🙂

majaokholm: @jalkire Thanks again for the feedback! But I am afraid I need to ask for a bit of help once more; to fix the last styling differences I need to update css for the report-collapsible-container (because the default browser styling for h4 elements is being applied now), but I can’t figure out which components uses which stylesheet, or how to add a new one. It should be so simple but I spend so much time trying to resolve this that I figured out it is time to ask for help. Can you please guide me once more? Thanks again!

jalkire: Sure thing! Unfortunately this particular styling looks a little messy on our end--tab-stops-failed-instance-section.scss, result-section.scss, automated-checks-report.scss, and collapsible-url-result-section.scss all set properties on title-container in report-collapsible-container.tsx. I might be missing something myself, but you're probably best off applying your fix to all of those.

This somewhat relates to the last failing PR check, e2e-report-tests--once you get that css updated, you can get the test passing by running `yarn build:package:report` and then ` yarn test:report:e2e -- -- -- -u` to update report snapshots.

Thanks for putting in all this effort, you're almost there!

majaokholm: I found a way to make it work (I believe) but am not convinced it's the most elegant solution. 

From my understanding, the title-container styling in the files you refered (tab-stops-failed-instance-section.scss, result-section.scss, automated-checks-report.scss, and collapsible-url-result-section.scss) are not applied in the case of the report-collapsible-container.tsx as the title-container styling is nested inside other classes in all the cases and neither of these classes are applied in the report-collapsible-container.tsx case. So I ended up appling the change to the content.scss file, which seems to work, but maybe a new file would have been better? 

And thanks for the guidance - again, again :-)

jalkire: Thank you for taking a look at that! With the content.scss changes, I'm still seeing the default `margin-block` values causing extra spacing to appear in the fastpass report:
screenshot showing extra spacing in the updated fastpass report between the newly updated H3 and H4 elements
In the old report, the h3 and h4 labelled elements (lowercase h indicating that these aren't native H<x> elements) are stacked with no extra spacing. In the updated report, those are now proper H3 and H4 elements, but there's now some extra spacing. 

Instead of putting the change in content.scss or the my other previous suggestions, I was able to get things looking right by adding your css changes to results-container.scss, something like:
```
.results-container {
    margin-top: 56px;

    .title-container {
        font-weight: unset;
        margin-block: 0;
        margin-inline: 0;
    }
}
```

src/popup/Styles/popup.scss
```diff
@@ -115,7 +115,7 @@ html {
     -moz-box-sizing: border-box;
 }
 
-#popup-container [role='heading'] {
+#popup-container :is(h1, h2, h3, h4, h5, h6) {
     width: fit-content;
     padding-left: 0;
 }
```
jalkire, line 121: The newly updated popup heading could use a `margin-block: 0;` to remove some extra spacing🙂

majaokholm: Thanks a lot for the suggestions, I have implemented both however the e2e report is still failing when I run it locally. I am struggeling to find the source of the issue, and in my digging I found another '<div role=heading' element being used in the rule-details-group in the DetailsView which I believe comes from the rules-with-instance.tsx component. However I am not sure how to change this to a native element. This also doesn't make sense as the source of the failing tests as this file hasn't been changed for a long time. So long story short, I am a bit lost again, but will have a look again tomorrow -hopefully with fresh eyes

majaokholm:
@jalkire is there a way to run the web-e2e-reliability test locally?

jalkire:
> @jalkire is there a way to run the web-e2e-reliability test locally?

That just runs the normal e2e tests repeatedly via `yarn test:e2e --ci`--as far as I can tell, it's failing for reasons unrelated to your change. Just so long as the CI / e2e-web-tests build passes, you don't have to worry about it 🙂

src/popup/Styles/popup.scss
jalkire: Thank you for looking into that! For the report e2e test, I'm still able to get that working with `yarn build:package:report` and then `yarn test:report:e2e -- -- -- -u`, and the snapshot updates look reasonable to me. 

For the remaining role=heading `div`, now that you mention it, I'm seeing a few other role=heading based headings that were a bit trickier to find in the code. Sorry for not finding those earlier! I've found [one with the props defined elsewhere](https://github.com/microsoft/accessibility-insights-web/blob/63a0325b99debe7f2a81aa8294f3b78efb342b87/src/common/components/cards/collapsible-component-cards.tsx#L79) (probably the one you're seeing that gets headingLevel from rules-with-instance.tsx), and then a few using `span` rather than `div`: [1](https://github.com/microsoft/accessibility-insights-web/blob/63a0325b99debe7f2a81aa8294f3b78efb342b87/src/DetailsView/components/adhoc-tab-stops-test-view.tsx#L166), [2](https://github.com/microsoft/accessibility-insights-web/blob/63a0325b99debe7f2a81aa8294f3b78efb342b87/src/DetailsView/components/requirement-instructions.tsx#L18), [3](https://github.com/microsoft/accessibility-insights-web/blob/63a0325b99debe7f2a81aa8294f3b78efb342b87/src/electron/views/tab-stops/tab-stops-testing-content.tsx#L47).

Given how much you've already put into this already, it would be totally fair to wrap up this PR without adding any of those, or just addressing that last `div` heading. Whatever you'd like to do!

majaokholm: I am looking into it and will get back - not sure if I will be able to fix it all so might make sense to scope some out but will have a closer look first

majaokholm:
I am afraid I will not be able to get any further with this. I fixed the collapsible-component-cards.tsx but I am still think there are a few more somewhere that are better hidden, but I have not been able to get further. 
All the tests worked when running locally but it seems that the tests are blocked because the web-e2e-reliability check is failing. It looks like this test is some kind of cross platform test so makes sense that can’t be run locally. I would love to get the last test to work so we can close the PR here but I don’t know what to do to get the web-e2e-reliability check to work. Do you have any advice here?
Thanks again :-)

jalkire:
> I am afraid I will not be able to get any further with this. I fixed the collapsible-component-cards.tsx but I am still think there are a few more somewhere that are better hidden, but I have not been able to get further. All the tests worked when running locally but it seems that the tests are blocked because the web-e2e-reliability check is failing. It looks like this test is some kind of cross platform test so makes sense that can’t be run locally. I would love to get the last test to work so we can close the PR here but I don’t know what to do to get the web-e2e-reliability check to work. Do you have any advice here? Thanks again :-)

Thanks for getting that component fixed up! The issues with the e2e-reliability tests are not actually related to your changes, so you should be good to go here 🙂

jalkire (approved):
Thank you for all the time that you put into this contribution!

PR #5201, opened 2022-02-23
feat(tabstops-auto-target-page-vis): Add and populate tabbed item type for automated tabbing results

sfoslund:
#### Details

Add the `itemType` property to `TabStopVisualizationInstance` and `TabbedItem`. This property differentiates between the different types of elements/failures that are passed to the svg-drawer to visualize and is populated based off the content of the `requirementResults`. 

##### Motivation

Feature work.

##### Context

A future PR will take advantage of the new itemType property in the svg-drawer for tab stops failure visualization. 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/visualization/get-visualization-instances-for-tab-stops.ts
+++ b/src/injected/visualization/get-visualization-instances-for-tab-stops.ts
@@ -45,6 +45,10 @@ export const GetVisualizationInstancesForTabStops = (
                 selectorToVisualizationInstanceMap[selector].requirementResults[requirementId] = {
                     instanceId: instance.id,
                 };
+                selectorToVisualizationInstanceMap[selector].itemType =
+                    requirementId === 'keyboard-navigation'
+                        ? TabbedItemType.MissingItem
+                        : TabbedItemType.ErroredItem;
                 return;
             }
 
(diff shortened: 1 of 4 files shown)


src/injected/visualization/tabbed-item.ts
```diff
@@ -8,4 +8,11 @@ export interface TabbedItem {
     focusIndicator?: FocusIndicator;
     tabOrder: number;
     shouldRedraw?: boolean;
+    itemType?: TabbedItemType;
+}
+
+export enum TabbedItemType {
+    RegularItem,
```
waabid, line 15: I think we don't need the RegularItem: many tab stop instances (in the Assessment case) will be created without one such that specifying one only for this case doesn't help.

src/injected/visualization/tabbed-item.ts
```diff
@@ -8,4 +8,11 @@ export interface TabbedItem {
     focusIndicator?: FocusIndicator;
     tabOrder: number;
     shouldRedraw?: boolean;
+    itemType?: TabbedItemType;
+}
+
+export enum TabbedItemType {
```
waabid, line 14: On the use of enums: we try not to use enums as the below example because their underlying values are basically dependent on position in the enum type. That is to say MissingItem is 1 and ErroredItem is 2 but if we switch the ordering it changes.

So there's two options:
1. Use enums but specify the string, like so: `MissingItem = "MissingItem"`.
2. Don't use enums but use types. I think this is the preferred approach now because it is easier to reference when using them. For example, (enum) `someType === TabbedItemType.MissingItem` vs (type) `someType === "MissingItem"`.

src/injected/visualization/get-visualization-instances-for-tab-stops.ts
```diff
@@ -64,5 +65,23 @@ export const GetVisualizationInstancesForTabStops = (
         });
     });
 
+    forOwn(
```
waabid, line 68: Why not just do this when we create the instances in the previous loop?

Thinking about this now ... I was wondering what is the hierarchy for the selected TabbedItemType. I think we don't need to care about it _for now_ because as long as we show an error, it is fine (and the scenario where we have a missing tabbed issue AND another error would need to be manually manufactured).

sfoslund: Yeah, I was thinking about the same thing. I arbitrarily chose to have the missing item config take precedence over errored, but that can easily be changed.

src/injected/visualization/get-visualization-instances-for-tab-stops.ts
```diff
@@ -44,6 +45,10 @@ export const GetVisualizationInstancesForTabStops = (
                 selectorToVisualizationInstanceMap[selector].requirementResults[requirementId] = {
                     instanceId: instance.id,
                 };
+                selectorToVisualizationInstanceMap[selector].itemType =
```
waabid, line 48: nit: you can likely do this on line 42 and just use it here and on line 66.

waabid (approved):
LGTM, minor comment.

PR #5199, opened 2022-02-23
feat(tabstops-auto-target-page-vis): Add errored and missing tab stops visualizations

sfoslund:
#### Details

Add svg-drawer logic to display different types of focus indicators depending on the item type (regular, missing, or errored). The regular circle is the same as the existing focus indicator and the missing/ errored circles are added in this PR and shown below. Note that the screenshots below are generated when testing and these visualizations are not yet included in any production visualizations.

Missing circle (screenshot of red circle with dotted line, red 'x' in the center, and red '!' in the top right corner):
MissingCircle

Errored circle (screenshot of red circle with solid line, red '1' in the center, and red '!' in the top right corner):
ErrorCircle

##### Motivation

Feature work.

##### Context

At the moment the itemType property is not populated. This means that it will always be undefined and as a result we will have the same behavior as before (createErrorCircle will not be called). A follow up PR will add logic to populate the itemType property. 

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [x] (UI changes only) Added screenshots/GIFs to description above
- [ not yet included in real UI ] (UI changes only) Verified usability with NVDA/JAWS

PR #5213, opened 2022-02-25
feat(tabstops-auto-target-page-vis): SVG drawer draws visualizations for failure instances

waabid:
#### Details

SVG drawer will now use data coming in to draw different kinds of errors for tab stops.

##### Motivation

feature work.

##### Context

This work is still incomplete: there's some undesired behavior on when lines are connecting elements and when they are not. That will be tinkered with in a future PR.

Screen grab of what the visualizations look like below:
image

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/injected/visualization/svg-drawer.ts
```diff
@@ -228,6 +236,7 @@ export class SVGDrawer extends BaseDrawer {
     }
 
     private createLinesInTabOrderVisualization(
+        items: TabbedItem[],
```
lisli1, line 239: Are we adding this parameter to be clearer about which TabbedItem array we are using?

src/injected/visualization/svg-drawer.ts
```diff
@@ -265,11 +274,44 @@ export class SVGDrawer extends BaseDrawer {
...
+
+        const circleConfiguration =
+            item.itemType === TabbedItemType.ErroredItem
+                ? drawerConfig.erroredCircle
+                : drawerConfig.missingCircle;
+
+        const newCircle = this.svgShapeFactory.createCircle(centerPosition, circleConfiguration);
+        const newLabel = this.svgShapeFactory.createTabIndexLabel(
+            centerPosition,
+            drawerConfig.tabIndexLabel,
```
sfoslund, line 299: This doesn't necessarily need to be done in this PR, but eventually we will want to use a different config for error labels since we will want the text on them to be red. See the erroredTabIndexLabel config I added here: https://github.com/microsoft/accessibility-insights-web/pull/5199/files#diff-00053e82d4b379e30958fd1ca42842a8fadbd20d903faaa76bc26e1397656017R66

src/injected/visualization/svg-drawer.ts
```diff
@@ -265,11 +274,44 @@ export class SVGDrawer extends BaseDrawer {
...
+            item.tabOrder,
+        );
+
+        const focusIndicator: FocusIndicator = {
+            circle: newCircle,
+            tabIndexLabel: newLabel,
```
sfoslund, line 305: Again, doesn't have to be in this PR but we will want a failure label here too: https://github.com/microsoft/accessibility-insights-web/pull/5199/files#diff-da79de6a20c6af768f69e58a29187f44ccf98e439534ee29d54dffcf8265f27cR306-R316

src/tests/unit/tests/injected/visualization/svg-drawer.test.ts
```diff
@@ -55,16 +55,39 @@ describe('SVGDrawer', () => {
     }
 
     test('initialize', () => {
-        fakeDocument.body.innerHTML = "<div id='id1'></div>";
```
sfoslund, line 58: It seems like we're mostly updating existing tests here. Are we going to add test coverage for the new logic?

src/tests/unit/tests/injected/visualization/svg-drawer.test.ts
```diff
@@ -701,14 +767,16 @@ describe('SVGDrawer', () => {
         drawerUtilsMock.verifyAll();
 
         expect(circles.length).toBe(2);
-        expect(lines.length).toBe(1);
+        expect(lines.length).toBe(0);
```
lisli1, line 770: Confused why we expect 0 lines when the test says "with solid line"?

lisli1: Is this related to the undesired behavior mentioned in the description?

src/tests/unit/tests/injected/visualization/svg-drawer.test.ts
```diff
@@ -772,9 +858,9 @@ describe('SVGDrawer', () => {
 
         drawerUtilsMock.verifyAll();
 
-        expect(circles.length).toBe(2);
-        expect(lines.length).toBe(0);
-        expect(labels.length).toBe(1);
+        expect(circles.length).toBe(3);
```
lisli1, line 861: Does this count only tabbedElements that have `isFailure: true`?

src/injected/visualization/svg-drawer.ts
waabid: Yeah, technically this doesn't need to happen in this PR but I have a refactor that will make use of the fact that it is agnostic to which array is used.

src/injected/visualization/svg-drawer.ts
waabid: Good catch! I'll look into it!

src/tests/unit/tests/injected/visualization/svg-drawer.test.ts
waabid: It's hard to parse because of how git shows the diffs but I removed a test (test was not necessary) and added a new test that contains failures and such specifically. All new code is covered (can be verified using `yarn watch:test --collect-coverage` and making a line deletion change to the file to cause it to run tests). There's some code in main that is not tested but not pertaining to this PR.

src/tests/unit/tests/injected/visualization/svg-drawer.test.ts
waabid: No, these are all circles created.

In general: my personal opinion of this is that testing in this file is lackluster but since this needs a significant refactor and improving tests without that refactor would be work that would be undone/redone when refactored.

src/tests/unit/tests/injected/visualization/svg-drawer.test.ts
waabid: I need to check on this again but this seems to be a part of the weird diff behavior: I didn't modify this test case but it makes it seem like I did because of the test deletion.

src/injected/visualization/svg-drawer.ts
waabid: Added!

src/tests/unit/tests/injected/visualization/svg-drawer.test.ts
```diff
@@ -1077,6 +1084,11 @@ describe('SVGDrawer', () => {
         return labels;
     }
 
+    function findFailureLabels(): NodeListOf<Element> {
+        const circles = shadowContainer.querySelectorAll('.insights-svg-failure-label');
```
lisli1, line 1088: nit: should this be called labels instead of circles?

lisli1 (approved):
Small comment but LGTM

sfoslund (approved):

PR #5202, opened 2022-02-23
fix(tabstops-auto-target-page-vis): keyboard navigation errors not adding to store

lisli1:
#### Details

Automatically detected keyboard navigation failure instances were not being added to the visualization-scan-result-store because the needToCollectTabbingResults flag was set to false before teardown. This PR sets the flag after teardown, so the keyboard navigation instances are added to the store and show up in the failure instance list.

##### Motivation

Part of feature work

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [ ] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #5219, opened 2022-03-01
feat(tabstops-auto-target-page-vis): Add TabStopsRequirementResultProcessor

lisli1:
#### Details

Create TabStopsRequirementResultProcessor to take some of the responsibility of the result processing away from the TabStopsAnalyzer and AnalyzerController.

##### Motivation

Part of feature work.

##### Context

AnalyzerController included a function that was used only in the TabStops case and was calling teardown() when it should not have. TabStopsAnalyzer also had too much responsibility in processing results and managing the requirement runner. This PR moves these functionalities to a separate class, TabStopsRequirementResultProcessor, to fix these issues. This also fixes an issue that #5202 tried to address, in which the keyboard navigation errors weren't being added to the store/failure instances list.

#### Pull request checklist

- [x] Addresses an existing issue: feat 1872858
- [x] Ran `yarn fastpass`


src/injected/main-window-initializer.ts
```diff
@@ -328,11 +329,26 @@ export class MainWindowInitializer extends WindowInitializer {
...
+            this.visualizationScanResultStoreProxy,
+        );
+
+        const tabStopsRequirementResultProcessorForFocusTracking =
+            new TabStopsRequirementResultProcessor(
+                this.featureFlagStoreProxy,
+                null,
+                tabStopRequirementActionMessageCreator,
+                this.visualizationScanResultStoreProxy,
+            );
```
lisli1, line 345: At the moment, can't think of a way to initialize just one TabStopsRequirementResultProcessor with a tabStopRequirementRunner or null to pass into AnalyzerProvider.

sfoslund: Maybe we can always have tabStopRequirementRunner be non-null, but pass something in start() to differentiate between the case when we want to use it and when we don't?

src/injected/analyzers/tab-stops-requirement-result-processor.ts
```diff
@@ -0,0 +1,92 @@
...
+import { BaseStore } from 'common/base-store';
+import { FeatureFlags } from 'common/feature-flags';
+import { FeatureFlagStoreData } from 'common/types/store-data/feature-flag-store-data';
+import { VisualizationScanResultData } from 'common/types/store-data/visualization-scan-result-data';
+import { TabStopRequirementActionMessageCreator } from 'DetailsView/actions/tab-stop-requirement-action-message-creator';
+import { AllFrameRunner } from 'injected/all-frame-runner';
+import { AutomatedTabStopRequirementResult } from 'injected/tab-stop-requirement-result';
+import { isEqual } from 'lodash';
+
+export class TabStopsRequirementResultProcessor {
```
sfoslund, line 13: I like this approach overall, nice to be able to separate out this logic 👍

sfoslund (approved):
Looks good to me overall! If possible, I'd love to only have to initialize one tabStopsRequirementResultProcessor but I don't see that as blocking.

src/injected/analyzers/tab-stops-analyzer.ts
```diff
@@ -50,37 +47,14 @@ export class TabStopsAnalyzer extends BaseAnalyzer {
...
-        if (
-            this.tabStopRequirementRunner &&
-            this.featureFlagStore.getState()[FeatureFlags.tabStopsAutomation] === true
-        ) {
-            this.seenTabStopRequirementResults = [];
+        if (this.featureFlagStore.getState()[FeatureFlags.tabStopsAutomation] === true) {
             this.tabStopsDoneAnalyzingTracker.reset();
-            this.tabStopRequirementRunner.topWindowCallback = this.processTabStopRequirementResults;
-            this.tabStopRequirementRunner.start();
+            this.tabStopsRequirementResultProcessor.start(this.needsRequirementRunner);
```
waabid, line 52: Why do this instead of the previous approach of checking for the null runner?

src/injected/analyzers/tab-stops-requirement-result-processor.ts
```diff
@@ -0,0 +1,95 @@
...
+        if (this.isStopped) {
+            return;
+        }
+
+        if (this.needsRequirementRunner) {
+            this.tabStopRequirementRunner.stop();
+        }
+
+        const state = this.visualizationResultsStore.getState();
+        if (state.tabStops.tabbingCompleted && state.tabStops.needToCollectTabbingResults) {
```
waabid, line 67: What's the scenario where this condition is not true? Or specifically, a scenario where we would be calling stop but would not want us to make these calls to the action message creator.

src/injected/analyzers/tab-stops-requirement-result-processor.ts
```diff
@@ -0,0 +1,95 @@
...
+            this.tabStopRequirementRunner.topWindowCallback = this.processTabStopRequirementResults;
+            this.tabStopRequirementRunner.start();
+        }
+
+        this.isStopped = false;
+    };
+
+    public listenToStore(): void {
+        this.visualizationResultsStore.addChangedListener(this.onStateChange);
+        this.featureFlagStore.addChangedListener(this.onStateChange);
```
waabid, line 46: Since the usage of this class is under a feature flag in the analyzer, I don't think we need to check the feature flag (or listen to the store). Unless I'm missing something?

src/injected/analyzers/tab-stops-analyzer.ts
lisli1: I didn't really like having to create two `TabStopsRequirementResultProcessor`s in the MainWindowInitializer and passing them both into the AnalyzerProvider but if there's a better way to do it, I can change it

src/tests/unit/tests/injected/analyzers/tab-stops-requirement-result-processor.test.ts
```diff
@@ -0,0 +1,329 @@
...
+
+describe('TabStopsRequirementResultProcessor', () => {
+    let tabStopRequirementRunnerMock: IMock<AllFrameRunner<AutomatedTabStopRequirementResult>>;
+    let requirementResultRunnerCallback: (
+        requirementResult: AutomatedTabStopRequirementResult,
+    ) => void;
+    let featureFlagStoreMock: IMock<FeatureFlagStore>;
+    let tabStopRequirementActionMessageCreatorMock: IMock<TabStopRequirementActionMessageCreator>;
+    let visualizationScanResultsStoreMock: IMock<VisualizationScanResultStore>;
+    let needsRequirementRunner;
```
waabid, line 23: Missing type.

src/injected/analyzers/tab-stops-analyzer.ts
waabid: I think another approach might be a wrapper/factory around the creation of it that the analyzer-provider can use.

waabid: But I'm still not sure I understand why we can't use the null approach from before: I think we only need the `TabStopsRequirementResultProcessor` in TabStopsAnalyzer (and not in FocusTrackingAnalyzer).

src/injected/analyzers/tab-stops-requirement-result-processor.ts
lisli1: We want to send telemetry only after the first time the user finished tabbing and then stop collecting tabbing results

src/injected/analyzers/tab-stops-requirement-result-processor.ts
```diff
@@ -0,0 +1,75 @@
...
+
+        this.seenTabStopRequirementResults = [];
+        this.tabStopRequirementRunner.topWindowCallback = this.processTabStopRequirementResults;
+        this.tabStopRequirementRunner.start();
+
+        this.isStopped = false;
+    };
+
+    public listenToStore(): void {
+        this.visualizationResultsStore.addChangedListener(this.onStateChange);
```
waabid, line 34: The current implementation means we end up listening to all visualization store changes but don't think we need to do that.

I think we shouldn't be listening to the store unless the class has been _started_. I think moving the `addChangedListener` to `start` seems like a good idea (and similarly adding a `removeChangedListener` in `stop`).

waabid (approved):

PR #5242, opened 2022-03-07
feat(tabstops-auto-target-page-vis): use instance id instead of selectors for tab stops

waabid:
#### Details

Instead of using selectors as the key for tab stop instances, this uses instance ids.

##### Motivation

feature work

##### Context

We want to have more than one tab value pertaining to the same element (since elements can be re-visited). If we want to do that, we cannot use a unique selector of the element as the identifier in the result dictionary (lest we make greater modifications to code in the SVG drawer to handle multiple tab stops per element). Using a unique-id to do that gives the desired experience.

As such, the changes here pertain to newer code in the data-processing layer (and attaching instance id to tab events) that would likely not (and could not) impact other scenarios that use the SVG drawer.

I smoke tested the bugs we've found recently around this feature as well to ensure there were no regressions there.

One regression we do have as a result of this change however is this: the tab values for errors won't be visible/displayed with the error circles. To see these, I've made them transparent.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/visualization/tab-stops-formatter.ts
+++ b/src/injected/visualization/tab-stops-formatter.ts
@@ -46,7 +46,7 @@ export class TabStopsFormatter implements Formatter {
             erroredCircle: {
                 stroke: '#E81123',
                 strokeWidth: '2',
-                fill: 'transparent',
+                fill: '#ffffff',
                 ellipseRy: '16',
                 ellipseRx: ellipseRx.toString(),
             },
--- a/src/injected/visualization/tab-stops-formatter.ts
+++ b/src/injected/visualization/tab-stops-formatter.ts
@@ -46,7 +46,7 @@ export class TabStopsFormatter implements Formatter {
             erroredCircle: {
                 stroke: '#E81123',
                 strokeWidth: '2',
-                fill: '#ffffff',
+                fill: 'transparent',
                 ellipseRy: '16',
                 ellipseRx: ellipseRx.toString(),
             },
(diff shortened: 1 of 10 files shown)


src/injected/visualization/tab-stops-formatter.ts
```diff
@@ -46,7 +46,7 @@ export class TabStopsFormatter implements Formatter {
             erroredCircle: {
                 stroke: '#E81123',
                 strokeWidth: '2',
-                fill: '#ffffff',
+                fill: 'transparent',
```
sfoslund, line 49: Have we checked with Fer if displaying these circles as transparent is okay, as it doesn't match the figma? 

Overall, this approach seems like it can get us into some weird states. For example, when smoke testing for a minute I noticed that at times we can get into a state where the focus circle has a tab index in it and where other tab index labels disappear suddenly. See the screenshot below, in which focus is on item 4 and, prior to the last tab, both of the error circles had tab index labels which have now disappeared. 
[Screenshot of 4 tabbed elements, 2 red error circles, one focus circle, and one regular circle.]
tabstopstesting

waabid: Discussed offline.

sfoslund (approved):
After discussing offline looks good to me.

src/tests/end-to-end/tests/details-view/automated-tabstops.test.ts
```diff
@@ -85,7 +85,9 @@ describe('Automated TabStops Results', () => {
 
             expect(ruleDetails).toHaveLength(1);
 
-            await verifyTargetPageVisualization(2, 1, 1, 0);
+            // The keyboard trap is triggered by an element re-receiving focus progromattically.
```
lisli1, line 88: nit:
```suggestion
            // The keyboard trap is triggered by an element re-receiving focus programmatically.
```

PR #5259, opened 2022-03-10
chore: make snapshots of typemoq mocks less brittle

dbjorge:
#### Details

This PR implements and configures a new `typemoq-snapshot-serializer`, a Jest Snapshot Serializer which causes typemoq mock objects that get captured by test snapshots to be serialized as `[typemoq mock object]` instead of as a big blob of all the objects' properties.

Example before:

```
      Object {
        "deps": Object {
          "storesHub": proxy {
            "___id": "BCDF5CE5-F0DF-40B7-8BA0-69DF395033C8",
            "stores": undefined,
          },
        },
      }
```

Same example after:

```
      Object {
        "deps": Object {
          "storesHub": [typemoq mock object],
        },
      }
```

##### Motivation

1. This makes tests less brittle. In general, we don't want to pin the individual properties of mocked objects in snapshots; from the perspective of the test performing the snapshot, they're implementation details (otherwise it wouldn't be mocked!), so we shouldn't be validating them directly
2. This simplifies work that @katydecorah and I are looking at to upgrade our `@swc/core` dependency (that upgrade would otherwise change how these proxy objects get serialized in a way which is an implementation detail to all these tests, but which would have required we update all these snapshots anyway)
3. Works around a long-standing issue where snapshotting the results of `enzyme.shallow(<Component prop={someMock}/>).getElement()` would fail with a confusing looking `MockException` (the one that the new snapshot serializer catches when trying to detect a mock function)

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -41,6 +41,7 @@ module.exports = {
         ],
     ],
     setupFilesAfterEnv: [`${__dirname}/src/tests/common/flush-promises-after-each-test.ts`],
+    snapshotSerializers: [`${__dirname}/src/tests/common/typemoq-snapshot-serializer.ts`],
     testEnvironment: 'node',
     testMatch: ['**/*.spec.[tj]s', '**/*.test.[tj]s'],
     testPathIgnorePatterns: ['/dist/', '/out/'],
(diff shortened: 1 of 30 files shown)


katydecorah (approved):
Looks good! I also appreciate how the serializer improves the readability of these snapshots.

PR #5265, opened 2022-03-11
feat(tabstops-auto-target-page-vis): Add 'don't show again' checkbox to auto detected failures dialog

sfoslund:
#### Details

Add `Don't show again` checkbox to new auto detected failures dialog shown on tab stops automated failures. Checking this box ensures that the auto detected failures dialog will not pop up on future runs. This PR also adds a new setting to specify whether to show the dialog or not in the settings tab.

##### Motivation

Address stretch goal not completed in feature.

##### Context

By default, the `Don't show again` checkbox is unchecked and the `Enable automatically detected failures dialog` setting is turned on.

[Screenshot of existing auto detected failures dialog, now with a checkbox labeled `Don't show again`.]
DontShowAgainCheck

[Screenshot of new toggle in settings bar labeled `Enable automatically detected failures dialog`.]
EnableDiagSetting

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/adhoc-tab-stops-test-view.tsx
```diff
@@ -43,6 +44,7 @@ import * as Markup from '../../assessments/markup';
 
 export type AdhocTabStopsTestViewDeps = {
     detailsViewActionMessageCreator: DetailsViewActionMessageCreator;
+    userConfigMessageCreator: UserConfigMessageCreator;
```
waabid, line 47: Would be better to cvreate/add this to the deps for AutoDetectedFailuresDialog rather than being passed in as a prop.

src/DetailsView/components/auto-detected-failures-dialog.tsx
```diff
@@ -39,7 +53,14 @@ export class AutoDetectedFailuresDialog extends React.Component<
             Object.entries(this.props.visualizationScanResultData.tabStops.requirements).some(
                 ([_, data]) => data.instances.length > 0 && data.status === 'fail',
             );
-        if (tabbingJustFinished && autoDetectedFailuresExist && !prevState.dialogEnabled) {
+        const dialogEnabled = this.props.userConfigurationStoreData.showAutoDetectedFailuresDialog;
```
waabid, line 56: nit: the naming here is a bit off since it collides with the `dialogEnabled` state variable, which refers to something else. I think either renaming this to `showAgainSetting` (or similar) OR to rename the other state variable to something else might be best.

src/tests/unit/tests/DetailsView/components/auto-detected-failures-dialog.test.tsx
```diff
@@ -70,6 +83,16 @@ describe('AutoDetectedFailuresDialog', () => {
             wrapper.find(Dialog).prop('onDismiss')();
             expect(wrapper.getElement()).toMatchSnapshot();
         });
+
+        it('box appears checked when "dont show again" box is clicked', () => {
+            userConfigMessageCreatorMock
+                .setup(ucmcm => ucmcm.setAutoDetectedFailuresDialogState(true))
+                .verifiable(Times.once());
+
+            wrapper.find(Checkbox).simulate('change', undefined, true);
```
waabid, line 92: nit: should probably have a test for undefined `checked` as well.

src/DetailsView/components/auto-detected-failures-dialog.tsx
lisli1: +1 on this

sfoslund:
Fer has approved this UI and asked for there to be telemetry on the checkbox being set, so I'll update this PR with that change and PR feedback in coming days.

sfoslund:
@waabid @lisli1 this should be ready for re-review whenever you have time

src/common/telemetry-data-factory.ts
```diff
@@ -499,4 +500,16 @@ export class TelemetryDataFactory {
...
+    public forSetAutoDetectedFailuresDialogState(
+        enabled: boolean,
+    ): AutoDetectedFailuresDialogStateTelemetryData | undefined {
+        if (enabled === undefined) {
+            return undefined;
+        }
+
+        return {
+            enabled,
+        };
```
lisli1, line 513: What is the reason for explicitly checking if `enabled` is undefined if we are returning the value of `enabled` in any case?

lisli1 (approved):
Minor comment but LGTM

src/common/telemetry-data-factory.ts
sfoslund: My understanding is that this way (returning undefined) we won't send telemetry when enabled is undefined, which is essentially an error state that we don't expect to get to. If we return `{ enabled: undefined }` here then that gets sent as telemetry, right? But this is pretty useless telemetry, so I figure we might as well not send it.

PR #5284, opened 2022-03-18
refactor: refactor SVG drawer focus indicator creation to FocusIndicatorCreator

waabid:
#### Details

Felt like a good break-point between objects. Note: does not use the FocusIndicatorCreator in this class. 

##### Motivation

SVG drawer is very large and needs to be a bit more manageable.

##### Context

**Reviewing by commit** might be easier since I copied out the methods first, wrote the tests, and then tried to make it more readable (and thus ensuring if tests are passing, that the logic hasn't changed).

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/injected/visualization/focus-indicator-creator.ts
```diff
@@ -21,110 +22,125 @@ export class FocusIndicatorCreator {
...
-        const newLabel =
-            isLastItem || !showTabIndexedLabel
-                ? null
-                : this.svgShapeFactory.createTabIndexLabel(
-                      centerPosition,
-                      drawerConfig.tabIndexLabel,
-                      item.tabOrder.toString(),
-                  );
+        if (isLastItem) {
+            return this.createIndicatorWithFocusedConfigurations(
```
sfoslund, line 44: nit: just for readability I'd prefer to add an else here and move the rest of the logic to another method as well (maybe `createIndicatorWithNormalConfig`? Mostly just a preference thing though.

src/injected/visualization/focus-indicator-creator.ts
```diff
@@ -159,10 +175,7 @@ export class FocusIndicatorCreator {
         return focusIndicator;
     };
 
-    private shouldBreakGraph(items: TabbedItem[], curElementIndex: number): boolean {
-        return (
-            curElementIndex === 0 ||
-            items[curElementIndex - 1].tabOrder !== items[curElementIndex].tabOrder - 1
-        );
+    private shouldBreakGraph(item: TabbedItem, prevItem: TabbedItem): boolean {
+        return prevItem == null || prevItem.tabOrder !== item.tabOrder - 1;
```
sfoslund, line 179: This change looks great to me. I seem to remember there being quite a few places where we referenced the order of the items in the array instead of the items' tab order (for example, the old implementation of this method). Have all of those instances been removed with this change?

sfoslund (approved):
LGTM!

src/injected/visualization/focus-indicator-creator.ts
waabid: Good point! I was thinking that too. I'll add it!

src/injected/visualization/focus-indicator-creator.ts
waabid: Somewhat. A fair bit of those were around the `shouldRedraw` method and those are simplified but not totally gone.

Issue #4778, opened 2021-10-06
Custom widgets test crashes details view in Dev and Canary extensions


jalkire:
## Describe the bug
In Canary and Dev extensions (both built with [devConfig in webpack](https://github.com/jalkire/accessibility-insights-web/blob/main/webpack.config.js#L136-L153)), the Custom widgets test causes details view to crash due to a circular reference error. This issue does not impact Prod/Insider builds.

**To Reproduce**

1. Open the Canary extension
2. Go to Custom widgets -> Design pattern and allow it to finish finding instances
3. Go to Custom widgets -> Instructions and allow it to finish finding instances
4. Details view turns blank 

## Expected behavior
Details view displays content 

## Context (please complete the following information)

- OS Name & Version: Windows 10 19043.1237
- AI-Web Version & Environment: AI-Web Canary 2021.10.6.2138 (I also reproed this with locally built dev extensions all the way back through the latest prod release, 2.28.0)
- Browser Version: Edge Version 94.0.992.38 (Official build) (64-bit)
- Target Page: Any

## Additional context
Details view console output:
```
assessment-data-formatter.ts?d527:10 
        
       Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'HTMLDivElement'
    |     property '__reactInternalInstance$28r42fudfk8' -> object with constructor 'FiberNode'
    --- property 'stateNode' closes the circle
    at JSON.stringify (<anonymous>)
    at AssessmentDataFormatter.formatAssessmentData (assessment-data-formatter.ts?d527:10)
    at Object.getSaveButtonForAssessment [as SaveAssessmentButton] (save-assessment-button-factory.tsx?dfba:28)
    at DetailsViewCommandBar.renderSaveAssessmentButton (details-view-command-bar.tsx?ff17:239)
    at DetailsViewCommandBar.renderCommandButtons (details-view-command-bar.tsx?ff17:163)
    at DetailsViewCommandBar.renderFarItems (details-view-command-bar.tsx?ff17:156)
    at DetailsViewCommandBar.render (details-view-command-bar.tsx?ff17:124)
    at finishClassComponent (react-dom.development.js?61bb:17160)
    at updateClassComponent (react-dom.development.js?61bb:17110)
    at beginWork (react-dom.development.js?61bb:18620)
AssessmentDataFormatter.formatAssessmentData @ assessment-data-formatter.ts?d527:10
[...]

PR #5289, opened 2022-03-23
fix: Fix custom widget crash json cycle bug

sfoslund:
#### Details

Fix bug in which custom widgets test would cause the details view to crash following a circular reference error. 

##### Motivation

Addresses issue #4778

##### Context

Alternative approaches could be to either (1) ensure that there are no cycles in the structure to begin with or to (2) use an algorithm to search and remove the whole structure for cycles. I didn't go with the first option because this would require possibly significant changes to the way data is stored in the assessment-store, which would have widespread effects. I didn't go with the second option because searching the whole structure is complicated and could be a performance hit. Instead, I added logic to ignore one of the fields in the structure which sometimes creates a cycle when stringifying into JSON. The `_owner` field was not shown anyways, so this has no impact on the resulting json. The trade off with this approach is that it will only catch this specific type of cycle, and we might have the same bug in the future if another cycle is introduced elsewhere in the data structure. 

#### Pull request checklist

- [x] Addresses an existing issue: #4778
- [x] Ran `yarn fastpass`


sfoslund:
@jalkire this PR fixes a bug you filed a while ago. If you remember the context, do you mind reviewing whenever you get a chance?

jalkire (approved):
👍

Issue #5305, opened 2022-03-28
Tabstops automation with modal dialogs


ferBonnin:
## Describe the bug

If there is a modal dialog and you run Tabstops, all elements behind the modal dialog will be marked as failures

**To Reproduce**

1. Go to https://www.washington.edu/accesscomputing/AU/after.html
2. Click on the cheat sheet of a11y issues
3. Run tabstops over the modal dialog
4. See that all other elements in the page are marked as potential failures

## Expected behavior
Only the interactable elements should be marked as potential failures

## Screenshots

![a modal dialog with tabstops, all the elements on the back are marked as potential failures](https://user-images.githubusercontent.com/9062104/160463419-a19f91ee-c4e4-4d8b-af47-9c763dacfeff.png)

ferBonnin:
triaged with Peter. Even though issues are added as potential / needs review, this could confuse users and make them not keep the focus in the modal only

sfoslund:
I think the root cause of this is an oddity in tabbable- see its [readme](https://github.com/focus-trap/tabbable#more-details) which states: `Only elements hidden with display: none or visibility: hidden are left out`. It looks like it ignores `aria-hidden=true` and includes those elements in its list of tabbable elements, which causes us to mark them as missing when the user is unable to tab to them.

@waabid were you aware of this behavior previously? Do you think it's something we can file a bug on them for or something we'll have to adapt to from our end?

waabid:
Unfortunately, this is news to me as well. I guess we'll have to self-filter after getting the list of elements. Filtering for components that are under aria-hidden (and aria-modal, per WCAG 1.1) should be doable. Still, I do think there might be other modals (usually ones that are not accessible per WCAG since my understanding is one of aria-hidden or aria-modal attributes are required) that exist that can cause this issue.

Wonder if it makes sense to maybe add a note for this scenario?

sfoslund:
@waabid makes sense, I made a PR to do that filtering: https://github.com/microsoft/accessibility-insights-web/pull/5308

sfoslund:
[...]

PR #5308, opened 2022-03-30
fix: Ignore hidden elements when determining tabbable elements

sfoslund:
#### Details

Fix automated tab stops bug so that elements behind an open modal are not marked as missing failures. 

[Screenshot of site with modal open and no missing tab stops marked in background]
ModalWithoutMissingTabs

##### Motivation

Addresses issue #5305.

##### Context

Tabbable does not take `aria-hidden` or `aria-modal` into account when determining which elements are tabbable, meaning before this change we were marking them as missing since the user can't tab to them. This PR adds logic to filter out these elements from the tabbable results so they are not considered when determining missing elements.

#### Pull request checklist

- [x] Addresses an existing issue: #5305
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/tabbable-element-getter.ts
+++ b/src/injected/tabbable-element-getter.ts
@@ -23,11 +23,11 @@ export class TabbableElementGetter {
     ) => {
         const hiddenSelectors = ['div[aria-hidden="true"]', 'div[aria-modal="true"]'];
         hiddenSelectors.forEach(selector => {
-            const hiddenElements = this.doc.querySelectorAll<FocusableElement>(
-                `${selector}, ${selector} *`,
-            );
+            const hiddenElements = this.doc.querySelectorAll(selector);
             if (hiddenElements) {
-                elements = elements.filter(e => !Array.from(hiddenElements).includes(e));
+                elements = elements.filter(
+                    e => !Array.from(hiddenElements).some(hidden => hidden.contains(e)),
+                );
             }
         });
         return elements;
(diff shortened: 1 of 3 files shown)


src/injected/tabbable-element-getter.ts
```diff
@@ -18,14 +18,31 @@ export class TabbableElementGetter {
     ) {}
 
     public get: () => TabbableElementInfo[] = () => {
-        return this.getTabbableElements(this.doc.documentElement).map((elem, index) => ({
+        return this.getRawElements().map((elem, index) => ({
```
waabid, line 21: I think we don't use this API so we can probably get rid of it.

src/injected/tabbable-element-getter.ts
```diff
@@ -18,14 +18,31 @@ export class TabbableElementGetter {
...
+        const filteredElements = this.filterHiddenElements(tabbableElements);
+        return filteredElements;
+    };
+
+    private filterHiddenElements: (elements: FocusableElement[]) => FocusableElement[] = (
+        elements: FocusableElement[],
+    ) => {
+        const hiddenSelectors = ['div[aria-hidden="true"]', 'div[aria-modal="true"]'];
+        hiddenSelectors.forEach(selector => {
+            const hiddenElements = this.doc.querySelectorAll<FocusableElement>(
```
waabid, line 39: I looked into this briefly and seems like there's a native way (node.Contains) to check whether an element is a child of another (and will also evaluate to true on itself).

So we can simplify a bit by querying the hidden selectors and then the filter method can use hiddenNode.contains(element).

I suspect that will likely faster/better since it's a browser API.

waabid (approved):
LGTM

PR #5343, opened 2022-04-08
chore(deps-dev): update to node 16

dbjorge:
#### Details

This PR bumps our default/recommended build environment to Node 16.

There aren't any real breaking changes affecting us here, except that our unit tests appear to be much slower (timing out after 10m instead of taking 6-7m in Node 14). We could work around this by just sharding the unit tests like we do for e2e tests already, but it's not ideal; besides "still being slower for local development", it also breaks code coverage reporting (unless we add some new job to merge the results before publishing coverage).

The most relevant looking Jest/Node/V8 issues seem to be covered in the discussion of https://github.com/facebook/jest/issues/11956 - it would be good to experiment with some of the options on that thread to see if we can get the unit test time back down without needing to shard them. Some options to explore:

* Test Node 16.10.0 vs 16.11.0 (we wouldn't want to pin to 16.10.0, but it would be good to try it out to verify that the slowness really is caused by the V8 update in question on that issue)
* Try out the assorted node/v8 command line argument combinations suggested by that thread

Alternatively, we could explore more generic ways of making the tests faster; I think https://github.com/nicolo-ribaudo/jest-light-runner would be promising to look into.

##### Motivation

Keep dependencies up to date; 16 has been an LTS release for several months now.

Also, `accessibility-insights-service` updated to Node 16 yesterday, and we want to ensure that the packages we publish for it to consume are tested in the same environment.

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn fastpass`


karanbirsingh:
I'm looking into the test timing.

dbjorge:
Thanks @karanbirsingh for recording the different test cases! I've implemented a "sharding + remove max-old-space-size" approach (this involved adding a new separate job for publishing merged code coverage results).

Even with these mitigations in place, it's still slower than Node 14, but not by too much (adds ~90s to the average total runtime).

karanbirsingh (approved):

PR #5337, opened 2022-04-08
refactor(mv3-part-2): Add persistent store

sfoslund:
#### Details

Factor `PersistentStore` out of `AssessmentStore` and `UserConfigurationStore`. `PersistentStore` manages persisting store data using the indexed database API. 

##### Motivation

Feature work.

##### Context

This PR aims to refactor only and not make any behavioral changes. In future PRs, we will take advantage of the new `PersistentStore` class to make sure all other store data is persisted as well. This will be necessary when we convert to manifest v3 because we will no longer be able to depend on the background for persisting data, since it is being depreciated in favor of service workers which can potentially stop and lose store data. 

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #5356, opened 2022-04-11
feat(mv3-part-2): Convert background stores data to be persistent

sfoslund:
#### Details

Convert background stores to extend `PersistentStore` instead of `BaseStore` so that the store data will be backed up in the indexed database. This will be necessary when we convert to manifest v3, which replaces the background with service workers which don't allow us to rely on store data to persist without being backed up. 

##### Motivation

Feature work.

##### Context

This PR converts all background stores that do not have an existing way to persist data. This does not include stores that already persist their data (the feature flag and launch panel stores) or stores that are not in the background (the tab stops view store). For consistency it might make sense to convert the stores that have their own custom methods for persisting data to use this new standard approach, but for now leaving them as is. 

Note that with these change the store data is only backed up, it is never read except in the assessment and user configuration stores (see https://github.com/microsoft/accessibility-insights-web/pull/5337). In a future PR we will need to add logic to either initialize stores with their default data (the results of `getDefaultState`, which is used now) or initialize with the persisted data when the service worker stopped and is restarting. 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/IndexedDBDataKeys.ts
+++ b/src/background/IndexedDBDataKeys.ts
@@ -5,4 +5,19 @@ export class IndexedDBDataKeys {
     public static readonly userConfiguration: string = 'userConfiguration';
     public static readonly installation: string = 'installationData';
     public static readonly unifiedFeatureFlags: string = 'featureFlags';
+    public static readonly cardSelectionStore: string = 'cardSelectionStore';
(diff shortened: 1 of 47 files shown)


pownkel (approved):
LGTM

PR #5366, opened 2022-04-13
feat(mv3-part-2): Add AlarmUtils to replace setTimeout

madalynrose:
#### Details
This PR creates:
 * `AlarmUtils`: a class that manages alarms and allows for setTimeout functionality using the chrome.alarms API, which persists across service worker lifetimes.
 *  `createAlarmPromiseFactory`: this will be used by the service worker's browser adapter instead of `createDefaultPromiseFactory`, which uses `setTimeout`.

##### Motivation

Feature work

##### Context

This logic will not have an impact on the manifest v2 extension but will be necessary for the manifest v3 extension, which does not have access to the window global and would produce inconsistent results from `setTimeout` because in-memory timers aren't available across service worker lifetimes. 

A later PR will implement using AlarmUtils instead of WindowUtils (for sending the injection started message) inside of the service worker's browser adapter.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/unit/mock-helpers/chrome-alarms-api-mock.ts
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// This must be the first import, otherwise importing webextension-polyfill from within
+// adapter implementations will fail
+import 'tests/unit/common/webextension-polyfill-setup';
+import { IMock, It, Mock, Times } from 'typemoq';
+import { Alarms, Events } from 'webextension-polyfill';
+
+export class ChromeAlarmsAPMIMock {
```
sfoslund, line 9: Wow, you weren't kidding about lots of mock logic 😆 looks great!

src/tests/unit/mock-helpers/chrome-alarms-api-mock.ts
```diff
@@ -0,0 +1,85 @@
...
+    private underlyingMock: IMock<Alarms.Static> = Mock.ofType<Alarms.Static>();
+    private alarms: Alarms.Alarm[] = [];
+    private alarmListener: (name: Alarms.Alarm) => void;
+
+    public setupAddListener(callback: (name: Alarms.Alarm) => void, times = 1) {
+        const mockOnAlarm: IMock<Events.Event<(name: Alarms.Alarm) => void>> =
+            Mock.ofType<Events.Event<(name: Alarms.Alarm) => void>>();
+        Mock.ofType<(callback: (name: Alarms.Alarm) => void) => void>();
```
sfoslund, line 17: This line seems to have no effect (based on running the tests with it commented out)- is it necessary?

src/common/promises/promise-factory.ts
```diff
@@ -25,3 +27,19 @@ export const createDefaultPromiseFactory = (): PromiseFactory => {
         timeout: createTimeout,
     };
 };
+
+export const createAlarmPromiseFactory = (alarmUtils: AlarmUtils): PromiseFactory => {
+    return {
+        timeout: <T>(promise: Promise<T>, delayInMilliseconds: number) => {
```
sfoslund, line 33: nit: I think this would be more readable if we followed the pattern used for `createDefaultPromiseFactory` and separate the timeout function into a separate variable.

sfoslund (approved):
LGTM!

src/background/alarm-utils.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export class AlarmUtils {
+    protected alarmCallbacks = {};
+    constructor(protected alarmAPI?) {
+        this.alarmAPI ||= chrome.alarms;
```
pownkel, line 6: nit: this could be simplified as
```suggestion
    constructor(protected readonly alarmAPI = chrome.alarms) {
```

src/tests/unit/mock-helpers/chrome-alarms-api-mock.ts
madalynrose: nice catch!

src/background/alarm-utils.ts
madalynrose: ooh good call I'm going to try this locally and make sure the chrome global doesn't upset tests but this looks much nicer.

PR #5370, opened 2022-04-14
feat(mv3-part-2): Persist target page and details view controllers data

sfoslund:
#### Details

Adding logic to persist target page and details view controller backing data. This change has no effect on the manifest v2 version of the extension but will be necessary for the manifest v3 version.

##### Motivation

Feature work.

##### Context

The target page controller actually deals with a map from known tab IDs to tab context objects, which could not be saved in the indexed database as is. As a result, this PR adds logic to save a list of the known tab IDs so that, when the service worker restarts, it knows to only instantiate a new context object to replace the lost one and not to broadcast a `ExistingTabUpdated` message when the url remains the same.

#### Pull request checklist

- [x] Ran `yarn fastpass`


pownkel (approved):
LGTM

PR #5375, opened 2022-04-15
feat(mv3-part-2): Add missing key to persisted data mapping overrides

sfoslund:
#### Details

Follow up to https://github.com/microsoft/accessibility-insights-web/pull/5370, which added logic to persist controllers data. That PR missed updating `keyToPersistedDataMappingOverrides` with the new keys introduced by those changes. Without this change, the controllers will always be initialized with empty objects, instead of persisted data, if any exists. 

##### Motivation

Feature work.

##### Context

This change does not have an effect on the manifest v2 extension because it will always be initialized with empty objects anyways. This change is necessary for the manifest v3 extension, which will need to be initialized with persisted data after a service worker restarts. 

#### Pull request checklist

- [x] Ran `yarn fastpass`


madalynrose (approved):
LGTM!

PR #5380, opened 2022-04-15
feat(mv3-part-2): Browser Adapter Event Manager

madalynrose:
#### Details

This PR creates:
 * `SetTimeoutManager`: class that calls either `windowUtils.setTimeout` or `alarmUtils.createAlarmWithCallback` depending on what it is passed. This will be used when instantiating `TabContextFactory` and `InjectionController` in init scripts.
 * `BrowserAdapterEventManager`: will allow the browser adapter used in background scripts to manage its own events by:
   * maintaining an internal list of application-level listeners
   * synchronously registering an adapter-level listener that calls registered application-level listeners for the given event.
   *  deferring events for which a listener is not found until one is registered
   *  preventing watchdog timeouts by wrapping listener calls in a 4-minute promise race
   * ensuring suspected asynchronous calls have time to finish by setting a 2-minute timer and waiting when an async response is received

##### Motivation

feature work

##### Context

Incorporates work from #5366.  A later PR will use these classes inside of the `BrowserAdapter` and  `TabContextFactory` used in the service worker.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/background/set-timeout-manager.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class SetTimeoutManager {
+    constructor(private utils: { alarmUtils?: AlarmUtils; windowUtils?: WindowUtils }) {}
+
+    public setTimeout(callback: Function, delayInMs: number, name?: string): void {
+        if (name && this.utils.alarmUtils) {
+            this.utils.alarmUtils.createAlarmWithCallback(name, delayInMs, callback);
+        } else {
+            this.utils.windowUtils?.setTimeout(callback, delayInMs);
+        }
```
sfoslund, line 15: nit: Should we check that windowUtils is not null, just in case? Maybe add the check and log an error when neither the alarm or window utils are non-null?

src/tests/unit/tests/background/set-timeout-manager.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import { WindowUtils } from 'common/window-utils';
+import { IMock, Mock, Times } from 'typemoq';
+
+describe(`setTimeout`, () => {
+    const testTimeoutId = `timeout-promise-${Date.now()}`;
+    const delayInMs = 50;
+    const callback = () => true;
+
+    describe('SetTimeoutManager', () => {
+        describe('with alarmUtils', () => {
```
sfoslund, line 14: Can we add some error test cases here, for example, when both alarm and window utils are null, or when name and window utils are both null?

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -0,0 +1,132 @@
...
+    private async forwardEventToApplicationListener(
+        listener: ApplicationListener,
+        eventArgs: any[],
+    ): Promise<any> {
+        const result = listener(...eventArgs);
+        if (!!result && typeof result.then === 'function') {
+            if (this.promiseFactory) {
+                //wrapping application listener promise responses in a 4-minute promise race
+                //prevents the service worker going idle before a response is sent
+                await this.promiseFactory.timeout(result, 240000);
```
sfoslund, line 68: nit: Can we save 240000 as a constant instead of using it as a magic number? (Same with 12000 below)

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -0,0 +1,132 @@
...
+            }
+        } else {
+            // it is possible that this is the result of a fire and forget listener
+            // wrap in 2-minute timeout to ensure it completes during service worker lifetime
+            if (this.setTimeoutManager) {
```
sfoslund, line 73: Is there ever a case when we reasonably expect the setTimeoutManager to be null? If not, can we do this check in the constructor or make it non-nullable?

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -0,0 +1,132 @@
...
+        const allListeners = this.eventsToApplicationListenersMapping[eventType];
+        const applicationListenerIndex = allListeners.findIndex(listener);
+        if (applicationListenerIndex < 0) {
+            console.error('no application listeners found for ', eventType);
+        } else {
+            if (allListeners.length === 1) {
+                delete this.eventsToApplicationListenersMapping[eventType];
+            } else {
+                this.eventsToApplicationListenersMapping[eventType] =
+                    allListeners.splice(applicationListenerIndex);
```
sfoslund, line 101: I'm confused by this logic- if the listener we want to remove is in the middle of allListeners, won't this remove the following items in allListeners as well?

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -0,0 +1,132 @@
...
+            }
+        }
+    }
+
+    public removeListener(
```
sfoslund, line 106: Is there any test coverage for this method?

sfoslund (commented):
Several questions/ comments

src/common/browser-adapters/browser-adapter-event-manager.ts
madalynrose: There isn't right now because it isn't used anywhere. None of the unregister or remove methods are. I'd plan on adding tests once we determine what graceful teardown looks like.

src/common/browser-adapters/browser-adapter-event-manager.ts
madalynrose: oh my you're right! I need to add the argument for how many to remove (just the one!).

src/common/browser-adapters/browser-adapter-event-manager.ts
madalynrose: good point. this depends on implementation details for when we actually use this inside browser adapters and really is out of scope for this PR. Going to make them required and get rid of this logic.

src/common/browser-adapters/browser-adapter-event-manager.ts
madalynrose: I ended up including the remove function from lodash to make this cleaner

src/common/browser-adapters/browser-adapter-event-manager.ts
sfoslund: Makes sense, yeah I'd prefer to omit them for now just to make sure we don't add them now then forget to add tests later

src/background/set-timeout-manager.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { AlarmUtils } from 'background/alarm-utils';
+import { WindowUtils } from 'common/window-utils';
+
+export class SetTimeoutManager {
+    constructor(private utils: { alarmUtils?: AlarmUtils; windowUtils?: WindowUtils }) {}
```
pownkel, line 8: It's a little unclear to me when/why AlarmUtils or WindowUtils would be undefined. Is WindowUtils a temporary measure during the mv3 transition, or will we need to use this class in a context where window is defined? If we need both paths long-term, could this class be separated into something like BackgroundSetTimeoutManager and BrowserSetTimeoutManager?

madalynrose: Oh that's a good point. I'll split this in two similar to how the `PromiseFactory` stuff is structured.

src/common/browser-adapters/browser-adapter-event-manager.ts
madalynrose: I ended up writing tests for them after remembering where they are used later. We'll need them for my next PR anyhow.

src/common/timeouts/timeout-factory.ts
```diff
@@ -0,0 +1,38 @@
...
+export enum TimeoutType {
+    Window,
+    Alarm,
+}
+export interface TimeoutFactory {
+    timeoutType: TimeoutType;
+    createTimeout: Timeout;
+}
+
+export class WindowTimeoutFactory implements TimeoutFactory {
```
sfoslund, line 17: This structure looks much better, great suggestion @pownkel!

sfoslund (approved):
LGTM!

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -0,0 +1,128 @@
...
+        this.deferredEvents = this.deferredEvents.filter(
+            event => this.processEvent(event.eventType, event.eventArgs, true) === false,
+        );
+    }
+
+    private async forwardEventToApplicationListener(
+        listener: ApplicationListener,
+        eventArgs: any[],
+    ): Promise<any> {
+        const result = listener(...eventArgs);
```
dbjorge, line 63: This should probably be defensive about the `listener` throwing

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -0,0 +1,128 @@
...
+const TWO_MINUTES = 120000;
+const FOUR_MINUTES = 2 * TWO_MINUTES;
+
+export class BrowserAdapterEventManager {
+    protected deferredEvents: EventDetails[] = [];
+    protected eventsToApplicationListenersMapping: DictionaryStringTo<ApplicationListener[]> = {};
+    protected adapterListener: AdapterListener =
+        (eventType: string) =>
+        (...eventArgs: any[]) => {
+            this.processEvent(eventType, eventArgs);
```
dbjorge, line 30: If this is intended to be the thing that gets attached to browser events directly, it needs to be returning the promise generated by `forwardEventToApplicationListener`, no?

dbjorge: (and probably creating some sort of deferred-with-timeout Promise in the event that there are no current app listeners)

PR #5363, opened 2022-04-12
feat(mv3-part-2): Initialize stores with persisted data

sfoslund:
#### Details

Follow up to #5356, which converted background stores to persist their data. This PR adds logic to initialize stores with the persisted data, if it exists. 

##### Motivation

Feature work.

##### Context

This logic will not have an impact on the manifest v2 extension but will be necessary for the manifest v3 extension, which may initialize stores after a service worker stopped and lost the in process data. 

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #5381, opened 2022-04-18
fix: Do not initialize with persisted data for manifest v2 extension

sfoslund:
#### Details

Fix bug discovered in canary caused by https://github.com/microsoft/accessibility-insights-web/pull/5363.

##### Motivation

Fixes bug in canary.

##### Context

The issue in canary seems to be caused with an underlying bug in the way details view store data is persisted. This PR does not address that issue, but instead it stops initializing stores/ controllers with persisted data for the manifest v2 extension. This will allow us to unblock canary but leave the persisting logic in so that it can be fixed and called from the manifest v3 extension.

#### Pull request checklist

- [ ] Ran `yarn fastpass`


madalynrose (approved):
I built locally and compared to building the main branch and this fixes the bug. Looks good!

PR #5387, opened 2022-04-18
fix: update async dependency where possible

jlperkins:
#### Details

This PR partially addresses [this dependabot alert](https://github.com/microsoft/accessibility-insights-web/security/dependabot/18) about using a vulnerable version of "async." Where possible, it updates the dependencies we have that rely on async and bumps them to use a current, secure version of the package.  For the case where this is not possible, choosing to leave the too-low version in for now, as there are two upstream fixes that are currently being worked on and should hit soon. 

"Nested" dependencies were updated by removing and re-adding the async@^2.0 entry of the lockfile, since our caret version, when refreshed, would bump up to a secure version of async (2.6.4).  Also, electron-builder and app-builder-lib were removed and then re-added from the package.json file, which refreshed some other respective dependencies that rely on async, but also bumped up the version for quite a few other files. The version bump to async has been highlighted by a comment so it is easier to find among the changes in this PR. 

##### Motivation

Keeping dependencies up-to-date.

##### Context

This fixes all but one of the dependencies we have on async.  However, that remaining dependency (we use app-builder-lib, which uses ejs, which uses jake, which uses a vulnerable version of async) has two upstream fixes that are currently getting attention. Secure versions of async are specifically 2.6.4, and then 3.2.3 and above. Jake currently relies on async@0.9.2.  However, there is [a recent commit in jake's repo](https://github.com/jakejs/jake/pull/411) which updates to use async@3.2.3, as well as [a pending commit in ejs](https://github.com/mde/ejs/pull/645) to remove its dependency on jake, period. Once either of those changes are included in a new release, we will only be dependent upon secure versions of async. This "wait for them to fix it" approach was chosen since the fixes are being actively addressed, and likely by the time we did work to create fixes ourselves, those fixes would be released already. 

#### Pull request checklist

- [x] Ran `yarn fastpass`


yarn.lock
```diff
@@ -3503,9 +3503,9 @@ async@0.9.x:
   integrity sha1-rqdNXmHB+JlhO/ZL2mbUx48v0X0=
 
 async@^2.6.0:
-  version "2.6.3"
-  resolved "https://registry.yarnpkg.com/async/-/async-2.6.3.tgz#d72625e2344a3656e3a3ad4fa749fa83299d82ff"
-  integrity sha512-zflvls11DCy+dQWzTW2dzuilv8Z5X/pjfmZOWba6TNIVDm+2UDaJmXSOXlasHKfNBs8oo3M0aT50fDEWfKZjXg==
+  version "2.6.4"
```
jlperkins, line 3506: This is the change in question; version 2.6.4 is not vulnerable.

dbjorge (approved):

PR #5389, opened 2022-04-19
feat(mv3-part-2): Include tab ID in tab specific persisted data keys

sfoslund:
#### Details

Add logic to include the tab ID in the indexed database key for stores that have multiple instances for each tab. These changes fix the bug that #5381 worked around. 

##### Motivation

Feature work.

##### Context

This PR leaves the workaround from #5381 in so that these changes do not effect the manifest v2 extension. 

Including the tab ID in the database key caused the following changes to be necessary:

- Update the keys to be functions that accept the tabID- see changes in IndexedDBDataKeys 
- Pass tab ID through to each store- see the changes in the stores, TabContextFactory, TargetPageController, and TabContextStoreHub
- Add logic to teardown stores to ensure that when tabs are closed we clean up the storage associated with them- see the changes in PersistentStore, TabContext, and TargetPageController
- Update get-persisted-data so that it knows how to read tab specific data from the database
- Update background and renderer init so that they only get the global persisted data that they will actually use.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/get-persisted-data.ts
+++ b/src/background/get-persisted-data.ts
@@ -23,26 +23,32 @@ import { IndexedDBDataKeys } from './IndexedDBDataKeys';
 import { InstallationData } from './installation-data';
 
 export interface PersistedData {
+    tabData: {
+        [tabId: number]: TabSpecificPersistedData;
+    };
     assessmentStoreData: AssessmentStoreData;
     userConfigurationData: UserConfigurationStoreData;
     installationData: InstallationData;
     featureFlags: FeatureFlagStoreData;
+    commandStoreData: CommandStoreData;
+    permissionsStateStoreData: PermissionsStateStoreData;
+    scopingStoreData: ScopingStoreData;
+    knownTabIds: number[];
+    tabIdToDetailsViewMap: DictionaryStringTo<number>;
+}
+
+export interface TabSpecificPersistedData {
     cardSelectionStoreData: CardSelectionStoreData;
     detailsViewStoreData: DetailsViewStoreData;
     devToolStoreData: DevToolStoreData;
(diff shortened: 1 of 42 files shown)


src/background/get-persisted-data.ts
```diff
@@ -63,12 +88,35 @@ export function getPersistedData(
...
+    const persistedData = { tabData: {} } as PersistedData;
+
+    let promises = getGlobalPersistedPromises(
+        IndexedDBDataKeys.globalKeys,
+        indexedDBInstance,
+        persistedData,
+    );
+
+    const knownTabIds: number[] = await indexedDBInstance.getItem(IndexedDBDataKeys.knownTabIds);
+    if (knownTabIds && knownTabIds.length > 0) {
```
madalynrose, line 106: This bit of code is doing a lot of heavy lifting. It might be good to break it out into a couple separate pure functions that we can call here to make it more readable and test individually.

madalynrose (approved):
Just one suggestion so I'm approving with comments! This looks awesome!

src/background/get-persisted-data.ts
sfoslund: Thanks Madalyn! I've pushed a merge to make sure this doesn't break the service-init you added yesterday and also separated out into functions. Do you think it's necessary to test all the helper functions separately? I'd prefer to not export them and I'm not sure if there's a way to test without exporting.

madalynrose: this looks great actually! there are ways to test non-exported functions by creating a "testable" version of the class inside the test file that extends the original class and has methods that act as handles to the internals but it's not blocking.

PR #5396, opened 2022-04-19
feat(mv3-part-2): Clean up BrowserAdapterEventManager

madalynrose:
#### Details

This addresses outstanding comments from @dbjorge on #5380 that weren't seen before it got merged.

This PR:
* removes AlarmUtils as they are unnecessary as long as we ensure setTimeout can finish
* updates the BrowserAdapterEventManager to register a single listener per eventType
* updates logic around event deferral to create a promise to be resolved externally, return pending promise, and resolve promise when the deferred event is handled by an application listener
* adds a function to promiseFactory for creating externally resolved promises

##### Motivation

feature work

##### Context

None of this code is called anywhere yet but will be used by the browser adapter used in the service worker.

Note that since we're tracking one application listener per eventType, we will need to create a single application listener to react to and dispatch events for eventTypes that have multiple calls to addListener (e.g. onMessage will need to respond to different types of messages).

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -22,24 +24,21 @@ const TWO_MINUTES = 120000;
...
     public registerEventToApplicationListener = (
         eventType: string,
         callback: ApplicationListener,
     ): void => {
-        if (this.eventsToApplicationListenersMapping[eventType]) {
-            this.eventsToApplicationListenersMapping[eventType].push(callback);
-        } else {
-            this.eventsToApplicationListenersMapping[eventType] = [callback];
-        }
+        this.eventsToApplicationListenersMapping[eventType] = callback;
```
dbjorge, line 41: Maybe have this be defensive about multiple registration (ie, detect and throw, rather than silently overriding)?

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -22,24 +24,21 @@ const TWO_MINUTES = 120000;
...
 
 export class BrowserAdapterEventManager {
-    protected deferredEvents: EventDetails[] = [];
-    protected eventsToApplicationListenersMapping: DictionaryStringTo<ApplicationListener[]> = {};
+    protected deferredEvents: DeferredEventDetails[] = [];
+    protected eventsToApplicationListenersMapping: DictionaryStringTo<ApplicationListener> = {};
     protected adapterListener: AdapterListener =
         (eventType: string) =>
         (...eventArgs: any[]) => {
             this.processEvent(eventType, eventArgs);
```
dbjorge, line 32: I think this needs to pass through the return value of `this.processEvent`

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
...
-                };
-                this.deferredEvents.push(eventDetails);
+                });
             }
-            return false;
         }
     }
 
     private unregisterAdapterListener(event: Events.Event<any>, eventType: string) {
         event.removeListener(this.adapterListener(eventType));
```
dbjorge, line 127: `this.adapterListener` isn't memoized, so I don't think this is going to have the desired effect

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
...
 
-    private unregisterApplicationListenerForEvent(
-        eventType: string,
-        listener: ApplicationListener,
-    ) {
-        const allListeners = this.eventsToApplicationListenersMapping[eventType];
-
-        if (allListeners.length === 1) {
+    private unregisterApplicationListenerForEvent(eventType: string) {
+        this.eventsToApplicationListenersMapping[eventType] &&
```
dbjorge, line 131: This check isn't necessary, `delete` is idempotent in javascript.

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
...
             // it is possible that this is the result of a fire and forget listener
             // wrap in 2-minute timeout to ensure it completes during service worker lifetime
-            return this.timeoutFactory.timeoutType === TimeoutType.Alarm
-                ? this.timeoutFactory.createTimeout(
-                      () => Promise.resolve(result),
-                      TWO_MINUTES,
-                      `wrapper-timeout-${Date.now()}`,
-                  )
-                : this.timeoutFactory.createTimeout(() => Promise.resolve(result), TWO_MINUTES);
+            globalThis.setTimeout(() => {
```
dbjorge, line 88: This block isn't quite right; it needs to be returning a promise which will resolves after two minutes. This is instead returning a promise which resolves immediately, and setting up a `setTimeout` which will create an unrelated Promise after two minutes and then immediately throw it away.

dbjorge: To make it easier to test this without actually needing to wait for timeouts, I'd suggest creating a `promiseFactory.delay` which does something like `return new Promise((resolve) => { this.setTimeout(() => resolve(), delayInMillis); });`

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
...
         listener: ApplicationListener,
         eventArgs: any[],
     ): Promise<any> {
-        const result = listener(...eventArgs);
+        let result: any;
+        try {
+            result = listener(...eventArgs);
+        } catch (error) {
+            console.error('Error thrown in application listener: ', error);
+            throw error;
```
dbjorge, line 79: I don't think rethrowing like this is quite right; I think that even if the listener throws an error, it's probably safest for us to still apply the rest of the logic and return a promise to the browser event (otherwise, a single misbehaving event handler is liable to cause the whole service worker to be torn down early).

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
...
+        }
         if (!!result && typeof result.then === 'function') {
             //wrapping application listener promise responses in a 4-minute promise race
             //prevents the service worker going idle before a response is sent
-            await this.promiseFactory.timeout(result, FOUR_MINUTES);
+            return this.promiseFactory.timeout(result, FOUR_MINUTES);
         } else {
```
dbjorge, line 85: Do we have any cases where an application handler will be returning a non-undefined, non-promise result? I'm not sure the "delay for 2 minutes" approach would be a good idea in that case, we probably don't want to delay actual responses for 2 minutes. I don't think we have a great solution for that case - we don't have a way to simultaneously report a response back and also indicate the 2 minute deferal.

I think we should probably fail fast and log an error if that happens, and update any such cases that occur in practice by auditing them for any fire and forget behavior, updating them to ensure any such behavior is captured within a Promise, and then returning a Promise with the original result.

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
     };
 
     private processDeferredEvents(): void {
-        this.deferredEvents = this.deferredEvents.filter(
-            event => this.processEvent(event.eventType, event.eventArgs, true) === false,
-        );
+        const stillDeferred: DeferredEventDetails[] = [];
+        this.deferredEvents.forEach(deferredEvent => {
```
dbjorge, line 54: Prefer using for-of loops rather than `Array.forEach` - the former gives better stack traces from `Error`s and is more straightforward to step through in a debugger

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
...
+    }
+
+    public processEvent(eventType: string, eventArgs: any[]): Promise<any> {
+        const applicationListener = this.eventsToApplicationListenersMapping[eventType];
+        if (applicationListener) {
+            return this.forwardEventToApplicationListener(applicationListener, eventArgs);
         } else {
-            if (!isDeferred) {
-                const eventDetails: EventDetails = {
+            const alreadyDeferred = this.deferredEvents.find(
```
dbjorge, line 111: This `find` call gives `processDeferedEvents` an `O(deferredEvents.length ^ 2)` time complexity, and we do see thousands of events in normal usage of the app. "Are we in an already-deferred" path is already known in each of the 2 `processEvent` callsites, and one of them doesn't actually use the return value of `processEvent`; maybe instead of the current contents of this else branch, just return `null` and update L32 to be something along the lines of `return processEvent(e, ...args) ?? deferEvent(e, ...args);`?

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
...
-        return result;
     }
 
-    public processEvent(eventType: string, eventArgs: any[], isDeferred?: boolean): boolean {
-        if (this.eventsToApplicationListenersMapping[eventType]) {
-            this.eventsToApplicationListenersMapping[eventType].forEach(applicationListener => {
-                this.forwardEventToApplicationListener(applicationListener, eventArgs);
-            });
-            return true;
+    private deferEvent(eventDetails: EventDetails): Promise<any> {
```
dbjorge, line 95: I don't think this has the desired effect in the case where the browser sends us an event we never register a handler for. I think what we'd want in that case would be to time out the promise after 4 minutes, but I think this would just pend the promise indefinitely and get our worker torn down after 5 minutes

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -51,78 +50,90 @@ export class BrowserAdapterEventManager {
...
+            result = listener(...eventArgs);
+        } catch (error) {
+            console.error('Error thrown in application listener: ', error);
+            throw error;
+        }
         if (!!result && typeof result.then === 'function') {
             //wrapping application listener promise responses in a 4-minute promise race
             //prevents the service worker going idle before a response is sent
-            await this.promiseFactory.timeout(result, FOUR_MINUTES);
+            return this.promiseFactory.timeout(result, FOUR_MINUTES);
```
dbjorge, line 84: `PromiseFactory.timeout` _rejects_ after the given timeout, and I'm not sure we actually want that here as opposed to resolving after the given timeout. Rejecting is more accurate, but if the browser responds to a rejected promise by killing the service worker, we probably don't want to do that. I don't know if the Browser treats resolved vs rejected promises the same way.

dbjorge (changes requested):
Thanks for going back for the comments! Reviewed inline

src/tests/unit/tests/common/promises/promise-factory.test.ts
```diff
@@ -46,6 +46,33 @@ describe(`promiseFactory`, () => {
...
+
+            expect(result).resolves.toEqual(actual);
+        });
+
+        it("propogates an underlying Promise's reject", async () => {
+            const reason = 'rejecting!';
+            const rejecting = new Promise((resolve, reject) => {
+                setTimeout(() => reject(reason), 10);
+            });
+            expect(rejecting).rejects.toEqual(reason);
```
dbjorge, line 66: This is missing a `testObject` invocation, no?

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -76,19 +95,27 @@ export class BrowserAdapterEventManager {
...
             console.error('Error thrown in application listener: ', error);
-            throw error;
+            result = error;
         }
         if (!!result && typeof result.then === 'function') {
-            //wrapping application listener promise responses in a 4-minute promise race
+            //wrapping application listener promise responses in a 4-minute promise resolution
             //prevents the service worker going idle before a response is sent
-            return this.promiseFactory.timeout(result, FOUR_MINUTES);
+            return this.promiseFactory.delay(result, FOUR_MINUTES);
```
dbjorge, line 103: I think `delay` is right for the behavior on L108, but I don't think it's quite what what we want here - if an application handler returns a `Promise` that resovles in a timely fashion, I think we should probably trust it to encapsulate the required async work and resolve as soon as it does, not delay it for 4 minutes. I think the 4 minutes should be a limit, not a minimum delay.

I think a straightforward way to achieve that but still avoid rejecting the promise would be something like

```ts
return this.promiseFactory.timeout(result, FOUR_MINUTES)
    .catch(() => {
        // We want to avoid rejecting the Promise we give back to the browser event handler
        // because doing so is liable to cause the browser to tear down our Service Worker
        // immediately, even if other events are in progress.
        return undefined;
    });
```

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -76,19 +95,27 @@ export class BrowserAdapterEventManager {
             result = listener(...eventArgs);
         } catch (error) {
             console.error('Error thrown in application listener: ', error);
```
dbjorge, line 97: Can this inject a `Logger` dependency and use `logger.error`, rather than calling `console.error` directly? Even if the message isn't important to test, doing that and using a noop logger avoids unnecessary console spew during test execution.

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -26,10 +25,13 @@ const FOUR_MINUTES = 2 * TWO_MINUTES;
 export class BrowserAdapterEventManager {
```
dbjorge, line 25: It would maybe be helpful to give this class a comment explaining the problem it's solving; I worry that it would be easy for a maintainer without context to remove a subtly important timeout

src/common/browser-adapters/browser-adapter-event-manager.ts
madalynrose: good call!

src/tests/unit/tests/common/promises/promise-factory.test.ts
madalynrose: 🤦🏻‍♀️ you're right! thanks!

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -14,115 +13,150 @@ export interface EventDetails {
...
-    protected eventsToApplicationListenersMapping: DictionaryStringTo<ApplicationListener[]> = {};
-    protected adapterListener: AdapterListener =
+    protected deferredEvents: DeferredEventDetails[] = [];
+    protected eventsToApplicationListenersMapping: DictionaryStringTo<ApplicationListener> = {};
+    protected eventsToAdapterListenersMapping: DictionaryStringTo<AdapterListener> = {};
+    protected handleEvent: (eventType: string) => AdapterListener =
         (eventType: string) =>
         (...eventArgs: any[]) => {
-            this.processEvent(eventType, eventArgs);
+            const processedResults = this.processEvent(eventType, eventArgs); //guard in case we return undefined for a timed out promise resoponse
```
dbjorge, line 43: I'm not sure I understand the comment here; the new behavior for the 4 minute timeout promise turns a "timeout" from a event that rejects the promise to an event that resolves the Promise as undefined, but the value that `processedResults` gets set to here is still going to be a Promise, not `undefined`.

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -14,115 +13,150 @@ export interface EventDetails {
...
-        listener: ApplicationListener,
-    ) {
-        const allListeners = this.eventsToApplicationListenersMapping[eventType];
+        this.deferredEvents.push({
+            deferredResolution: deferredResolution,
+            ...eventDetails,
+        });
+        //resolve the promise in four minutes in case a listener is never registered
+        setTimeout(() => {
+            deferredResolution.resolveHook('no listener registered');
```
dbjorge, line 148: I think it would be good for us to log an error in this case as well; I don't think this is expected in any scenario

src/common/browser-adapters/browser-adapter-event-manager.ts
```diff
@@ -14,115 +13,150 @@ export interface EventDetails {
...
+
     private processDeferredEvents(): void {
-        this.deferredEvents = this.deferredEvents.filter(
-            event => this.processEvent(event.eventType, event.eventArgs, true) === false,
-        );
+        const stillDeferred: DeferredEventDetails[] = [];
+        for (const deferredEvent of this.deferredEvents) {
+            if (this.eventsToApplicationListenersMapping[deferredEvent.eventType]) {
+                try {
+                    this.processEvent(deferredEvent.eventType, deferredEvent.eventArgs);
```
dbjorge, line 88: I don't think it's quite right for this to ignore the return value of `this.processEvent`; I think this needs to set it up so that if `processEvent` returns a `Promise`, the `deferredResolution.resolveHook` gets invoked only after that promise settles, not immediately.

dbjorge (dismissed):
Thanks for the updates! I've gone back through and marked resolved all the original comments. Left another few (the one about the `processEvents`/`deferredResolution` relationship is blocking, I think)

dbjorge:
I've cleaned up the outstanding issues. I also ended up redoing the unit test strategy; I created simulators for the browser events and underlying promise factory to avoid needing to test implementation details.

I would recommend the next reviewer review the changes as a whole, rather than just the delta since Madalyn's last commit - this new behavior is complex enough that I'd like a second pair of eyes on the whole thing anyway.

sfoslund (commented):
Looks good overall, a few nits/ questions!

src/common/browser-adapters/browser-event-manager.ts
```diff
@@ -0,0 +1,196 @@
...
+        const applicationListener = this.eventsToApplicationListenersMapping[eventType];
+        return applicationListener
+            ? this.forwardEventToApplicationListener(applicationListener, eventType, eventArgs)
+            : null;
+    }
+
+    private processDeferredEvents(): void {
+        const stillDeferred: DeferredEventDetails[] = [];
+        for (const deferredEvent of this.deferredEvents) {
+            if (deferredEvent.isStale) {
```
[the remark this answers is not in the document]
dbjorge: I think it's preferable to avoid processing an already-timed-out event. If a still-deferred event ever becomes stale, it means that it timed out while it was still in a deferred state, and we've already indicated back to the browser that this event should no longer be preventing the worker from being torn down. If that were to happen, I expect debugging the resulting timeout error log on its own will probably be less misleading than simultaneously debugging a timeout error log plus also whatever exciting result happens when the worker gets torn down mid-handler-execution.

src/common/browser-adapters/browser-event-manager.ts
```diff
@@ -0,0 +1,196 @@
...
+
+interface DeferredEventDetails extends EventDetails {
+    deferredResolution: ExternalResolutionPromise;
+    isStale: boolean;
+}
+
+// As of writing, Chromium maintains its own 5 minute event timeout and will tear down our
+// service worker if this is exceeded, even if other work is outstanding. To avoid this, our
+// own timeout MUST be shorter than Chromium's.
+const EVENT_TIMEOUT_MS = 4 * 60 * 1000; // 4 minutes
```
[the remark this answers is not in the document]
dbjorge: Excellent question. Added a link to the relevant Chromium docs to the class-level comment

src/common/browser-adapters/browser-event-manager.ts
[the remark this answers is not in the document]
dbjorge: (added a comment to this effect)

sfoslund (approved):
Looks great! I appreciate all the detailed comments!

PR #5388, opened 2022-04-19
feat(axe-core-4.4): Update axe-core@4.4.1

katydecorah:
#### Details

This PR will update axe-core to 4.4.1. This new version includes bug fixes, 17 false-positive fixes, 1 rule change, and 1 general improvement.

- Omit [color-contrast-enhanced, wcag146](https://github.com/dequelabs/axe-core/blob/f583c70b191513f5bd116a3ed091e937ea2cd65d/lib/rules/color-contrast-enhanced.json) in tests. Being AAA, this rule is out of scope.
- Update axe config to replace `branding` property value to be a string.
- Remove webpack configuration for fallback of the "crypto" import, the dependency was removed in axe-core.
- Remove manual typings that are now in axe-core.
- Exclude all rules tagged with `wcag2aaa`. The product does not include AAA rules and by filtering them out it will prevent `identical-links-same-purpose` from being included. In 4.4, some [rules were retagged](https://github.com/dequelabs/axe-core/commit/828e526fd06ee596df73f4825e750aad459ca75e), which removed the `best-practices` tag from this rule (the `best-practices` tag previously excluded `identical-links-same-purpose`).

##### To do

- [x] Understand how we should handle identical-links-same-purpose
- [x] Check if we can improve our code from the changes: 
    - [x] include/exclude property for AxeOptions has been enhanced 
    - [x] allow passing a NodeList to ElementContext 

##### Motivation

Feature work

##### Context

axe-core [4.4.0](https://github.com/dequelabs/axe-core/releases/tag/v4.4.0) and [4.4.1](https://github.com/dequelabs/axe-core/releases/tag/v4.4.1) release notes

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #5397, opened 2022-04-19
feat(mv3-part-2): minimum changes needed to produce a working MV3 extension

madalynrose:
#### Details

This includes changes required to stand up a working manifest V3 extension:
* include crypto backup in webpack
* replace references to window with references to globalThis
* update deprecated API calls:
  *  chrome.extension.getUrl --> chrome.runtime.getUrl
  * chrome.tabs.insertCSS --> chrome.scripting.insertCSS 
  * chrome.tabs.executeScript --> chrome.scripting.executeScript

##### Motivation

feature work

##### Context

Note that the only place the background script references window is where the InjectionController calls windowUtils.setTimeout. For now, windowUtils.setTimeout has been updated to call globalThis.setTimeout instead of window.setTimeout. This should be made more robust later.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -437,7 +437,7 @@ module.exports = function (grunt) {
         grunt.file.write(configJSONPath, configJSON);
         const copyrightHeader =
             '// Copyright (c) Microsoft Corporation. All rights reserved.\n// Licensed under the MIT License.\n';
-        const configJS = `${copyrightHeader}window.insights = ${configJSON}`;
+        const configJS = `${copyrightHeader}globalThis.insights = ${configJSON};`;
         grunt.file.write(configJSPath, configJS);
     });
 
@@ -461,15 +461,86 @@ module.exports = function (grunt) {
             // Settings that are specific to MV3
             merge(manifestJSON, {
                 action: {
+                    default_popup: 'popup/popup.html',
                     default_icon: {
                         20: config.options.icon16,
                         40: config.options.icon48,
                     },
                 },
+                permissions: [
+                    'alarms',
+                    'notifications',
+                    'scripting',
+                    'storage',
+                    'tabs',
+                    'webNavigation',
+                ],
                 background: {
                     service_worker: 'bundle/serviceWorker.bundle.js',
                 },
-                host_permissions: [],
+                host_permissions: ['*://*/*'],
+                web_accessible_resources: [
+                    {
+                        resources: [
+                            'insights.html',
+                            'assessments/*',
+                            'injected/*',
+                            'background/*',
+                            'common/*',
+                            'DetailsView/*',
+                            'bundle/*',
+                            'NOTICE.html',
+                        ],
+                        matches: ['<all_urls>'],
+                    },
+                ],
+                commands: {
+                    _execute_browser_action: {
+                        suggested_key: {
+                            windows: 'Alt+Shift+K',
+                            mac: 'Alt+Shift+K',
+                            chromeos: 'Alt+Shift+K',
+                            linux: 'Alt+Shift+K',
+                        },
+                        description: 'Activate the extension',
+                    },
+                    '01_toggle-issues': {
+                        suggested_key: {
+                            windows: 'Alt+Shift+1',
+                            mac: 'Alt+Shift+1',
+                            chromeos: 'Alt+Shift+1',
+                            linux: 'Alt+Shift+1',
+                        },
+                        description: 'Toggle Automated checks',
+                    },
+                    '02_toggle-landmarks': {
+                        suggested_key: {
+                            windows: 'Alt+Shift+2',
+                            mac: 'Alt+Shift+2',
+                            chromeos: 'Alt+Shift+2',
+                            linux: 'Alt+Shift+2',
+                        },
+                        description: 'Toggle Landmarks',
+                    },
+                    '03_toggle-headings': {
+                        suggested_key: {
+                            windows: 'Alt+Shift+3',
+                            mac: 'Alt+Shift+3',
+                            chromeos: 'Alt+Shift+3',
+                            linux: 'Alt+Shift+3',
+                        },
+                        description: 'Toggle Headings',
+                    },
+                    '04_toggle-tabStops': {
+                        description: 'Toggle Tab stops',
+                    },
+                    '05_toggle-color': {
+                        description: 'Toggle Color',
+                    },
+                    '06_toggle-needsReview': {
+                        description: 'Toggle Needs review',
+                    },
+                },
             });
         } else {
             // Settings that are specific to MV2. Note that many of these settings--especially the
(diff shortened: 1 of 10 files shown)


src/background/service-worker-init.ts
```diff
@@ -1,20 +1,230 @@
...
-}
+    const urlValidator = new UrlValidator(browserAdapter);
+    const indexedDBInstance: IndexedDBAPI = new IndexedDBUtil(getIndexedDBStore());
+    const indexedDBDataKeysToFetch = [
+        IndexedDBDataKeys.assessmentStore,
+        IndexedDBDataKeys.userConfiguration,
+    ];
+
+    // // These can run concurrently, both because they are read-only and because they use different types of underlying storage
+    const persistedDataPromise = getPersistedData(indexedDBInstance, indexedDBDataKeysToFetch);
```
sfoslund, line 67: My changes in https://github.com/microsoft/accessibility-insights-web/pull/5389 will conflict with this/ change some of this logic, so we should probably try to get this PR in before merging that one.

sfoslund (approved):
LGTM!

Gruntfile.js
```diff
@@ -461,15 +461,86 @@ module.exports = function (grunt) {
...
+                            'background/*',
+                            'common/*',
+                            'DetailsView/*',
+                            'bundle/*',
+                            'NOTICE.html',
+                        ],
+                        matches: ['<all_urls>'],
+                    },
+                ],
+                commands: {
```
dbjorge, line 497: It looks like the `commands` sections are identical between v2 and v3; should that be moved to the base `manifest.json` instead of being repeated?

package.json
```diff
@@ -184,6 +184,7 @@
         "axe-core": "4.3.2",
         "axios": "^0.26.1",
         "classnames": "^2.3.1",
+        "crypto-browserify": "^3.12.0",
```
dbjorge, line 187: This should become unnecessary once #5388 merges

webpack.config.js
```diff
@@ -102,7 +102,7 @@ const commonConfig = {
         extensions: ['.tsx', '.ts', '.js', '.json'],
         // axe-core invokes require('crypto'), but only in a path we don't use, so we don't need a polyfill
         // See https://github.com/dequelabs/axe-core/issues/2873
-        fallback: { crypto: false },
+        fallback: { crypto: require.resolve('crypto-browserify'), stream: false },
```
dbjorge, line 105: This should become unnecessary once #5388 merges (axe-core removed their crypto dependency, so the whole `fallback` config goes away in that PR)

src/common/browser-adapters/webextension-browser-adapter.ts
```diff
@@ -91,12 +91,22 @@ export abstract class WebExtensionBrowserAdapter
         details: ExtensionTypes.InjectDetails,
     ): Promise<any[]> {
         this.verifyPathCompatibility(details.file);
-        return browser.tabs.executeScript(tabId, details);
+        return typeof browser.tabs.executeScript === 'function'
+            ? browser.tabs.executeScript(tabId, details)
+            : chrome.scripting.executeScript({
+                  target: { tabId, allFrames: details.allFrames },
+                  files: [details.file],
+              } as chrome.scripting.ScriptInjection);
```
dbjorge, line 99: Is the `as chrome.scripting.ScriptInjection` necessary here (and on L109)? I'd have expected the option block we're passing to match the expected type without requiring a cast

src/common/browser-adapters/webextension-browser-adapter.ts
```diff
@@ -206,7 +216,9 @@ export abstract class WebExtensionBrowserAdapter
     }
 
     public getUrl(urlPart: string): string {
-        return chrome.extension.getURL(urlPart);
+        return typeof chrome.extension?.getURL === 'function'
```
dbjorge, line 219: Is falling back to `chrome.extension.getURL` necessary here? I think `browser.extension.getURL` is documented as working in all cases we support, regardless of manifest version

src/common/window-utils.ts
```diff
@@ -25,7 +25,7 @@ export class WindowUtils {
     }
 
     public setTimeout(handler: Function, timeout: number): number {
-        return window.setTimeout(handler, timeout);
+        return globalThis.setTimeout(handler, timeout);
```
dbjorge, line 28: It feels a little off to be setting up `WindowUtils` to be only partially compatible with a `window`-free enviornment like this. The only place the service worker uses it is to pass it via `TabContextFactory` to `InjectorController`; perhaps it might make more sense to replace the `WindowUtils` dependency in those two classes with a `setTimeout` dependency and just pass `globalThis.setTimeout` from `service-worker-init`, to avoid having to create `WindowUtils` there at all?

dbjorge (changes requested):
Looks reasonable overall! A few comments inline

src/common/configuration/global-variable-configuration.ts
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dbjorge, line 1: It looks like this file is intended to be a rename of `window-variable-configuration.ts`, but it looks like the PR doesn't actually remove the old file (or its test file); is that intentional? It doesn't look like the old file is still used anywhere after this PR.

package.json
dbjorge: (I've just merged #5388 to make it easier to resolve this)

webpack.config.js
dbjorge: (I've just merged #5388 to make it easier to resolve this)

src/common/window-utils.ts
madalynrose: oh good call! I like this solution.

src/common/configuration/global-variable-configuration.ts
madalynrose: good catch! I'll remove those.

src/common/window-utils.ts
dbjorge, line 28: I think this line can be reverted for consistency with the rest of `WindowUtils` now that you've made the change elsewhere to pass `setTimeout` directly instead

Gruntfile.js
```diff
@@ -461,15 +461,39 @@ module.exports = function (grunt) {
...
+                            'injected/*',
+                            'background/*',
+                            'common/*',
+                            'DetailsView/*',
+                            'bundle/*',
+                            'NOTICE.html',
+                        ],
+                        matches: ['<all_urls>'],
+                    },
+                ],
```
dbjorge, line 496: I think the MV2 section for `commands` down on L527 in this file can also be dropped now that it's in the base `manfiest.json`

dbjorge (approved):
Looks good overall! I left two new comments, but they're pretty minor; approving with them still outstanding

PR #5407, opened 2022-04-22
(feat: mv3-part-2): Consolidate background message distribution logic

pownkel:
#### Details

Currently, we register separate onMessage event handlers in MessageDistributor and PostMessageContentHandler. This PR combines these into one onMessage handler in MessageDistributor.

##### Motivation

As noted in #5396, BrowserAdapterEventManager only allows one event handler per event, so we will need to combine message distribution logic into one ApplicationHandler. Most of it is already consolidated into MessageDistributor, so this PR moves the remaining onMessage registrations in the background context into MessageDistributor.

##### Context

Multiple onMessage handlers are also registered by StoreProxies. However, StoreProxies are only created in contexts that don't have a MessageDistributor (i.e. non-background contexts), so those event handlers will be consolidated in a different class and a different PR.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/background-init.ts
+++ b/src/background/background-init.ts
@@ -174,9 +174,16 @@ async function initialize(): Promise<void> {
     );
     keyboardShortcutHandler.initialize();
 
+    const postMessageContentRepository = new PostMessageContentRepository(
+        DateProvider.getCurrentDate,
+    );
+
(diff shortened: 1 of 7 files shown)


madalynrose (approved):
This is a great solution! Nice usage of functionality we already have!

PR #5410, opened 2022-04-22
feat(mv3-part-2): Do not persist data for mv2 extension

sfoslund:
#### Details

Previous PRs added logic to persist store/controller data, which will be necessary for the mv3 extension. Those PRs made it so that the mv2 extension also persists data but does not read from the persisted data. @dbjorge suggested that we do not persist data at all in the mv2 extension to avoid backward compat issues in the future, so this PR turns data persist-ing off by default. I will follow up in a future PR to turn data persisting on only for the mv3 extension.

##### Motivation

Feature work.

##### Context

Confirmed by running dev mv2 extension locally that we no longer persist data (other than assessment, user config, etc- which were already persisted prior to this change).

Note that test file diffs look big primarily because of whitespace changes, might be easier to review that code locally. 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/target-page-controller.ts
+++ b/src/background/target-page-controller.ts
@@ -67,8 +67,8 @@ export class TargetPageController {
     private removeKnownTabId = async (tabId: number, context: TabContext) => {
         if (this.knownTabIds.includes(tabId)) {
             this.knownTabIds.splice(this.knownTabIds.indexOf(tabId, 0), 1);
+            context.teardown();
             if (this.persistStoreData) {
-                context.teardown();
                 await this.idbInstance.setItem(IndexedDBDataKeys.knownTabIds, this.knownTabIds);
             }
         }
--- a/src/background/target-page-controller.ts
+++ b/src/background/target-page-controller.ts
@@ -22,6 +22,7 @@ export class TargetPageController {
         private readonly logger: Logger,
         private readonly knownTabIds: number[],
         private readonly idbInstance: IndexedDBAPI,
+        private persistStoreData = false,
     ) {}
 
     public async initialize(): Promise<void> {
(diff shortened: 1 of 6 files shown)


src/common/flux/persistent-store.ts
```diff
@@ -12,13 +12,16 @@ export abstract class PersistentStore<TState> extends BaseStoreImpl<TState> {
...
     ) {
         super(storeName);
     }
 
     protected async persistData(storeData: any): Promise<boolean> {
-        return await this.idbInstance.setItem(this.indexedDBDataKey, storeData);
+        if (this.persistStoreData) {
+            return await this.idbInstance.setItem(this.indexedDBDataKey, storeData);
+        }
+        return Promise.resolve(true);
```
dbjorge, line 24: nit: it's preferable to just `return true;` here and let the async machinery do the `Promise`-wrapping - it results in more readable stack traces if the `Promise` ends up involved in an `Error` path.

src/background/target-page-controller.ts
```diff
@@ -57,15 +58,19 @@ export class TargetPageController {
...
         }
     };
 
     private removeKnownTabId = async (tabId: number, context: TabContext) => {
         if (this.knownTabIds.includes(tabId)) {
             this.knownTabIds.splice(this.knownTabIds.indexOf(tabId, 0), 1);
-            context.teardown();
-            await this.idbInstance.setItem(IndexedDBDataKeys.knownTabIds, this.knownTabIds);
+            if (this.persistStoreData) {
+                context.teardown();
```
dbjorge, line 71: Should this be outside the `if`? I think it should probably be the context's responsibility to decide whether its `teardown` work involves persisting store data or not. (I think in practice it just defers to stores' `teardown`s, which should already handle deciding whether to skip idb updates or not appropriately with the `PersistentStore.teardown()` change you've already made

dbjorge (approved):
Built the MV2 version locally and verified that I no longer see any unexpected values in idb! This looks good to me modulo the two nit comments

PR #5415, opened 2022-04-25
feat(mv3-part-2): Consolidate store proxy message handling

pownkel:
#### Details

Creates StoreUpdateMesssageDistributor to distribute update messages to the appropriate StoreProxy, if one is registered.

##### Motivation

As mentioned in #5396 and #5407, we will eventually need to combine message handling into one event listener (per context) that distributes messages to their respective handlers. Currently, each StoreProxy registers its own separate onMessage handler. This change creates one central message listener that distributes store update messages to the appropriate StoreProxy.

##### Context

This continues a change from #5407. After this PR, all contexts will have only one registered onMessage listener at the browser level.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/constants/generic-store-messages-types.ts
+++ b/src/common/constants/generic-store-messages-types.ts
@@ -1,5 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-export class GenericStoreMessageTypes {
-    public static readonly storeStateChanged: string = 'STORE_CHANGED';
+export enum GenericStoreMessageTypes {
+    storeStateChanged = 'STORE_CHANGED',
 }
--- a/src/common/constants/generic-store-messages-types.ts
+++ b/src/common/constants/generic-store-messages-types.ts
@@ -1,5 +1,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-export enum GenericStoreMessageTypes {
-    storeStateChanged = 'STORE_CHANGED',
-}
--- a/src/common/state-dispatcher.ts
+++ b/src/common/state-dispatcher.ts
@@ -3,8 +3,7 @@
 import { StoreHub } from 'background/stores/store-hub';
 import { Logger } from 'common/logging/logger';
 import { BaseStore } from './base-store';
-import { GenericStoreMessageTypes } from './constants/generic-store-messages-types';
-import { StoreUpdateMessage } from './types/store-update-message';
+import { StoreUpdateMessage, storeUpdateMessageType } from './types/store-update-message';
 
 export class StateDispatcher {
     constructor(
@@ -28,7 +27,7 @@ export class StateDispatcher {
             this.broadcastMessage({
                 isStoreUpdateMessage: true,
                 storeId: store.getId(),
-                messageType: GenericStoreMessageTypes.storeStateChanged,
+                messageType: storeUpdateMessageType,
                 storeType: this.stores.getStoreType(),
                 payload: store.getState(),
             }).catch(this.logger.error);
--- a/src/common/state-dispatcher.ts
+++ b/src/common/state-dispatcher.ts
@@ -25,7 +25,6 @@ export class StateDispatcher {
     private getDispatchStateUpdateEvent = (store: BaseStore<any>): (() => void) => {
         return () => {
             this.broadcastMessage({
-                isStoreUpdateMessage: true,
                 storeId: store.getId(),
                 messageType: storeUpdateMessageType,
                 storeType: this.stores.getStoreType(),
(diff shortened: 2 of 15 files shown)


src/common/store-proxy.ts
```diff
@@ -1,33 +1,25 @@
...
+        this.messageDistributor.registerStoreUpdateListener(storeId, this.onChange);
     }
 
     private onChange = (message: StoreUpdateMessage<TState>): void => {
-        if (!this.isValidMessage(message)) {
-            return;
-        }
-
         if (
             message.messageType === GenericStoreMessageTypes.storeStateChanged &&
```
dbjorge, line 24: It feels off for this to be handled by the `StoreProxy` when the new usage expectation is that the distributor has already validated it's a store update message before passing it here. It looks like this check and the `message.isStoreUpdateMessage` check in L46 of the distributor are redundant; maybe move this check over to there and update the `StoreUpdateMessage` type to drop `isStoreUpdateMessage` entirely and retype `messageType` as the constant `GenericStoreMessageTypes.storeStateChanged` instead of `string`?

src/common/store-update-message-distributor.ts
```diff
@@ -0,0 +1,58 @@
...
+            message.isStoreUpdateMessage &&
+            message.storeId &&
+            message.payload &&
+            (this.isMessageForCurrentTab(message) || message.storeType === StoreType.GlobalStore)
+        );
+    }
+
+    private isMessageForCurrentTab(message: StoreUpdateMessage<any>): boolean {
+        return (
+            this.tabId == null || message.tabId === this.tabId // tabid will be null on inital state in content script of target page
```
dbjorge, line 55: I think this comment is outdated and should be removed

* `tabId` will be `undefined`, not `null`, in the cases where it's unspecified
* `tabId` is `undefined` in more contexts than just content scripts
* There's no mechanism by which `tabId` will be updated later in content scripts (or any other case where it isn't provided)

dbjorge (changes requested):
Looks good overall! A few minor suggestions inline

src/common/store-update-message-distributor.ts
```diff
@@ -43,16 +45,14 @@ export class StoreUpdateMessageDistributor {
...
             message.payload &&
             (this.isMessageForCurrentTab(message) || message.storeType === StoreType.GlobalStore)
         );
     }
 
     private isMessageForCurrentTab(message: StoreUpdateMessage<any>): boolean {
-        return (
-            this.tabId == null || message.tabId === this.tabId // tabid will be null on inital state in content script of target page
-        );
+        return _.isEmpty(this.tabId) || message.tabId === this.tabId;
```
dbjorge, line 56: I don't think `isEmpty` is right here; it returns `true` for any integer. I think the old check was correct, just the old comment was off.

If this passes tests I think it indicates that there's a missing test case, though.

src/common/constants/generic-store-messages-types.ts
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-export class GenericStoreMessageTypes {
-    public static readonly storeStateChanged: string = 'STORE_CHANGED';
+export enum GenericStoreMessageTypes {
```
dbjorge, line 3: I wouldn't recommend using a TypeScript enum for a type that has to pass over a browser serialization boundary; even if it does work, I don't want every maintainer that thinks about that to have to wonder "wait, does this work?"

I'd recommend just using a string constant, like what `src/common/messages.ts` does. (I think moving this to `messages.ts` would be reasonable, if you wanted)

src/tests/unit/tests/common/store-proxy.test.ts
```diff
@@ -78,17 +77,16 @@ describe('StoreProxyTest', () => {
             tabId: 1,
             storeType: StoreType.TabContextStore,
             storeId: 'TestStore',
-            isStoreUpdateMessage: true,
             payload: 'store state',
-        } as StoreUpdateMessage<string>);
+        } as unknown as StoreUpdateMessage<string>);
```
dbjorge, line 81: I think these last 2 test cases should probably just be removed, they're the distributor's responsibility now.

src/common/state-dispatcher.ts
```diff
@@ -28,7 +27,7 @@ export class StateDispatcher {
             this.broadcastMessage({
                 isStoreUpdateMessage: true,
```
dbjorge, line 28: It looks like there's still a few dangling references to the removed `isStoreUpdateMessage` (here, `state-dispatcher.test.ts`, `store-proxy.test.ts`)

dbjorge (approved):

PR #5426, opened 2022-04-26
feat(mv3-part-2): Only use persisted tab info if tabs are still valid after service worker restart

sfoslund:
#### Details

Adds logic to determine if a tab url has changed between service worker shut down and restart. This ensures that we correctly handle url changes even if they happen with the service worker was down.

##### Motivation

Feature work.

##### Context

This change required us to restructure knownTabIds such that it now stores and persists both the tab ID and url. This allows us to check that the urls haven't changed between service worker shut down and restart.

This change does not effect the mv2 extension because it does not persist any data and therefore it never has any known tabs on target page controller initialization.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/background/target-page-controller.test.ts
+++ b/src/tests/unit/tests/background/target-page-controller.test.ts
@@ -592,7 +592,6 @@ describe('TargetPageController', () => {
 
             describe('onConnect', () => {
                 it('should not have any observable effect', () => {
-                    idbInstanceMock.reset();
                     setupDatabaseInstance(Times.never());
 
                     mockBrowserAdapter.notifyOnConnect({
@@ -672,10 +671,6 @@ describe('TargetPageController', () => {
             });
 
             describe('onTabRemoved', () => {
-                beforeEach(() => {
-                    idbInstanceMock.reset();
-                });
-
                 it('should ignore removals of non-tracked tabs', () => {
                     setupDatabaseInstance(Times.never());
                     setupTeardownInstance(Times.once());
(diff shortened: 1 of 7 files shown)


src/tests/unit/tests/background/target-page-controller.test.ts
```diff
@@ -512,7 +592,8 @@ describe('TargetPageController', () => {
 
             describe('onConnect', () => {
                 it('should not have any observable effect', () => {
-                    setupDatabaseInstance(null, Times.never());
+                    idbInstanceMock.reset();
```
pownkel, line 595: Do we need to reset the mock here? it looks like it's already handled in the beforeEach one level up

src/tests/unit/tests/background/target-page-controller.test.ts
```diff
@@ -586,15 +672,19 @@ describe('TargetPageController', () => {
             });
 
             describe('onTabRemoved', () => {
+                beforeEach(() => {
+                    idbInstanceMock.reset();
```
pownkel, line 676: Same comment here

src/tests/unit/tests/background/target-page-controller.test.ts
```diff
@@ -632,7 +722,8 @@ describe('TargetPageController', () => {
 
             describe('onDetailsViewTabRemoved', () => {
                 beforeEach(() => {
-                    setupDatabaseInstance(null, Times.never());
+                    idbInstanceMock.reset();
```
pownkel, line 725: Same comment here

pownkel (approved):
LGTM, one minor nit

PR #5436, opened 2022-04-27
feat(mv3-part-3): Move tabIdToContextMap out of TargetPageController

pownkel:
#### Details

Moves tabIdToContextMap out of TargetPageController and into a new TabContextManager object

##### Motivation

This is part of a larger refactor of TargetPageController and ExtensionDetailsViewController, with the goal of simplifying the tangled dependency between them and removing duplicate tab event listeners that are registered in both.

##### Context

Currently, TargetPageController depends on ExtensionDetailsViewController for the sole purpose of passing ExtensionDetailsViewController a callback function, onDetailsViewTabRemoved. This function is only called from and used by ExtensionDetailsViewController, so it's strange for it to live in TargetPageController as a private method. It's there because it needs access to the TabContext to send a message the target page, and tabIdToContextMap lives in TargetPageController. Its only function is to use that TabContext's interpreter to interpret a message.

This PR moves tabIdToContextMap into another object, TabContextManager, which wraps the current uses of tabIdToContextMap. Currently, TargetPageController has sole access to and responsibility of TabContextManager, but a future PR will add it to ExtensionDetailsViewController so they don't need to depend on each other and pass callbacks back and forth.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/tab-context-manager.ts
+++ b/src/background/tab-context-manager.ts
@@ -1,22 +1,23 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { TabContextFactory } from 'background/tab-context-factory';
 import { Message } from '../common/message';
-import { TabContext, TabToContextMap } from './tab-context';
+import { TabToContextMap } from './tab-context';
 
 export class TabContextManager {
     constructor(private readonly targetPageTabIdToContextMap: TabToContextMap) {}
 
-    public addTabContextIfNotExists(tabId: number, tabContext: TabContext): void {
-        if (this.targetPageTabIdToContextMap[tabId] === undefined) {
-            this.targetPageTabIdToContextMap[tabId] = tabContext;
+    public addTabContextIfNotExists(tabId: number, tabContextFactory: TabContextFactory): void {
+        if (!(tabId in this.targetPageTabIdToContextMap)) {
+            this.targetPageTabIdToContextMap[tabId] = tabContextFactory.createTabContext(tabId);
         }
     }
 
     public async deleteTabContext(tabId: number): Promise<void> {
-        const tabContext = this.targetPageTabIdToContextMap[tabId];
-        if (tabContext) {
+        if (tabId in this.targetPageTabIdToContextMap) {
+            const tabContext = this.targetPageTabIdToContextMap[tabId];
             delete this.targetPageTabIdToContextMap[tabId];
-            await tabContext.teardown();
+            await tabContext?.teardown();
         }
     }
 
--- a/src/background/tab-context-manager.ts
+++ b/src/background/tab-context-manager.ts
@@ -33,7 +33,7 @@ export class TabContextManager {
         const tabContext = this.targetPageTabIdToContextMap[tabId];
         if (tabContext) {
             delete this.targetPageTabIdToContextMap[tabId];
-            await tabContext.teardown(); // Make sure this doesn't need to happen in a specific order with removeKnownTab
+            await tabContext.teardown();
         }
     }
 
--- a/src/background/tab-context-manager.ts
+++ b/src/background/tab-context-manager.ts
@@ -1,31 +1,14 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { BrowserAdapter } from '../common/browser-adapters/browser-adapter';
 import { Message } from '../common/message';
-import { BrowserMessageBroadcasterFactory } from './browser-message-broadcaster-factory';
-import { ExtensionDetailsViewController } from './extension-details-view-controller';
-import { TabToContextMap } from './tab-context';
-import { TabContextFactory } from './tab-context-factory';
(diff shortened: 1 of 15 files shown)


src/background/tab-context-manager.ts
```diff
@@ -0,0 +1,47 @@
...
+                tabId,
+                this.persistStoreData,
+            );
+        }
+    }
+
+    public async deleteTabContext(tabId: number): Promise<void> {
+        const tabContext = this.targetPageTabIdToContextMap[tabId];
+        if (tabContext) {
+            delete this.targetPageTabIdToContextMap[tabId];
```
sfoslund, line 35: nit: can we ever get into a situation where we have a tab ID key mapping to null/ undefined? If so, this would leave the key in the map instead of removing it, is that what we want? Same with in `addTabContextIfNotExists`, the if check doesn't check if the key exists or not, which seems to be a subtle change in behavior.

src/background/target-page-controller.ts
```diff
@@ -194,14 +168,6 @@ export class TargetPageController {
     }
 
     private sendTabVisibilityChangeAction(tabId: number, isHidden: boolean): void {
-        if (!this.hasTabContext(tabId)) {
-            return;
-        }
-        const tabContext = this.targetPageTabIdToContextMap[tabId];
-        if (tabContext == null) {
```
sfoslund, line 201: Not sure if this code is just really old or not, but here we used to be checking *both* that the tabID was a key in the map (in the first if statement) *and* that the key's value is non-null. This makes me think that we should be doing the same thing in the new code as well, but it doesn't appear that we are right now.

src/background/tab-context-manager.ts
```diff
@@ -0,0 +1,47 @@
...
+import { TabContextFactory } from './tab-context-factory';
+
+export class TabContextManager {
+    constructor(
+        private readonly targetPageTabIdToContextMap: TabToContextMap,
+        private readonly broadcasterFactory: BrowserMessageBroadcasterFactory,
+        private readonly browserAdapter: BrowserAdapter,
+        private readonly detailsViewController: ExtensionDetailsViewController,
+        private readonly tabContextFactory: TabContextFactory,
+        private persistStoreData = false,
```
sfoslund, line 17: Can we make this non-optional? It looks like we're not passing it in anywhere at the moment, when I think we do want to pass in true from the service worker init. Either way, I think it's more clear to explicitly specify in the init files if we're persisting or not.

src/tests/unit/tests/background/tab-context-manager.test.ts
```diff
@@ -0,0 +1,121 @@
...
+            )
+            .returns(() => tabContextMock.object)
+            .verifiable();
+
+        testSubject.addTabContextIfNotExists(tabId);
+
+        expect(tabToContextMap[tabId]).toBe(tabContextMock.object);
+    });
+
+    it('Does not recreate tab context if already exists', () => {
```
sfoslund, line 81: Again, not sure if we ever get into this state, but might be worthwhile adding a test case for when `tabToContextMap[tabId] = undefined`, if nothing else to define the behavior.

src/tests/unit/tests/background/target-page-controller.test.ts
```diff
@@ -1,40 +1,31 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import {
```
sfoslund, line 3: Yay for making this test file so much smaller!!! 🎉

sfoslund (commented):
Looks great overall, the test changes in particular are great- just a few nits/ questions

src/background/tab-context-manager.ts
pownkel: Good catch, I didn't realize those were different. I can't think of a situation where we would expect an undefined TabContext in this map, but I'll change it for consistency with what we had before.

src/background/tab-context-manager.ts
pownkel: Moved this flag into TabContextFactory where it's actually used and made it non-optional

src/tests/unit/tests/background/dev-tools-listener.test.ts
```diff
@@ -17,19 +16,19 @@ import { PortOnMessageMock } from '../../mock-helpers/port-on-message-mock';
...
 
     beforeEach(() => {
-        tabId1InterpreterMock = Mock.ofType(Interpreter, MockBehavior.Strict);
-        tabId2InterpreterMock = Mock.ofType(Interpreter, MockBehavior.Strict);
-        tabIdToContextMap = {
-            1: new TabContext(tabId1InterpreterMock.object, null),
-            2: new TabContext(tabId2InterpreterMock.object, null),
-        };
+        tabContextManagerMock = Mock.ofType<TabContextManager>();
+        // tabIdToContextMap = {
```
dbjorge, line 23: Can this comment be removed?

dbjorge (approved):
LGTM modulo one stray comment in a test!

PR #5453, opened 2022-05-03
feat(mv3-part-3): Replace debug tools telemetry connection with messages

pownkel:
#### Details

Remove the use of connect() for telemetry in debug tools and use messages instead.

Related, minor refactors in this PR:
- Rename MessageDistributor to BackgroundMessageDistributor (to distinguish it from DebugToolsMessageDistributor)
- Rename StoreUpdateMessageDistributor to StoreUpdateMessageHub and don't have it register its own message listener (because some contexts only use the StoreUpdateMessageHub message listener, but the debug tools context needs to listen for store update messages as well as telemetry messages, and we need to keep one listener per event)

##### Motivation

When we update to MV3, open connections are likely to trigger a watchdog timeout that will kill our service workers after five minutes. Replacing these open connections with messaging will avoid this.

##### Context

Other uses of connect() will be addressed in future PRs.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/telemetry/debug-tools-telemetry-client.ts
+++ b/src/background/telemetry/debug-tools-telemetry-client.ts
@@ -1,55 +1,44 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { FeatureFlagChecker } from 'background/feature-flag-checker';
 import { ApplicationTelemetryDataFactory } from 'background/telemetry/application-telemetry-data-factory';
 import { TelemetryClient } from 'background/telemetry/telemetry-client';
 import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
-import { ConnectionNames } from 'common/constants/connection-names';
+import { FeatureFlags } from 'common/feature-flags';
+import { Messages } from 'common/messages';
 
 export class DebugToolsTelemetryClient implements TelemetryClient {
-    private connections: chrome.runtime.Port[] = [];
+    private featureFlagChecker: FeatureFlagChecker;
 
     constructor(
         private readonly browserAdapter: BrowserAdapter,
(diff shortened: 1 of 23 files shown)


src/background/telemetry/debug-tools-telemetry-client.ts
```diff
@@ -1,55 +1,44 @@
...
-        }
-
-        const finalProperties = {
-            ...properties,
-            ...this.telemetryDataFactory.getData(),
-        };
-
-        this.connections.forEach(connection =>
-            connection.postMessage({
+        if (this.featureFlagChecker?.isEnabled(FeatureFlags.debugTools)) {
```
pownkel, line 31: This check is not strictly necessary, but will reduce unnecessary messages. The current state (before this PR) only sends if there is a connection open, which we can no longer check for.

src/tests/unit/tests/debug-tools/debug-tools-message-distributor.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    let testSubject: DebugToolsMessageDistributor;
+
+    beforeEach(() => {
+        browserAdapterMock = Mock.ofType<BrowserAdapter>();
+        telemetryListenerMock = Mock.ofType<TelemetryListener>();
+        storeUpdateHubMock = Mock.ofType<StoreUpdateMessageHub>();
+
+        browserAdapterMock
+            .setup(b => b.addListenerOnMessage(It.isAny()))
+            .returns(listener => (registeredListener = listener));
```
sfoslund, line 26: nit: doesn't matter functionally since we check if the registered listener is defined later, but it might be more clear if you verify that this is called exactly once after init too.

sfoslund (approved):
LGTM!

PR #5465, opened 2022-05-09
feat(mv3-part-3): DevToolsMonitor

pownkel:
#### Details

Creates DevToolsMonitor to poll active devtools for their status every 500ms until they close. It will only poll if there are active devtools. With this PR, DevToolsMonitor will poll alongside the browser connection that we currently use to detect devtools opening and closing, but the plan is for it to eventually replace that connection.

##### Motivation

Because of the switch to MV3, we will need to replace uses of browser.connect() with message passing. However, since we use the devtools connection to detect when a devtool has closed in addition to when it opened, there isn't an easy 1:1 replacement and we'll need to implement another mechanism to detect when a devtool is no longer open. This PR implements a polling system for that purpose.

##### Context

Future PRs will:
- Persist the list of active devtools in DevToolsMonitor in case the service worker dies unexpectedly
- Remove the browser connection, send a message when a DevTool is initialized, and only use DevToolsMonitor to detect when it closes

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/dev-tools-monitor.ts
+++ b/src/background/dev-tools-monitor.ts
@@ -1,0 +1,96 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { BrowserAdapter } from 'common/browser-adapters/browser-adapter';
+import { Messages } from 'common/messages';
+import { PromiseFactory, TimeoutError } from 'common/promises/promise-factory';
+import { DevToolsStatusResponse } from 'common/types/dev-tools-messages';
+import _ from 'lodash';
+
+export class DevToolsMonitor {
+    constructor(
+        private readonly browserAdapter: BrowserAdapter,
+        private readonly promiseFactory: PromiseFactory,
+        private readonly activeDevtoolTabIds: number[],
+        private readonly messageTimeoutMilliseconds = 5000,
+        private readonly pollIntervalMilliseconds = 1000,
+    ) {}
+
+    public initialize() {
+        // TODO: initialize with persisted activeDevtoolTabIds, start monitor if needed
+    }
+
+    public monitorActiveDevtool(tabId: number): void {
+        this.addActiveDevtool(tabId);
+
+        if (this.activeDevtoolTabIds.length === 1) {
+            // Do not await, we want to continue running other synchronous code
+            this.waitForAllDevtoolsClosed();
+        }
+    }
+
+    private addActiveDevtool(tabId: number): void {
+        if (!_.isNil(tabId) && !this.activeDevtoolTabIds.includes(tabId)) {
+            this.activeDevtoolTabIds.push(tabId);
+        }
+        // TODO: persist activeDevtoolTabIds
+    }
+
+    private removeActiveDevtool(tabId: number): void {
+        const index = this.activeDevtoolTabIds.indexOf(tabId);
+        if (index >= 0) {
+            this.activeDevtoolTabIds.splice(index, 1);
+        }
+        // TODO: persist activeDevtoolTabIds
+    }
+
+    private async waitForAllDevtoolsClosed(): Promise<void> {
+        while (this.activeDevtoolTabIds.length > 0) {
+            await this.promiseFactory.delay(null, this.pollIntervalMilliseconds);
+            await this.removeInactiveDevtools();
+        }
+    }
+
+    private async removeInactiveDevtools(): Promise<void> {
+        const idsToRemove: number[] = [];
+
+        await Promise.all(
+            this.activeDevtoolTabIds.map(async tabId => {
+                const isOpen = await this.isDevtoolOpen(tabId);
+                if (!isOpen) {
+                    idsToRemove.push(tabId);
+                }
+            }),
+        );
+
+        idsToRemove.forEach(tabId => {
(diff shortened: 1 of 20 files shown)


src/background/dev-tools-monitor.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private removeActiveDevtool(tabId: number): void {
+        const index = this.activeDevtoolTabIds.indexOf(tabId);
+        if (index >= 0) {
+            this.activeDevtoolTabIds.splice(index, 1);
+        }
+    }
+
+    protected async monitorActiveDevtools(): Promise<void> {
```
sfoslund, line 44: Why does this need to be protected? I don't see any classes extending this one, is that something that will be added in the future?

src/background/dev-tools-monitor.ts
```diff
@@ -0,0 +1,97 @@
...
+import { DevToolsStatusResponse } from 'common/types/dev-tools-messages';
+import _ from 'lodash';
+
+export class DevToolsMonitor {
+    constructor(
+        private readonly browserAdapter: BrowserAdapter,
+        private readonly promiseFactory: PromiseFactory,
+        protected readonly activeDevtoolTabIds: number[],
+        private readonly tabContextManager: TabContextManager,
+        private readonly messageTimeoutMilliseconds = 5000,
```
sfoslund, line 18: The PR description says that the timeout will be 500 ms but this has an extra 0 on the end- which is the correct timeout?

src/background/dev-tools-monitor.ts
```diff
@@ -0,0 +1,97 @@
...
+        this.addActiveDevtool(tabId);
+
+        if (this.activeDevtoolTabIds.length === 1) {
+            // Do not await, we want to continue running other synchronous code
+            this.monitorActiveDevtools();
+        }
+    }
+
+    private addActiveDevtool(tabId: number): void {
+        if (!_.isNil(tabId) && !this.activeDevtoolTabIds.includes(tabId)) {
```
sfoslund, line 32: nit: Personal preference, but I don't love the `_.` syntax, would prefer to change the lodash import so we can call isNil directly without this.

src/tests/unit/tests/background/dev-tools-monitor.test.ts
```diff
@@ -0,0 +1,237 @@
...
+import _ from 'lodash';
+import { flushSettledPromises } from 'tests/common/flush-settled-promises';
+import { IMock, It, Mock, Times } from 'typemoq';
+import { DictionaryNumberTo } from 'types/common-types';
+
+class TestDevToolsMonitor extends DevToolsMonitor {
+    public activeDevtoolTabIds: number[];
+
+    public async testPollLoop(): Promise<void> {
+        await super.monitorActiveDevtools();
```
sfoslund, line 24: Ah, I see now why this has to be protected... I don't love that we can't keep this private but I don't see an alternative if we want to have good test coverage. :/

src/tests/unit/tests/background/dev-tools-monitor.test.ts
```diff
@@ -0,0 +1,237 @@
...
+        promiseFactory = {
+            timeout: timeoutMock.object,
+            delay: delayMock.object,
+        } as PromiseFactory;
+        tabContextManagerMock = Mock.ofType<TabContextManager>();
+
+        testSubject = new TestDevToolsMonitor(
+            browserAdapterMock.object,
+            promiseFactory,
+            [],
```
sfoslund, line 58: Can we add some test coverage for when there are known tabs ids? We can also wait and include this in the next PR, when this info actually starts persisting.

sfoslund (approved):
A couple of comments but nothing blocking- looks good!

src/background/dev-tools-monitor.ts
pownkel: My mistake, I changed some of the timeouts while I was testing and forgot to put them back

src/tests/unit/tests/background/dev-tools-monitor.test.ts
pownkel: I'll add this to the next PR, since right now there's no functional difference between passing the list to the constructor and changing the list directly after the object is initialized

src/tests/unit/tests/background/dev-tools-monitor.test.ts
pownkel: The main reason is to give us a way to await the polling loop and/or catch exceptions in tests, since the public function is intentionally non-asynchronous. I agree that it's awkward, but I also couldn't think of another way to do this without pushing the unawaited promise somewhere else where the intention might be less clear. I'm open to alternatives if there are any suggestions!

src/background/dev-tools-monitor.ts
pownkel: This function is overwritten to be public in a test class (see comment below)

src/tests/unit/tests/background/dev-tools-monitor.test.ts
sfoslund: Makes sense, I can't think of a better way either so this looks fine.

PR #5490, opened 2022-05-12
feat(mv3-part-3): Register browser listeners on init

sfoslund:
#### Details

Add logic to register browser event listeners synchronously on startup and incorporate the browser-event-manager (added in #5380 and #5396) to maintain the set of listeners. 

##### Motivation

Feature work

##### Context

This change is necessary for the mv3 extension and effects both the mv2 and mv3 extensions, although it should not effect the user facing behavior of the mv2 extension. 

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/store-update-message-hub.ts
```diff
@@ -22,15 +22,16 @@ export class StoreUpdateMessageHub {
         this.registeredUpdateListeners[storeId] = listener;
     }
 
-    public readonly handleMessage = (message: StoreUpdateMessage<any>): void => {
+    public readonly handleMessage = (message: StoreUpdateMessage<any>): Promise<void> => {
```
pownkel, line 25: It feels odd to me to artificially return a promise from this method when it isn't always a top-level message listener (ex. in DevToolMessageDistributor). Could this function stay synchronous and be wrapped in an async function instead when it's the only listener?

src/background/background-message-distributor.ts
```diff
@@ -43,6 +43,8 @@ export class BackgroundMessageDistributor {
 
         if (response) {
             return response;
+        } else {
+            return Promise.resolve();
```
dbjorge, line 47: This is going to return `Promise.resolve()` for all `interpreter` paths. I don't think that's what we want to do, since that's masking the fact that those paths are all fire-and-forget cases we haven't yet async-ified.

Instead, I think it would be preferable to only invent a `Promise.resolve()` for the case where we don't know how to interpret it at all (L41), and leave the interpreter paths returning `undefined` until we come up with a mechanism to enable starting to make them async.

src/injected/client-store-listener.ts
```diff
@@ -41,7 +41,7 @@ export class ClientStoreListener {
 
     private onChangedState = (): void => {
         const storeData = this.storeHub.getAllStoreData();
-        if (storeData == null) {
+        if (storeData == null || !storeData.visualizationStoreData) {
```
dbjorge, line 44: nit:

```suggestion
        if (storeData?.visualizationStoreData == null) {
```

src/common/browser-adapters/browser-adapter-factory.ts
```diff
@@ -1,18 +1,41 @@
...
-            return new FirefoxAdapter();
+            return new FirefoxAdapter(browserEventManager, browserEvents);
         } else {
-            return new ChromiumAdapter();
+            return new ChromiumAdapter(browserEventManager, browserEvents);
         }
     }
+
+    protected getDefaultBrowserEvents(): DictionaryStringTo<Events.Event<any>> {
+        const browserEvents: DictionaryStringTo<Events.Event<any>> = {
```
dbjorge, line 25: This specific list is tightly coupled to the `webextension-browser-adapter` implementation referencing the keys, and it isn't used anywhere else, so I think it would make more sense for this to live there.

src/background/service-worker-init.ts
```diff
@@ -46,7 +47,15 @@ import { cleanKeysFromStorage } from './user-stored-data-cleaner';
 async function initialize(): Promise<void> {
     const userAgentParser = new UAParser(globalThis.navigator.userAgent);
     const browserAdapterFactory = new BrowserAdapterFactory(userAgentParser);
-    const browserAdapter = browserAdapterFactory.makeFromUserAgent();
+    const logger = createDefaultLogger();
+    const promiseFactory = createDefaultPromiseFactory();
+    const FIRE_AND_FORGET_EVENT_DELAY_MS = 2 * 60 * 1000; // 2 minutes
```
dbjorge, line 52: Without the comment explaining its purpose near it, this feels like a magic number and it's hard to understand why 2 minutes makes sense for the value. I feel like it's easier for a reason to understand whether 2 minutes sounds reasonable if it's explained in the context of the rest of the browser event manager code, so my preference would be to leave the constant there and to use a boolean `isServiceWorker` for the ctor parameter.

src/common/browser-adapters/webextension-browser-adapter.ts
```diff
@@ -9,6 +18,20 @@ import { StorageAdapter } from './storage-adapter';
...
+    private browserEventManager: BrowserEventManager;
+
+    constructor(
+        browserEventManager: BrowserEventManager,
+        browserEvents: DictionaryStringTo<Events.Event<any>>,
+    ) {
+        this.browserEventManager = browserEventManager;
+        for (const [eventType, event] of Object.entries(browserEvents)) {
+            if (event !== undefined) {
+                this.browserEventManager.addBrowserListener(event, eventType);
```
dbjorge, line 30: It would be good leave a comment explaining that it's important that these calls happen synchronously from a service worker initializer; that probably won't be obvious to a future maintainer

src/common/store-update-message-hub.ts
dbjorge: Yes, I think for our purposes it will be important to expand the return types of our assorted intermediate message handlers (eg, this and `Interpreter`) to be able to distinguish between 3 cases:

* "I didn't handle this message"
* "I handled this message via an async handler; here is the `Promise` tracking its work"
* "I handled this message via a fire-and-forget handler"

This is essentially what `PostMessageContentHandler` already does by returning a `BackchannelMessageResponse`; I would support generalizing and re-using it.

src/common/store-update-message-hub.ts
```diff
@@ -22,15 +22,16 @@ export class StoreUpdateMessageHub {
...
         if (!this.isValidMessage(message)) {
-            return;
+            return Promise.resolve();
         }
 
         const listener = this.registeredUpdateListeners[message.storeId];
         if (listener) {
             listener(message);
```
dbjorge, line 32: Particularly, this path through a fire-and-forget `listener` *must not* return `Promise.resolve()`; it needs to be triggering the fire-and-forget delay for now.

dbjorge (changes requested):
Thanks Sarah! Some suggestions inline

dbjorge:
#5491 removes the `connect()` method from `BrowserAdapter` along with its corresponding listener, FYI that that will be a semantic merge conflict for this PR

src/injected/focus-change-handler.ts
```diff
@@ -47,13 +47,17 @@ export class FocusChangeHandler {
 
     private automatedChecksDataIsPopulated(storeData: TargetPageStoreData): boolean {
         return (
+            storeData.cardSelectionStoreData &&
+            storeData.unifiedScanResultStoreData &&
             storeData.cardSelectionStoreData.focusedResultUid !== null &&
```
pownkel, line 52: nit: Can these null checks all use the same syntax? (either `if(thing)` or `if(thing !== null)` for all three) (same comment for lines 59-61 in this file)

src/common/browser-adapters/browser-event-manager.ts
```diff
@@ -46,18 +54,17 @@ export class BrowserEventManager {
...
-    ) {}
+        isServiceWorker: boolean = false,
+    ) {
+        if (isServiceWorker) {
+            this.fireAndForgetEventDelayMs = FIRE_AND_FORGET_EVENT_DELAY_MS;
+        }
+    }
 
     public addApplicationListener = (eventType: string, callback: ApplicationListener): void => {
         if (this.eventsToApplicationListenersMapping[eventType]) {
```
dbjorge, line 70: I think there should be a check in here verifying that a browser listener already exists before registering a corresponding application listener; otherwise, if we miss adding a browser event to the provider, it'll result in silently dropping events rather than an actionable error

src/background/service-worker-init.ts
```diff
@@ -49,13 +50,12 @@ async function initialize(): Promise<void> {
...
     const logger = createDefaultLogger();
     const promiseFactory = createDefaultPromiseFactory();
-    const FIRE_AND_FORGET_EVENT_DELAY_MS = 2 * 60 * 1000; // 2 minutes
-    const browserEventManager = new BrowserEventManager(
-        promiseFactory,
-        logger,
-        FIRE_AND_FORGET_EVENT_DELAY_MS,
+    const browserEventProvider = new BrowserEventProvider();
+    const browserEventManager = new BrowserEventManager(promiseFactory, logger, true);
+    const browserAdapter = browserAdapterFactory.makeFromUserAgent(
```
dbjorge, line 55: It would be good to leave a comment here explaining why it is important that this happens before any usage of `await` inside the initializer

src/background/service-worker-init.ts
```diff
@@ -52,6 +52,7 @@ async function initialize(): Promise<void> {
     const promiseFactory = createDefaultPromiseFactory();
     const browserEventProvider = new BrowserEventProvider();
     const browserEventManager = new BrowserEventManager(promiseFactory, logger, true);
+    // Create browser adapter before any async calls to ensure that browser listeners are registered first
```
dbjorge, line 55: This comment is explaining what is happening, but not *why* it's happening, which I think is generally much more valuable information for future maintainers to have available. A reader can already see that the adapter is being created before any async calls, but a reader without context won't know that the reason why this is important is that if the service worker does not register all of its browser listeners *synchronously* during worker initialization, the browser may tear down the worker as being "done" before we ever register any listeners.

```suggestion
    // It is important that the browser adapter gets initialized *before* any "await" statement.
    //
    // If a service worker does not register all of its browser listeners *synchronously* during worker initialization,
    // the browser may decide that the worker is "done" as soon as the synchronous part of initialization finishes
    // and tear down the worker before we tell it which events to wake us back up for.
```

PR #5491, opened 2022-05-12
feat(mv3-part-3): replace last use of connect() with sendMessage()

dbjorge:
#### Details

This PR removes the last usage of `browser.runtime.connect()` from the extension. 

This usage was in our injected content scripts. We didn't use it for message-passing; it was used strictly so we could use an onDisconnect handler to detect when the extension had become disabled and clean up any active visualizations appropriately.

This PR replaces that with a mechanism that polls for the availability of `browser.runtime.id` once per second and considers it "disposed" if the browser doesn't respond with an id. This works because once the extension is disabled/uninstalled, the browser disconnects the running content script from its extension context, and any query for `browser.runtime.id` after that point returns undefined.

##### Motivation

Removing usage of `connect()` is important generally because indefinitely-lived `Port`s established by `connect()` are treated by the browser the same as an unresponded message; after 5 minutes of the port being open, the browser will consider it to be "timed out" and tear down the service worker, even if the worker has outstanding work in progress.

This specific usage of `connect()` would have been broken in MV3 regardless of that, though; it would detect disconnections when the service worker goes to sleep even if the extension isn't disabled, so it would have disabled visualizations prematurely.

##### Context

I originally implemented the polling mechanism in terms of sending a "ping" message to the background instead of querying for `browser.runtime.id`. I think the id query feels a little jankier, but the sendMessage version would have had the undesirable side effect of keeping the background worker awake indefinitely as long as any injected content script is present in any open browser tab, which I worry would have risked hiding other lifetime error cases.

#### Pull request checklist

- [x] Addresses an existing issue: 1943378
- [x] Ran `yarn fastpass`


pownkel (approved):
LGTM, glad we can poll for this without messaging the background/service worker!

PR #5512, opened 2022-05-17
feat(mv3-part-4): Fix tab message errors

pownkel:
#### Details

- In injected scripts, handle a scenario where the storeHub contains stores, but not all stores have their state initialized
- Change the signature of getAllStoreData() to return a Partial instead of assuming all storeData are present

##### Motivation

This fixes frequent error messages in the MV3 extension when a scan is started. The same error occurs silently in the MV2 extension, but does not affect the functionality of the extension.

##### Context

The cause of this bug is that the injected content scripts register some store update listeners that assume the store's state already exists, which isn't the case when the scripts are first initialized, so the initial store update message throws an error. This PR assumes that none of the listeners registered by ClientStoreListener need to be called before the store state is initialized (the listener that updates the StoreProxy state is called separately).

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/components/with-store-subscription.tsx
+++ b/src/common/components/with-store-subscription.tsx
@@ -17,16 +17,16 @@ export type WithStoreSubscriptionDeps<T> = {
 
 export function withStoreSubscription<P extends WithStoreSubscriptionProps<S>, S>(
     WrappedComponent: React.ComponentType<P>,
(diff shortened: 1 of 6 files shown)


sfoslund (approved):
LGTM!

PR #5517, opened 2022-05-18
chore: turn on no-floating-promises for most directories

pownkel:
#### Details

Turns on the [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md) eslint rule for all directories except background/ and injected/, and fixes the errors that were caught.

This PR fixes 26 issues caught by this rule out of 48 in total. The issues fixed in this PR are fairly minor, and the remaining issues in background/ and injected/ are more complex and involve changes that propagate more to other files. Those issues will be addressed in future PRs.

##### Motivation

The no-floating-promises rule was turned off when we migrated from tslint to eslint due to a large number of failures, and was never reenabled. This would be an extremely useful rule to have, especially with the ongoing migration to Manifest V3, to make sure all promises are correctly handled and we aren't unintentionally orphaning promises.

##### Context

As stated above, the remaining 22 warnings for no-floating-promises will be fixed in future PRs. It is likely that, in the course of the MV3 migration, more floating promises will need to be added before they are eventually resolved. Many of the current method signatures don't accurately reflect when async work is happening, and that will need to be changed. I expect this eslint rule to stay off for background/ and injected/ while that work continues.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/eslintrc.base.js
+++ b/eslintrc.base.js
@@ -63,7 +63,7 @@ module.exports = {
         '@typescript-eslint/no-unnecessary-type-assertion': 'off',
         '@typescript-eslint/no-misused-promises': 'off',
         '@typescript-eslint/restrict-template-expressions': 'off',
-        '@typescript-eslint/no-floating-promises': 'off',
+        '@typescript-eslint/no-floating-promises': ['warn'],
         '@typescript-eslint/require-await': 'off',
         '@typescript-eslint/no-implied-eval': 'off',
         '@typescript-eslint/prefer-regexp-exec': 'off',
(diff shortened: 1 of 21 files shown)


eslintrc.base.js
```diff
@@ -92,5 +92,11 @@ module.exports = {
                 'security/detect-eval-with-expression': 'off',
             },
         },
+        {
+            files: ['src/background/**/*', 'src/injected/**/*'],
+            rules: {
+                '@typescript-eslint/no-floating-promises': ['warn'],
```
dbjorge, line 98: I think we should probably leave this `off` rather than `warn` until we address the other issues; I don't think it makes sense to pollute unrelated PRs with warnings that aren't relevant to them

dbjorge (changes requested):
The code changes look reasonable to me! Left a few comments on the config changes

dbjorge (approved):

Issue #5412, opened 2022-04-23
Engineering: explore migration to esbuild


waabid:
I'm creating this issue to track an exploration into migrating this project from webpack to esbuild.

Latest branch: https://github.com/waabid/accessibility-insights-web/tree/esbuildTry2

Current status: local dev build is working in this branch (yarn build)

Some stats based on `yarn build` on my machine:
- esbuild: 15s on average.
- webpack: 45s on average.

## Changes that were made for esbuild to work
- All style imports must now be represented as: import stylesFromX from 'path/to/X.scss';
    - Previously, our imports were mostly like ```import * as stylesFromX from 'path/to/X.scss'```; esbuild had issues parsing this format. I believe this has to do with how the modules are created (only have default exports) but not 100% sure if this change is required. **If** you would like to look into this, make sure to validate the produced JS bundles/local dev build: it will build cleanly but the class names may not transfer from the scss parsing. This can also be seen when the logLevel is set to `debug`.
    - Similarly, changed how UAParser is imported.
- Created `style-plugin.js` which is now the home to our css-modules solution. 
- Embedding styles into our source map files fixed issues that generally plagued our sourcemap usage but more so for esbuild.
- Using dynamic imports in injected code prevented issues that seemed to be prevented by multiple injections (coincidentally happens in cross-frame e2e test).
- Esbuild doesn't support multiple entry-points creating one bundle file. To get around this, we: 
  - moved code from stylesheet-init.ts to client-init.ts
  - created new files pertaining to react-devtools where we would import the devtools and our init files and in our esbuild configuration we now specify these new init files when react-devtools is present.

## TODO

- [x] Appropriately name commands: currently it's all still named after webpack.
- [x] Figure out when/where we want to run type-checking: esbuild **does not** type check when compiling so we'll want to do that somewhere.
- [x] esbuild with tsc running sync: 35s on average. While this is only a modest decrease, I imagine we likely only run this for our CI/PR builds and likely do not include it in the local dev build script anyways.
- Ensuring the other packages being built by webpack can be built as well.
    - [x] web extension
    - [x] report package
    - [ ] ui package
    - [ ] unified
        - Running into errors at runtime. See here: https://github.com/microsoft/accessibility-insights-web/issues/5412#issuecomment-1128338484

## Nice to have
- Implement watch for our style-plugin
- Implement caching for our style-plugin

waabid:
Update: with a minor change, was able to validate the branch on Windows as well.

[...]

PR #5522, opened 2022-05-20
feat(esbuild): change to use default imports for styles

waabid:
#### Details

Changes all our style imports to use default exports/imports. Most of the changes are simply just import changes and then the others are to build default export typings and to load css through default import/exports.

##### Motivation

Necessary for migration to esbuild.

##### Context

This is the biggest necessary change for esbuild migration. As a result, I'm doing this separately such that it's a bit easier to review. Doing this separately means we can always switch back to webpack with ease (the point here being that these changes work with both webpack and esbuild).

I tested dev, prod, unified, and the report package locally to verify that everything looked good (and e2e tests).

#### Pull request checklist

- [x] Addresses an existing issue: #5412 
- [x] Ran `yarn fastpass`


dbjorge (approved):
LGTM! Thanks for breaking this out into its own PR

PR #5524, opened 2022-05-20
feat(mv3-part-4): Fix or suppress no-floating-promise issues in background directory

pownkel:
#### Details

Fixes or suppresses no-floating-promises issues in src/background and enables the eslint rule for that directory.

##### Motivation

Continuing work from #5517

##### Context

See #5517. This PR fixes 4 issues out of 22.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/telemetry/app-insights-telemetry-client.ts
+++ b/src/background/telemetry/app-insights-telemetry-client.ts
@@ -36,7 +36,7 @@ export class AppInsightsTelemetryClient implements TelemetryClient {
         this.updateTelemetryState();
     }
 
-    public async trackEvent(name: string, properties?: { [name: string]: string }): Promise<void> {
+    public trackEvent(name: string, properties?: { [name: string]: string }): void {
         if (this.enabled) {
             this.applicationInsights.trackEvent({ name }, properties);
         }
--- a/src/background/telemetry/app-insights-telemetry-client.ts
+++ b/src/background/telemetry/app-insights-telemetry-client.ts
@@ -36,7 +36,7 @@ export class AppInsightsTelemetryClient implements TelemetryClient {
         this.updateTelemetryState();
     }
 
-    public trackEvent(name: string, properties?: { [name: string]: string }): void {
+    public async trackEvent(name: string, properties?: { [name: string]: string }): Promise<void> {
         if (this.enabled) {
             this.applicationInsights.trackEvent({ name }, properties);
         }
--- a/src/background/telemetry/telemetry-event-handler.ts
+++ b/src/background/telemetry/telemetry-event-handler.ts
@@ -28,6 +28,7 @@ export class TelemetryEventHandler {
 
         const flattenTelemetryInfo: DictionaryStringTo<string> =
             this.flattenTelemetryInfo(telemetryInfo);
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         this.telemetryClient.trackEvent(eventName, flattenTelemetryInfo);
     }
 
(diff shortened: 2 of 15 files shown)


src/background/telemetry/telemetry-event-handler.ts
```diff
@@ -28,6 +28,7 @@ export class TelemetryEventHandler {
 
         const flattenTelemetryInfo: DictionaryStringTo<string> =
             this.flattenTelemetryInfo(telemetryInfo);
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         this.telemetryClient.trackEvent(eventName, flattenTelemetryInfo);
```
pownkel, line 32: Eventually, `publishTelemetry` should become async so this call can be awaited. However, `publishTelemetry` is called by almost every ActionMessageCreator, as well as some actions and store update listeners, so we'll need to make significant changes to the flux infrastructure code first. For now, we'll just disable the rule here, so it's more easily searchable than marking the promise `void`.

src/background/telemetry/app-insights-telemetry-client.ts
```diff
@@ -36,7 +36,7 @@ export class AppInsightsTelemetryClient implements TelemetryClient {
         this.updateTelemetryState();
     }
 
-    public trackEvent(name: string, properties?: { [name: string]: string }): void {
+    public async trackEvent(name: string, properties?: { [name: string]: string }): Promise<void> {
```
dbjorge, line 39: This doesn't contain any await call, is that intended?

src/background/telemetry/app-insights-telemetry-client.ts
```diff
@@ -36,7 +36,7 @@ export class AppInsightsTelemetryClient implements TelemetryClient {
...
         if (this.enabled) {
             this.applicationInsights.trackEvent({ name }, properties);
```
dbjorge, line 41: My understanding is that this is synchronous because it adds the telemetry to a background batching/queuing system. I'm not sure that's a great idea for us, since we might lose telemetry that way

dbjorge: (but I'd also be very hesitant to make it flush every time and await `trackEvent` every time without being very careful about the order in which we're invoking `trackEvent` vs other business logic - making sure the service worker lifetime waits for the network event would be good, but network calls can be arbitrarily slow, I wouldn't want the app responsiveness to ever be awaiting for telemetry to flush)

src/background/telemetry/telemetry-event-handler.ts
dbjorge: This makes sense! Can you add the context from this PR comment as a comment in the code next to this line?

src/background/telemetry/app-insights-telemetry-client.ts
pownkel: The original issue was that trackEvent in DebugToolsTelemetryClient sends a runtime message, and making trackEvent async there required a change to the TelemetryClient interface that affected everything that implements it. It sounds like changing the other telemetry clients could create more problems than keeping DebugToolsTelemetryClient fire-and-forget, and I'd rather we lose messages in the debug tools than in app insights, so I'll revert this change.

pownkel: Per offline discussion, this is an existing issue in the extension which is not affected by whether or not this function is async. A potential solution is to flush app insights telemetry regularly in the background, for instance each time we receive a message. This will be addressed in a later PR.

dbjorge (approved):

PR #5528, opened 2022-05-21
feat(esbuild): use esbuild to build web extension

waabid:
#### Details

Replaces usage of webpack with esbuild for the web extension. Additionally sets up but **does not** use esbuild for the unified and report packages (future PRs).

##### Motivation

#5412 

##### Context

Context can be found in the issue above.

#### Pull request checklist

- [x] Addresses an existing issue: #5412
- [ ] Ran `yarn fastpass`


style-plugin.js
```diff
@@ -0,0 +1,88 @@
...
+                            generateScopedName: '[local]' + (!isProd ? '--[hash:base64:5]' : ''),
+                            localsConvention: 'camelCaseOnly',
+                            getJSON(_, json) {
+                                singleModuleCssJSON = JSON.stringify(json);
+                            },
+                        }),
+                    ]).process(source, { from: undefined });
+
+                    return {
+                        contents: `import "css-module:${args.path}";\nexport default ${singleModuleCssJSON};`,
```
dbjorge, line 65: This needs to be escaped appropriately for a js string (this is why the `build-reliability` job is failing on all Windows builds right now)

Gruntfile.js
```diff
@@ -167,8 +167,9 @@ module.exports = function (grunt) {
             },
         },
         exec: {
-            'webpack-dev': `"${webpackPath}" --config-name dev`,
-            'webpack-dev-mv3': `"${webpackPath}" --config-name dev-mv3`,
+            'esbuild-dev': `node esbuild.js`,
+            'esbuild-dev-mv3': `node esbuild.js --env devM3`,
+            'esbuild-prod': `node esbuild.js --env prod`,
             'webpack-prod': `"${webpackPath}" --config-name prod`,
```
dbjorge, line 173: Did you intend to leave `webpack-prod` here?

package.json
```diff
@@ -81,15 +81,18 @@
         "test:e2e:docker": "npm-run-all --serial test:e2e:docker:build \"test:e2e:docker:run {@}\" --",
         "test:unified": "cross-env NODE_OPTIONS='--unhandled-rejections=strict' jest --projects src/tests/electron --runInBand --forceExit --detectOpenHandles",
         "test:report:e2e": "lerna --scope accessibility-insights-report-e2e-tests run test",
+        "tbuild": "yarn build && yarn type-check",
+        "tbuild:all": "yarn build:all && yarn type-check",
+        "type-check": "tsc -noEmit",
```
dbjorge, line 86: I think this should be added to fastpass

package.json
```diff
@@ -176,6 +179,7 @@
         "webpack-node-externals": "^3.0.0"
     },
     "dependencies": {
+        "@esbuild-plugins/node-resolve": "^0.1.4",
```
dbjorge, line 182: All of the new deps are devDependencies, not prod dependencies, no?

esbuild.js
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dbjorge, line 1: Is it intentional that this PR doesn't include removing the old webpack configs it obsoletes?

esbuild.js
```diff
@@ -0,0 +1,127 @@
...
+
+        // minification of the details-view bundle leads to a collision with the axe-core package,
+        // so we do not minify that bundle specifically.
+        config.entryPoints = {
+            detailsView: `${src}/DetailsView/details-view-initializer.ts`,
+        };
+        config.minify = false;
+        esbuild.build(config).catch(() => process.exit(1));
+    })
+    .catch(() => process.exit(1));
```
dbjorge, line 127: This catch handler (and the equivalent one on L125) feels like it would throw away error information that should probably be emitted to the console. Maybe easier to just use `buildSync` instead?

esbuild.js
```diff
@@ -0,0 +1,127 @@
...
+};
+
+esbuild
+    .build(config)
+    .then(() => {
+        if (!isProd) {
+            return;
+        }
+
+        // minification of the details-view bundle leads to a collision with the axe-core package,
```
dbjorge, line 119: This strategy is going to cause problems for `--watch`, no?

dbjorge: It also doubles the size of the production details view bundle, which isn't really thrilling from a user experience perspective

esbuild.js
```diff
@@ -0,0 +1,127 @@
...
+
+const electronEntryFiles = {
+    renderer: [path.resolve(__dirname, 'src/electron/views/renderer-initializer.ts')],
+    main: [path.resolve(__dirname, 'src/electron/main/main.ts')],
+};
+const unifiedOutdir = path.join(__dirname, 'extension/unifiedBundle');
+
+const webExtensionEntryFiles = {
+    background: `${src}/background/background-init.ts`,
+    injected: `${src}/injected/client-init.ts`,
```
dbjorge, line 22: In webpack, the injected entry point was defined as using both `client-init.ts` and `stylesheet-init.ts` as roots. Is omitting that here intentional? I'd have expected to need an update folding `stylesheet-init` into `client-init`

esbuild.js
```diff
@@ -0,0 +1,127 @@
...
+
+const config = {
+    plugins,
+    entryPoints: entryFiles,
+    format,
+    platform,
+    external,
+    outbase: src,
+    bundle: true,
+    minify: isProd,
```
dbjorge, line 100: For other reviewers: The MPL requires us to preserve axe-core's license notice even in our minified prod bundles. When I was reviewing, I saw that esbuild's prod bundles didn't have the same legal comments at the top of the file that webpack's prod bundles did and thought this would be a problem, but it turns out that [esbuild just defaults to putting them at the end of the file](https://esbuild.github.io/api/#legal-comments) instead of the beginning (this is fine).

style-plugin.js
```diff
@@ -0,0 +1,88 @@
...
+
+            Note: the stylesheet only needs to be read once, in the first onLoad method, and can be
+            passed from there to the onResolve method, and from there to the second onLoad method.
+            */
+            build.onLoad(
+                {
+                    filter: /\.s?css$/,
+                },
+                async args => {
+                    const source = await sass.renderSync({ file: args.path }).css.toString();
```
dbjorge, line 52: It would maybe be good to include a comment explaining why this is using `renderSync` instead of `render` despite being in an async context (I'm assuming it's for performance reasons based on the "Heads up!" aside in [`render`'s docs](https://sass-lang.com/documentation/js-api/modules#render))

style-plugin.js
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+const postcss = require('postcss');
+const postCssModules = require('postcss-modules');
+const sass = require('sass');
+
+const CreateStylePlugin = isProd => {
```
dbjorge, line 8: it might be helpful to include some links if there are any references you started from while designing this

package.json
```diff
@@ -81,15 +81,18 @@
         "test:e2e:docker": "npm-run-all --serial test:e2e:docker:build \"test:e2e:docker:run {@}\" --",
         "test:unified": "cross-env NODE_OPTIONS='--unhandled-rejections=strict' jest --projects src/tests/electron --runInBand --forceExit --detectOpenHandles",
         "test:report:e2e": "lerna --scope accessibility-insights-report-e2e-tests run test",
+        "tbuild": "yarn build && yarn type-check",
```
dbjorge, line 84: It would be good to update `building-web.md` with explanations of when to use `tbuild` vs `build`

dbjorge (changes requested):
Awesome work Wahaj, thanks for continuing this effort! Left some comments inline.

I also suspect that the e2e failure is legitimate - I am able to reproduce it locally, but haven't debugged beyond that.

Gruntfile.js
waabid: Yes; we're using it in the report build. I did verify that report package was working but didn't want to include that in this PR so decided to leave that unchanged (and hence leaving the webpack-prod command here).

package.json
waabid: Good catch! I must have missed the appropriate flag when adding all my changes to this branch.

esbuild.js
waabid: Yes ... but I was torn. Ultimately, it's probably best to remove them (they're always there in the history if we want to revert or something) but figured that could be a future PR, after adding the report package which would completely omit the need for any web extension related configs.

esbuild.js
```diff
@@ -0,0 +1,127 @@
...
+    sourcemap: !isProd,
```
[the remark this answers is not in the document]
waabid: I'm open to it but we've not had source maps in prod in the past. Historically, we usually just debug in other environments (like Insider, which is often the same as Prod but not necessarily) and that works okay. I think my preference would still be to go that way to reduce the bundle size for prod (goes up roughly 400% IIRC).

Thoughts?

karanbirsingh:
Nice start! I tried building this locally and realized it's failing on Windows. You can see more in the `build-reliability` run: https://dev.azure.com/accessibility-insights/accessibility-insights-web/_build/results?buildId=35569&view=logs&j=45ac6fad-22a2-55b6-51dc-72f9d5547440&t=823f074f-56d9-59bb-f2eb-8c7d33a103cc

The error message is a little bizarre: 
```js
✘ [ERROR] Syntax error "r"

    src/common/components/cards/urls-card-row.scss:1:93:
      1 │ ...s-module:C:\Users\karansin\accessibility-insights-web\src\common\components\cards\urls-card-row.scss";
        ╵                                                                                       ^

1 error
```

The output is not immediately helpful, I'm almost wondering if the \u is interpreted as some sort of escape or unicode sequence?

esbuild.js
waabid: On the bundle size portion, I mention that here: https://github.com/microsoft/accessibility-insights-web/issues/5412#issuecomment-1129457320.

It's a difference of 15.7MB to 12.6MB (vs. 13.1MB with webpack). It felt like to me that was _okay_ for now and ideally we can file an issue on esbuild/hopefully get it resolved at some point (since it was due to a naming collision).

Thoughts?

waabid: Will look into the watch again ... seemed to be working prior fine. Will report back.

esbuild.js
waabid: Good catch! I missed adding that to the client-init.ts file in this branch.

esbuild.js
```diff
@@ -0,0 +1,127 @@
...
+    serviceWorker: `${src}/background/service-worker-init.ts`,
+    debugTools: `${src}/debug-tools/initializer/debug-tools-init.tsx`,
+};
+const devWebExtensionOutdir = path.join(__dirname, 'extension/devBundle');
+const devWebExtensionM3Outdir = path.join(__dirname, 'extension/devMv3Bundle');
+
+const prodWebExtensionOutDir = path.join(__dirname, 'extension/prodBundle');
+
+// Default behavior; builds dev extension.
+let entryFiles = webExtensionEntryFiles;
```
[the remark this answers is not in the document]
waabid: Good catch! I missed this from prior. I'm not sure necessarily how to resolve this.

esbuild doesn't currently support multiple entry-points per a single bundle the way webpack does. For stylesheet-init, it is fairly easy to just add that to the client-init ... but for this conditional, I'm not sure. Would be curious to hear if you have any ideas/thoughts?

esbuild.js
dbjorge: The issue isn't with the total size of all bundles, but with the size of the details view bundle specifically; just the extra time in the initial load + evaluate of the script is user-perceptible (+80ms) on my relatively powerful dev machine.

I don't think that's a big enough regression to consider it blocking compared to the development cycle time benefits, but it's enough that I think it would be worth investigating if there's maybe a middle ground we could approach. For example, did you already explore whether the axe-core issue still repros with `minify: true, keepNames: true` (gets ~85% of the minification) or with `minifyWhitespace: true, minifySyntax: true, minifyIdentifiers: false` (gets ~60% of the minification)?

esbuild.js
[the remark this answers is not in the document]
dbjorge: The first option that comes to my mind is to add a `details-view-init-with-devtools.ts` that just imports both react-devtools and the normal init file, and then have some conditional in the esbuild config that uses that as the dev entry point if react devtools is present (+ same for popup)

esbuild.js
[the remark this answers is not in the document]
dbjorge: Leaving it out of scope for this PR is reasonable, given that we don't currently ship them

esbuild.js
waabid: Gotcha!

Yeah, I did look into those options but wasn't fully satisfied with the result because it seems like a transient issue on my end: currently, minifying everything normally **doesn't** produce this issue. That's without changing anything (i.e. not including the workaround or using the sub-minify variables).

I'm trying to get the error to repro again so I can appropriately test these options but reading through the bundles made it seem like the variable names were still likely to collide ... when the stars align. It's all very odd.

esbuild.js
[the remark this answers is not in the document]
waabid: I really like this approach. I'll give it a go! Thanks!

esbuild.js
waabid: I was able to go through my history to find a moment in time that was able to repro the error*. With that, I was able to test a bit more and it seems like minifyIdentifiers set to false leads to a bundle that is workable (I think previously when I tested this, I forgot to unset minify).

\* still don't know what is causing this to happen.

style-plugin.js
```diff
@@ -0,0 +1,88 @@
...
+        name: 'style-plugin',
+
+        setup(build) {
+            build.onResolve(
+                {
+                    filter: /^css-module:/,
+                },
+                args => {
+                    const prefixLen = 'css-module:'.length;
+                    const suffixLen = '.scss'.length;
```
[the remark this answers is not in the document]
waabid: I can change it to be more generic for sure but I think it would not be possible to get here without either `.scss` or `.css` in the import.

Let me know if you had something else in mind.

esbuild.js
waabid: Good catch! Changing to console.error for the catch.

On `buildSync`: it doesn't work when using plugins sadly.

style-plugin.js
waabid: Thanks for the code-snippet! I didn't include the error because we'd likely see an error from trying to read the file using sass, which I figure is good enough?

style-plugin.js
waabid: Good point! Turns out `render` and `renderSync` are now deprecated so I've switched to `compile`, which is synchronous and faster than `compileAsync`.

src/injected/client-init.ts
```diff
@@ -17,3 +38,5 @@ if (!window.windowInitializer) {
 
     window.windowInitializer.initialize().catch(logger.error);
```
dbjorge, line 39: Now that this is in an async context, probably preferable to just await it and move the `.catch(logger.error)` up a level to L42

src/injected/client-init.ts
```diff
@@ -1,11 +1,32 @@
...
+        return;
+    }
+
+    const { Stylesheet } = await import('@fluentui/merge-styles');
+    const { createDefaultLogger } = await import('common/logging/default-logger');
+    const { initializeFabricIcons } = await import('../common/fabric-icons');
+    const { MainWindowInitializer } = await import('./main-window-initializer');
+    const { WindowInitializer } = await import('./window-initializer');
+
+    const stylesheet = Stylesheet.getInstance();
```
dbjorge, line 24: can/should `stylesheet-init.ts` be deleted now that it's incorporated here?

karanbirsingh (commented):
Overall the changes look good to me as well. Building locally works; however, I'm having trouble debugging components in the DetailsView from Chrome DevTools (e.g. ctrl+P to open a component like Toast.tsx, setting a breakpoint fails)

waabid:
> Overall the changes look good to me as well. Building locally works; however, I'm having trouble debugging components in the DetailsView from Chrome DevTools (e.g. ctrl+P to open a component like Toast.tsx, setting a breakpoint fails)

This has been fixed with the embedded style changes; good catch! 😄

packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
```diff
@@ -129,6 +129,7 @@
         --nav-link-selected: var(--communication-tint-40);
         --nav-link-expanded: transparent;
       }
+
```
waabid, line 132: **Note:** The changes in the snapshots for the report are only new line changes. This is due to changing the way we manage new line characters in our embedded styles (which itself was necessary for functioning source maps in the details view).

esbuild.js
```diff
@@ -136,21 +136,4 @@ const config = {
...
-        config.entryPoints = {
-            detailsView: `${src}/DetailsView/details-view-initializer.ts`,
-        };
-        config.minify = false;
-        config.minifyWhitespace = true;
-        config.minifySyntax = true;
-        esbuild.build(config).catch(console.error);
-    })
-    .catch(console.error);
+esbuild.build(config).catch(console.error);
```
dbjorge, line 139: I think this needs to *both* emit to console.error *and* invoke `process.exit(nonzero)` - right now, build failures don't actually cause webpack (or a build pipeline) to bail

Gruntfile.js
```diff
@@ -347,7 +347,7 @@ module.exports = function (grunt) {
             'embed-styles': {
                 [targetName]: {
                     cwd: path.resolve(extensionPath, bundleFolder),
-                    src: '**/*bundle.js',
+                    src: '**/*bundle.js**',
```
dbjorge, line 350: Maybe a little easier to understand to instead use `['**/*bundle.js', '**/*bundle.js.map']`?

dbjorge (approved):
LGTM modulo the check failure on `initClient();`. That's a semantic merge conflict with #5552, the fix should just be to use `void initClient();` instead

PR #5537, opened 2022-05-23
feat(mv3-part-4): Fix most no-floating-promises issues in injected directory

pownkel:
#### Details

Fix no-floating-promises issues around the tab stops visualizer and DrawingController.

##### Motivation

Continuing work from #5517 and #5524

##### Context

See #5517. This PR fixes 17 of the remaining 18 issues.

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #5538, opened 2022-05-23
feat(mv3-part-4): Enable interpreter to return a promise

pownkel:
#### Details

Instead of calling a fire-and-forget handler and returning a boolean flag, let Interpreter return an object that contains the boolean flag and a nullable promise that can be awaited if the handler is async. This PR also updates all places where the interpreter is called and makes sure those promises are being awaited.

##### Motivation

Feature work

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/background/background-message-distributor.ts
```diff
@@ -41,8 +50,11 @@ export class BackgroundMessageDistributor {
             this.logger.log('Unable to interpret message - ', message);
         }
 
-        if (response) {
-            return response;
+        if (globalContextResult || tabContextResult || backchannelMessageResponse) {
+            return this.createResponsePromise(
```
dbjorge, line 54: Is it actually the case that multiple interpreters ever respond to the same message? I know the `postMessageContentHandler`-bound messages should always be mutually exclusive with interpreter-handled messages, but I'm not sure without investigation whether there are any cases where both a global and a tab-specific interpreter have a handler for the same message.

If a message is only ever bound for exactly one of the 3 interpreters, I think it simplifies this significantly

src/background/interpreter.ts
```diff
@@ -13,11 +13,16 @@ export class Interpreter {
...
+            return {
+                success: true,
+                result: this.messageToActionMapping[message.messageType](
+                    message.payload,
+                    message.tabId,
+                ),
+            };
         }
-        return false;
+        return { success: false };
```
dbjorge, line 26: When I was reading the usage of this property in the message distributor, the term `success` felt a little ambiguous to me - it wasn't clear whether it implied just "this interpreter handled this message vs not" or "the handler for this message worked without error vs not". If it's strictly the former, maybe `handledMessage` or something would read more clearly?

src/background/target-page-controller.ts
```diff
@@ -93,7 +95,7 @@ export class TargetPageController {
             });
 
             for (const activeTab of activeTabs) {
-                this.sendTabVisibilityChangeAction(
+                await this.sendTabVisibilityChangeAction(
```
dbjorge, line 98: Can this use `Promise.all` instead of awaiting in sequence? (ideally across all windows, too, not just across tabs per-window)

src/background/target-page-controller.ts
```diff
@@ -142,29 +144,34 @@ export class TargetPageController {
...
-                });
-            },
-            () => {
-                this.logger.log(
-                    `sendTabUrlUpdatedAction: tab with ID ${tabId} not found, skipping action message`,
-                );
-            },
-        );
+    private async sendTabUrlUpdatedAction(tabId: number): Promise<void> {
+        await new Promise<void>((resolve, reject) => {
```
dbjorge, line 152: With this change, half the callsites for `BrowserAdapter.getTab` are now doing this promise-wrapping dance - I wonder if it would be more readable/concise to instead switch `BroserAdapter.getTab` to be promise-based

src/tests/end-to-end/common/page-controllers/background-page.ts
```diff
@@ -26,14 +26,14 @@ export class BackgroundPage extends Page {
     public async setHighContrastMode(enableHighContrast: boolean): Promise<void> {
         await this.waitForInitialization();
         await this.evaluate(enable => {
-            window.insightsUserConfiguration.setHighContrastMode(enable);
+            void window.insightsUserConfiguration.setHighContrastMode(enable);
```
dbjorge, line 29: I think it would be preferable not to void this; Playwright's evaluate is Promise-aware and will await the result if an evaluated function returns a promise, I think it would be preferable to let it do so

dbjorge: (same for `setTelemetryState`)

dbjorge (changes requested):
Thanks Kelsey! Left a few comments inline. It would also be good to rebase this once #5537 and #5549 merge, I'd like to test these changes in combination with them.

src/background/target-page-controller.ts
pownkel: I added a getTabAsync function to BrowserAdapter to wrap this call in a promise, but currently there's one getTab use in AssessmentStore that's complicated to replace. Switching over to use a promise there would make some of the action listeners async and complicate the testing code quite a bit. It will be much easier to replace once the planned changes to stores and actions have been made.

src/background/background-message-distributor.ts
pownkel: There are a few messages that have listeners registered with both the global and tab interpreters (AssessmentScanCompleted, StartOverAllAssessments, StartOverTest, and DetailsView.select)

dbjorge:
After some local testing, I found that the original logic for merging the interpreter messages needed to be more complex than the original PR's use of `Promise.all`. I created a new `EventResponseFactory` class to contain the merging logic. It felt strange to keep adding more and more things to wire up to all the initializers that would never need it, so I also changed the non-background initializer strategy to avoid needing to create unnecessary promise factories, event response factories, etc.

I think this is closer to correct, but I'm seeing issues with the current implementation where it looks like it's somehow broken the tab stops visualizer (whether triggered ad-hoc or via an assessment requirement). Needs further investigation on Monday

dbjorge:
Adding the passthrough event manager stuff to this PR was making it unduly large, so I extracted that out into #5591. Will rebase this on top of that once it merges to cut down on the PR size.

dbjorge:
Confirmed locally that the issue with the tab stops visualizer was resolved as expected by merging the changes from #5623

src/background/background-message-distributor.ts
```diff
@@ -29,44 +30,33 @@ export class BackgroundMessageDistributor {
...
+                return Promise.resolve(postMessageResponse.response);
+            } else {
+                return; // fire and forget
+            }
         }
 
-        return null;
-    }
+        return Promise.reject(
+            new Error(`Unable to interpret message - ${JSON.stringify(message)}`),
```
pownkel, line 57: Do we want to return a promise in this case? If the message isn't handled by the background, it may still be processed by a different context.

dbjorge (approved):
Approved the changes Kelsey originally authored, will wait on merging until Kelsey approves my changes since then

pownkel:
Approved!

PR #5544, opened 2022-05-24
feat(mv3-part-4): Improve browser event timeout debugability

dbjorge:
#### Details

This PR makes 2 improvements to ease debugging of `browser-event-manager` timeouts:

* It reduces the default timeout/fnf delay from 4min/2min to 1min/30s, respectively
    * The original timeouts were chosen mostly based on "being slightly less than the browser's 5min timeout". However, these new timeout values are still plenty high enough that anything hitting them in practice is a bug we should fix, while cutting the time required for a human to sit around waiting to trigger an issue by a factor of 4
* It updates `promiseFactory.timeout` to support an `errorContext` parameter to include in `TimeoutError`s, and updates `BrowserEventManager` to use that to inject details of the event that timed out into the message
    * Previously, a browser event timeout message only showed that "some event's handler timed out", but didn't clarify which event or whether the timeout was due to no handler being registered vs a handler taking too long. The new context info in the error message enables distinguishing these cases.

##### Motivation

This PR addresses 2 pain points I hit while trying to root cause why we're seeing browser-event-manager timeout error spew in context scripts on pages with iframes. Actually fixing that issue is out of scope/for a separate PR, this is just improving debugability.

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/browser-adapters/browser-event-manager.ts
+++ b/src/common/browser-adapters/browser-event-manager.ts
@@ -163,7 +163,11 @@ export class BrowserEventManager {
         if (isPromise(result)) {
             // Wrapping the ApplicationListener promise responses in a 4-minute timeout
             // prevents the service worker going idle before a response is sent
-            return await this.promiseFactory.timeout(result, EVENT_TIMEOUT_MS);
+            const timeoutErrorContext = `[browser event listener: ${JSON.stringify({
+                eventType,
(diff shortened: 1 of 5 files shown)


sfoslund (approved):
Looks good, just 2 nits!

PR #5549, opened 2022-05-25
feat(mv3-part-4): Avoid using stale data in CardSelectionStore

pownkel:
#### Details

- Make CardSelectionStoreData.rules nullable and set it to null by default
- Reset CardSelectionStoreData to the default state when the target tab URL changes
- Do not try to get a selectorMap if CardSelectionStoreData.rules is null

##### Motivation

This fixes an issue exposed by #5537. When the Automated checks or Needs review ad hoc visualizers are enabled, the DrawingController receives two requests, one with an empty list, and one with the correct list of elements to highlight. This is caused by two consecutive store updates, one to UnifiedScanResultStore and one to CardSelectionStore, which both trigger a visualization update. Between these two store updates, stale data in CardSelectionStore causes DrawingController to receive an empty list of elements. While working on #5537, we saw these two visualization requests racing in an E2E test, which caused the visualization to be wiped clean if the invalid request with the empty list was processed after the request with the correct element list.

This PR fixes the issue by setting CardSelectionStoreData.rules to null be default, to distinguish between situations where no scan has yet been run vs when a scan has completed and no issues were found. It also removes stale data from the store if the tab is reloaded or the URL changes. The visualization is not updated unless we have valid data from both the CardSelectionStore and the UnifiedScanResultStore.

##### Context

This PR does not address the underlying issue of store updates racing and being processed out of order, but this should be fixed when the rest of the flux architecture is converted from fire-and-forget to promises.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/stores/card-selection-store.ts
+++ b/src/background/stores/card-selection-store.ts
@@ -57,6 +57,7 @@ export class CardSelectionStore extends PersistentStore<CardSelectionStoreData> 
 
     public getDefaultState(): CardSelectionStoreData {
         const defaultValue: CardSelectionStoreData = {
+            rules: null,
             visualHelperEnabled: false,
             focusedResultUid: null,
         };
@@ -81,11 +82,11 @@ export class CardSelectionStore extends PersistentStore<CardSelectionStoreData> 
     };
 
     private toggleRuleExpandCollapse = (payload: RuleExpandCollapsePayload): void => {
-        if (!payload || !this.state.rules || !this.state.rules[payload.ruleId]) {
+        if (!payload || !this.state.rules?.[payload.ruleId]) {
             return;
         }
 
-        const rule = this.state.rules![payload.ruleId];
+        const rule = this.state.rules[payload.ruleId];
 
         rule.isExpanded = !rule.isExpanded;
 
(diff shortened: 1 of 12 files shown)


src/background/stores/card-selection-store.ts
```diff
@@ -79,11 +81,11 @@ export class CardSelectionStore extends PersistentStore<CardSelectionStoreData>
     };
 
     private toggleRuleExpandCollapse = (payload: RuleExpandCollapsePayload): void => {
-        if (!payload || !this.state.rules[payload.ruleId]) {
+        if (!payload || !this.state.rules || !this.state.rules[payload.ruleId]) {
             return;
         }
 
-        const rule = this.state.rules[payload.ruleId];
+        const rule = this.state.rules![payload.ruleId];
```
dbjorge, line 88: You shouldn't need the `!` on `rules!` here (or any of the other usages in this file) - TypeScript is smart enough to narrow the typings based on the conditional checks you're making earlier in the function.

src/background/stores/card-selection-store.ts
```diff
@@ -79,11 +81,11 @@ export class CardSelectionStore extends PersistentStore<CardSelectionStoreData>
     };
 
     private toggleRuleExpandCollapse = (payload: RuleExpandCollapsePayload): void => {
-        if (!payload || !this.state.rules[payload.ruleId]) {
+        if (!payload || !this.state.rules || !this.state.rules[payload.ruleId]) {
```
dbjorge, line 84: nit: You can shorten this (and similar checks throughout the file) with `?.` syntax:

```suggestion
        if (!payload || !this.state.rules?.[payload.ruleId]) {
```

src/background/stores/card-selection-store.ts
pownkel: Removed most instances, a few are still necessary to avoid strict null checks errors after we explicitly set it to a non-null value

PR #5550, opened 2022-05-25
feat(mv3-part-4): Add exception telemetry listener

sfoslund:
#### Details

Adding `ExceptionTelemetryListener`, which listens to for unhandled exceptions and promise rejections and sends telemetry information.

##### Motivation

Feature work.

##### Context

This PR only adds exception listening for the background, follow up PRs will add logic to also send telemetry information from other contexts. 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/background-init.ts
+++ b/src/background/background-init.ts
@@ -8,6 +8,7 @@ import { TabContextManager } from 'background/tab-context-manager';
 import { TabEventDistributor } from 'background/tab-event-distributor';
 import { ConsoleTelemetryClient } from 'background/telemetry/console-telemetry-client';
 import { DebugToolsTelemetryClient } from 'background/telemetry/debug-tools-telemetry-client';
+import { ExceptionTelemetryListener } from 'background/telemetry/exception-telemetry-listener';
 import { createToolData } from 'common/application-properties-provider';
 import { BrowserAdapterFactory } from 'common/browser-adapters/browser-adapter-factory';
 import { BrowserEventManager } from 'common/browser-adapters/browser-event-manager';
@@ -122,6 +123,9 @@ async function initialize(): Promise<void> {
     const usageLogger = new UsageLogger(browserAdapter, DateProvider.getCurrentDate, logger);
 
     const telemetryEventHandler = new TelemetryEventHandler(telemetryClient);
+
+    const exceptionTelemetryListener = new ExceptionTelemetryListener(telemetryEventHandler);
+    exceptionTelemetryListener.initialize(logger);
 
     const browserSpec = new NavigatorUtils(window.navigator, logger).getBrowserSpec();
 
--- a/src/background/telemetry/exception-telemetry-listener.ts
+++ b/src/background/telemetry/exception-telemetry-listener.ts
@@ -121,7 +121,7 @@ export class ExceptionTelemetryListener {
         source?: string,
     ): void => {
         const telemetry: UnhandledExceptionTelemetryData = {
-            message,
+            message: message.toString(),
             stackTrace,
             source,
             errorType,
@@ -149,7 +149,10 @@ export class ExceptionTelemetryListener {
         if (telemetryData.stackTrace && telemetryData.stackTrace.length > this.MAX_STACK_CHARS) {
             telemetryData.stackTrace = telemetryData.stackTrace.substring(0, this.MAX_STACK_CHARS);
         }
-        if (telemetryData.message?.includes('http') || telemetryData.stackTrace?.includes('http')) {
+        if (
+            (telemetryData.message && telemetryData.message.includes('http')) ||
+            (telemetryData.stackTrace && telemetryData.stackTrace.includes('http'))
+        ) {
             return undefined;
         }
         return telemetryData;
--- a/src/background/telemetry/exception-telemetry-listener.ts
+++ b/src/background/telemetry/exception-telemetry-listener.ts
@@ -9,7 +9,11 @@ import { UnhandledExceptionTelemetryData } from '../../common/extension-telemetr
 export class ExceptionTelemetryListener {
     constructor(private readonly telemetryEventHandler: TelemetryEventHandler) {}
 
-    public initialize(logger: Logger): void {
+    public initialize(
+        logger: Logger,
+        extWindow: Window = window,
+        extConsole: Console = console,
+    ): void {
         const sendExceptionTelemetry = (message: string, stackTrace?: string): void => {
             const telemetry: UnhandledExceptionTelemetryData = { message, stackTrace };
             const payload: BaseActionPayload = {
@@ -23,7 +27,7 @@ export class ExceptionTelemetryListener {
         };
 
         // Catch top level synchronous errors
-        window.onerror = function (
+        extWindow.onerror = function (
             message: string,
             source: string,
             lineno: number,
@@ -35,31 +39,31 @@ export class ExceptionTelemetryListener {
         };
 
         // Catch errors thrown in promises
-        window.onunhandledrejection = function (event: PromiseRejectionEvent) {
+        extWindow.onunhandledrejection = function (event: PromiseRejectionEvent) {
             sendExceptionTelemetry(event.reason);
             return false;
         };
 
         // Catch errors written to console.error
-        var consoleError = console.error;
-        console.error = function (message?: any, ...optionalParams: any[]) {
+        const consoleError = extConsole.error;
+        extConsole.error = function (message?: any, ...optionalParams: any[]) {
             const err = optionalParams[0] as Error;
             if (message || err) {
                 sendExceptionTelemetry(message, err?.stack);
             }
 
-            consoleError(message, ...optionalParams);
+            consoleError(message, err);
         };
 
         // Catch errors written to logger.error
-        var loggerError = logger.error;
+        const loggerError = logger.error;
         logger.error = function (message?: any, ...optionalParams: any[]) {
             const err = optionalParams[0] as Error;
             if (message || err) {
                 sendExceptionTelemetry(message, err?.stack);
             }
 
-            loggerError(message, ...optionalParams);
+            loggerError(message, err);
         };
     }
 }
--- a/src/background/telemetry/exception-telemetry-listener.ts
+++ b/src/background/telemetry/exception-telemetry-listener.ts
@@ -16,6 +16,17 @@ enum ErrorType {
 export class ExceptionTelemetryListener {
     private readonly MAX_MESSAGE_CHARS = 300;
     private readonly MAX_STACK_CHARS = 5000;
+    private readonly EXCLUDED_PROPERTIES = [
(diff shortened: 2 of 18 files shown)


src/background/telemetry/exception-telemetry-listener.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        // Catch errors written to logger.error
+        const loggerError = logger.error;
+        logger.error = function (message?: any, ...optionalParams: any[]) {
+            const err = optionalParams[0] as Error;
+            if (message || err) {
+                sendExceptionTelemetry(message, err?.stack);
+            }
+
+            loggerError(message, err);
```
pownkel, line 66: Are the other optional parameters intentionally thrown away here, and if so could you note that in a comment?

src/background/background-init.ts
```diff
@@ -123,6 +124,9 @@ async function initialize(): Promise<void> {
 
     const telemetryEventHandler = new TelemetryEventHandler(telemetryClient);
 
+    const exceptionTelemetryListener = new ExceptionTelemetryListener(telemetryEventHandler);
```
dbjorge, line 127: You mentioned in the PR description that only adding to "the background" was intentional - I wanted to clarify whether that meant "only the manifest v2 background page", or if this is intended to also be included in `service-worker-init`

src/background/telemetry/exception-telemetry-listener.ts
```diff
@@ -0,0 +1,69 @@
...
+export class ExceptionTelemetryListener {
+    constructor(private readonly telemetryEventHandler: TelemetryEventHandler) {}
+
+    public initialize(
+        logger: Logger,
+        extWindow: Window = window,
+        extConsole: Console = console,
+    ): void {
+        const sendExceptionTelemetry = (message: string, stackTrace?: string): void => {
+            const telemetry: UnhandledExceptionTelemetryData = { message, stackTrace };
```
dbjorge, line 18: It would be good to set this up to include a `source` property - we know we're going to want to be able to distinguish "from background" vs "from details view", etc

src/background/telemetry/exception-telemetry-listener.ts
```diff
@@ -0,0 +1,69 @@
...
+            sendExceptionTelemetry(event.reason);
+            return false;
+        };
+
+        // Catch errors written to console.error
+        const consoleError = extConsole.error;
+        extConsole.error = function (message?: any, ...optionalParams: any[]) {
+            const err = optionalParams[0] as Error;
+            if (message || err) {
+                sendExceptionTelemetry(message, err?.stack);
```
dbjorge, line 52: I think it's a bit misleading to call the console.error and logger.error cases "unhandled exceptons" - they aren't necessarily unhandled exceptions. It would probably be good to be able to distinguish the source of the error (window error vs unhandled rejection vs logger error, etc) as part of the telemetry.

src/background/telemetry/exception-telemetry-listener.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export class ExceptionTelemetryListener {
+    constructor(private readonly telemetryEventHandler: TelemetryEventHandler) {}
+
+    public initialize(
+        logger: Logger,
+        extWindow: Window = window,
+        extConsole: Console = console,
+    ): void {
+        const sendExceptionTelemetry = (message: string, stackTrace?: string): void => {
```
dbjorge, line 17: I wonder if there are any patterns we can predict ahead of time that would be a good idea to sanitize preemptively. Off the top of my head, I'm wondering if we should:

* Impose a size cap on message/stackTrace past which we truncate it (I've seen error messages be 100kb long if they manage to embed an entire axe-core result blob, for example)
* Avoid sending messages that include anything that looks like a URL
* Avoid sending messages that include any of the property names we or axe-core use to represent css selectors and/or HTML snippets

src/background/telemetry/exception-telemetry-listener.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        // Catch errors written to console.error
+        const consoleError = extConsole.error;
+        extConsole.error = function (message?: any, ...optionalParams: any[]) {
+            const err = optionalParams[0] as Error;
+            if (message || err) {
+                sendExceptionTelemetry(message, err?.stack);
+            }
+
+            consoleError(message, err);
```
dbjorge, line 55: It took me some investigation to convince myself that this wouldn't be double-sending logger.error results (which go to console.error by default). It would be nice to include a comment here explaining why that is (and probably also a comment in `createDefaultLogger()` that explains why it's now functionally important that it use `error: console.error` and not `error: (...args) => console.error(...args)`).

src/background/telemetry/exception-telemetry-listener.ts
```diff
@@ -0,0 +1,69 @@
...
+            };
+
+            this.telemetryEventHandler.publishTelemetry(
+                TelemetryEvents.UNHANDLED_EXCEPTION,
+                payload,
+            );
+        };
+
+        // Catch top level synchronous errors
+        extWindow.onerror = function (
```
dbjorge, line 30: I think all of these hooks should be taking care to avoid reentrancy (ie, to avoid infinite recursion in the event that `sendExceptionTelemetry` somehow ends up logging an error/throwing an unhandled exception/floating a rejecting promise). The browser makes *some* attempts at avoiding infinite recursion here (eg, neither Chromium nor Firefox will trigger an `onerror` handler a second time if the `onerror` handler itself throws an error), but it isn't absolute and it isn't mandated by the standard, so we shouldn't rely on it (eg, Firefox doesn't protect against `onerror` and `onunhandledrejection` bouncing back and forth forever).

I would recommend protecting against this by having the listener maintain a `hookIsActive` boolean and having each individual hook be of the form:

```ts
if (this.hookIsActive) { return; }
this.hookIsActive = true;
try {
    // insert current hook implementation
} finally {
    this.hookIsActive = false;
}
```

src/background/telemetry/exception-telemetry-listener.ts
sfoslund: Makes sense to me. Can you expand on that last item- do you have any examples of property names we wouldn't want to send and how we can generate a full list?

sfoslund: I've gone ahead and addressed all the feedback in my most recent commits, except the 3rd item in this list.

dbjorge: I'd suggest starting with the following properties (it's okay if for it to be incomplete/best-effort): `html`, `target`, `url`, `path`, `snippet`, `selector`, `elementSelector`, `cssSelector`

sfoslund: Cool, just added that, thanks!

src/background/telemetry/exception-telemetry-listener.ts
```diff
@@ -0,0 +1,183 @@
...
+        telemetryData: UnhandledExceptionTelemetryData,
+    ): UnhandledExceptionTelemetryData => {
+        if (telemetryData.message && telemetryData.message.length > this.MAX_MESSAGE_CHARS) {
+            telemetryData.message = telemetryData.message.substring(0, this.MAX_MESSAGE_CHARS);
+        }
+        if (telemetryData.stackTrace && telemetryData.stackTrace.length > this.MAX_STACK_CHARS) {
+            telemetryData.stackTrace = telemetryData.stackTrace.substring(0, this.MAX_STACK_CHARS);
+        }
+        if (
+            (telemetryData.message && telemetryData.message.includes('http')) ||
```
dbjorge, line 164: I think it would be preferable for this to look for `://` instead of `http`; it's possible for URLs to use different protocols (particularly, `file://`)

src/background/telemetry/exception-telemetry-listener.ts
```diff
@@ -0,0 +1,183 @@
...
+            (telemetryData.stackTrace && telemetryData.stackTrace.includes('http'))
+        ) {
+            return undefined;
+        }
+        if (
+            (telemetryData.message &&
+                this.EXCLUDED_PROPERTIES.some(property =>
+                    telemetryData.message.includes(property),
```
dbjorge, line 172: I think this is probably over-excluding; I'd recommend something that looks specifically for quoted versions of the excluded properties (indicating a JSON-serialized object using those props as a key), rather than also picking up error messages that just happen to include those words.

It would probably be easiest to build up a regex to do that. It would probably be most performant to try to compile the regex just one statically, along these lines:

```ts
function makeQuestionableSubstringRegex(): RegExp {
    const urlPattern = ':\\/\\/';
    const questionablePropertyNamePattern =
        this.EXCLUDED_PROPERTIES.map(escapeRegExp).join('|');
    const questionablePropertyPattern = `['"](${questionablePropertyNamePattern})['"]`;
    const questionableSubstringPattern = `${urlPattern}|${questionablePropertyPattern}`;

    // This argument *is* a constant built from literals, it's just built up from parts
    // eslint-disable-next-line security/detect-non-literal-regexp
    return new RegExp(questionableSubstringPattern);
}
```

This is complicated enough that it probably warrants moving the sanitization logic into its own independently testable component

src/common/extension-telemetry-events.ts
```diff
@@ -283,6 +284,13 @@ export type AutoDetectedFailuresDialogStateTelemetryData = {
     enabled: boolean;
 };
 
+export type UnhandledExceptionTelemetryData = {
+    message: string;
+    stackTrace?: string;
+    source: string;
```
dbjorge, line 290: This should probably be `TelemetryEventSource`

src/common/extension-telemetry-events.ts
```diff
@@ -283,6 +284,13 @@ export type AutoDetectedFailuresDialogStateTelemetryData = {
...
+    errorType: string;
```
dbjorge, line 291: This can probably be a string enum, rather than a generic string (eg `'logger.error' | 'console.error' | 'unhandled error' | 'unhandled promise rejection'`)

src/common/extension-telemetry-events.ts
```diff
@@ -80,6 +80,7 @@ export const LEFT_NAV_PANEL_EXPANDED: string = 'leftNavPanelExpanded';
 export const NEEDS_REVIEW_TOGGLE: string = 'NeedsReviewToggled';
 export const NAVIGATE_TO_NEW_CARDS_VIEW: string = 'NavigateToNewCardsView';
 export const SET_AUTO_DETECTED_FAILURES_DIALOG_STATE: string = 'setAutoDetectedFailuresDialogState';
+export const UNHANDLED_EXCEPTION: string = 'unhandledException';
```
dbjorge, line 83: I think it would be more accurate for this to read `ERROR` than `EXCEPTION` (same for the type name on L287)

dbjorge (approved):

Issue #4616, opened 2021-08-25
[Windows HC] More actions menu icon is not visible when selected


lisli1:
## Describe the bug

1. When the "more actions" menu icon is selected, the icon is no longer visible. 
2. Similarly, when the menu items are hovered, there does not seem to be sufficient contrast.

* In Windows High Contrast 1, (1) is present.
* In Windows High Contrast 2, (1) is present.
* In Windows High Contrast Black, (1) and (2) are present. For (2), the text of the menu items has insufficient contrast.
* In Windows High Contrast White, (2) is present. The icons of the menu items have insufficient contrast.

**To Reproduce**

1. Turn on Windows HC mode (not from within AI-Web)
2. Run automated checks
3. Open the cards view
4. Select the "more actions" menu on the footer of a card
5. Hover the resulting menu items

## Expected behavior

In all Windows high contrast themes, the (1) menu icon and (2) menu items should still be visible and have sufficient contrast when selected/hovered.

## Screenshots

* AI Web high contrast (**not** Windows high contrast) shows menu icon and menu items visible, with sufficient contrast:
![image](https://user-images.githubusercontent.com/48259897/130871142-d768c7a3-9777-463e-ae92-021e130a9100.png)

* Windows High Contrast 1 shows menu icon not visible:
![image](https://user-images.githubusercontent.com/48259897/130871697-de191c66-1b96-4446-8d28-69d113365e33.png)

* Windows High Contrast 2 shows menu icon not visible:
![image](https://user-images.githubusercontent.com/48259897/130871699-b72a5546-97e8-4274-8cbb-f913c7c80997.png)

* Windows High Contrast Black shows menu icon not visible and text of menu items with insufficient contrast:
![image](https://user-images.githubusercontent.com/48259897/130870325-a1b6af37-b4ea-4c87-9727-8d28cf702827.png)

* Windows High Contrast White shows icon of menu items with insufficient contrast:
![image](https://user-images.githubusercontent.com/48259897/130871688-52f21ae1-7c70-4f38-a517-8acdd52d5275.png)

## Context (please complete the following information)

- OS Name & Version:  Windows Version 10.0.19043 Build 19043
- AI-Web Version & Environment:  2.28.0 Insider
- Browser Version:  Edge Version 92.0.902.62 (Official build) (64-bit)
- Target Page:  https://www.washington.edu/accesscomputing/AU/before.html

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

[...]

PR #5568, opened 2022-06-01
fix: Low contrast issues in kebab menu when using high contrast themes

katydecorah:
#### Details

To fix the low contrast text in the kebab menu when using high contrast mode in web and unified, we could leverage FluentUI styling of the context menu and kebab button to remove added scss to handle high contrast mode.

These changes will have slight UI changes:

- Each menu item's icons are colored blue which is consistent with other FluentUI elements.
- The context menu has a slightly less border-radius as defined by FluentUI.
- FluentUI's context menu items do not have hover state in high contrast mode by default. (🎀Note to self: Open issue in FluentUI). This pull request will set a hover state with `color: highlight` (a pattern in FluentUI) instead of setting `background-color`. 

The table below shows screenshots of the current kebab menu and proposed to illustrate the UI changes described above:

Current | Proposed | Proposed in high contrast
---|---|---
 |  | 

The table below shows screenshots of the fixed high contrast issues in each theme for web and unified. The first menu item is hovered. I also checked that the hovered text has a passing color contrast ratio:

High contrast theme | Web | Unified
|---|---|---
Aquatic |   | 
Desert |  | 
Dusk |   | 
Night sky |  | 

##### Additional changes

- Replace deprecated `name` with `text` in menu items.
- Remove unused color variables.
- By using FluentUI's styling of the context menu, there is no longer gaps in the border edges  in high contrast mode:
Screenshot of context menu where the white border is clipped at the rounded edges

##### Motivation

Addresses #4616

##### Context

- I considered using FluentUI's `MoreVertical` icon and while it would reduce tech debt, the dots in the icon were smaller and felt lower in contrast to the current icon overall. 

#### Pull request checklist

- [x] Addresses an existing issue: #4616
- [x] Ran `yarn fastpass`


src/common/components/cards/card-kebab-menu-button.scss
```diff
@@ -1,86 +1,14 @@
...
-        }
+    :global(.ms-Icon) {
+        font-size: 18px; // Increase kebab icon size
     }
 
-    > span {
-        justify-content: center;
+    &:hover,
+    &:global(.is-expanded) {
+        background-color: #edebe9;
```
katydecorah, line 12: Note: It's not clear to me why these styles aren't cascading from [FluentUI button](https://developer.microsoft.com/en-us/fluentui#/controls/web/button). Without this line the background color is set to `transparent`.

katydecorah: Note to self: swap for a color variable.

src/common/icons/more-actions-menu-icon.tsx
```diff
@@ -10,6 +10,7 @@ export const MoreActionsMenuIcon = NamedFC('MoreActionsMenuIcon', () => (
         height="16"
         viewBox="0 0 10 16"
         fill="none"
+        stroke="currentColor"
```
katydecorah, line 13: By setting `currentColor`, the SVG will inherit the color from the parent -- this works great for high contrast mode.

JGibson2019 (approved):

JGibson2019:
@katydecorah in HC do we need the border around the kebab menu button itself or is it fine to just have the high contrast color applied to the three dots? I noticed in non-HC mode, both the original and proposed UI feature a grey button that provides some contrast to the white background.

katydecorah:
> @katydecorah  in HC do we need the border around the kebab menu button itself or is it fine to just have the high contrast color applied to the three dots? I noticed in non-HC mode, both the original and proposed UI feature a grey button that provides some contrast to the white background.


@JGibson2019 I checked high contrast mode on the [FluentUI buttons](https://developer.microsoft.com/en-us/fluentui#/controls/web/button) page and the hover/active states are only a color change to the text and icons. I haven't found a source about best practices for high contrast mode with regards to hover, so it's not clear to me if we need an additional hover indicator (in this case highlighting the button border).

JGibson2019:
> > @katydecorah  in HC do we need the border around the kebab menu button itself or is it fine to just have the high contrast color applied to the three dots? I noticed in non-HC mode, both the original and proposed UI feature a grey button that provides some contrast to the white background.
> 
> @JGibson2019 I checked high contrast mode on the [FluentUI buttons](https://developer.microsoft.com/en-us/fluentui#/controls/web/button) page and the hover/active states are only a color change to the text and icons. I haven't found a source about best practices for high contrast mode with regards to hover, so it's not clear to me if we need an additional hover indicator (in this case highlighting the button border).

Gotcha! Do you think we'll need to verify this before closing out #4616?

katydecorah:
>> > @katydecorah  in HC do we need the border around the kebab menu button itself or is it fine to just have the high contrast color applied to the three dots? I noticed in non-HC mode, both the original and proposed UI feature a grey button that provides some contrast to the white background.
> 
>> @JGibson2019 I checked high contrast mode on the [FluentUI buttons](https://developer.microsoft.com/en-us/fluentui#/controls/web/button) page and the hover/active states are only a color change to the text and icons. I haven't found a source about best practices for high contrast mode with regards to hover, so it's not clear to me if we need an additional hover indicator (in this case highlighting the button border).

> Gotcha! Do you think we'll need to verify this before closing out #4616? 

Confirmed with PM that we will not need to add a border on hover since on hover the cursor pointer changes and the color change contrast is 3:1 😎

PR #5570, opened 2022-06-01
feat(mv3-part-4): Split exception listener into sending and forwarding listeners

sfoslund:
#### Details

Follow up to #5550, which added the `ExceptionListener` to the background. This PR adds exception listeners to other contexts and separates the exception listener into `SendingExceptionListener` for contexts that will send their own telemetry (background, pop up, details view) and `ForwardingExceptionListener` for contexts that will forward their telemetry to the background to send (target page). 

##### Motivation

Feature work.

##### Context

This PR adds logic for the pop up and details view to send their own telemetry instead of forwarding to the background as usual, since we want to be able to send any error information related to connecting/ communicating with the background. 

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/details-view-initializer.ts
```diff
@@ -235,6 +246,38 @@ if (tabId != null) {
                 storeUpdateMessageHub,
             );
 
+            const telemetryLogger = new TelemetryLogger(logger);
+            const applicationTelemetryDataFactory = getApplicationTelemetryDataFactory(
```
dbjorge, line 250: I think this initialization is unsafe, for 2 reasons:

1. I think passing `{} as InstallationData` will end up causing the underlying `InstallDataGenerator` to *always* consider its installation data to be "stale" and to *always* generate a fresh installation ID and overwrite any existing one. This is an extremely bad result; it would subtly our ability to correlate all other telemetry together, not just this new telemetry.

Anytime I find myself writing something like `stubValue as SomeLargerType` outside of a test case, I always treat it as a red flag; it almost always means that I'm breaking some assumption that the callee is making, leading to subtle bugs.

2. Even if this was fixed to read current installation data instead of passing a stub, `getApplicationTelemetryDataFactory` attempts to generate a new installation ID if one isn't already persisted and persists the new one afterwards if one is generated. This strategy isn't thread-safe when two contexts (background vs details or background vs popup) are racing one another, which is likely to happen in practice in MV3 since service worker init will commonly end up racing popup/details init. Particularly, if the installation data is actually stale, those two contexts could each check the current state concurrently, see that it's stale, each generate *separate* new installation IDs, and then have inconsistent IDs in telemetry from the same session.

Anytime I find myself adding an "import" on something in `/src/background` from a non-background context, that's also something to treat as a red flag. Many background components make assumptions that they're the only instance of a given bit of functionality; anytime you're taking a previously background-only component and running it elsewhere, it's important to audit it (and all its transitive dependencies!) for assumptions like that.

src/background/telemetry/forwarding-exception-telemetry-listener.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dbjorge, line 1: Components intended for use in non-background contexts shouldn't live in `/src/background`. I'd recommend `/src/common/telemetry` instead.

src/DetailsView/details-view-initializer.ts
```diff
@@ -235,6 +246,38 @@ if (tabId != null) {
...
+            const debugToolsTelemetryClient = new DebugToolsTelemetryClient(
+                browserAdapter,
+                applicationTelemetryDataFactory,
+            );
+            const telemetryClient = getTelemetryClient(applicationTelemetryDataFactory, [
+                consoleTelemetryClient,
+                debugToolsTelemetryClient,
+            ]);
+
+            const telemetryEventHandler = new TelemetryEventHandler(telemetryClient);
```
dbjorge, line 269: There would need to be a `telemetryStateListener` here making sure we don't send telemetry unless the user has explicitly configured the opt-in.

Thinking about this, and looking at the complexity this is adding to the page, it's making me re-think wanting to send telemetry directly from non-background contexts at all. If a non-background context can't communicate with the background, that means it might not be up to date with telemetry opt-in settings, so it's probably not safe for us to send telemetry in that case. Given that, I think it might make more sense to just have *all* non-background contexts use the `ForwardingExceptionTelemetryListener`, and just accept that if they truly can't communicate with the background at all that we'll miss those cases. I'd rather miss events than risk sending telemetry for users that haven't opted into it.

src/popup/popup-initializer.ts
```diff
@@ -1,12 +1,24 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { loadTheme } from '@fluentui/react';
+import { getStoreProxyFeatureFlagChecker } from 'background/feature-flag-checker';
```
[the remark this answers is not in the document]
dbjorge: (Same comment for details view init)

src/common/telemetry/sending-exception-telemetry-listener.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { BaseActionPayload } from 'background/actions/action-payloads';
+import { TelemetryEventHandler } from 'background/telemetry/telemetry-event-handler';
```
dbjorge, line 4: I think this `sending` listener *does* belong in the background, unlike the base class and the forwarding listener - it has dependencies on other background-only components and is only used from the background.

src/common/telemetry/exception-telemetry-listener.ts
```diff
@@ -1,13 +1,11 @@
...
 // Licensed under the MIT License.
-import { BaseActionPayload } from 'background/actions/action-payloads';
-import { TelemetryEventHandler } from 'background/telemetry/telemetry-event-handler';
 import { Logger } from 'common/logging/logger';
 import { escapeRegExp } from 'lodash';
 import * as TelemetryEvents from '../../common/extension-telemetry-events';
 import { UnhandledErrorTelemetryData } from '../../common/extension-telemetry-events';
 
-export class ExceptionTelemetryListener {
+export abstract class ExceptionTelemetryListener {
```
dbjorge, line 8: In general, it's usually better to prefer composition over inheritance for a situation like there where two instances of a component operate the same way modulo one inner call's behavior. Using `protected abstract` inheritance like this works, but:

* It's substantially more overhead to write and to test. Consider how this PR would look if `ExceptionTelemetryListener` instead accepted a constructor parameter like `private readonly publishTelemetry: (eventName: string, data: UnhandledErrorTelemetryData) => void`. The entire `ForwardingExceptionTelemetryListener` class and its tests could be eliminated, the `SendingExceptionTelemetryListener` would be cut in half, and its tests would become substantially simpler (no longer needing to jump through a test subclass)
* The overhead gets even worse if you ever need to add a second dependency like this; if that happens, the inheritance-based approach needs `2^numDeps` concrete classes, where the extra code required for the composition-based approach scales linearly

The book [Design Patterns](https://www.amazon.com/Design-Patterns-Elements-Reusable-Object-Oriented/dp/9332555400#customerReviews) is an oldie-but-goodie for a deeper exploration of the tradeoffs (you can probably get a digital copy from the MS Library, there might also be a copy in the team room)

src/debug-tools/initializer/debug-tools-init.tsx
```diff
@@ -105,13 +122,11 @@ const createStoreProxies = (storeUpdateMessageHub: StoreUpdateMessageHub) => {
...
 
-const getStoreActionMessageCreator = (browserAdapter: BrowserAdapter, stores: BaseStore<any>[]) => {
-    const actionMessageDispatcher = new RemoteActionMessageDispatcher(
-        browserAdapter.sendMessageToFrames,
-        null,
-        createDefaultLogger(),
-    );
-
+const getStoreActionMessageCreator = (
+    browserAdapter: BrowserAdapter,
```
dbjorge, line 126: nit: this impl doesn't actually use `browserAdapter` anymore. It might be cleaner to just inliine it, it's very simple and only used once.

src/tests/unit/tests/common/telemetry/exception-telemetry-sanitizer.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    });
+
+    describe('it returns undefined on invalid data', () => {
+        afterEach(() => {
+            const sanitizedTelemetry = testSubject.sanitizeTelemetryData(telemetryStub);
+
+            expect(sanitizedTelemetry).toBeUndefined();
+        });
+
+        test('it does not send data that includes urls', () => {
```
dbjorge, line 81: It would be good to include more robust test coverage of the complicated regex matching. A technique I like to use when trying to think of test cases is to consider any bugs I find when I'm testing stuff manually (or bugs that code review finds) as good cases to add regression tests for. I often try to write the regression test before I fix the bug in those cases (that's a good verification that the test works).

For example, some test cases that would be good to add which could act as regression tests for issues we've found/discussed during the review process might be:

* It should reject cases that include `file://` URLs, extension URLs for other extensions, or non-`chrome-extension://` URLs with our extension ID as the domain
* It should accept cases where excluded properties like html appear as part of a larger string, rather than enquoted like a JSON object key
* It should reject cases that quote excluded properties in single-quotes

src/common/telemetry/exception-telemetry-sanitizer.ts
```diff
@@ -0,0 +1,55 @@
...
+        telemetryData: UnhandledErrorTelemetryData,
+    ): UnhandledErrorTelemetryData | undefined {
+        if (telemetryData.message && telemetryData.message.length > this.MAX_MESSAGE_CHARS) {
+            telemetryData.message = telemetryData.message.substring(0, this.MAX_MESSAGE_CHARS);
+        }
+        if (telemetryData.stackTrace && telemetryData.stackTrace.length > this.MAX_STACK_CHARS) {
+            telemetryData.stackTrace = telemetryData.stackTrace.substring(0, this.MAX_STACK_CHARS);
+        }
+
+        const exclusionRegex = this.generateExclusionRegex();
```
dbjorge, line 33: Generating the regex is both stateless and expensive; it would be good to do it just once, rather than with every event. I'd probably do that by calling this from the constructor and storing the result in a private readonly field.

dbjorge (approved):

Issue #5187, opened 2022-02-17
Consider enforcing scss values


katydecorah:
## Is your feature request related to a problem? Please describe.

While working on https://github.com/microsoft/accessibility-insights-web/pull/5158, I noticed that the values of `font-weight` for semibold text varied: `600` or `$fontWeightSemiBold`. This does not present a bug, but by using variables (`$fontWeightSemiBold`) we have the potential for reducing human error and making the stylesheets more adaptable to global changes. 

## Describe the desired outcome

Introduce patterns to writing scss that are ideally enforced through tests, such as [stylelint](https://stylelint.io/). For example, with [stylelint-declaration-strict-value](https://github.com/AndyOGo/stylelint-declaration-strict-value) we could enforce that `font-weight` always uses a variable.

## Describe alternatives you've considered

- Manually update property values to a consistent pattern. (However, without tests we risk regression.)
- We could do nothing! As described, this work would be an investment that could help us write scss more efficiently (re: not needing to manually update values because swapped in variables) in the future.

## Additional context

I haven't yet investigated if there are other property values that should enforce a value. I think performing an audit of scss files and variables would be a good first step in determining scope, especially if we want introduce tooling like stylelint.

DaveTryon:
Make this a guideline for SCSS changes (already in OneNote)

PR #5584, opened 2022-06-06
refactor: Create shared ChoiceGroupPassFail component for high contrast improvements & stylelint

katydecorah:
#### Details

While enabling the stylelint rule [`selector-class-pattern`](https://stylelint.io/user-guide/rules/list/selector-class-pattern/) which enforces kebab case for all selectors, only two classes were failing, `.PASS` and `.FAIL`, which are used by the `TextStatusChoiceGroup` component. In examining the component, I found a few improvements, especially around high contrast:

- Add hover and active state in the default and high contrast themes. By switching to use the `IconButton` component, rather than `Link`, the undo button will be more consistent with other UI elements in the extension.
- Add hover state for the radio buttons in high contrast mode.
- Modernize the css by using flexbox.
- Rework css to avoid using nth selectors since they rely on the radio button options always having the same order. While unlikely, should the order of the options switch then the colors will swap: green = fail, red = pass. By assigning a specific class, we can improve the code readability and make the component more amenable to future changes.
- Fix a tiny code typo.

I also noticed a possible discrepancy in the snapshot test for the component. If I'm understanding correctly the prop `isLabelVisible` controls whether the label (`text`) or the aria label are visible at a given time. In the snapshots, both fields have a value when in practice only one or the other will have a value. To help with this, I set a default prop `isLabelVisible: false`. I also removed the `onRenderLabel` prop/function as it may no longer be needed.

#### Screenshots

The following table compares the default state of the component.  The current undo button icon is text colored. While the proposed undo button's icon is colored blue, which is consistent with FluentUI buttons. Elements in the proposed are also more horizontally centered and the width is slightly wider, but still fits when the component is found inside a table.

Current | Proposed
---|---
 | 
 | 

The following table compares the hover state of the undo button. The current undo button has no hover state, while the proposed will have a background color change as prescribed by all FluentUI buttons.

Current | Proposed
---|---
 | 

The following table compares the undo button in high contrast mode. The current undo button color is purple, while the proposed undo button color is yellow, which is consistent with the theme's button color. In both screenshots, a radio button is selected to show the active state which is the radio button's default color (no change between versions).

Current | Proposed
---|---
 | 

The following table compares the undo button's hover state in high contrast mode. The current undo button has no state change, while the proposed undo button color changes.

Current | Proposed
---|---
 | 

The following table compares a radio button hover in high contrast mode. The current radio buttons have no hover state, while the current will have a border and background state change.

Current | Proposed
---|---
 | 

##### Motivation

#### Pull request checklist

- [x] Addresses an existing issue: #5187 
- [x] Ran `yarn fastpass`


madalynrose:
Nice catch @katydecorah! It bothered me that the undo button didn't seem to obey by the same button styles as others but I couldn't figure out why. Would using this component for the Tab Stops failure instances in Fast Pass fix some of the oddness there? I'm not sure if we can actually use it there because it includes extra UI for adding descriptions of instances but it would be awesome if we could make the behavior consistent.

katydecorah:
> Would using this component for the Tab Stops failure instances in Fast Pass fix some of the oddness there? I'm not sure if we can actually use it there because it includes extra UI for adding descriptions of instances but it would be awesome if we could make the behavior consistent

@madalynrose Aha! I think there is an opportunity here, I'll take a closer look. Thank you!

madalynrose (approved):
This looks so much tidier in code and consistent in the UI! Awesome work!

PR #5591, opened 2022-06-06
feat(mv3-part-4): implement a passthrough event manager for use with non-background contexts

dbjorge:
#### Details

This PR changes our browser event handling strategy for non-background contexts to use a new `PassthroughBrowserEventManager` instead of going through the same event management logic that the background requires.

##### Motivation

A followup PR (#5538) involves enabling sharing of some browser event manager code (for creating fnf delays/etc) with the background interpreter in a way that is easiest if it can be allowed to use background-only types. This made it awkward to create the event manager in non-background contexts.

This also simplifies most of the context initializers substantially; it enables moving the interesting logic out of the initializers (which are hard to test) and into separate components (which now have tests in front of them.

Finally, it removes some risk/complexity from places where it isn't needed. For example, doing this earlier would have prevented the issue #5582 fixed.

##### Context

This code was extracted from #5538, since it was getting uncomfortably large.

Technically, we could also use the passthrough event manager from the mv2 background page context. I chose to keep the background event manager there to minimize differences between mv2 and mv3, for the sake of catching issues with the event manager earlier.

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/common/browser-adapters/webextension-browser-adapter.ts
```diff
@@ -14,29 +17,37 @@ import browser, {
...
-    public initialize(browserEvents: DictionaryStringTo<Events.Event<any>>): void {
-        // Add browser listeners synchronously on initialization
-        for (const [eventType, event] of Object.entries(browserEvents)) {
-            this.browserEventManager.addBrowserListener(event, eventType);
-        }
-    }
-
     public abstract getManageExtensionUrl(): string;
 
+    public allRequiredEvents(): DictionaryStringTo<Events.Event<any>> {
```
sfoslund, line 27: If we're moving the list of required events back here then we can remove the browser-event-provider.

src/common/browser-adapters/webextension-browser-adapter.ts
```diff
@@ -14,29 +17,37 @@ import browser, {
...
+        return {
+            TabsOnActivated: browser.tabs?.onActivated,
```
sfoslund, line 29: We originally restructured things in part to avoid having all of these undefined checks that could cause odd errors, right? I would prefer to keep the browser-event-provider or something analogous so we don't have to have all of these `browser.foo?...` checks here.

sfoslund (commented):
Looks good but would like to make sure we don't leave the browser event provider dangling before I approve.

src/common/browser-adapters/webextension-browser-adapter.ts
dbjorge: The `?` checks weren't actually necessary here, they were a relic from an earlier attempt that didn't use a containing function. Removed them!

sfoslund (approved):

PR #5598, opened 2022-06-07
feat(mv3-part-4): revert async-ification of trackEvent

dbjorge:
#### Details

This PR reverts a part of #5524 that involved making our telemetry clients' `trackEvent` async in the interest of propogating the return value of the `browserAdapter.sendRuntimeMessage` call used by `debug-tools-telemetry-client.ts`.

The issue with that strategy is that it doesn't cooperate with the new unhandled promise rejection hook added by #5550 - particularly, the hook ends up entering an infinite loop if that `sendRuntimeMessage` call fails for any reason. You can see this by clean-installing a dev build of the extension, opening the background page devtools from the extension management page, and enabling the console telemetry logging and debug tools feature flags from the background console (never opening any other extension page/popup UI).

Instead, this PR:

* Reverts `trackEvent` back to synchronous (to make it part of the `TelemetryClient` contract that the onerror hook anti-reentrancy logic will work)
* Moves the promise-voiding to be internal to the `debug-tools-telemetry-client` with a comment explaining why
* Adds a no-op catch handler to the voided `sendRuntimeMessage` promise, to ensure that it won't enter an infinite loop with the rejected promise hook that... tries to send error telemetry.

##### Motivation

Avoid an infinite loop found during #5591 testing (thanks @sfoslund for tracking down repro steps!)

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/telemetry/app-insights-telemetry-client.ts
+++ b/src/background/telemetry/app-insights-telemetry-client.ts
@@ -36,7 +36,7 @@ export class AppInsightsTelemetryClient implements TelemetryClient {
         this.updateTelemetryState();
     }
 
-    public async trackEvent(name: string, properties?: { [name: string]: string }): Promise<void> {
+    public trackEvent(name: string, properties?: { [name: string]: string }): void {
         if (this.enabled) {
             this.applicationInsights.trackEvent({ name }, properties);
         }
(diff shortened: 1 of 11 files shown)


pownkel (approved):
LGTM, great work tracking down this bug!

sfoslund (approved):

PR #5615, opened 2022-06-11
fix(electron-orca-a11y-bugs): Announce which item in a dropdown list is selected

pownkel:
#### Details

Adds an aria-label to the selected item in dropdown lists to indicate that that item is selected. This affects the details view switcher in AI Web and the Azure boards issue filing menu in electron (pictured below).

details view switcher

select a field for issue details dropdown

##### Motivation

The fluentUI Dropdown component does correctly set aria-selected on dropdown items. However, Orca and JAWS do not announce which items in any dropdown or combobox are selected, so a JAWS or Orca user navigating through the dropdown menu cannot tell what item is currently selected, since selection does not follow focus.

##### Context

NVDA announces "not selected" after a non-selected list item, but does not announce "selected" for a selected item. This PR only adds "selected" after the selected item, so the state will not be read twice with NVDA. This does not affect screen reader behavior when the dropdown is closed.

#### Pull request checklist

- [x] Addresses an existing issue: #1944170
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/switcher.tsx
+++ b/src/DetailsView/components/switcher.tsx
@@ -1,8 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { Dropdown, IDropdownOption } from '@fluentui/react';
+import { IDropdownOption } from '@fluentui/react';
 import { Icon } from '@fluentui/react';
 import { ResponsiveMode } from '@fluentui/react';
+import { InsightsDropdown } from 'common/components/insights-dropdown';
 import { DetailsViewPivotType } from 'common/types/details-view-pivot-type';
 import * as React from 'react';
 
@@ -80,7 +81,7 @@ export class Switcher extends React.Component<SwitcherProps, SwitcherState> {
     public render(): JSX.Element {
         return (
             <div className={styles.leftNavSwitcher} role="region" aria-label="activity">
-                <Dropdown
+                <InsightsDropdown
                     className={styles.leftNavSwitcherDropdown}
                     ariaLabel="select activity"
                     responsiveMode={ResponsiveMode.large}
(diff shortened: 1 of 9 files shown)


dbjorge:
This feels like something that would be preferable to report/contribute upstream to Fluent UI. It looks like the last two times this was brought up there were https://github.com/microsoft/fluentui/issues/16490, where the issue was ignored for a year and then soft-closed, and https://github.com/microsoft/fluentui/issues/22086, where @smhigley gave an explanation that suggested that the JAWS behavior was the same as the NVDA behavior (announcing "not selected" on the not-selected cases).

Can you confirm that JAWS and Orca do *not* read "not selected" for the non-selected cases? If so, we should update/reopen the upstream bug (we might still apply this workaround here, but we should reference the upstream bug in a comment)

pownkel:
@dbjorge Yes, I've confirmed that neither JAWS nor ORCA reads "not selected." This seems to be an issue with the screenreaders themselves, since I also see it happen with [non-fluent UI dropdowns](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html).

I've submitted a technical support request to JAWS, and I believe [#171](https://gitlab.gnome.org/GNOME/orca/-/issues/171) in the ORCA repo is related. The discussion thread suggests that a fix has already been merged, but unfortunately ORCA 3.36 is the highest version supported on Ubuntu 20.04/GNOME 3.36, so new fixes aren't available without an OS upgrade or manual GNOME upgrade.

dbjorge:
Thanks for investigating further, @pownkel! I agree that #171 seems related. I would recommend following the process to resolve the bug as external against Orca, link #171, and note in the comments that it is already resolved in the latest version of Orca and that Orca's design decision for this case is to distinguish the "selected" and "not selected" cases by announcing the "not selected" cases and not the "selected" cases, similar to NVDA's behavior.

@ferBonnin, is there any special handling for a case like this where we think the issue is already resolved in a more recent version of the AT in question than the trusted testers tested against? Are we able to ask them to verify against the more recent version without triggering an entirely new pass?

PR #5623, opened 2022-06-13
feat(mv3-part-4): prevent spurious getStoreState messages

dbjorge:
#### Details

This PR fixes an issue where the initial messages used to seed data in `StoreProxy`s were sent:
* Multiple times in cases where multiple components in a page use `withStoreSubscription` (eg, popup sent 3 copies of each message)
* For non-proxy stores that aren't backed by the background page (eg, the `TabStopsViewStore`)

The second case is the biggest motivator - with the changes in #5538, those unnecessary messages reach the background page and cause errors there because there are (rightfully) no handlers in the background to receive them.

To fix this, I moved the "send initial getStoreState messages" logic out of `with-store-subscription` and into `StoreUpdateMessageHub`, the same component responsible for receiving and distributing the responses to those messages already. This also enabled removing a few unnecessary components (`StoreActionMessageCreator` and `StoreActionMessageCreatorFactory`)

I decided that the easiest way to avoid having to double-register things or risk missing sending the seed messages was to just send them at registration. This removed a bunch of initializer boilerplate, but had the side effect of moving the sending of those messages earlier; to make sure that didn't break anything, I added some extra error handling to ensure that the `StoreUpdateMessageHub` registers its browser listener *before* individual store proxies register with the hub.

I also included some clean up in related areas as part of the PR:

* Consolidated the confusingly-named `BaseClientStoresHub` and `ClientStoresHub` (the "Base" one was the only subclass of the non-"Base" one)
* Refactored repeated message distribution logic in `DevToolMessageDistributor` and `DebugToolsMessageDistributor` into a shared `BrowserMessageDistributor`
* Merged the insights page's `dependencies.ts` into its initializer - the test wasn't verifying anything valuable and became harder to set up correctly with the new side effects happening pre-render, and containing the initialization felt more consistent with other initializers

##### Motivation

Unblocks #5538 by fixing an issue with Tab Stops that happens because of the spurious `TabStopsViewStore` messages

##### Context

This is a relatively dangerous PR because it substantially changes initialization order, which isn't covered by unit tests. I've done some manual smoke testing on each of the impacted pages (+ unified) and some more thorough manual smoke testing with the debug tools, dev tools, and tab stops visualizers.

#### Pull request checklist

- [x] Ran `yarn fastpass`


pownkel (approved):
LGTM

PR #5638, opened 2022-06-16
fix: use maxWorkers=2 for GitHub Action CI build

waabid:
#### Details

Specifies maxWorkers to 2 for our CI build in GH action.

##### Motivation

Speeds up CI builds.

##### Context

I came upon this post: https://bharathvaj.me/blog/speedup-jest-github-actions.

TLDR: Linux VMs on GH Action have two cores (link to GH docs [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)) but jest defaults to only using one. This specifies using both. Our unit-tests were previously the slowest portion of our CI builds in GH action and this makes them see about a 40%-50% decrease in runtime. More importantly, this makes the e2e/unified tests our longest-running tasks (saving roughly 2-3 minutes per PR build, about a 25% decrease, based off looking at the last 10-15 successful builds without this change).

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,8 @@ jobs:
     - run: yarn install --frozen-lockfile --prefer-offline
       timeout-minutes: 10
 
+    # We set maxWorkers to 2 since GH Action VMs have 2 cores and by default Jest only uses 1.
+    # If this were to change in the future, the number here should change accordingly.
     - run: yarn test --maxWorkers=2 --ci --shard=${{ matrix.shard-index }}/${{ strategy.job-total }}
       timeout-minutes: 10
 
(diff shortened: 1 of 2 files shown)


brocktaylor7 (approved):
This is great, Wahaj! From what I can tell, we have separate jobs for the unit tests so this doesn't affect other jobs running concurrently in any way, is that correct? Assuming that's correct, this looks like a nice little win. Good work!

waabid:
> This is great, Wahaj! From what I can tell, we have separate jobs for the unit tests so this doesn't affect other jobs running concurrently in any way, is that correct? Assuming that's correct, this looks like a nice little win. Good work!

Yep! 😄

PR #5158, opened 2022-02-10
fix: Fix semibold font-weight rendering in Chrome

katydecorah:
#### Details

This PR will fix a `font-weight` rendering issue in Chrome where some text was not appearing as semibold. This fix includes:
- removing the semibold font stack as it appears to not have an effect on the weight of the font.
- setting `font-weight: $fontWeightSemiBold` wherever the semibold font stack was in use.

Current | Proposed
---|---
screenshot highlighting issues in Details View automated checks  | screenshot fixes issues in Details view automated checks
screenshot highlighting issues in FastPass report failed instances | screenshot fxies issues in FastPass report failed instances
screenshot highlighting issues in FastPass report passed checks | screenshot fixes issues in FastPass report passed checks

##### Motivation

Addresses #5131 

##### Context

Initially, I created a new scss mixin to bundle the `font-weight` and `font-family`, see commit: https://github.com/microsoft/accessibility-insights-web/pull/5158/commits/bed8c50ffff7724ae7738e8e430e7699443e4ebd. Upon further testing, it appears we no longer need to declare the semibold `font-family` stack, see comment: https://github.com/microsoft/accessibility-insights-web/pull/5158#issuecomment-1036539902.

I noticed throughout stylesheets that font-weight values can be either `600` and `$fontWeightSemiBold`. In a future pull request, we could consider introducing stylelint rules to make this value consistent.

## Pull request checklist

- [x] Addresses an existing issue: #5131
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [x] (UI changes only) Added screenshots/GIFs to description above
- [n/a?] (UI changes only) Verified usability with NVDA/JAWS

PR #5655, opened 2022-06-22
chore: Enable and fix `declaration-property-value-allowed-list` stylelint rule for `font-weight`

katydecorah:
#### Details

Enables [`declaration-property-value-allowed-list`](https://stylelint.io/user-guide/rules/list/declaration-property-value-allowed-list/) for `font-weight` where the value of must be one of `'unset', '100', '300', 'normal', '600', 'bold'`. The goal of this rule is to make font-weight values consistent.

To consider for the future: 100 and 300 are both used once (in launchpad title and footer, respectively). While I don't think we'll want to remove 100 as it's part of the extension's brand, we could consider removing the 300 as the visual difference between weight 300 and 400 in the launch pad footer is slight.

This pull request will also remove an unused selector rule -- I confirmed that it's not used in web or unified. I believe the `h2` (How to test) is now a div/span with an aria-level set and is styled by another selector now.

Related:
- https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#syntax

##### Motivation

##### Context

Initially, I thought we should use scss variables to define font-weights, but then I did a quick dive into how font-weights are currently set. I found that:

- Most font-weight variables were not used at all or barely used, with the exception of `font-weight-semi-bold`, but that's because I introduced the variables usage in https://github.com/microsoft/accessibility-insights-web/pull/5158 (😅). See table below for a full audit.
- Switching to require a variable value resulted in `116 additions and 106 deletions` whereas this pull request results in `63 additions and 75 deletions`. This is an additional signal that sticking with the current patterns (using css values) could be less disruptive.
- Variables for font-weights can be helpful if they are descriptive to its use rather than value (for example, `title-font-weight`). Since the codebase names font-weight variables for their value, we could use css values and cut out a dependency. (For example, if we switched to variables, we would need to import font.css in a few files that didn't have it before to support the variable.)

One downside to not using variables is if we decide to change weights. With variables, we'd have a single place to update, whereas with the proposed we'd need to find and replace. With how text is styled right now, I don't think we would see the benefits of switching to font-weight variables right now. As we continue to enable more stylelint rules, I think more patterns will emerge and opportunities to reduce redundancy and improve consistency with mixins.

Below is a table of all the values of `font-weight` and how many time the value is used on production:

Weight | Number of times used
---|---:
100 | 1
$font-weight-lighter (= 200) | 0
300 | 1
$normal-title-font-weight (= 300) | 1
400 | 6
normal (= 400) | 7
$font-weight-normal (= 400) | 2
$font-weight-semi-bold (= 600) | 34
600 | 18
700 | 4
bold (= 700) | 5
$font-weight-heavy (= 700) | 0

When setting the rule's values, I ultimately went with what was already used the most. Both `normal` and `bold` were ahead of their counterparts `400` and `700` by 1 additional usage. If we prefer to use all number values, for example use `400` instead of `normal`, I'd be ok with that as well! I definitely welcome suggestions to improve the ergonomics of this rule.

#### Pull request checklist

- [x] Addresses an existing issue: #5187 
- [x] Ran `yarn fastpass`

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -11,6 +11,9 @@ module.exports = {
         'scss/comment-no-empty': null, // Disabled to allow for paragraph breaks in longer comments which is not supported yet: https://github.com/stylelint-scss/stylelint-scss/issues/606
         'selector-class-pattern':
             '^(ms-([A-Z][a-z0-9]*)(-[a-z0-9]+)*)|(([a-z][a-z0-9]*)(-[a-z0-9]+)*)$', // Allows: kebab case and ms-Kebab-case
+        'declaration-property-value-allowed-list': {
+            'font-weight': ['unset', '100', '300', 'normal', '600', 'bold'],
+        },
 
         // TO BE ENABLED: Recommended fixes
         'no-descending-specificity': null,
@@ -18,9 +21,6 @@ module.exports = {
         // TO BE CONFIGURED: Limit shorthand for margin, padding
         // Stretch goal: border-width, border-radius, border-color, border-style, grid-gap
         // Example: 'declaration-property-max-values': { padding: 1 },
-
-        // TO BE CONFIGURED: Enforce variable values
-        // Example: 'declaration-property-value-allowed-list': { 'font-weight': ['/^\\$.*$/']},
 
         // Requires investigation
         'property-no-vendor-prefix': null,
--- a/packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
+++ b/packages/report-e2e-tests/src/examples/axe-results-with-issues.snap.html
@@ -458,7 +458,7 @@
         font-size: 11px;
         border-radius: 0 8px 8px 0;
         padding: 0 6px 0 10px;
-        font-weight: 700;
+        font-weight: bold;
         margin-left: -8px;
         height: 13px;
       }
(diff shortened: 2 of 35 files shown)


katydecorah:
I've been thinking about this some more and I am on the fence on whether we should go with css values or variables.  After looking at `font-size`, we will be in a similar situation, where the stylesheets use css values more often than the variables. Originally, I based my choice to follow the established patterns (css values), while that solves reducing contributor friction, it's unclear to me if we're optimizing the stylesheets as system.

If we put on our solve-root-problems-hat, the stylesheets style text in many different combinations of font-sizes, weights, and line-heights. When approaching this codebase, I don't see a pattern or guidance for how to style text. This pull request will not solve that, but I think as a whole, enabling stylelint rules will help us uncover patterns. Once we've moved through enabling stylelint, I anticipate that the work will us develop patterns so contributors can pull in reusable text styles (via mixins or other) and we can document them.

All that said, I think what we choose here is a small step towards consistency and may be adjusted later as we improve upon our typography system.

katydecorah:
@dbjorge shared the history of some of the typography variables with me and it sounds like many of them came from Figma. It could be advantageous to sync variable names with how they're defined in Figma, especially when building new features (example: our team can use the established scss variables that match the Figma mock). 

I'm converting this back to draft to allow some extra time to look into this new direction!

PR #5693, opened 2022-06-30
test(performance): improve mock-adb perf 30x by converting to Rust

dbjorge:
#### Details

This PR is an experiment to migrate the implementation of `mock-adb` from TypeScript to Rust.

##### Motivation

While debugging some test timeouts that @madalynrose and @ThanyaLeif were seeing as part of the Electron v18 update feature, we found that one of the underlying issues causing timeouts was that some tests (notably, any tests whose setup involves going through the whole device setup flow) made a lot of `adb.exe` calls, and those calls have a substantial amount of overhead.

On Tania's dev box, each individual call to mock-adb (with a warm cache) takes ~2.2s. On my dev box, they take 0.8s. Running the full test suite involves hundreds of these calls, and at 2.2s most of our per-test timeout allocation is spent there.

On my dev box, this rust version is about 30x faster; it takes ~500ms with a *cold* cache (mostly waiting to read the config file), and ~35ms with a warm cache. This speed increase is mostly because the Rust version avoids the overhead of spawning an entire Node.js runtime every time it starts up.

##### Context

Introducing a new language to the repo is a big change that we shouldn't take lightly. This PR is an experimental draft intended to see whether doing this really has a meaningful benefit on test reliability; my hope is that if it does, it might also help the timeouts we see when running the e2e tests in CI builds on windows/mac agents.

If we decide to accept this change, it might be reasonable for us to split mock-adb out into a separate repo to avoid the requirement for everyone working on unified to have an up-to-date Rust installation. That would probably warrant a bigger change that also pulls the test integration bits into the same separate package and produces npm packages per-architecture bundling the executable alongside the test integration library.

#### Pull request checklist

- [x] Ran `yarn fastpass`


dbjorge:
Closing this for now pending further engineering discussion

Issue #5317, opened 2022-04-01
Automated Tab Stops Focus Trap False Positive


sfoslund:
## Describe the bug

A false positive focus trap failure is being reported through the new automated tab stops checks.

**To Reproduce**

1. Go to https://www.washington.edu/accesscomputing/AU/after.html
2. Click on the cheat sheet of a11y issues
3. Run tabstops over the modal dialog
4. Tab at least 3 times, so that focus loops back to the close ('X') button
5. See that focus trap failure is incorrectly reported on the close button

## Screenshots

[Screenshot of modal open with focus trap failure visualization over close button in top right corner]
![FocusTrapFalsePositive](https://user-images.githubusercontent.com/16010855/161346781-9d4484a5-2960-42c4-9548-a39b52d548f6.png)

## Additional context

Issue found when investigating https://github.com/microsoft/accessibility-insights-web/issues/5305

ferBonnin:
reviewed with Peter

sfoslund:
Resolved by #5705

PR #5705, opened 2022-07-06
fix: Store last focused element for focus trap detector

pownkel:
#### Details

- Extracts TabStopsOrchestrator's keydown listener into its own class, FocusTrapsKeydownHandler
- Store the last focused element in FocusTrapsKeydownHandler and use that instead of dom.activeElement to determine whether we've reached a focus trap

##### Motivation

Fixes a race condition in TabStopsOrchestrator that causes a focus trap false positive. The root cause is that, in the keydown listener, we assume that the focus has not yet changed and use dom.activeElement as the "last focused element." However, sometimes the keyboard listener runs after focus has already changed, in which case the extension wrongly detects that the previous and currently focused elements are the same, and reports a focus trap.

This PR fixes the race condition by storing the last focused element in an instance variable that is updated after every tab press.

##### Context

I first tried using TabStopsOrchestrator's existing latestVisitedTabStop variable, but that created a similar race condition: latestVisitedTabStop is changed by the onfocus listener, which may run either before or after the keydown listener. Since each listener needs to track its previous tab stop separately, I thought it would be clearer to refactor the keydown listener into a new class with its own state instead of adding another variable with a confusingly similar name to TabStopsOrchestrator.

#### Pull request checklist

- [x] Addresses an existing issue: #5317
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/analyzers/focus-traps-keydown-handler.ts
+++ b/src/injected/analyzers/focus-traps-keydown-handler.ts
@@ -1,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { PromiseFactory } from 'common/promises/promise-factory';
+import { AutomatedTabStopRequirementResult } from 'injected/tab-stop-requirement-result';
+import { TabStopsRequirementEvaluator } from 'injected/tab-stops-requirement-evaluator';
+
+export class FocusTrapsKeydownHandler {
+    // TabStopsOrchestrator already stores latestVisitedTabStop,
+    // but that may be modified by other event listeners before
+    // this one runs, so we need to keep this state separate
+    protected lastFocusedElement: Element | null = null;
+
+    constructor(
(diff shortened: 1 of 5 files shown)


src/injected/analyzers/focus-traps-keydown-handler.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    public getResultOnKeydown = async (
+        e: KeyboardEvent,
+        dom: Document,
+    ): Promise<AutomatedTabStopRequirementResult | null> => {
+        if (e.key !== 'Tab' || dom.activeElement === dom.body) {
+            return null;
+        }
+
+        await this.promiseFactory.delay(null, this.keyboardTrapTimeout);
```
sfoslund, line 32: I have a vague memory of @waabid preferring to use set timeout instead of awaiting here but I don't remember the details/ I might be misremembering. Wondering if he has any thoughts here.

sfoslund (approved):
LGTM!

PR #5713, opened 2022-07-08
refactor: Break up TabStopsOrchestrator

pownkel:
#### Details

Extracts state management and focusin listener out of TabStopsOrchestrator and into a separate class, TabStopsHandler. After this PR, TabStopsOrchestrator's main function is to call methods on TabStopsHandler and FocusTrapsHandler at the appropriate times and reporting results. TabStopsHandler and FocusTrapsHandler contain the main logic for processing new tab stops or key presses and managing the associated state.

##### Motivation

Previously, TabStopsOrchestrator was in charge of most of the logic and state management for tab stops. This doesn't make a lot of sense when both event listeners need to track their own similar states, but also keep them separate. #5705 extracted the keyboard listener into its own class that manages its own state, but left the focus listener and associated state in TabStopsOrchestrator. This PR extracts that listener and its state as well for clarity and consistency. As a bonus, this also simplifies tests for TabStopsOrchestrator and makes it easier to more thoroughly test the state management of the focus listener.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/unit/tests/injected/analyzers/tab-stops-handler.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        );
+        testSubject.initialize();
+    });
+
+    afterEach(() => {
+        requirementEvaluatorMock.verifyAll();
+        tabbableElementGetterMock.verifyAll();
+    });
+
+    it('getTabbabbleFocusOrderResults() calls requirement evaluator with tabbabble elements', () => {
```
sfoslund, line 48: It looks like we don't have any tests for `getKeyboardNavigationResults`. It's exercised as a side effect of other tests but personally I'd like to have a standalone test for it as well.

PR #5743, opened 2022-07-13
refactor(mv3-part-5): SyncAction and AsyncAction

pownkel:
#### Details

- Rename Action to SyncAction
- Create a generic Action interface
- Create an AsyncAction, which accepts async listeners and returns a Promise from invoke()
- Refactor scoping/mutex logic out of SyncAction and share it with AsyncAction

##### Motivation

Feature work (Converting Flux logic to use promises as part of the migration to Manifest V3)

##### Context

Future work will include converting existing Actions/SyncActions into AsyncActions

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/flux/async-action.ts
+++ b/src/common/flux/async-action.ts
@@ -14,9 +14,10 @@ export class AsyncAction<TPayload> implements Action<TPayload, Promise<void>> {
         this.scopeMutex.tryLockScope(scope);
 
         try {
-            await Promise.all(
+            const results = await Promise.allSettled(
                 this.listeners.map((listener: (payload: TPayload) => void) => listener(payload)),
             );
+            this.throwCombinedRejections(results);
         } finally {
             this.scopeMutex.unlockScope(scope);
         }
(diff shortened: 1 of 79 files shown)


src/common/flux/async-action.ts
```diff
@@ -0,0 +1,44 @@
...
+        } finally {
+            this.scopeMutex.unlockScope(scope);
+        }
+    }
+
+    public addListener(listener: (payload: TPayload) => Promise<void>): void {
+        this.listeners.push(listener);
+    }
+
+    private throwCombinedRejections(promiseResults: PromiseSettledResult<void>[]): void {
```
sfoslund, line 30: My understanding was that we would combine listener results with the `EventResposeFactory`. This seems to be duplicating a lot of the logic in `mergeAsyncResponses` anyways, can we call that here instead?

src/tests/unit/tests/common/flux/async-action.test.ts
```diff
@@ -0,0 +1,124 @@
...
+                })
+                .verifiable(Times.once());
+            testObject.addListener(mock.object);
+        });
+
+        scopeMutexMock.setup(m => m.tryLockScope(undefined)).verifiable(Times.once());
+        scopeMutexMock.setup(m => m.unlockScope(undefined)).verifiable(Times.once());
+
+        try {
+            await testObject.invoke(testPayload);
```
sfoslund, line 106: Is it possible to use the `expect(...).toThrowError(...)` syntax here instead of try/ catch?

sfoslund (approved):

src/tests/unit/tests/common/flux/async-action.test.ts
pownkel: Resolving because this test was removed

src/common/flux/async-action.ts
pownkel: (Discussed offline with Dan) EventResponseFactory is a bit more involved than what we need here: it also adds a delay to fire-and-forget handlers, which we wouldn't need to do here, and takes some additional dependencies that we wouldn't need to use. I refactored EventResponseFactory to extract only the logic of mergeAsyncResponses so we can reuse that here.

sfoslund (approved):
Changes look good to me!

Issue #5609, opened 2022-06-09
[Usability] Reflow buttons in failure cards


ferBonnin:
## Describe the bug

currently the failure cards in the automated checks and needs review instances have a "..." button by default. When this UI was first implemented, we determined reflowing the buttons to file bugs and copy failure details was not feasible. This bug is to revisit this after we updated the fluent UI version used 

**To Reproduce**

1. [Go to accessible university demo site](https://www.washington.edu/accesscomputing/AU/before.html)
2. Run FastPass > automated checks
3. notice how the failure cards always have the "..." button

## Expected behavior

the two buttons inside that menu appear / disappear as reflow allows 

## Context (please complete the following information)

- OS Name & Version:  N/A
- AI-Web Version & Environment:  Prod 2.32

## Are you willing to submit a PR?
No

## Did you search for similar existing issues?
Yes

## Additional context
Usability. Exploratory bug; if we determine this is now feasible let's chat with PM

pownkel:
@ferBonnin I don't have much context for why this was originally infeasible, but I have a working version on [my own branch](https://github.com/microsoft/accessibility-insights-web/compare/main...pownkel:accessibility-insights-web:reflow-cards?expand=1) that reflows those buttons similarly to how it's done in the command bar and handles focus when the issue filing dialog closes
![gif showing the Automated Checks screen being expanded and the kebab button on cards expanding to show the "file issue" and "copy failure details" buttons](https://user-images.githubusercontent.com/55459788/178595961-d0135e0e-b275-4e09-9ef6-4f9ffe88c0f8.gif)
![gif showing the issue filing settings dialog being opened and closed at different widths. When the dialog is closed, focus returns to either the kebab button or the "file issue" button, depending on what is visible](https://user-images.githubusercontent.com/55459788/178596758-b0a5e5d3-5d81-459c-ac68-c3c310079af9.gif)

ferBonnin:
This is awesome, thanks @pownkel! We should do this change, please!

pownkel:
@ferBonnin this is complete and the change is in canary!

ferBonnin:
thanks @pownkel, looks great!

PR #5745, opened 2022-07-14
fix: Reflow buttons in failure cards

pownkel:
#### Details

Currently, the issue detail cards always have a "..." button in the corner that opens a menu with two buttons. This PR will expand those two buttons when the screen is wide enough to show them, and collapse them back into the "..." button when the screen is narrow:
178595961-d0135e0e-b275-4e09-9ef6-4f9ffe88c0f8

This PR also makes sure that focus correctly returns to the "..." button or the "file issue" button when the issue filing settings dialog is closed:
178596758-b0a5e5d3-5d81-459c-ac68-c3c310079af9

##### Motivation

Address #5609

##### Context

Still to be done on this draft PR:
- [x] Validate this change in report package and electron
- [x] Explore breaking up/refactoring CardFooterFarButtons, which was already a large component and got even larger with this change

Refactors deferred to a separate PR:
- Extract out getMenuItems and the onClick methods from CardFooterFarButtons (already done on another branch)
- Use store instead of component state to determine whether the instance filing setting dialog is open

Also worth mentioning is that this PR uses the width of the entire screen to determine if cards should reflow, rather than the width of the cards themselves. This is not ideal, especially in electron where both the left nav and the right-hand screenshot also resize as the screen resizes, and the card width is affected by more than just screen width. It doesn't make a huge difference from a user perspective, but in the future, we may want to consider refactoring NarrowModeDetector to be more generic (it currently assumes that it's measuring the width of the whole screen and sets its thresholds accordingly) and adding another NarrowModeDetector to the element containing the cards.

#### Pull request checklist

- [x] Addresses an existing issue: #5609
- [x] Ran `yarn fastpass`


pownkel:
I'm planning to break this work into several PRs, starting with #5852. Closing this one.

Issue #5753, opened 2022-07-14
TabStore crashes and breaks background initialization when given an invalid URL


dbjorge:
## Describe the bug

If `TabStore` ever tries to handle a tab-updated event that involves a URL which cannot be parsed by `new URL()`, it emits an unhandled exception which prevents the completion of background initialization. The stack looks like:

```
TypeError: Failed to construct 'URL': Invalid URL
    at TabStore.originsMatch (chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:89756:43)
    at TabStore.onExistingTabUpdated (chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:89737:19)
    at chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:86744:11
    at Array.forEach (<anonymous>)
    at _Action.invoke (chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:86743:24)
    at TabActionCreator.onExistingTabUpdated [as insights/tab/existingTabUpdated] (chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:89074:44)
    at Interpreter.interpret (chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:87138:67)
    at TabContextManager.interpretMessageForTab (chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:83838:30)
    at TargetPageController.interpretMessageAsync (chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:90460:47)
    at TargetPageController.sendTabUrlUpdatedAction (chrome-extension://hbcplehnakffdldhldncjlnbpfgogbem/bundle/background.bundle.js:90442:18)
```

Once in this state, the whole extension is broken and just shows loading spinners until the next time the background is torn down and re-initialized.

**To Reproduce**

Not sure how to reproduce reliably; I encountered the issue locally by accident. We see it pretty regularly in telemetry against insider/canary builds, though; team members with staging telemetry access can use this query to search for instances (there are 46 in the last 30d):

```kusto
customEvents
| where timestamp > ago(30d)
| where name == 'unhandledError'
[...]

PR #5754, opened 2022-07-14
fix: avoid crashing when a browser tab has an invalid URL

dbjorge:
#### Details

* Moves some URL-origin-parsing logic specific to `TabStore` out into `UrlParser` for ease of testing
* Adds support to the logic to handle invalid URL cases

`UrlParser` has no dependencies and there is no other component in the background that uses `UrlParser` today, so I just initialized it in the `TabStore` directly rather than piping it around everywhere.

##### Motivation

Addresses issue #5753

##### Context

I don't have a good way to test the change locally, since we don't have consistent repro steps, but the root cause seems easy enough to understand and it shows up commonly enough in insider/canary telemetry that we should be able to determine if the fix worked by checking there in a few days.

#### Pull request checklist

- [x] Addresses an existing issue: #5753
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/stores/tab-store.ts
+++ b/src/background/stores/tab-store.ts
@@ -23,7 +23,7 @@ export class TabStore extends PersistentStore<TabStoreData> {
         logger: Logger,
         tabId: number,
         persistStoreData: boolean,
-        private readonly urlParser: UrlParser = new UrlParser(),
+        private readonly urlParser: UrlParser,
     ) {
         super(
             StoreNames.TabStore,
--- a/src/background/stores/tab-store.ts
+++ b/src/background/stores/tab-store.ts
@@ -7,6 +7,7 @@ import { Tab } from 'common/itab';
 import { Logger } from 'common/logging/logger';
 import { StoreNames } from 'common/stores/store-names';
 import { TabStoreData } from 'common/types/store-data/tab-store-data';
+import { UrlParser } from 'common/url-parser';
 import { TabActions } from '../actions/tab-actions';
 import { VisualizationActions } from '../actions/visualization-actions';
 
@@ -22,6 +23,7 @@ export class TabStore extends PersistentStore<TabStoreData> {
         logger: Logger,
         tabId: number,
         persistStoreData: boolean,
+        private readonly urlParser: UrlParser = new UrlParser(),
     ) {
         super(
             StoreNames.TabStore,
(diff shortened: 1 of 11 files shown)


src/common/url-parser.ts
```diff
@@ -33,4 +48,26 @@ export class UrlParser {
     private areUrlProtocolsEqual(urlAObj: URL, urlBObj: URL): boolean {
         return urlAObj.protocol === urlBObj.protocol;
     }
+
+    private tryParseUrl(url?: string): URL | null {
```
waabid, line 52: Not sure if I understand why this parameter or the one below are optional? If you're worried about them being null/undefined, wouldn't it make sense to change the parameter type as such (assuming that's the case, I'd imagine we'd want to change the types on the public methods as well?).

waabid (approved):
LGTM, just a minor comment/question.

src/tests/unit/tests/background/stores/tab-store.test.ts
```diff
@@ -107,27 +121,33 @@ describe('TabStoreTest', () => {
     });
 
     test.each`
-        payloadUrl                               | initialIsOriginChanged | expectedIsOriginChanged
-        ${'https://original-host/original-path'} | ${false}               | ${false}
-        ${'https://original-host/new-path'}      | ${false}               | ${false}
```
lisli1, line 112: Are we removing the test cases for `https://original-host/new-path` here because those are being tested in `url-parser.test.ts` now?

lisli1 (approved):

src/background/stores/tab-store.ts
```diff
@@ -22,6 +23,7 @@ export class TabStore extends PersistentStore<TabStoreData> {
         logger: Logger,
         tabId: number,
         persistStoreData: boolean,
+        private readonly urlParser: UrlParser = new UrlParser(),
```
waabid, line 26: Should this be piped from Background -> TabContextFactory -> TabContextStoreHub -> TabStore instead?

It feels weird to create a UrlParser object for every tab, even though that piping feels cumbersome.

src/tests/unit/tests/background/stores/tab-store.test.ts
dbjorge: Yes - `https://original-host/original-path` and `https://original-host/new-path` are both the same origin as `https://original-host/original-path`, so those two cases collapse into one now that the tab-store logic is reduced to checking "are the two URLs the same origin or not?"

Issue #5757, opened 2022-07-15
a11y: Low contrast icon and links in high contrast message bar


patrickhlauke:
## Describe the bug

When Accessibility Insights is switched to the high contrast theme, there are contrast issues in the message bar: the icon appears white on light yellow, and any links inside the message are saturated yellow on light yellow.

This screenshot was provided by https://twitter.com/TkDodo/status/1547580473743794183?s=20&t=dWI73ySGWxujaTZ5oGxPkg

![Screenshot of the message bar in high contrast, showing a white icon and yellow links](https://user-images.githubusercontent.com/895831/179120463-949dddf3-54f3-4208-a10e-c6319f4198c2.png)

I was personally not able to trigger this specific message (as I think I've got sufficient privileges set for the iframe issue not to occur), but just brutally sticking a link (using devtools) into the message bar gets me to the same visual result:

![screenshot of AI's message bar, with devtools open showing that I added a link. the link appears yellow on light yellow](https://user-images.githubusercontent.com/895831/179120808-9fbb3c75-d940-4897-875d-0485cd2cb81b.png)

**To Reproduce**

1. Open the "Fast Pass" window
2. Set AI to high contrast (Settings > Enable high contrast)
3. Switch to another tab in the browser - the message bar appears "The target page is in a hidden state..."
4. Note the white info icon 
5. Trigger a message bar that also contains a link (it seems with insufficient permissions, a page with an `<iframe>` will do it), or just use devtools to add a link into the above "The target page..." message box, making sure to set the link to `<a href="..." class="insights-link"> ... </a>`

## Expected behavior

The icon and the links should have sufficient contrast (3:1 minimum for the icon, 4.5:1 minimum for the link)

## Context (please complete the following information)

- OS Name & Version: Windows 10.0.19044 Build 19044
- AI-Web Version & Environment: Accessibility Insights for Web - Canary 2022.7.13.1526
- Browser Version: Google Chrome 103.0.5060.114 (Official Build) (64-bit) (cohort: Stable)
- Target Page: any

## Are you willing to submit a PR?

I could give it a try

## Did you search for similar existing issues?

Yes

## Additional context

For the icon, it seems the problem is caused by the use of `color: $secondary-text` in `details-view-body.scss`

```
:global(.ms-MessageBar-icon) i {
    margin-left: 12px;
    color: $secondary-text;
}
```

[...]

PR #5762, opened 2022-07-15
fix: message bar icon in high contrast, and restyle message bar button/link

patrickhlauke:
#### Details

* set icon to use `$always-black` rather than `$secondary-text`, so it doesn't end up white on light yellow in high contrast mode
* restyle button/link in message bar to always use underline and to inherit the regular text color of the message bar - avoids yellow button/link on light yellow in high contrast, but also improves visibility of the link in regular contrast mode

##### Motivation

Closes https://github.com/microsoft/accessibility-insights-web/issues/5757

In regular contrast mode, any button/link inside the message bar is not very obvious - and the difference between the blue button/link and the adjacent static text is relatively low.

More importantly/critically, when set to high contrast, the little "i" info icon turns white on light yellow, and the button/link uses yellow text on light yellow making both completely illegible.

This PR attempts to fix these issues with minimal styling changes.

---

###### Screenshots

**Before:**

Regular contrast, icon and button/link (dark gray/nearly black icon and blue button/link on light yellow bar)

Regular contrast, icon and button/link

High contrast, icon and button/link (white icon and yellow button/link on light yellow message bar)

High contrast, icon and button/link

**After:**

Regular contrast, icon and button/link (black icon and dark gray/black button/link with underline on light yellow bar)

Regular contrast, icon and button/link

High contrast, icon and button/link (black icon and dark gray/black button/link with underline on light yellow bar)

High contrast, icon and button/link

##### Context

Admittedly the change to using dark text with underline for the button/link may be a bit opinionated, and a slight departure from links anywhere else in the extension. I'd argue though that even in those other cases, it's worth considering making at least the underline visible for the links so as not to just rely mostly on colour to distinguish them from static text.

#### Pull request checklist

- [x] Addresses an existing issue: #5757
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/details-view-body.scss
+++ b/src/DetailsView/details-view-body.scss
@@ -54,7 +54,12 @@
 
         :global(.ms-MessageBar-icon) i {
             margin-left: 12px;
-            color: $secondary-text;
+            color: $always-black;
+        }
+
+        :global(.ms-MessageBar-text .insights-link) {
+            color: inherit !important;
+            text-decoration: underline;
         }
 
         .view {
--- a/src/DetailsView/details-view-body.scss
+++ b/src/DetailsView/details-view-body.scss
@@ -55,6 +55,9 @@
         :global(.ms-MessageBar-icon) i {
             margin-left: 12px;
             color: $always-black;
+            @include ms-high-contrast-override {
+                color: inherit;
+            }
         }
 
         :global(.ms-MessageBar-text .insights-link) {
--- a/src/tests/end-to-end/tests/details-view/headings.test.ts
+++ b/src/tests/end-to-end/tests/details-view/headings.test.ts
@@ -36,12 +36,9 @@ describe('Details View -> Assessment -> Headings', () => {
             await headingsPage.closeNavTestLink('Headings');
         });
 
-        it.each([
-            [true, 1],
-            [false, 0],
-        ])(
-            'should pass accessibility validation with highContrastMode=%s, %i errors expected',
-            async (highContrastMode, errorsExpected) => {
+        it.each([true, false])(
+            'should pass accessibility validation with highContrastMode=%s',
+            async highContrastMode => {
                 await browser.setHighContrastMode(highContrastMode);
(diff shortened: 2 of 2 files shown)


src/DetailsView/details-view-body.scss
```diff
@@ -54,7 +54,12 @@
 
         :global(.ms-MessageBar-icon) i {
             margin-left: 12px;
-            color: $secondary-text;
+            color: $always-black;
```
dbjorge, line 57: This is an improvement with Accessibility Insights' in-app high contrast setting, but unfortunately it introduces a new issue with native OS forced color settings. For example, here is what I see with Windows 11's "Aquatic" contrast theme:

screenshot of message bar with black icon on dark yellow background

I'd recommend something like this to resolve it for all cases:

```suggestion
            color: $always-black;
            @include ms-high-contrast-override {
                color: inherit;
            }
```

dbjorge:
The failing end to end tests are legitimate, but good - they are failing because this has fixed a previously-grandfathered-in accessibility error caught by two test cases. I don't want to make you ramp up on our e2e test infrastructure just to make a CSS fix, so I'll go ahead and push an update to those tests to your branch here to stop them from expecting the now-fixed issue.

dbjorge (changes requested):
Thanks for the contribution, @patrickhlauke ! Left one suggestion on the `color` attribute, besides that I think we're good to merge.

Someday it would be nice for us to move the message bar styling out of these global details-view-wide styles and into a component stylesheet, but I don't think we need to block this PR's fixes on that, this is a good improvement as-is.

src/DetailsView/details-view-body.scss
patrickhlauke: ah, good point. didn't test with actual windows high contrast (on Win 10) itself

patrickhlauke:
thanks @dbjorge - merged your suggestion

dbjorge (approved):
Thanks for the contribution!

PR #5763, opened 2022-07-16
feat(mv3-part-5): convert CommandActions to use AsyncActions

pownkel:
#### Details

Convert the action in CommandActions from a SyncAction to an AsyncAction

##### Motivation

Feature work

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/unit/tests/background/global-action-creators/global-action-creator.test.ts
```diff
@@ -12,7 +12,7 @@ import { GlobalActionCreator } from 'background/global-action-creators/global-ac
 import { Interpreter } from 'background/interpreter';
 import { TelemetryEventHandler } from 'background/telemetry/telemetry-event-handler';
 import { Action } from 'common/flux/action';
-import { SyncAction } from 'common/flux/sync-action';
```
sfoslund, line 15: It might not matter in this case, but I notice we aren't changing command-store-test.ts to await the newly async listener. It looks like that will involve changing the store-tester, were you intentionally leaving that for a separate PR or was there another rationale for not updating those tests here?

Issue #5608, opened 2022-06-09
[Usability] Save and Load assessment dialog


ferBonnin:
## Describe the bug
After saving an assessment, users can be confused on how to load it.
Some users might try to load it directly from their downloads folder instead of using the _load assessment_ button

## Expected behavior

After downloading an assessment, show a dialog with a "don't show this again" option that lets user know the assessment can be loaded again using the "load assessment" feature in the command bar.

## Screenshots

dialog that says: Assessment saved
To load this assessment, use the **Load assessment** button in the Accessibility Insights Assessment command bar
_Load assessment_ should be bolded
<img width="336" alt="image" src="https://user-images.githubusercontent.com/9062104/172950472-5c7dbb95-12eb-4ca1-bca0-077caca5c4d5.png">

## Context (please complete the following information)

- OS Name & Version:  N/A
- AI-Web Version & Environment:  Prod 2.32
- Browser Version:  N/A
- Target Page:  N/A

## Are you willing to submit a PR?
No
## Did you search for similar existing issues?
Yes
## Additional context
There is a similar dialog in the Tabstops feature, when possible automated failures are found

ferBonnin:
verified in canary Version 2022.9.20.1622

PR #5770, opened 2022-07-18
fix: Show user an "Assessment saved" dialog after they save the assessment

katydecorah:
#### Details

This pull request will add an Assessment saved dialog after the users saves an assessment. In the dialog, the user can click "Don't show again" to dismiss the dialog. Like the tab stops dialog, they can reset this value from the settings panel.

The image below shows the "Assessment save dialog" with the option to "Don't show [dialog] again" and "Got it" button to dismiss the dialog:

The image below shows the settings panel, where the user can toggle to enable the save assessment dialog:
image

When building this feature, I followed the patterns with the tab stops dialog. I found one small improvement with the tab stops dialog that I made with saved assessment: use the `Stack` component to visually line-up the Do not show again and Got it buttons in the dialog footer. The table below shows screenshots of the current tab stops dialog and proposed:

Current | Proposed
---|---
image | image

##### Motivation

Addresses https://github.com/microsoft/accessibility-insights-web/issues/5608

#### Pull request checklist

- [x] Addresses an existing issue: #5608
- [x] Ran `yarn fastpass`


src/DetailsView/components/save-assessment-button.tsx
```diff
@@ -1,29 +1,86 @@
...
+export const SaveAssessmentButton = NamedFC<SaveAssessmentButtonProps>(
+    'SaveAssessmentButton',
+    props => {
+        const [hideDialog, { toggle: toggleHideDialog }] = useBoolean(true);
+        const [showDialogAgain, { toggle: toggleShowDialogAgain }] = useBoolean(
+            props.userConfigurationStoreData.showSaveAssessmentDialog,
+        );
+
+        function handleSaveAssessmentClick() {
+            props.deps.detailsViewActionMessageCreator.saveAssessment;
```
karanbirsingh, line 33: I think this needs to be `saveAssessment()` at the end; at the moment it downloads because of the `download` attribute, but this function is not being invoked.

src/DetailsView/components/save-assessment-button.tsx
```diff
@@ -1,29 +1,86 @@
...
     deps: SaveAssessmentButtonDeps;
+    userConfigurationStoreData: UserConfigurationStoreData;
 }
 
-export class SaveAssessmentButton extends React.Component<SaveAssessmentButtonProps> {
-    public render(): JSX.Element {
+export const SaveAssessmentButton = NamedFC<SaveAssessmentButtonProps>(
+    'SaveAssessmentButton',
+    props => {
+        const [hideDialog, { toggle: toggleHideDialog }] = useBoolean(true);
```
karanbirsingh, line 27: I haven't seen these fluent-UI hook utility functions before! I understand it now but had to trace it at first. I think(?) this is the first usage of fluent-ui/react-hooks so we might consider a quick ping to interested web-folks to validate that there are no concerns. I'm good either way 🙂

dbjorge:
At our lowest supported width (1280px wide by 400% zoom), the "Don't show again" text reflows kind of awkwardly:

screenshot of checkbox label reflow behavior

Ideally:
* The checkbox label shouldn't be right-aligned
* The checkbox should be vertically centered with the wrapped label text
* There should be some margin in between the label text and the "Got it" button

dbjorge:
The NVDA+Edge behavior for this new dialog feels a bit awkward because the announcement of the dialog gets talked over by the browser's announcement of the download completing; the resulting readout feels to me like it's implying that our dialog controls (the "don't show again" and "got it" controls) are elements of the browser's file download UX, not a page-specific dialog.

@RobGallo, would you mind evaluating the new dialog for usability with your JAWS setup?

dbjorge (changes requested):
Left a few comments on usability/UX of the new dialog, will take a look over the code once we've nailed down the user experience

katydecorah:
I received @RobGallo's evaluation (a few days ago, I was focused elsewhere😅) in which he agrees that the text is confusing, but comprehendible given the constraints that we have.
 

Another I option I explored was the [downloads.onChanged](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/onChanged#examples) event. The component would listen for when the file downloaded and then trigger the modal. When testing with NDVA, the dialog is announced. The implementation could something like:

```jsx
React.useEffect(() => {
    browser.downloads.onChanged.addListener(handleDownloadedAssessment);
    return () => {
        browser.downloads.onChanged.removeListener(handleDownloadedAssessment);
    };
}, []);

function handleDownloadedAssessment(event) {
    if (event.state && event.state.current === 'complete') {
        props.deps.detailsViewActionMessageCreator.saveAssessment(event);
        if (props.userConfigurationStoreData.showSaveAssessmentDialog) {
            showDialog();
        }
    }
}
```


However, this [API is limited to chromium](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/onChanged#browser_compatibility) and requires [manifest.json "downloads" permission](https://developer.chrome.com/docs/extensions/reference/downloads/#manifest). It's not clear to me if either of those are dealbreakers (or if we prefer not to use an event listener). @dbjorge what you think about using the `downloads.onChanged` here?

src/DetailsView/components/details-view-overlay/settings-panel/settings/save-assessment-dialog/save-assessment-dialog-settings.tsx
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dbjorge, line 1: It would be good to add the new file to the strict null checks config. I think you can actually add an include for the whole `"src/DetailsView/components/details-view-overlay/settings-panel/**/*"` directory tree (or all of `details-view-overlay/**/*` with 2 minor fixes to `details-view-overlay.tsx`)

src/tests/unit/tests/DetailsView/components/details-view-overlay/settings-panel/settings/save-assessment-dialog/save-assessment-dialog.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { SaveAssessmentDialogSettings } from 'DetailsView/components/details-view-overlay/settings-panel/settings/save-assessment-dialog/save-assessment-dialog-settings';
+import {
+    SettingsDeps,
+    SettingsProps,
+} from 'DetailsView/components/details-view-overlay/settings-panel/settings/settings-props';
+import { shallow } from 'enzyme';
+import * as React from 'react';
+import { Mock, Times } from 'typemoq';
+
+describe('SaveAssessmentDialogSettings', () => {
```
dbjorge, line 17: Mismatch on the filename vs the class under test?

src/tests/unit/tests/DetailsView/components/details-view-overlay/settings-panel/settings/save-assessment-dialog/save-assessment-dialog.test.tsx
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dbjorge, line 3: Double copyright header

src/tests/unit/tests/background/global-action-creators/user-configuration-action-creator.test.ts
```diff
@@ -213,6 +213,33 @@ describe('UserConfigurationActionCreator', () => {
         );
     });
 
+    it('should SetSaveAssessmentDialogState Message', async () => {
```
dbjorge, line 216: nit: this file's test naming seems off - ideally the test name would summarize the actual assertions the test is making (eg, `should publish telemetry and invoke the corresponding event in response to SetSaveAssessmentDialogState message`)

src/tests/unit/tests/DetailsView/components/should-show-report-export-button.test.ts
```diff
@@ -30,6 +32,20 @@ describe('ShouldShowReportExportButton', () => {
             .setup(m => m.getConfiguration(selectedTest))
             .returns(() => visualizationConfigurationMock.object);
         tabStoreData = { isChanged: false } as TabStoreData;
+
+        userConfigurationData = {
```
dbjorge, line 36: This feels like a lot of implementation detail for this particular test to encode, I don't think we want to expect folks to find and update this stub every time a new user configuration field is added. Can this use a more contained stub?

src/tests/unit/tests/DetailsView/components/save-assessment-button.test.tsx
```diff
@@ -1,47 +1,104 @@
...
+        });
+
+        it('renders when dialog is enabled', () => {
+            expect(wrapper.getElement()).toMatchSnapshot();
+        });
+
+        it('is dismissed when "got it" button is clicked', () => {
+            wrapper.find(PrimaryButton).simulate('click');
+
+            expect(wrapper.getElement()).toMatchSnapshot();
```
dbjorge, line 57: I think it would be preferable to separate the behavioral tests to avoid also testing rendering behavior (ie, have this check for the presence of a `Dialog` child rather than matching a snapshot). If this ever broke in the future, it would be really easy for a reviewer to skim past an important snapshot change.

dbjorge (changes requested):
Thanks for investigating the AT behavior with Rob, Katy! Per our discussion offline, I don't think the issue with the AT reading is enough to justify a new permission on the app, even if it does read a bit awkwardly.

Reviewed the code, left a few suggestions inline.

src/DetailsView/components/details-view-overlay/settings-panel/settings/save-assessment-dialog/save-assessment-dialog-settings.tsx
katydecorah: Thank you! Done in c0a03da14d9e55c3bd79f909942b33c7992e69ac

dbjorge (approved):
LGTM!

PR #5774, opened 2022-07-19
refactor(mv3-part-5): MockInterpreter

pownkel:
#### Details

Replace `createInterpreterMock` with a new `MockInterpreter` class in unit tests

##### Motivation

Currently, most tests for -ActionCreator classes use `createInterpreterMock`, which calls a callback as it is registered in the interpreter with `registerTypeToPayloadCallback`. This works fine for synchronous callbacks, but there is no way to await an async call back with this strategy, since `registerTypeToPayloadCallback` is synchronous. Currently, the few ActionCreators that register async callbacks just use `awaitSettledPromises()` to ensure that the async callback completes.

This PR creates a mock wrapper class that separates callback registration from actually calling the callback, so async callbacks can be awaited without using `awaitSettledPromises`. This also makes the sequence of events clearer in the tests.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/tests/background/global-action-creators/mock-interpreter.ts
+++ b/src/tests/unit/tests/background/global-action-creators/mock-interpreter.ts
@@ -1,0 +1,52 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { Interpreter } from 'background/interpreter';
+import { PayloadCallback } from 'common/message';
+import { IMock, It, Mock, Times } from 'typemoq';
+import { DictionaryStringTo } from 'types/common-types';
+
+export class MockInterpreter {
+    private internalMock: IMock<Interpreter>;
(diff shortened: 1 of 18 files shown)


src/tests/unit/tests/background/global-action-creators/mock-interpreter.ts
```diff
@@ -0,0 +1,52 @@
...
+            .callback((message, handler) => {
+                this.messageCallbacksMap[message] = handler;
+            });
+    }
+
+    public get object(): Interpreter {
+        return this.internalMock.object;
+    }
+
+    public setupRegisterCallbackForMessage(messageType: string): void {
```
waabid, line 28: Not sure if I follow why this method is necessary with the setup on line 17. What case would you want this for?

waabid (approved):
LGTM

PR #5782, opened 2022-07-21
feat(mv3-part-5): Convert AssessmentActions to AsyncActions

pownkel:
#### Details

Convert all SyncActions in AssessmentActions to AsyncActions

##### Motivation

Feature work

##### Context

Some of the messages that AssessmentActionCreator listens for also have other listeners registered in ActionCreator. Messages are interpreted concurrently using both the global context and the tab context, and since some of the messages affected by this change have listeners in both contexts, some of the now-async actions may be invoked concurrently. This shouldn't be a problem, except that the current Action classes have built-in measures to prevent two actions from running at the same time in the same "scope." To avoid spurious exceptions, this PR adds a "scope" parameter to Action.invoke() calls inside listeners that may run concurrently with other listeners (i.e. listeners for messages that have multiple listeners registered).

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/actions/assessment-action-creator.ts
+++ b/src/background/actions/assessment-action-creator.ts
@@ -36,12 +36,8 @@ const AssessmentMessages = Messages.Assessment;
 
 export class AssessmentActionCreator {
     // This is to be used as the scope parameter to invoke().
-    // If a message has multiple callbacks registered, all invoke() calls
-    // inside those callbacks must pass a scope parameter. Those message
-    // callbacks will run concurrently, and our Flux classes don't allow
-    // multiple invoke() calls to run in the same scope at the same time.
-    // Passing our own scope will allow multiple actions to be invoked
-    // concurrently as long as there are no infinite loops.
+    // Some callbacks in this class are registered to messages with
+    // multiple callbacks (see the comment in src/common/flux/scope-mutex.ts)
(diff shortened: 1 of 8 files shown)


src/background/actions/assessment-action-creator.ts
```diff
@@ -152,137 +154,159 @@ export class AssessmentActionCreator {
...
     };
 
-    private onChangeVisualizationStateForAll = (payload: ChangeInstanceSelectionPayload): void => {
+    private onChangeVisualizationStateForAll = async (
+        payload: ChangeInstanceSelectionPayload,
+    ): Promise<void> => {
         const eventName = TelemetryEvents.CHANGE_ASSESSMENT_VISUALIZATION_STATUS_FOR_ALL;
         this.telemetryEventHandler.publishTelemetry(eventName, payload);
-        this.assessmentActions.changeAssessmentVisualizationStateForAll.invoke(payload);
+        await this.assessmentActions.changeAssessmentVisualizationStateForAll.invoke(payload);
```
sfoslund, line 265: Why are some of the invoke calls using the executing scope and some aren't? Can we either standardize across the file or comment what the difference is?

sfoslund (approved):
LGTM, thanks!

PR #5786, opened 2022-07-22
chore: remove usages of GlobalMock to unblock swc upgrade

dbjorge:
#### Details

This PR updates/rewrites all tests that use typemoq's `GlobalMock`/`GlobalScope` to instead avoid using global mocks.

* For most tests of components that live at the boundary of how we interact with `axe-core` (`axe-utils`, all custom rules), I felt that the boundary tests were more valuable as integration tests with `axe-core` anyway, so I updated them to use actual JSDOM and axe-core behavior instead of mocking things like `axe.utils.isVisible` and `window.getComputedStyle`
  * The `css-content rule` tests are an exception and do still use a mock `window.getComputedStyle` because they rely on retrieving styles of pseudo-elements, which JSDom does not support (see https://github.com/jsdom/jsdom/issues/1928). For this case, I updated the mocking method to just override `window.getComputedStyle` directly instead of using `GlobalMock`.
* For the rest of the cases (`DialogRenderer`, `getElectronIconPath`), this involved updates to inject a previously-static dependency

Verified that this PR + `yarn upgrade @swc/core` to 1.2.218 unblocks all previously-failing tests against the current swc version.

##### Motivation

The primary motivation is to unblock `@swc/core` updates, which are currently blocked for versions >= 1.2.206 due to a variation of the issue being dicsussed in https://github.com/swc-project/swc/discussions/5151. The workaround suggested in [this comment](https://github.com/swc-project/swc/discussions/5151#discussioncomment-3149154) doesn't work for our case (I suspect because we're hitting it via typemoq's `GlobalMock`, which works slightly differently than the `jest.mock` motivating that discussion).

Secondary motivation is that I think this makes the tests more valuable overall; for components that live at the boundary of our interaction with external dependencies, I think it's generally more valuable to do integration tests than mock-heavy unit tests where possible.

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/tests/unit/tests/scanner/axe-utils.test.ts
```diff
@@ -324,132 +323,78 @@ describe('AxeUtils', () => {
...
 
         afterEach(() => {
             document.body.querySelector('#test-fixture').remove();
         });
 
-        it('has bg image', () => {
+        it('returns true with background: some-value', () => {
             fixture.innerHTML = `
-                <img id="el1" alt="" />
+                <img id="el1" alt="" style="background: \\"some-value\\"" />
```
katydecorah, line 337: I like that by removing the windowMock, it reduces the abstraction and improves the readability here.

katydecorah (commented):
Looks good to me. I left one inline comment suggestion to remove a commented-out import.

src/tests/unit/tests/scanner/axe-utils.test.ts
dbjorge: I agree, I think most of the tests ended up more readable overall.

katydecorah (approved):
(I meant to approve before!)

PR #5790, opened 2022-07-22
feat(mv3-part5): tabActions to use AsyncAction

waabid:
#### Details

Title covers it.

##### Motivation

Feature work.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/background/actions/tab-action-creator.ts
```diff
@@ -45,8 +47,8 @@ export class TabActionCreator {
         );
         this.interpreter.registerTypeToPayloadCallback(
             Messages.Tab.VisibilityChange,
-            (payload: PageVisibilityChangeTabPayload) =>
-                this.tabActions.tabVisibilityChange.invoke(payload.hidden),
+            async (payload: PageVisibilityChangeTabPayload) =>
+                await this.tabActions.tabVisibilityChange.invoke(payload.hidden),
```
sfoslund, line 51: I think we need to use scopes in this file too, when I run this branch I get an error in the background: 
```
Error while processing browser TabsOnActivated event:  Error: Cannot invoke an action with scope DEFAULT_SCOPE from inside another action with the same scope
    at _ScopeMutex.tryLockScope 
```

sfoslund (commented):
LGTM other than the scope error

src/background/actions/tab-action-creator.ts
pownkel: I was looking at this to figure out why this error would happen, because this is the only listener registered for this message. I think the culprit is [TargetPageController](https://github.com/microsoft/accessibility-insights-web/blob/main/src/background/target-page-controller.ts#L74), which invokes this action for several tabs concurrently and waits with Promise.all. Since those are all calling the same action listener, just adding a scope string won't fix the issue unless each TargetPageController has its own unique scope string depending on the tab context it serves (Or we could just remove the Promise.all in TargetPageController and await each promise sequentially)

waabid: > await each promise sequentially

The promise.all seems redundant as the loop will ensure we await each promise anyways. I'll update and remove it.

waabid: I've updated the PR and am no longer seeing the error. Please do the same locally since I'm getting some other transient issues and wanna make sure I did actually get rid of it 😄

sfoslund: Just tested locally and the error is gone for me!

src/background/target-page-controller.ts
```diff
@@ -71,13 +71,11 @@ export class TargetPageController {
...
 
         const tabs = await this.browserAdapter.tabsQuery({ windowId });
-        await Promise.all(
-            tabs.map(async tab => {
-                if (!tab.active) {
-                    await this.sendTabVisibilityChangeAction(tab.id, true);
-                }
-            }),
-        );
+        tabs.map(async tab => {
```
sfoslund, line 74: nit: now that we don't have the promise.all, do we want to use a foreach loop here instead of map for consistency?

sfoslund (approved):

sfoslund (approved):
LGTM!

PR #5792, opened 2022-07-25
chore: Enable react/jsx-key

katydecorah:
#### Details

When running `yarn test` I noticed a key prop error: `Warning: Each child in a list should have a unique "key" prop.` This pull request will enable (set to `error`) and make the required fixes.

##### Motivation

Reduce `console.errors` in test output and make sure [React items have a stable identity](https://reactjs.org/docs/lists-and-keys.html#keys).

##### Context

In looking back at the history for [react/jsx-key](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-key.md), I believe it was disabled when switching from tslint to eslint https://github.com/microsoft/accessibility-insights-web/pull/3273 as a means to unblock the migration. Please correct me if I'm wrong :)

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/eslintrc.base.js
+++ b/eslintrc.base.js
@@ -71,7 +71,7 @@ module.exports = {
         'react/prop-types': 'off',
         'react/display-name': 'off',
         'react/no-unescaped-entities': 'off',
-        'react/jsx-key': 'off',
+        'react/jsx-key': 'error',
         'react/no-access-state-in-setstate': 'error',
         'react/no-unused-state': 'error',
         'react/no-direct-mutation-state': 'off',
--- a/eslintrc.base.js
+++ b/eslintrc.base.js
@@ -42,6 +42,8 @@ module.exports = {
             },
         ],
         'no-throw-literal': 'error',
+        'react/no-access-state-in-setstate': 'error',
+        'react/no-unused-state': 'error',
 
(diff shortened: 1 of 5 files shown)


eslintrc.base.js
```diff
@@ -71,7 +71,7 @@ module.exports = {
         'react/prop-types': 'off',
         'react/display-name': 'off',
         'react/no-unescaped-entities': 'off',
-        'react/jsx-key': 'off',
+        'react/jsx-key': 'error',
```
dbjorge, line 74: Having entries in this code block that aren't disabling rules feels inconsistent with the comment at the top of the block. I would recommend instead deleting this line (since `react/jsx-key` is already an error by default in the `plugin:react/recommended` set we extend) and moving the other two `error` cases below it up to the "additional rules not enabled by default" section

katydecorah: Thanks for pointing those headings out, I skipped over them. Fixed in [560b41d](https://github.com/microsoft/accessibility-insights-web/pull/5792/commits/560b41d17fb8cd8e72e98bad3613078bfe998989)!

dbjorge (approved):

Issue #5735, opened 2022-07-12
Visualizations fail for targets with semicolon character


sfoslund:
## Describe the bug

Automated checks visualizations fail on web pages that include the character `;`. 

**To Reproduce**

1. Go to code pen: https://codepen.io/dbjorge/pen/OJvXaVg
2. Run fast pass automated checks
4. See message `unable to find frame to highlight button[data-attr="with ` in target page and no visualization. 

## Expected behavior

Visualization should be shown on pages with `;` character in targets.

## Context (please complete the following information)

- OS Name & Version: Windows 11
- AI-Web Version & Environment: 2.32.0 prod
- Browser Version: Chrome 103.0.5060.114
- Target Page: https://codepen.io/dbjorge/pen/OJvXaVg

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

Yes

ferBonnin:
I can't see the error message, however, the button in the codepen is not failing the Button-name rule also, changing the attr of the button to something without a semicolon makes it such that the frame-title rule fails whether as with the semicolon, no failures are found.

sfoslund:
> the button in the codepen is not failing the Button-name rule

@ferBonnin what build/ version are you reproing that with? I'm seeing the button failing the button-name rule as expected in the current prod.

> changing the attr of the button to something without a semicolon makes it such that the frame-title rule fails whether as with the semicolon, no failures are found.

I think this behavior is related to updating the codepen, not to changing the content of the code pen in this specific case. I'm seeing those frame-title failures regardless of how I change the code pen content both in this example and with other code pens.

ferBonnin:
investigated this with @sfoslund. We found it was an issue with the codepen and couldn't repro when testing html directly.

PR #5843, opened 2022-07-29
fix: Fix visualizations for elements that include semicolons/commas

sfoslund:
#### Details

Fixes visualizations for elements that include special characters in their selectors. This addresses issue #5735 as well as an instance @pownkel was running into in which automated checks visualizations were not showing up on some github PR web pages (for example, https://github.com/microsoft/accessibility-insights-web/pull/5745)

##### Motivation

Addresses issue #5735

##### Context

https://github.com/microsoft/accessibility-insights-web/pull/5728 allowed us to preserve the target in the state that we receive it from axe throughout our logic instead of serializing it into a string as we used to do. This PR removes one more instance of stringifying then converting the target string back into an array. We want to get rid of this logic because it doesn't work with elements that have special characters in their selectors, including semicolons and commas. 

We can't completely remove the `getTargetFromSelector` logic (which does the conversion from a target string to a target array) because of some unified logic that does not have the target array, but these changes ensure that `getTargetFromSelector` is not used in the web extension. 

#### Pull request checklist

- [x] Addresses an existing issue: #5735
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/types/store-data/unified-data-interface.ts
+++ b/src/common/types/store-data/unified-data-interface.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { BoundingRectangle } from 'electron/platform/android/android-scan-results';
+import { Target } from 'scanner/iruleresults';
 import { DictionaryStringTo } from 'types/common-types';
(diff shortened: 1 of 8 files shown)


pownkel (approved):
LGTM

PR #5844, opened 2022-07-29
chore: Update mac environment from 10.15 to 11

DaveTryon:
#### Details

MacOS 10.15 is being phased out as a virtual environment. We're getting the following warnings in our builds:
```
The macOS-10.15 environment is deprecated, consider switching to macos-11(macos-latest), macos-12 instead. For more details see https://github.com/actions/virtual-environments/issues/5583
```

This PR simply changes all instances of `macOS-10.15` to `macOS-11`.

##### Context

We could use `macOS-latest` to avoid having to touch these files in the future, but deprecations are infrequent and having an explicit version makes it easier to identify cases where tests potentially fail after changing OS versions.

We could use `macOS-12`, but that would leave downlevel clients with no testing in our pipeline. Since most of our internal testing will be on macOS 12 (required for security reasons), using `macOS-11` is a cheap and easy way to get version coverage.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -13,7 +13,7 @@ trigger:
 
 variables:
     windowsImage: 'windows-2019'
-    macImage: 'macOS-10.15'
+    macImage: 'macOS-11'
     linuxImage: 'ubuntu-18.04'
 
 jobs:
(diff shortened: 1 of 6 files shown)


waabid (approved):
LGTM

PR #5852, opened 2022-07-30
fix: Create reflowable card footer buttons

pownkel:
#### Details

Extracts out some of the work in #5745, plus additional refactors:
- Add a new narrowModeThreshold for reflowing card footers
- Create a new component, CardFooterFarButtons, a reflowable and refactored version of CardKebabMenuButton

The gifs below show the behavior or CardFooterFarButtons when it is resized, but this PR does **_not_** use CardFooterFarButtons anywhere in the UI. Replacing CardKebabMenuButton and passing NarrowModeStatus through to the cards will be done in a separate PR.

card footer buttons reflowing as window is resized

CardFooterFarButtons also makes sure that focus correctly returns to the "..." button or the "file issue" button when the issue filing settings dialog is closed:
demo of focus handling when issue filing settings is opened and closed

##### Motivation

Partially addresses #5609

##### Context

Deferred to future PRs:
- Further refactor CardFooterFarButtons to use store state instead of UI state to determine whether the issue filing dialog is open (this will allow renderIssueFilingSettingContent() to be extracted to a different file)
- Replace CardKebabMenuButton with CardFooterFarButtons in the cards UI
- Pass NarrowModeStatus through to the cards components

(copied from #5745) Also worth mentioning is that this PR uses the width of the entire screen to determine if cards should reflow, rather than the width of the cards themselves. This is not ideal, especially in electron where both the left nav and the right-hand screenshot also resize as the screen resizes, and the card width is affected by more than just screen width. It doesn't make a huge difference from a user perspective, but in the future, we may want to consider refactoring NarrowModeDetector to be more generic (it currently assumes that it's measuring the width of the whole screen and sets its thresholds accordingly) and adding another NarrowModeDetector to the element containing the cards.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/issue-filing-dialog.tsx
+++ b/src/DetailsView/components/issue-filing-dialog.tsx
@@ -27,6 +27,7 @@ export interface IssueFilingDialogProps {
     selectedIssueData: CreateIssueDetailsTextData;
     issueFilingServicePropertiesMap: IssueFilingServicePropertiesMap;
     onClose: (ev?: React.SyntheticEvent) => void;
+    afterClosed?: () => void;
 }
 
 export type IssueFilingDialogDeps = {
@@ -60,7 +61,7 @@ export class IssueFilingDialog extends React.Component<
     }
 
     public render(): JSX.Element {
-        const { onClose, isOpen, deps } = this.props;
+        const { onClose, afterClosed, isOpen, deps } = this.props;
         const { selectedIssueFilingService } = this.state;
         const selectedIssueFilingServiceData =
             this.state.selectedIssueFilingService.getSettingsFromStoreData(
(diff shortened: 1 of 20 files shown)


madalynrose (changes requested):
This looks awesome! What a great improvement!

The only thing I'm hesitant about is the name `CardFooterFarButtons`. What is the intent of the word `Far` here? I think we can come up with a more illustrative name.

Ideas:
* `CardFooterMenuItems` keeps naming aligned with `CardFooterMenuItemsBuilder` and brings to mind that this contains items that might be presented as a menu.
* `CardFooterInstanceActionButtons` captures that these are the actions a user can take on the failure instance (and also these are `ActionButton`s)

pownkel:
Thanks for the suggestion, @madalynrose! I changed the class name to `CardFooterInstanceActionButtons` to avoid confusion with the `CardFooterMenuItem` type.

madalynrose (approved):
Awesome! 🚀

PR #5877, opened 2022-08-03
feat(mv3-part-5): Allow event handler list and store to support async and sync listeners

sfoslund:
#### Details

Allow event handler list and store to support sync or async callbacks. This PR does not change specific stores but it enables us to convert specific stores to accept async callbacks in the future. 

##### Motivation

Feature work

##### Context

We were originally planning on making all stores use async callbacks as part of the mv3 effort, but this would mean having to make a lot of unnecessary changes to both web and unified code. Instead, we're planning to switch to this approach, which will allow us to only add async callback support to stores whose callbacks actually do async work. This will result in fewer unnecessarily async actions as well as fewer changes (and no functional changes) to unified code. 

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/flux/event-handler-list.ts
+++ b/src/common/flux/event-handler-list.ts
@@ -1,25 +1,43 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { mergePromiseResponses } from 'common/merge-promise-responses';
+import { isPromise } from 'common/promises/is-promise';
 import { FunctionPPR } from '../../types/common-types';
 
-export class EventHandlerList<TSender, TEventArgs> {
-    private handlers: FunctionPPR<TSender, TEventArgs, any>[] = [];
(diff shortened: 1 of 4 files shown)


src/common/flux/event-handler-list.ts
```diff
@@ -1,25 +1,41 @@
...
+    public invokeHandlers(sender: TSender, eventArgs?: TEventArgs): TReturn {
         const handlersCopy = [...this.handlers];
 
-        handlersCopy.forEach(handler => handler(sender, eventArgs));
+        const results = handlersCopy.map(handler => handler(sender, eventArgs));
+
+        const asyncResults = results.filter(isPromise);
+
+        if (asyncResults.length === 1) {
+            return asyncResults[0] as TReturn;
```
waabid, line 32: nit: this cast is not necessary.

waabid (approved):
LGTM. Minor nit around casting and title discussed offline.

PR #5884, opened 2022-08-06
feat(mv3-part6): make injector controller explicitly async

waabid:
#### Details

Makes inject async.
##### Motivation

Feature work.
##### Context

This method is async but not explicitly so. Changing it now such that we can make it clear moving forward which stores need to be async and which do not. Did some smoke testing in both mv3 and mv2 version of the extension; no errors were found.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/injector-controller.ts
+++ b/src/background/injector-controller.ts
@@ -10,8 +10,6 @@ import { TabStore } from './stores/tab-store';
 import { VisualizationStore } from './stores/visualization-store';
 
 export class InjectorController {
-    private static readonly injectionStartedWaitTime = 10;
-
     private oldInspectType = InspectMode.off;
 
     constructor(
@@ -20,7 +18,6 @@ export class InjectorController {
         private readonly interpreter: Interpreter,
         private readonly tabStore: TabStore,
         private readonly inspectStore: InspectStore,
-        private readonly setTimeout: (handler: Function, timeout: number) => number,
         private readonly logger: Logger,
     ) {}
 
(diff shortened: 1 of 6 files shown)


src/background/injector-controller.ts
```diff
@@ -42,22 +39,18 @@ export class InjectorController {
             inspectStoreInjectingRequested || visualizationStoreState.injectingRequested;
 
         if (isInjectingRequested && !visualizationStoreState.injectingStarted) {
-            this.setTimeout(() => {
-                this.interpreter.interpret({
-                    messageType: Messages.Visualizations.State.InjectionStarted,
-                    tabId: tabId,
-                });
-            }, InjectorController.injectionStartedWaitTime);
+            this.interpreter.interpret({
```
pownkel, line 42: Should this call be awaited?

src/background/injector-controller.ts
```diff
@@ -42,22 +39,18 @@ export class InjectorController {
...
-                    const response = this.interpreter.interpret({
+                    await this.interpreter.interpret({
                         messageType: Messages.Visualizations.State.InjectionCompleted,
                         tabId: tabId,
-                    });
-
-                    await response.result;
+                    }).result;
                 })
                 .catch(this.logger.error);
```
pownkel, line 55: I don't think .then() and .catch() actually wait for the promise in question, could this be a try/catch block instead?

src/background/injector-controller.ts
sfoslund, line 42: Why do we no longer want to set a timeout here?

waabid: This was done a long while back to avoid the store "same scope" error that could occur. This can no longer occur with the changes we've already made.

pownkel (approved):

PR #5894, opened 2022-08-09
feat(mv3-part6): Revert user config store actions to be sync

sfoslund:
#### Details

The changes in https://github.com/microsoft/accessibility-insights-web/pull/5877 allow us keep any actions that don't have any async work in any listeners to remain synchronous. As a result, we want to revert actions that act on the user configuration store to be sync, since that store does not have any async work.

##### Motivation

Feature work

##### Context

We considered/ were planning on converting all actions to be async (even if they had no async work) but decided that this caused too many unnecessary changes (particularly to the electron code) so went with this solution instead. 

#### Pull request checklist

- [x] Ran `yarn fastpass`


pownkel (approved):
LGTM

PR #5895, opened 2022-08-09
feat(mv3-part6): Opt into async handler for stores with async work

sfoslund:
#### Details

Follow up to https://github.com/microsoft/accessibility-insights-web/pull/5877, which allowed us to opt stores into supporting async handlers. These changes to not touch any stores that do not have async work, since by default stores support sync handlers. The stores changed in this file correspond to actions that were converted to async in previous PRs. 

##### Motivation

Feature work.

##### Context

These changes ensure that any async work in action handlers is properly awaited, which will protect from the manifest v3 service worker timing out when async work is still happening.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/background/completed-test-step-telemetry-creator.ts
```diff
@@ -40,7 +40,7 @@ export class CompletedTestStepTelemetryCreator {
         this.store.addChangedListener(this.onAssessmentChange);
     }
 
-    private onAssessmentChange = (): void => {
+    private onAssessmentChange = async (): Promise<void> => {
```
pownkel, line 43: Why does this method signature need to change? It doesn't look like we call anything async here

src/background/stores/base-store-impl.ts
```diff
@@ -1,10 +1,14 @@
...
 // Licensed under the MIT License.
+import { HandlerReturnType } from 'common/flux/event-handler-list';
 import { BaseStore } from '../../common/base-store';
 import { Store } from '../../common/flux/store';
 import { StoreNames } from '../../common/stores/store-names';
 
-export abstract class BaseStoreImpl<TState> extends Store implements BaseStore<TState> {
+export abstract class BaseStoreImpl<TState, TReturn extends HandlerReturnType = void>
+    extends Store<TReturn>
+    implements BaseStore<TState>
```
pownkel, line 10: Could `TReturn` also be added to the BaseStore interface? I noticed that BaseStore's `addChangedListener` takes a `Function` with unspecified return type, and I wonder if that's why async store listeners are getting silently cast to () => void

src/background/completed-test-step-telemetry-creator.ts
sfoslund: This is added as a listener on the assessment store, which after this change only accepts async listeners.

src/background/stores/base-store-impl.ts
sfoslund: Yeah, we can definitely do that! It looks like it will be a lot of small changes though (We'll have to change `BaseStore<Type>` to `BaseStore<Type, HanlderReturnType>` in quite a lot of places) so I would prefer to do it as a separate PR

src/electron/platform/android/setup/android-setup-start-listener.ts
```diff
@@ -11,14 +11,14 @@ export class AndroidSetupStartListener {
...
+    public async initialize(): Promise<void> {
         this.userConfigStore.addChangedListener(this.storeChanged);
         this.androidSetupStore.addChangedListener(this.storeChanged);
 
-        this.storeChanged();
+        await this.storeChanged();
     }
 
-    private storeChanged = () => {
+    private storeChanged = async () => {
```
pownkel, line 21: Instead of making this async, could it just be wrapped in an async arrow function where it needs to be passed as an async function? It looks like it's registered as a sync callback for AndroidSetupStore, so the promise is probably being floated there

pownkel: That change would also let initialize() stay as sync, so this PR wouldn't touch as many electron files

src/background/telemetry/telemetry-state-listener.ts
```diff
@@ -9,12 +9,12 @@ export class TelemetryStateListener {
...
 
-    public initialize(): void {
+    public async initialize(): Promise<void> {
         this.userConfigStore.addChangedListener(this.onStateChanged);
-        this.onStateChanged();
+        await this.onStateChanged();
     }
 
-    private onStateChanged = (): void => {
+    private onStateChanged = async (): Promise<void> => {
```
pownkel, line 17: Same comment here, if `onStateChanged` is sync then `initialize` can also stay sync

pownkel (approved):

PR #5918, opened 2022-08-15
chore(null): Add CI task to check for new files to add to tsconfig.strictNullChecks.json

madalynrose:
#### Details

Add an additional task to the `lints` CI job to run `yarn null:autoadd` and determine if there are new files to be added to `tsconfig.strictNullChecks.json` for the PR. If it fails, it will display an error message reminding the author to run `yarn null:autoadd` and commit the results.
 
##### Motivation

This will help us work towards #2869 as we add/change files. As things are now, if someone wants to focus on adding some files to `tsconfig.strictNullChecks.json` in a dedicated PR, they often find a whole host of additional files that could have been added earlier. This PR ensures we expand our strict null checks whenever there are new files to include.

##### Context

Another approach we identified for keeping `tsconfig.strictNullChecks.json` up to date involves flipping `tsconfig.strictNullChecks.json` to explicitly `exclude` files instead of using `include`, thus automatically including new files. This would require adjusting all of our strict null scripts (e.g., autoadd, find, etc.) to follow that convention, which is not an insignificant amount of work.

#### Pull request checklist
- [x] Ran `yarn fastpass`


.github/workflows/verify-unchanged-strict-null-checks.js
```diff
@@ -0,0 +1,41 @@
...
+
+if (isLockfileChanged) {
+    const gitDiffResult = child_process.execFileSync('git', ['diff', '--', lockfilePath]);
+    console.error(`
+Error: Running "yarn null:autoadd" identified new files to be included in null checks.
+
+To ensure these files are included:
+    1) Pull this branch
+    2) Run "yarn null:autoadd"
+    3) Commit and push the resulting change to tsconfig.strictNullChecks.json
```
sfoslund, line 29: I'm confused by this logic- the comment at the top of the file says that we're verifying that there are NO changes to tsconfig.strictNullChecks.json, but this error message suggests that we DO want to update that file. What is the actual expected outcome?

madalynrose: So this is running `yarn null:autoadd` in CI. CI will pass if it finds no changes and will fail if it finds changes, prompting the author to run it and add those changes manually.

sfoslund: Ah, okay, makes sense to me!

sfoslund (approved):

PR #5953, opened 2022-08-22
fix(mv3-part6): Clear old visualizations before rerunning axe scan

sfoslund:
#### Details

Addressing color-contrast e2e test flakiness as well as color contrast bug. Repro steps for bug before this change:

1. Open poor-color-contrast.html test resource
2. Run fastpass, see 1 color contrast requirement failed
3. Close details view (do not refresh target page)
4. Re-open details view and re-run fastpass
5. See no failures where there should be 1 color contrast requirement

##### Motivation

Bug fix

##### Context

@pownkel noticed that this bug was [introduced when updating to axe-core 4.4.1](https://github.com/microsoft/accessibility-insights-web/commit/79ff4e19e6a50daa49b60bf3879cbc17fead9267), so we suspect that it has something to do with the color-contrast/ bgOverlap changes they made in that release. It appears that when we run fastpass on the page multiple times we get different axe results because axe thinks our visualizations are overlapping with the page elements and it gets confused. This PR works around that issue by explicitly clearing our visualizations before re-running axe on the target page.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/analyzer-controller.ts
+++ b/src/injected/analyzer-controller.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { AssessmentsProvider } from 'assessments/types/assessments-provider';
+import { ShadowInitializer } from 'injected/shadow-initializer';
 import { BaseStore } from '../common/base-store';
 import { VisualizationConfigurationFactory } from '../common/configs/visualization-configuration-factory';
 import { EnumHelper } from '../common/enum-helper';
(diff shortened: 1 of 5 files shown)


src/injected/analyzer-controller.ts
```diff
@@ -65,8 +67,12 @@ export class AnalyzerController {
     };
 
     protected startScan = (id: string): void => {
+        this.shadowInitializer.removeExistingShadowHost();
```
pownkel, line 70: It looks like this is also called as part of `shadowInitializer.initialize()`, is there a reason it needs to be called before `analyzer.analyze()` (or for `initialize()` to be called after)?

sfoslund: Yeah, analyze actually runs axe-core, so this way we're clearing the visualizations beforehand so that they don't interfere with our axe results, then after we've gotten the results we can safely re-init visualizations

pownkel (approved):

PR #5957, opened 2022-08-23
fix(mv3-part6): Register exception telemetry listener event handlers synchronously

sfoslund:
#### Details

Right now the way we register the exception telemetry listener in the mv3 extension causes the following warnings:

> exception-telemetry-listener.ts:30 Event handler of 'error' event must be added on the initial evaluation of worker script.
initialize @ exception-telemetry-listener.ts:30
exception-telemetry-listener.ts:48 Event handler of 'unhandledrejection' event must be added on the initial evaluation of worker script.

This is a new warning in mv3 (does not repro with mv2) and is caused because the exception telemetry listener is [initialized in an async function](https://github.com/microsoft/accessibility-insights-web/blob/main/src/background/service-worker-init.ts#L124) and the [initialization adds global event listeners](https://github.com/microsoft/accessibility-insights-web/blob/main/src/common/telemetry/exception-telemetry-listener.ts#L48). The solution is to initialize the exception telemetry listener synchronously on service worker startup.

##### Motivation

Fix warning in mv3 extension.

##### Context

This warning seems to be a false alarm in this case and these changes should not have any functional effect besides removing the warning.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/service-worker-init.ts
+++ b/src/background/service-worker-init.ts
@@ -30,15 +30,17 @@ import { AxeInfo } from 'common/axe-info';
 import { BackgroundBrowserEventManager } from 'common/browser-adapters/background-browser-event-manager';
 import { BrowserAdapterFactory } from 'common/browser-adapters/browser-adapter-factory';
 import { EventResponseFactory } from 'common/browser-adapters/event-response-factory';
+import { WebExtensionBrowserAdapter } from 'common/browser-adapters/webextension-browser-adapter';
 import { WebVisualizationConfigurationFactory } from 'common/configs/web-visualization-configuration-factory';
 import { DateProvider } from 'common/date-provider';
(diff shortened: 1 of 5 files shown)


src/background/service-worker-init.ts
```diff
@@ -254,6 +266,7 @@ async function initialize(): Promise<void> {
     globalThis.insightsUserConfiguration = globalContext.userConfigurationController;
 }
 
-initialize()
+initializeSync();
+initializeAsync()
     .then(() => console.log('Background initialization completed successfully'))
     .catch((e: Error) => console.error('Background initialization failed: ', e));
```
pownkel, line 272: Are there any exceptions in initializeSync that we might need to catch?

pownkel (approved):
Left one question, other than that LGTM

PR #5974, opened 2022-08-30
feat(mv3-part7): fix issue where popup closes before dispatching message

waabid:
#### Details

Creates a new API on remote action message dispatcher to expose the async nature of dispatch message.

##### Motivation

Fixes an issue/race-condition where the popup may close before a message is dispatched, leading to that action never being acted on (i.e. opening FastPass from the launch pad).

##### Context

A general summary since there are multiple possible approaches for this fix:
- Make dispatchMessage async, which it really is, but is obfuscated (i.e. doesn't cause lint errors) due to the catch statement. This is a high-cost solution as many dependent APIs would then need to become async as well. Technically, this is "correct" but requires lots of changes.
- Make window.close async (via wrapper). This will also work because it will be added to queue after the dispatch async call. But it's unintuitive and technically there are no guarantees for the caller. This is low-cost but hacky.
- Make a new API where we expose the async nature and use async/await here. This is the chosen option because it's clear what is happening but doesn't require massive amounts of change.
- Remove window.close from popup-action-message-creator and trigger it from background which can only happen through dispatchMessage. This approach would work but requires a lot of piping for something that shouldn't have piping.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/popup/actions/popup-action-message-creator.ts
+++ b/src/popup/actions/popup-action-message-creator.ts
@@ -9,9 +9,8 @@ import { ActionMessageDispatcher } from 'common/message-creators/types/dispatche
 import { Tab } from 'common/types/store-data/itab';
 import { LaunchPanelType } from 'common/types/store-data/launch-panel-store-data';
 import * as React from 'react';
-
-import { TelemetryEventSource } from '../../common/extension-telemetry-events';
 import * as TelemetryEvents from '../../common/extension-telemetry-events';
+import { TelemetryEventSource } from '../../common/extension-telemetry-events';
 import { Messages } from '../../common/messages';
 import { SupportedMouseEvent, TelemetryDataFactory } from '../../common/telemetry-data-factory';
 import { DetailsViewPivotType } from '../../common/types/store-data/details-view-pivot-type';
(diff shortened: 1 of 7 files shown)


src/popup/actions/popup-action-message-creator.ts
```diff
@@ -57,19 +56,19 @@ export class PopupActionMessageCreator {
         });
     }
 
-    public openDetailsView(
+    public async openDetailsView(
```
sfoslund, line 59: Do you have an understanding of why this message creator needs to be async but it's (presumably) safe to leave the other ones as sync? It feels like dispatchMessage being sync might end up causing a lot of hard to diagnose bugs if we leave it as is.

sfoslund (commented):
One question, and it looks like there are some failing tests.

waabid:
> One question, and it looks like there are some failing tests.

Looked into the errors; seemed transient and couldn't repro on my machine. Seem to be green now.

src/popup/actions/popup-action-message-creator.ts
waabid: I'm not sure if leaving `dispatchMessage` as is would cause a lot of issues: dispatchMessage is where we cross boundaries between popup/details-view/etc. to background so the async work is really just making sure the message is sent and after that, nothing else is guaranteed (the rest of which is async as well).

In this particular case, the reason we run into an issue is that even just the dispatch of the message is not completed in the action-message-creator before we kill the popup, resulting in it never being completed.

sfoslund: Makes sense. Since there are so many references to dispatchMessage I assume it might not be a good use of time to audit all of them to check that there isn't another situation like this... but might be good to keep in mind going forward.

sfoslund (approved):

PR #5975, opened 2022-08-30
feat(mv3-part7): Convert mv3 extension to use build generated JSON validator

sfoslund:
#### Details

Fixes mv3 bug with loading assessments. Repro steps:

1. Open mv3 canary/ dev assessment
2. Try to load assessment file
3. See nothing happens in details view, see following errors in console:
```
exception-telemetry-listener.ts:87 Error compiling schema, function code: const schema2 = scope.schema[2];const schema1 = scope.schema[1];return function validate1(data, {instancePath="", parentData, parentDataProperty, rootData=data}={}){let vErrors = null;let errors = 0;const _errs1 = errors;if(!(((typeof data == "number") && (!(data % 1) && !isNaN(data))) && (isFinite(data)))){validate1.errors = [{instancePath,schemaPath:"#/definitions/nonNegativeInteger/type",keyword:"type",params:{type: "integer"},message:"must be integer"}];return false;}if(errors === _errs1){if((typeof data == "number") && (isFinite(data))){if(data < 0 || isNaN(data)){validate1.errors = [{instancePath,schemaPath:"#/definitions/nonNegativeInteger/minimum",keyword:"minimum",params:{comparison: ">=", limit: 0},message:"must be >= 0"}];return false;}}}validate1.errors = vErrors;return errors === 0;}
```
```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".
```

##### Motivation

Feature work

##### Context

The root cause of this bug is that AJV, the library we use to validate the JSON input of assessment files, requires `unsafe-eval` in in the app's content security policy (see their docs: https://ajv.js.org/security.html#content-security-policy). This isn't a problem in mv2, we just add the appropriate line to the manifest: https://github.com/microsoft/accessibility-insights-web/blob/main/Gruntfile.js#L518. But with mv3, chrome no longer allows extensions to include `unsafe-eval` in their manifest CSP, so this is no longer an option. As a solution, this PR migrates us to use AJV's standalone validation code process instead: https://ajv.js.org/standalone.html.

A few changes were necessary to convert to the standalone process:

1. Add json generator package. This allows us to build the validator generation code separately from the rest of the codebase.
2. Add extra steps to mv3 build process to (A) build the validator generation code, (B) run the validation generator, and (C) place the generated validator in the expected file such that it can be accessed by the rest of the code.
3. Add extra logic to `load-assessment-data-validator.ts` to determine if the build generated validator is populated or if we should fallback on using the old, dynamically generated validator. We theoretically could convert all products to use the new, build generated validator and remove the old code, but I decided to only convert the mv3 extension to minimize changes to the other products that share this logic.
5. In the case that we use the build generated validator, also check that the assessments data matches the enabled feature flags. With the dynamically generated validator, we are able to check on the fly which feature flags are enabled/ disabled, and as a result, the schema that the validator checks against excludes any assessments that are behind disabled feature flags. We can't check which flags are disabled when we're generating the validator on build, so instead we include all assessments in the build generated validator schema and separately check in `LoadAssessmentDataValidator.uploadedDataIsValid` that none of the assessments in the uploaded data are behind disabled feature flags.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`


src/packages/assessment-validator/index.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
pownkel, line 1: Is there a reason this needs to be separate from the `packages/validator` directory? It took me a while to understand the build flow because I was confused by there not being any code in `packages/validator`, and the connection to this script isn't obvious

sfoslund: I decided to do it this way to follow the `accessibility-insights-ui` package pattern, but yeah, if you think it's cleaner I don't see any blockers to moving this to same directory as the build logic.

sfoslund: Looking at this again, I think the ui package code is in its existing location so that it can easily reference/ reuse the code in the rest of the repo. Since we need to do the same thing with the validator package, I think the index file will need to stay in this location.

esbuild.js
```diff
@@ -113,6 +113,28 @@ switch (argsObj.env) {
...
 
+    case 'validator':
+        entryFiles = {
+            validator: `${src}/packages/assessment-validator/index.ts`,
+        };
+        outdir = path.join(__dirname, 'packages/validator/bundle');
+        platform = 'node';
+
+        plugins.push(
+            NodeResolve.NodeResolvePlugin({
```
waabid, line 124: Let's extract this out to re-use here and report.

Gruntfile.js
```diff
@@ -761,6 +777,16 @@ module.exports = function (grunt) {
         console.log(`package is in ${packageUIDropPath}`);
     });
 
+    grunt.registerTask('package-validator', function () {
+        const mustExistPath = path.join(packageValidatorBundlePath, 'validator.bundle.js');
+
+        mustExist(mustExistPath, 'Have you run webpack?');
```
waabid, line 783: typo: webpack to esbuild

packages/validator/README.md
```diff
@@ -0,0 +1,18 @@
+<!--
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+-->
+
+# accessibility-insights-assessment-validator
+
+This library produces a tool that generates a AJV JSON validator for the accessibility insights assessment data schema.
```
waabid, line 8: typo: "[...] a**n** AJV JSON [...]"

waabid: Also think Accessibility Insights should be capitalized.

Gruntfile.js
```diff
@@ -807,6 +833,8 @@ module.exports = function (grunt) {
     grunt.registerTask('build-dev-mv3', [
         'clean:intermediates',
         'exec:generate-scss-typings',
+        'build-package-validator',
```
waabid, line 836: I think this should also be in _somewhere_ in build:all as well, no?

waabid: As far as I can tell, it does not currently run there?

src/DetailsView/components/load-assessment-data-schema-provider.ts
```diff
@@ -0,0 +1,199 @@
...
+                schema,
+                requirement.requirementKey,
+                requirement.assessmentKey,
+            );
+        });
+        return schema;
+    }
+
+    private setAssessmentBaseProperties(schema: any, assessmentKey: string) {
+        const assessments = schema.properties.assessmentData.properties.assessments.properties;
```
waabid, line 47: nit: this chain is so long and used so frequently, I almost think it might be worth writing a private "getAssessments(schema)" method that does this for you, to make accessing this more readable.

waabid (commented):
Largely LGTM: do think this needs two reviewers though.

Gruntfile.js
sfoslund: I figured it would be redundant to add build-package-validator explicitly to build:all since it's already implicitly included through build-dev-mv3, do you think it's worthwhile to add is explicitly so it runs multiple times?

brocktaylor7 (approved):
This looks like it was a fun one to have to work around, good work! I think there will be a `yarn format:fix` required, but outside of that LGTM!

waabid (approved):
LGTM!

PR #5980, opened 2022-08-30
feat(implement-new-functionality-links): Allow step-specific telemetry

DaveTryon:
#### Details

Our feature adds step-specific telemetry to the `CHANGE_OVERALL_REQUIREMENT_STATUS` data point. This PR adds an optional `Requirement.getCompletedStepDetailsForTelemetry` method that gets queried to create additional telemetry for that `Requirement`. The new data is placed in a `stepDetails` object (the name is a bit generic because a similar approach could be used to add telemetry to _any_ telemetry data point, not just this one). Telemetry is unmodified unless the optional method is provided. This new method gets a copy of the `AssessmentStoreData` so that it has access to the data when creating the additional telemetry.

##### Motivation

Part of current feature work.

##### Context

A future PR will actually use this data in the label-in-name assessment for links.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`


src/common/telemetry-data-factory.ts
```diff
@@ -523,4 +538,43 @@ export class TelemetryDataFactory {
...
+            step.key,
+            newStatus[step.key].stepFinalResult === ManualTestStatus.PASS,
+            numInstances,
+            step.getCompletedStepDetailsForTelemetry
+                ? step.getCompletedStepDetailsForTelemetry(assessment, storeData)
+                : undefined,
+        );
+    }
+
+    private getNumInstances(step: Requirement, assessmentData: AssessmentData): number {
```
waabid, line 563: I think this should remain in completed-test-step-telemetry as this is not really relevant for telemetry-data-factory.

src/common/telemetry-data-factory.ts
```diff
@@ -523,4 +538,43 @@ export class TelemetryDataFactory {
             enabled,
         };
     }
+
+    public forCompletedTestStep(
```
waabid, line 542: Why are we moving this to the telemetry data factory? This seems very relevant to the "completed-test-step-telemetry" work where it constructs that telemetry data and the telemetry-data-factory can be responsible for putting it together.

src/common/telemetry-data-factory.ts
```diff
@@ -523,4 +538,43 @@ export class TelemetryDataFactory {
...
+            .getVisualizationConfiguration()
+            .getAssessmentData(storeData);
+        const numInstances = this.getNumInstances(step, assessmentData);
+        const newStatus = storeData.assessments[assessment.key].testStepStatus;
+        return this.forRequirementStatus(
+            assessment.visualizationType,
+            step.key,
+            newStatus[step.key].stepFinalResult === ManualTestStatus.PASS,
+            numInstances,
+            step.getCompletedStepDetailsForTelemetry
```
waabid, line 557: I think it would be better to add this default logic to AssessmentBuilder rather than here as I don't know if it's helpful for the telemetry data factory to know so much about assessment data.

waabid: I think if we also did do some sort of null pattern in the AssessmentBuilder we can likely avoid that if block on line 214 as well.

src/common/telemetry-data-factory.ts
DaveTryon: > I think this should remain in completed-test-step-telemetry as this is not really relevant for telemetry-data-factory.

Agreed. Making it public in TelemetryDataFactory is what led to the null:check problems. The new iteration limits TelemetryDataFactory changes to adding a new optional parameter to `forRequirementStatus`

src/common/telemetry-data-factory.ts
DaveTryon: Agreed

src/background/completed-test-step-telemetry-creator.ts
```diff
@@ -111,6 +117,16 @@ export class CompletedTestStepTelemetryCreator {
         return numInstances;
     }
 
+    private getStepDetails(
+        step: Requirement,
+        assessment: Assessment,
+        storeData: AssessmentStoreData,
+    ): any {
```
madalynrose, line 124: The usage of the `any` type feels a bit off to me, since it's generally avoided in our code base. While I know the data returned is surely going to vary from step to step, it might be worth making a flexible type that includes common data like the result status (pass/fail/incomplete) for the step, similar to types we use for `telemetryData` and `payload` objects, and then make a specific `LabelInNameStepDetails` type that lists exactly what we'll be sending over for this test step.

DaveTryon: For payloads, we have a strict contract since we need to programmatically produce and consume these items. I'm also not convinced that we have a generic base that needs to always be in this type. In this case, it's just an object. I could use `Object`, but that's only marginally improved from `any`. A quick scan for `: any` reveals that we use it in a lot of places, so I'm not fully convinced that stronger typing is needed.

waabid (approved):

PR #5965, opened 2022-08-25
feat(mv3-part7): Add command/ service worker logic to run e2e tests on mv3 extension

sfoslund:
#### Details

At the moment we are only running e2e tests on the mv2 extension, despite having dev and canary builds for the mv3 version. This PR adds service worker/mv3 specific testing logic alongside the mv2 specific testing logic, so we can run the existing e2e tests on either version.

Note that if you run the new mv3 e2e test command locally (`yarn test:e2e:mv3`) there are 3 failing test files. Based on preliminary investigations at least 2 of them seem like legitimate mv3 extension bugs which we will need to investigate further and fix in future PRs. 

##### Motivation

Feature work.

##### Context

This PR only adds the logic and command to be able to run these tests locally. More work needs to be done to fix some test failures and add a mv3 e2e test leg to CI. This work will be done in separate, future PRs. 

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #5984, opened 2022-08-30
feat(mv3-part7): Fix mv3 e2e test failures

sfoslund:
#### Details

Fix 2 mv3 bugs uncovered by running e2e tests on the mv3 extension (added in https://github.com/microsoft/accessibility-insights-web/pull/5965). The 2 bugs are described below and addressed in the 2 commits separately.

##### Motivation

Feature work.

##### Context

Failing cross origins iframe tests:

- This bug can be repro'ed by running mv3 extension manually. Repro steps: (1) Open all-cross-origin-iframe.html test resource (2) Run fastpass (3) See details view seems to be scanning forever, errors in service worker:
```
error injecting /bundle/injected.css: Error: Cannot access contents of url "file:///C:/code/accessibility-insights-web/src/tests/end-to-end/test-resources/all-cross-origin-iframe.html". Extension manifest must request permission to access this host. Error: Cannot access contents of url "file:///C:/code/accessibility-insights-web/src/tests/end-to-end/test-resources/all-cross-origin-iframe.html". Extension manifest must request permission to access this host.
```
```
error injecting /bundle/injected.bundle.js: Error: Cannot access contents of url "file:///C:/code/accessibility-insights-web/src/tests/end-to-end/test-resources/all-cross-origin-iframe.html". Extension manifest must request permission to access this host. Error: Cannot access contents of url "file:///C:/code/accessibility-insights-web/src/tests/end-to-end/test-resources/all-cross-origin-iframe.html". Extension manifest must request permission to access this host.
```

- Another side effect of this bug is that the yellow banner asking the user to grant permission to access iframes is not appearing at all in the mv3 extension at the moment.
- The cause was the permissions we were setting in the mv3 manifest. For mv2, we always request the activeTab permission (https://github.com/microsoft/accessibility-insights-web/blob/main/src/manifest.json#L12) and conditionally gain `optional_permissions: ['*://*/*']` if a user grants access to iframes from the yellow banner. Prior to this change, we were not requesting `activeTab` permissions for mv3 and we were requiring host permissions instead of making them optional. This PR fixes those mv3 permissions such that they are now analogous to those in the mv2 extension.
- The differing permission formats in mv2 and mv3 also necessitated changes to how we handle editing the manifest in out e2e test logic, which ended up being the bulk of the changes in this PR. 

Failing first time dialog tests: 
- This is a test specific bug, it does not repro when running the mv3 extension manually.
- Tests were attempting to run a script that referenced `chrome.tabs` before the service worker was fully started up and while `chrome.tabs` was still undefined. Fix was to check to ensure `chrome.tabs` is defined before running script.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/end-to-end/common/browser.ts
+++ b/src/tests/end-to-end/common/browser.ts
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { setTimeout } from 'timers/promises';
 import * as Playwright from 'playwright';
 import { ChromiumBrowserContext } from 'playwright';
 import { BackgroundContext } from 'tests/end-to-end/common/page-controllers/background-context';
@@ -18,18 +19,16 @@ import { Page } from './page-controllers/page';
 import { PopupPage, popupPageRelativeUrl } from './page-controllers/popup-page';
 import { TargetPage, targetPageUrl, TargetPageUrlOptions } from './page-controllers/target-page';
 
-import { setTimeout } from 'timers/promises';
-
 export class Browser {
     private memoizedBackgroundPage: BackgroundPage;
     private memoizedServiceWorker: ServiceWorker;
     private pages: Array<Page> = [];
     private underlyingBrowserContext: Playwright.BrowserContext | null;
-    private readonly MANIFEST_VERSION = process.env.WEB_E2E_TARGET;
 
     constructor(
         private readonly browserInstanceId: string,
         underlyingBrowserContext: Playwright.BrowserContext,
+        private readonly manifestV3: boolean,
         private readonly onClose?: () => Promise<void>,
     ) {
         this.underlyingBrowserContext = underlyingBrowserContext;
@@ -61,7 +60,7 @@ export class Browser {
     }
 
     public async background(): Promise<BackgroundContext> {
-        if (this.MANIFEST_VERSION?.includes('mv3')) {
+        if (this.manifestV3) {
             if (this.memoizedServiceWorker) {
                 return this.memoizedServiceWorker;
             }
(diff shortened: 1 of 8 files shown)


src/tests/end-to-end/common/browser.ts
```diff
@@ -189,6 +190,12 @@ export class Browser {
 
     private async getActivePageTabId(): Promise<number> {
         const backgroundPage = await this.background();
+
+        // Check chrome.tabs is initialized
+        while (await backgroundPage.evaluate(() => chrome.tabs == null, null)) {
```
pownkel, line 195: Can this loop have a max timeout in case something goes wrong and chrome.tabs is never initialized?

src/tests/end-to-end/common/mv3-manifest-instance.ts
```diff
@@ -0,0 +1,33 @@
...
+        return parsedManifest;
+    }
+
+    public constructor(private readonly content: chrome.runtime.ManifestV3) {}
+
+    public addTemporaryPermission(permissionToAdd: string): ManifestV3Instance {
+        if (this.content.host_permissions == null) {
+            this.content.host_permissions = [];
+        }
+        this.content.host_permissions.push(permissionToAdd);
```
pownkel, line 23: should this be optional_host_permissions?

sfoslund: My understanding here is that when we call `addTemporaryPermissions` in e2e tests we're essentially mocking the user giving us extra permissions. So in this case I think we would want it to be non-optional.

pownkel (approved):

PR #6012, opened 2022-09-08
fix(mv3-part-7): Fix assessment view updater bug

sfoslund:
#### Details

Fix mv3 bug in which partially automated requirements pages (ex. Landmarks > landmark roles) blink between results and spinner before showing results again. 

##### Motivation

Fix mv3 bug

##### Context

The root cause of this bug was that we were sending too many enable visualization messages. We send these messages in `assessment-view-update-handler` based on if the assessment store data thinks that a step has been scanned or not (`props.assessmentData.testStepStatus[step].isStepScanned`). We were seeing this bug because:

- The `assessment-view-update-handler` is called every time any store data updates
- Multiple stores have listeners for the scanCompleted action
- If any store besides the assessment store completes its updates first, it causes the `assessment-view-update-handler` to run its check against assessment store data that is in the process of being updates/ already out of date. 
- This can create a situation where the assessment store is finished with the scan but the `assessment-view-update-handler` thinks it hasn't, so it re-sends the enable visualization message, which causes a re-scan.

The solution I implemented is to only send enable visualization messages is the store update came from the assessment store. This means that we won't be confused by unrelated store updates anymore. 

#### Pull request checklist

- [ n/a] Addresses an existing issue: #0000
- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/assessment-view-update-handler.ts
+++ b/src/DetailsView/components/assessment-view-update-handler.ts
@@ -8,6 +8,7 @@ import {
 } from 'common/types/store-data/assessment-result-data';
 import { Tab } from 'common/types/store-data/itab';
 import { DetailsViewActionMessageCreator } from 'DetailsView/actions/details-view-action-message-creator';
+import { isEqual } from 'lodash';
 import { VisualizationType } from '../../common/types/visualization-type';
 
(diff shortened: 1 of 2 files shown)


pownkel (approved):
LGTM

PR #6015, opened 2022-09-08
fix: improve automated tab stops instance generation by batching getUniqueSelector

waabid:
#### Details

Creates a new axe-util API that allows us to getSelectors based on an array, meaning setup and teardown of axe is only done once.

##### Motivation

Significantly improves perf when many instances are generated by automated tab stops.

##### Context

Before:
canaryTabStops

After:
fixedDevTabStops

I believed this perf issue existed because, upon stopping, we loop through each tabbable item we didn't see, in this case hundreds of tabbable elements, and send a separate action message to create an instance for each one. Obviously, that wouldn't be performant and we could batch this. This was kind of correct and does improve performance but for scenarios which could take 20-30s, it would only improve that by 1-2s.
 
In actuality, the main performance culprit is `getUniqueSelector`. Basically, we need to setup axe each time we use that API but the cost of doing so is high. Doing so hundreds of times, like we can do in this case, is what causes the hanging.

I did not investigate whether there are other places where we use getUniqueSelector in quick succession as there might be perf gains to be had.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/tab-stops-requirement-evaluator.ts
+++ b/src/injected/tab-stops-requirement-evaluator.ts
@@ -1,8 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
 import { HTMLElementUtils } from 'common/html-element-utils';
-import { getUniqueSelector } from 'scanner/axe-utils';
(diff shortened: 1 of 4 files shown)


pownkel (approved):
LGTM

Issue #5931, opened 2022-08-16
Tab stops breaks if some iframes do not load


pownkel:
## Describe the bug

Currently, if we try to run any visualizer on a page where some iframes fail to load, the command (correctly) times out after 30 seconds and prints an error to the target page console. For automated checks and needs review, the visualization still shows in the rest of the page, so the failure isn't noticeable. However, for tab stops, the injected script throws an exception when trying to start the tab stops listener, and the visualizer appears to hang forever. If running tab stops from ad hoc tools, the spinner spins forever. If running tab stops from fast pass, the page appears frozen and the toggle cannot be disabled.

Before #5537, the timeout exception was ignored, but #5537 introduced awaits for the async processes that throw the timeout exception, so now a timeout stops execution before the tab stops visualizer can be enabled. Ideally, we would be able to detect which iframes are valid before sending a command, but a quick google search didn't yield any results for how to do that for iframes with a different domain origin. Some other short-term patches could be to improve error handling to let the tab stops visualizer proceed after the timeout (currently 30 seconds), or float the promises that throw the timeout exception (which is how the extension used to work).

Interesting to note is that this only seems to happen with our own FrameMessenger, not with axe-core's, hence why Automated Checks and Needs Review can still run as expected (barring an error message in the console). I ran out of time to investigate this before figuring out the exact reason, but I've learned the following from [axe-core's FrameMessenger documentation ](https://github.com/dequelabs/axe-core/blob/develop/doc/frame-messenger.md)and some experimentation on pages where I was seeing this bug:
- axe-core first pings each frame with a short timeout (500 milliseconds) to make sure they have the right version of axe injected
- If the frame does not respond, axe-core reports a failure of the `frame-tested` rule
- I've only seen `frame-tested` failures on one of the two pages I was using to investigate this bug, so this doesn't seem to be a reliable way of determining whether we'll be able to message a frame
- Our own messaging timeout is 30 seconds, which is much more noticeable to a user than axe-core's 500ms timeout. However, this doesn't seem to be the cause of the differences in behavior, as increasing axe's ping timeout does not make automated checks take longer.

From these results, it appears that axe-core does have some way to detect and skip problematic iframes, at least in some cases, although it would take more investigation to figure out what that is.

**To Reproduce**

1. Open src/tests/end-to-end/test-resources/all-cross-origin-iframe.html
2. When opened as a file without hosting a local server, the iframes on this page should show an error with "localhost refused to connect."
[...]

PR #6028, opened 2022-09-12
fix: Ping iframes before attempting to message them

pownkel:
#### Details

Creates AllFramesMessenger and uses that for the tab stops visualizer. Before sending any messages, AllFramesMessenger sends a ping to all iframes with a 500 ms timeout. It then only messages frames that responded to the ping. This is to prevent the extension from attempting to message a frame that can't respond (for example, because an adblocker has prevented the frame from loading, or because the script has not been injected)

##### Motivation

Addresses #5931 

##### Context

This is similar to the approach that axe-core uses to test iframes. Another solution I considered was to wrap all frame messages in an "ignored timeout" similar to the one introduced in #5933. However, in that case, we would still be sending spurious messages to frames that won't respond.

#### Pull request checklist

- [x] Addresses an existing issue: #5931
- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/frameCommunicators/all-frames-messenger.ts
+++ b/src/injected/frameCommunicators/all-frames-messenger.ts
@@ -79,39 +79,38 @@ export class AllFramesMessenger {
             return;
         }
 
-        const promises: Promise<unknown>[] = [];
-        for (let i = 0; i < allIFrameElements.length; i++) {
-            promises.push(
-                this.promiseFactory.timeout(
-                    this.singleFrameMessenger.sendMessageToFrame(allIFrameElements[i], {
-                        command: this.pingCommand,
-                    }),
-                    this.pingTimeoutMilliseconds,
-                ),
-            );
-        }
+        const promises: Promise<unknown>[] = Object.entries(allIFrameElements).map(([key, value]) =>
+            this.promiseFactory.timeout(
+                this.singleFrameMessenger.sendMessageToFrame(value, {
+                    command: this.pingCommand,
+                }),
+                this.pingTimeoutMilliseconds,
+            ),
+        );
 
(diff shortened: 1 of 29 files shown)


src/injected/frameCommunicators/all-frames-messenger.ts
```diff
@@ -0,0 +1,118 @@
...
+            'iframe',
+        ) as HTMLCollectionOf<HTMLIFrameElement>;
+
+        if (allIFrameElements.length === 0) {
+            this.responsiveFrames = [];
+            return;
+        }
+
+        const promises: Promise<unknown>[] = [];
+        for (let i = 0; i < allIFrameElements.length; i++) {
```
sfoslund, line 83: nit: any reason not to use a for each loop here? Or maybe `Object.entries(allIFrameElements).map(element => this.promiseFactory.timeout(...`

src/injected/frameCommunicators/all-frames-messenger.ts
```diff
@@ -0,0 +1,118 @@
...
+                // Throw if the error is anything else.
+                if (!(error instanceof TimeoutError)) {
+                    throw error;
+                }
+            }
+        });
+
+        if (!allFramesSucceeded) {
+            this.logger.error(
+                `Some iframes could not be reached within ${this.pingTimeoutMilliseconds} milliseconds. Those frames will be ignored.`,
```
sfoslund, line 114: nit: might be useful for debugging purposes to hold onto the timeout errors and include them in the logger.error call here, so that we have a stack trace/ original error message?

sfoslund (approved):

PR #6090, opened 2022-09-28
fix(mv3-part-8): Determine json validator on build

sfoslund:
#### Details

Discovered a bug in mv3 canary/ insider builds in which assessments could not be loaded because the fixes from https://github.com/microsoft/accessibility-insights-web/pull/5975 were not included in those build steps. 

##### Motivation

Fix mv3 bugs.

##### Context

The approach implemented in https://github.com/microsoft/accessibility-insights-web/pull/5975 involved checking an empty json validator file into the repo and replacing its content during the build for any builds that could not use the dynamically generated json validator (see https://github.com/microsoft/accessibility-insights-web/pull/5975 for details). This works for building items one by one (ex calling `yarn build:dev:mv3` then `yarn build:prod:mv3` sequentially), but does not work for `yarn build:all` because in this case we run all builds concurrently and therefore can't keep replacing the content of the validator file.

The solution this PR implements is to: 
- Continue to check in empty json validator file into repo and import it in source code (this PR changed that file name from `validate-assessment-json` to `empty-validate-assessment-json`).
- When generating the json validator at build time, place it in a separate file named `generated-validate-assessment-json`.
- When building extensions that need to use the build generated validator, update the import statement to import `generated-validate-assessment-json` instead of `empty-validate-assessment-json` and for extensions that can dynamically generate a validator, do nothing and keep the `empty-validate-assessment-json` import as is.
- This results in both the empty and build generated validators being included in the built extensions, but each extension uses the correct one based on its import statement.

I considered several alternative approaches, including importing the build generated validator file conditionally. As far as I could see, there was no practical way to implement this, especially considering our bundling process, so I discarded this idea.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`


pownkel (approved):
LGTM, great solution!

PR #6099, opened 2022-09-30
fix(mv3-part-8): Only persist store data when changed

sfoslund:
#### Details

Addressing assessment perf issues caused by MV3 changes to `emitChanged`. During MV3 testing we noticed that perf seems to degrade as you go through an assessment because the assessment data structure is growing and it therefore takes more time to persist data in the indexedDB. This PR mitigates this issue by minimizing the calls to indexedDB. 

##### Motivation

Address MV3 perf issue

##### Context

EmitChanged is often called for stores when no store data was actually changed, which causes us to make unnecessary calls to IndexedBD to persist data that is already saved in the database. This PR adds logic to keep track of the data we last saved to the database and only update the database when that data has changed. 

Testing results:
Scenario: Clean install of extension (nothing saved in stores) -> open new assessment against https://markreay.github.io/AU/before.html -> run assessment automated checks -> average `emitChanged` performance for each store across all calls made. Repeated 3 times for each branch and averaged.
Main branch (no caching of last persisted data):
Avg time for `emitChanged` in AssessmentStore: 184.5 ms
VisualizationStore: 159.9 ms
Pr branch (with caching of last persisted data):
AssessmentStore: 81.9 ms
VisualizationStore: 6.3 ms

These changes make more of a difference for some stores than others depending on how often the store data is actually updated. 

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/stores/assessment-store.ts
+++ b/src/background/stores/assessment-store.ts
@@ -54,10 +54,10 @@ export class AssessmentStore extends PersistentStore<AssessmentStoreData> {
         private readonly assessmentDataConverter: AssessmentDataConverter,
         private readonly assessmentDataRemover: AssessmentDataRemover,
         private readonly assessmentsProvider: AssessmentsProvider,
-        protected readonly idbInstance: IndexedDBAPI,
-        protected readonly persistedData: AssessmentStoreData,
+        idbInstance: IndexedDBAPI,
+        persistedData: AssessmentStoreData,
         private readonly initialAssessmentStoreDataGenerator: InitialAssessmentStoreDataGenerator,
-        protected readonly logger: Logger,
+        logger: Logger,
     ) {
         super(
             StoreNames.AssessmentStore,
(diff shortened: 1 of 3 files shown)


pownkel (approved):
LGTM, I ran an assessment with this branch and the UI (anecdotally) feels much faster now!

src/background/stores/assessment-store.ts
```diff
@@ -54,10 +54,10 @@ export class AssessmentStore extends PersistentStore<AssessmentStoreData> {
         private readonly assessmentDataConverter: AssessmentDataConverter,
         private readonly assessmentDataRemover: AssessmentDataRemover,
         private readonly assessmentsProvider: AssessmentsProvider,
-        protected readonly idbInstance: IndexedDBAPI,
-        protected readonly persistedData: AssessmentStoreData,
+        idbInstance: IndexedDBAPI,
```
peterdur, line 57: Do we actually need to remove `protected readonly` from all of these parameter properties?

waabid (approved):

src/background/stores/assessment-store.ts
sfoslund: No, it's not functionally necessary, but I noticed that they weren't being used anywhere other than in this constructor so I figured it would make sense to minimize their scope.

peterdur: Makes sense - I'd only looked at changed usage outside the constructor, and missed that there was no existing need for these to be properties before the change. Thanks!

Issue #6122, opened 2022-10-11
english a/an in error text


Bram-diederik:
My Dutch translation picked up some new tests that i believe are in error.

I did a google search and i state that I'm correct https://www.a-or-an.com/a_an/undefined

test messages:
"A undefined element must be coded correctly as a data table or a layout table"
"The headers attribute of a undefined element must reference the correct undefined element(s)"

madalynrose:
Thank you for letting us know about this issue, @Bram-diederik! Could you point us to where you are seeing these messages?

Bram-diederik:
Its a test case string somewhere in a html or json result export. My translation dictionary picked them up when i added the json format into my parser. 

Any html or json result export contains the error and can be found there.

madalynrose:
Thank you for finding this issue @Bram-diederik! It turned out to be a bug in the JSON export process where we parse the JSX description for requirements. `undefined` was being inserted into the final description instead of the element name (e.g. `header`, or `td`) in some cases and your attention to grammar highlighted this issue though it is present in more than those two places. A fix has been merged into main and will go out in the next release of the web extension!

PR #6132, opened 2022-10-14
fix: update json export to properly parse tagName elements

madalynrose:
#### Details

Our Assessment JSON export mechanism surfaces the rule id and description for each requirement. The description originates as JSX and then gets converted into a string. This works fine for the majority of our requirements, where the JSX `props` contain `children` that mostly include innerText. This is not the case for several requirements that include `Markup.Tag` tags.

For example, the Semantics Assessment's "Header Attribute" Requirement includes two `Markup.Tag` tags for `td` and `th` that do not have `children`. 

```jsx
const headersAttributeDescription: JSX.Element = (
    <span>
        The <Markup.CodeTerm>headers</Markup.CodeTerm> attribute of a <Markup.Tag tagName="td" />{' '}
        element must reference the correct <Markup.Tag tagName="th" /> element(s).
    </span>
);
```

When our JSON exporter encounters tags without children, it appends "undefined" to the string:
```json 
{
   "requirementKey": "headersAttribute",
   "requirementDescription": "The headers attribute of a undefined element must reference the correct undefined element(s)"
 } 
```

This fix adapts our description builder to look for the `tagName` property if `children` does not exist.

```json
{
   "requirementKey": "headersAttribute",
   "requirementDescription": "The headers attribute of a td element must reference the correct th element(s)"
}
```

##### Motivation

Addresses issue #6122 

#### Pull request checklist

- [x] Addresses an existing issue: #6122
- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/reports/assessment-json-export-json-builder.ts
+++ b/src/reports/assessment-json-export-json-builder.ts
@@ -247,7 +247,7 @@ function processRequirementDescription(descriptionProps): string {
             }
         });
     } else {
-        requirementDescription += descriptionProps.children;
+        requirementDescription += descriptionProps.children ?? descriptionProps.tagName;
     }
     return requirementDescription;
 }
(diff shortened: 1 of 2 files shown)


katydecorah (approved):
Nice fix, @madalynrose!

PR #6151, opened 2022-10-26
fix: address needs review accessibility failures

madalynrose:
#### Details

This PR:
* addresses a `label-content-name-mismatch` true positive failure in:
  * FastPass automated checks failure instances
    * screenshot of developer tools showing label with words in wrong order for the 'label' rule failure in automated checks
    * the original label was in the order of `rule ID + # + "failed" + rule description` but the order on screen is `icon showing # failed + rule  ID + rule description`. this changes the label to match the visible order of information.
  * FastPass Tab Stops instances
  * Report failure instances
* updates end-to-end tests accessibility checks to include Needs Review failures 

##### Motivation

Accessibility Compliance

##### Context

This `label-content-name-mismatch` failure fix is technically compliant but is likely not a good user experience for dictation users. We should prioritize user testing with dictation users to ensure that our controls are labelled properly.

As we enable more needs review rules that return failures rather than incomplete, we should run those rules on our extension for end-to-end tests. If any tests produce needs review failures, we should check for false/true positives and then create a snapshot for results that have false positives. This pattern has been established in two files here.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`


src/tests/end-to-end/tests/details-view/headings.test.ts
```diff
@@ -46,7 +46,11 @@ describe('Details View -> Assessment -> Headings', () => {
                     headingsPage,
                     detailsViewSelectors.mainContent,
                 );
-                expect(results).toStrictEqual([]);
+                // this results object has a false positive for label-content-name-misatch
+                // where the "show all visualizations" content does not match the label
+                // because the content is a checkbox. because this is symbolic, this criteria
```
waabid, line 51: nit: I think it might be clearer to say "It is a false-positive because the checkbox is symbolic and so the criteria doesn't apply".

src/tests/electron/common/scan-for-accessibility-issues.ts
```diff
@@ -27,13 +28,15 @@ async function scanForAccessibilityIssues(
...
 async function runAxeScan(client: Page, selector?: string): Promise<Result[]> {
     await injectAxeIfUndefined(client);
+    const needsReviewRulesConfig = getNeedsReviewRulesConfig();
+
     const axeRunOptions = {
         runOnly: {
             type: 'tag',
             values: ['wcag2a', 'wcag21a', 'wcag2aa', 'wcag21aa'],
         },
+        rules: needsReviewRulesConfig,
```
waabid, line 38: Would this run all the default + needs review rules or just the needs review rules?

My understanding of the axe run options indicates the latter but I think we want the former, right?

waabid (commented):
Largely LGTM, but have a couple notes.

waabid (approved):
LGTM

PR #6163, opened 2022-11-02
feat(medium-pass-1): refactor analyzers to use messages specified in message configuration

waabid:
#### Details

Analyzers do not use the analyzer configurations, given at construction, to decide message types. Instead, they use given to them at time of analysis. Allows to share usage of analyzers between Assessment and Medium Pass.

##### Motivation

Feature work.
##### Context
In Requirement configuration objects, we specify how an analyzer should be created and specify the message type it should use upon scan completion or other events during analysis (relevant for tab stop specific stuff).

```
    getAnalyzer: provider =>
        provider.createFocusTrackingAnalyzer({
            key: visibleFfocusOrderTestStep.focusOrder,
            testType: VisualizationType.VisibleFocusOrderAssessment,
            getAnalyzerMessageTypes: Messages.Assessment.AssessmentScanCompleted,
            analyzerProgressMessageType: Messages.Assessment.ScanUpdate,
            analyzerTerminatedMessageType: Messages.Assessment.TrackingCompleted,
        } (this object is an AnalyzerConfiguration),
```

These message types are generic (do not specify a single test completed, i.e. there is not a "HeadingsLevelScanCompleted" message type) wrt to the test but are not generic to Assessment vs. older/adhoc visualization. As a result, action creators listening to those specific message types only update those specific stores and not the medium pass store we will introduce.

How do we solve this? Some considerations to make for any possible solution:

1. Ideally, I think it makes sense to not "override" the message type specified in the configuration at runtime. It sets a precedent I'm not excited by: that configs can be overwritten/overridden at any time during operation.

2. Creating entirely new analyzer configurations for this is possible but how we manage analyzers is currently tied to VisualizationType; creating new VisualizationType values essentially leads to creating new assessment objects and duplicating all assessment objects, as far as I can tell. Obviously this is expensive and something we sought out to avoid for that reason and others.

Ideas:
1. Create a new method on the analyzer configuration to replace the static text properties for these message types. This method can take some input (either store data or a summary) to determine which message type to use (Assessment or Medium Pass).
2. **This was the approach taken in this PR:** Remove the message type properties completely from Requirement config objects and also potentially Visualization Configuration objects. At runtime determine which message type should be used: as shown in the above example, these messages are largely generic anyways so it doesn't make sense they are duplicated in each configuration. This wouldn't violate the first issue I listed because we aren't overriding anything in the configuration. Largely, this was taken because the message types are duplicated across the configuration objects for no benefit. Moving them to a single place seems better as we can remove some properties from configuration objects along the way as well. **Deleting the older message types present in the configurations is not in this PR** but will be in the future.

Note: I also did testing with MV3 to ensure there were no issues (including some MV3 focused testing around letting the service worker stop).

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`


src/injected/analyzers/get-analyzer-message-types.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export const GetAnalyzerMessageTypes = (detailsViewPivot: DetailsViewPivotType) => {
+    switch (detailsViewPivot) {
+        case DetailsViewPivotType.assessment:
+            return AssessmentVisualizationMessageTypes;
+
+        case DetailsViewPivotType.fastPass:
+            return AdhocVisualizationMessageTypes;
+    }
```
madalynrose, line 45: Now that the `DetailsViewPivotType` has been updated in main, we can include it here.
```suggestion
        case DetailsViewPivotType.mediumPass:
            return MediumPassVisualizationMessageTypes;
    }
```

madalynrose (changes requested):
One small comment but looks great!

madalynrose (approved):
LGTM!

Issue #5802, opened 2022-07-26
Scanning in Edge Internet Explorer Mode


lisli1:
## Describe the bug

When loading a page in Edge Internet Explorer Mode and running a scan such as FastPass, the scanning/loading circle continues looping, even though the scan isn't working. No error messages appear in the console.

**To Reproduce**

1. Open Edge browser
2. Go to any test target page, such as https://www.washington.edu/accesscomputing/AU/before.html
3. Select "Settings and more" menu in upper right corner of the browser and select "Reload in Internet Explorer mode" (near the bottom of the menu)
4. Open AI Web and run FastPass
5. See the scanning circle spin continuously in the resulting details page

## Expected behavior

Instead of scanning/loading circle, it would be helpful for the extension to instead display a message that AI Web doesn't work in Edge IE mode.

## Context (please complete the following information)

- OS Name & Version:  Windows 11 Version 10.0.22000 Build 22000
- AI-Web Version & Environment:  Production 2.32.0
- Browser Version:  Edge Version 103.0.1264.49 (Official build) (64-bit), Internet Explorer mode
- Target Page:  https://www.washington.edu/accesscomputing/AU/before.html

## Are you willing to submit a PR?

yes

## Did you search for similar existing issues?
yes

## Additional context

It seems to be possible to detect Edge IE mode: https://stackoverflow.com/a/66379147

sfoslund:
Addressed in https://github.com/microsoft/accessibility-insights-web/pull/6183

lisli1:
Verified fix in Canary Version 2022.12.30.1724

<img width="1112" alt="image" src="https://user-images.githubusercontent.com/48259897/210476538-52eef4ca-ba88-49b4-9b35-23d4897aad5c.png">
(screenshot of FastPass details view page with banner at the top, as well as a Windows notification on the bottom right corner, indicating the error)

PR #6183, opened 2022-11-15
fix: Add logic to detect failed injection in edge IE mode

sfoslund:
#### Details

Add logic to detect failed injections into the target page, retry, and if failure continues, sends a notification to the user.

##### Motivation

Addresses issue https://github.com/microsoft/accessibility-insights-web/issues/5802

##### Context

When investigating https://github.com/microsoft/accessibility-insights-web/issues/5802, it appears that we're failing because we attempt to inject the content scripts into the target page, the injection fails, and we never handle the error, which causes the details view to show that it's 'scanning' forever. To address this, this PR adds logic to handle an injection failure and, if it persists, send a notification to the user.

I'm open to suggestions for how to display the error when the injection fails. Right now I'm sending a notification since it feels minimally intrusive, but originally I implemented an 'injection failed' view similar to the ['no content' view](https://github.com/microsoft/accessibility-insights-web/blob/main/src/DetailsView/components/no-content-available/no-content-available.tsx) to be displayed in the details view. The problem with this was that it prevented users from accessing settings/ the left nav bar when the target page cannot be communicated with. This causes issues for our e2e tests as well because we grant minimal permissions for most of those tests, meaning often we can't communicate with the target page because of these limited permissions and as a result the injection failed path is triggered. This is fine when we're just sending a notification on injection failed, but would mean significant changes to e2e tests if we were to use a more intrusive view. 

#### Pull request checklist

- [x] Addresses an existing issue: [#5802](https://github.com/microsoft/accessibility-insights-web/issues/5802)
- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`


dbjorge:
@ferBonnin what is your preference for the UX in this case? Would you prefer this notification strategy or some sort of in-page error message?

ferBonnin:
> @ferBonnin what is your preference for the UX in this case? Would you prefer this notification strategy or some sort of in-page error message?

question: is this something we can detect when the user selects the extension and the pop-up dialog opens? (if yes, because it would be an ideal way to say we cannot scan here:
Accessibility Insights for Web popup dialog saying it cannot run on this url

sfoslund:
> is this something we can detect when the user selects the extension and the pop-up dialog opens?

I originally was thinking that UX would be ideal as well, but I haven't found a way to make it work 😞 The problem is that we can't tell if a page is in IE mode until we try to inject our content scripts into it, which we don't do until we start trying to scan

ferBonnin:
> > is this something we can detect when the user selects the extension and the pop-up dialog opens?
> 
> I originally was thinking that UX would be ideal as well, but I haven't found a way to make it work 😞 The problem is that we can't tell if a page is in IE mode until we try to inject our content scripts into it, which we don't do until we start trying to scan

Makes sense, no worries! In that case, I would suggest following our pattern of adding a banner at the top of the Accessibility Insights page 🙂

sfoslund:
> In that case, I would suggest following our pattern of adding a banner at the top of the Accessibility Insights page

A yellow banner across the top of the details view? 2 questions then:

- Are we okay with the main details view content continuing to spin and say 'scanning' forever then?
- What should the experience be for ad hoc tests (since no details view is open in that case)?

sfoslund:
Thanks @ferBonnin and @dbjorge, this PR is now up to date with the UX described in https://github.com/microsoft/accessibility-insights-web/pull/6183#issuecomment-1343027259 :)

src/background/injector-controller.ts
```diff
@@ -52,9 +57,22 @@ export class InjectorController {
...
         this.oldInspectType = inspectStoreState.inspectMode;
     };
+
+    private handleInjectionError = async (err: any): Promise<void> => {
+        this.logger.error(err);
+        const attempts = (this.visualizationStore.getState().injectionAttempts ?? 0) + 1;
+        const payload = {
+            failedAttempts: attempts,
+            injectionFailed: attempts > 3,
+        } as InjectionFailedPayload;
```
dbjorge, line 72: It would be preferable to use `const payload: InjectionFailedPayload = { /* contents */ };` here rather than `const payload = { /* contents */ } as InjectionFailedPayload;`. The former tells typescript "this variable may only be assigned objects which are of type `InjectionFailedPayload`", where the latter tells typescript "cast the object I've created here to an `InjectionFailedPayload` type, and infer the type of the variable to be the same type".

In practice, the latter gives fewer type protections because the `as` cast is telling typescript "yes, really treat this object as the type I'm telling you, *even if it's missing required properties*". In the former pattern, forgetting a required property (eg, accidentally omitting `injectionFailed` here) is an error, but in the latter pattern with `as`, it's not an error. That's why you often see the `as` pattern in unit tests, where we're often intentionally giving underdefined inputs (eg, `deps` that are intentionally missing unnecessary bits). But it's usually a bug to use `as` outside of unit tests - if a non-unit test really does need to omit part of a type, it's usually better to update the type to indicate that that part of the type might be null/undefined.

src/background/actions/injection-action-creator.ts
```diff
@@ -19,11 +23,22 @@ export class InjectionActionCreator {
...
     }
 
     private injectionStarted = async (): Promise<void> =>
         await this.injectionActions.injectionStarted.invoke(null);
 
     private injectionCompleted = async (): Promise<void> =>
         await this.injectionActions.injectionCompleted.invoke(null);
+
+    private injectionFailed = async (payload: InjectionFailedPayload): Promise<void> => {
+        if (payload.injectionFailed) {
```
dbjorge, line 39: The naming of the message vs the `injectionFailed` property makes this consuming code a bit confusing to read - the message is named `InjectionFailed`, but it only sometimes indicated that injection failed?

Maybe instead something like `failedPermanently`, or invert it as `shouldRetry`/`willRetry`?

src/common/types/store-data/visualization-store-data.ts
```diff
@@ -22,6 +22,8 @@ export interface VisualizationStoreData {
     selectedDetailsViewPivot: DetailsViewPivotType;
     injectingRequested: boolean;
     injectingStarted: boolean;
+    injectionFailed?: boolean;
```
dbjorge, line 25: It feels awkward to use 3 booleans for `injectingRequested`, `injectingStarted`, `injectionFailed` when only 4 of the 8 combinations of them represent valid states (`fff`/`tff`/`ttf`/`fft`). Perhaps it would be easier to understand the state machine if these 3 were replaced by a single `injectingState` enum?

src/background/stores/visualization-store.ts
```diff
@@ -248,6 +250,16 @@ export class VisualizationStore extends PersistentStore<VisualizationStoreData>
         await this.emitChanged();
     };
 
+    private onInjectionFailed = async (payload: InjectionFailedPayload): Promise<void> => {
+        this.state.injectionAttempts = payload.failedAttempts;
```
dbjorge, line 254: Should this get reset on successful injection?

src/tests/unit/tests/background/injector-controller.test.ts
```diff
@@ -116,6 +118,43 @@ describe('InjectorControllerTest', () => {
...
+            .setupTabStore({ id: tabId })
+            .setupVizStoreGetState(visualizationData)
+            .setupInspectStore({ inspectMode: InspectMode.off })
+            .setupInjectScriptsCall(tabId, 0);
+        validator.buildInjectorController().initialize();
+        await validator.inspectInjectCallback();
+        validator.verifyAll();
+    });
+
+    test('inject sends injection failed message when injection errors', async () => {
```
dbjorge, line 137: Missing a case for the `attempts > 3` transition that should set `injectionFailed: true`?

dbjorge (changes requested):
Thanks @sfoslund , apologies for the review delay! A few comments inline.

src/background/injector-controller.ts
sfoslund: Sounds good, thanks for the detailed explanation, that makes sense!

sfoslund:
@dbjorge this is updated now and ready for another round of review, thanks! One question- do you think there's any possibility of [switching from a bool to an enum](https://github.com/microsoft/accessibility-insights-web/pull/6183/files#diff-e96bcce8d8759d4cc4bba064974c1f709261877a00065c7c38ece7b1211815a7R23) causing a backwards compat issue with  indexeddb storage?

dbjorge:
> 


> @dbjorge this is updated now and ready for another round of review, thanks! One question- do you think there's any possibility of [switching from a bool to an enum](https://github.com/microsoft/accessibility-insights-web/pull/6183/files#diff-e96bcce8d8759d4cc4bba064974c1f709261877a00065c7c38ece7b1211815a7R23) causing a backwards compat issue with indexeddb storage?

Yes, we'll need a corresponding change to the code that handles loading persisted data for that store which translates from the old booleans to the new state variable.

sfoslund:
> Yes, we'll need a corresponding change to the code that handles loading persisted data for that store which translates from the old booleans to the new state variable.

@dbjorge great, just implemented this!

src/background/translate-persisted-data.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function translatePersistedData(persistedData: PersistedData): PersistedData {
+    forEach(persistedData.tabData, tabData => {
+        translateInjectionState(tabData.visualizationStoreData);
+    });
+
+    return persistedData;
+}
+
+function translateInjectionState(data: VisualizationStoreData) {
```
dbjorge, line 18: There's already a layer that exists to handle merging in old persisted state as a new initial store state (`initial-visualization-store-data.ts`) - we should reuse that instead of creating a new one.

src/background/translate-persisted-data.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    if (data['injectingStarted']) {
+        data.injectingState = InjectingState.injectingStarted;
+    } else {
+        data.injectingState = data['injectingRequested']
+            ? InjectingState.injectingRequested
+            : InjectingState.notInjecting;
+    }
+    delete data['injectingRequested'];
```
dbjorge, line 31: In general, a function named `translate` that returns the same type it takes in shouldn't be mutating the original copy of the data - that's extremely misleading based on the name and an easy source of bugs. If mutating it makes sense for a given application, it makes it a lot harder to misunderstand the consuming code if both the name and typings are clear that it's mutating the original (eg, using a verb like `update` instead of `translate` and returning `void`).

src/background/translate-persisted-data.ts
sfoslund: Fixed, thanks!

dbjorge (approved):

PR #6202, opened 2022-11-22
refactor(medium-pass-1): add forEachConfig and medium pass provider to VisualizationConfigurationFactory

waabid:
#### Details

Heavily suggest reviewing by commit due to signature changes causing many files to be changed.

This PR can roughly be broken into 3 things:

1. Use constructor to pass in assessment and medium pass provider: this requires the most amount of file changes as each usage/creation of the factory needs to be changed.
2. Add the ability to loop through each configuration from the factory instead of components being responsible for this. The benefit here is that each component doing this no longer needs to be aware of the difference between assessment and not but also the structure of the assessment data.
3. Consumption of this forEachConfig method in the appropriate locations.

##### Motivation

feature work.

##### Context

1. `getIdentifier` is refactored out of AssessmentBuilder. This allows us to use test mode as a part of generating the identifier allowing for it to be unique across medium pass and assessment; removing it is the 3rd commit but adding it to config factory is in the first commit.
2. The mediumPass provider is currently optional so as not to cause issues/errors as the work on that is in the next PR. This will be removed in that PR as well.
3. I did verify that the async behavior in `VisualizationStateChangeHandler` was kept the same.
4. There is obviously a lot of potential refactoring work around the configuration factory (many tests don't mock it, we could build configs only once, we could break the building from the using/getting, etc.) but this PR is large enough as it is and I tried to maintain what we were doing in the main branch as best as I could.

#### Pull request checklist

- [ ] Addresses an existing issue: #0000
- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [ ] (UI changes only) Added screenshots/GIFs to description above
- [ ] (UI changes only) Verified usability with NVDA/JAWS

PR #6224, opened 2022-12-01
feat(medium-pass-1): add visualization types for quick assess/medium pass

waabid:
#### Details

Adding new visualization types necessitates changes in many places. Generally speaking the changes fall into the following camps:
- VisualizationConfigurationFactory should not throw errors when a new visualization type is used for `getConfiguration`. This requires the medium pass provider be implemented and as such many changes in this PR are signature updates for the creation of the factory.
- `getStoreData` needs to be updated as this is another place where Assessment and QuickAssess will differ. As such, they've been removed from assessment-builder and moved to the config factory.
- Consequently this also means we must update the visualization store to contain mediumPass data.
- The requirement filter is modified to amend the VisualizationType on the configs and now uses a requirement/testStep to VisualizationType map to do so.
- Another minor change: checking the TestMode to be Adhoc instead of Assessment as QuickAssess and Assessment will share the same behavior. Don't feel super strong about this change but it made sense to me to at least rename these methods where applicable.

##### Motivation

Feature work.

##### Context

Overwriting the VisualizationType in the written Assessment configurations is not ideal. Ultimately, landed with this because we should move to an architecture where the assessment objects themselves do not contain VisualizationType but are also added/amended with some mapping. The changes made here do not make that work harder in the future so I went ahead with this for now.

This still uses medium pass in the naming because it was written before the quick assess name change; should not be an issue to change later.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/details-view-initializer.ts
+++ b/src/DetailsView/details-view-initializer.ts
@@ -5,7 +5,11 @@ import { loadTheme, setFocusVisibility } from '@fluentui/react';
 import { AssessmentDefaultMessageGenerator } from 'assessments/assessment-default-message-generator';
 import { Assessments } from 'assessments/assessments';
 import { assessmentsProviderWithFeaturesEnabled } from 'assessments/assessments-feature-flag-filter';
-import { MediumPassRequirementKeys } from 'assessments/medium-pass-requirements';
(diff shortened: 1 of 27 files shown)


madalynrose (approved):
These changes are fairly straightforward after #6202. Looks great!

PR #6225, opened 2022-12-01
feat(medium-pass-1): Create base analyzer configuration in analyzer-controller and pipe into getAnalyzer

waabid:
#### Details

Our configuration objects statically link analyzers and their communication/lifecycle methods to visualization type, message types, and requirement keys. Since these are used by our communication methods, this effectively amounts to the configuration objects, a place for business logic, containing architectural logic which we should avoid.

As such, this PR changes it such that the analyzer controller is responsible for this duty. This PR does not contain:
1. Modifying the configuration objects to remove mention of visualization type, message types, and requirement keys. This will be done in the next PR.
2. Modifying the analyzers to NOT use the message configuration passed in during `analyze`; it makes sense to re-factor this out with this change but in a future PR.

To summarize, this PR makes the analyzer controller responsible for piping the `testType`, the requirement key, and the message types through getAnalyzer but does not use this newly piped object (except for the automated checks assessment, which is updated in this PR); the only reason Automated Checks was modified was because the test was breaking and it made sense to implement while fixing.

This is useful for quick assess because we then do not need to worry about over-writing something else in the assessment definitions, except the bare-minimum (visualizationType).

##### Motivation

Feature work.

##### Context

How are these 3 properties used:
1. `testType` and `requirementKey` are used for telemetry and for creating the notification when the scan is completed.
2.  `analyzerMessageType` is used to for an action message type to signify that a scan is completed.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/assessments/assessment-builder.tsx
+++ b/src/assessments/assessment-builder.tsx
@@ -8,12 +8,12 @@ import {
     UniquelyIdentifiableInstances,
 } from 'background/instance-identifier-generator';
 import { AssessmentVisualizationConfiguration } from 'common/configs/assessment-visualization-configuration';
-import { Messages } from 'common/messages';
 import { InstanceIdToInstanceDataMap } from 'common/types/store-data/assessment-result-data';
 import { FeatureFlagStoreData } from 'common/types/store-data/feature-flag-store-data';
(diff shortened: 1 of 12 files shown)


lisli1 (approved):
LGTM

PR #6231, opened 2022-12-02
fix: ensure default store data is merged with persisted data during VisualizationStore initialization

madalynrose:
#### Details

When there is persisted data in IndexedDB for the `VisualizationStore` and `storeData.tests` is missing the `mediumPass` key, the `WebConfigurationVisualizationFactory` hits an error. This PR overrides `generateDefaultState` for `VisualizationStore` (called during `initialize()`), allowing us to ensure that default state is generated for test modes not present in `persistedData`. This implementation is based on `AssessmentStore`, which is the only other store that overrides this method. It pulls initial data generation logic into a separate class to facilitate testing. 

##### Motivation

This PR addresses an issue seen in Canary where adhoc visualizations are nonfunctional and the service worker logs the following error: 
```Error while processing browser RuntimeOnMessage event:  TypeError: Cannot read properties of undefined (reading 'keyboardInteractionAssessment')```

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/initial-visualization-store-data-generator.ts
+++ b/src/background/initial-visualization-store-data-generator.ts
@@ -1,0 +1,60 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { VisualizationConfigurationFactory } from 'common/configs/visualization-configuration-factory';
+import { EnumHelper } from 'common/enum-helper';
+import { DetailsViewPivotType } from 'common/types/store-data/details-view-pivot-type';
+import {
+    TestsEnabledState,
+    VisualizationStoreData,
+} from 'common/types/store-data/visualization-store-data';
+import { VisualizationType } from 'common/types/visualization-type';
+
+export class InitialVisualizationStoreDataGenerator {
+    constructor(
+        private readonly visualizationConfigurationFactory: VisualizationConfigurationFactory,
+    ) {}
+
+    public generateInitialState(
+        persistedData: VisualizationStoreData = null,
+    ): VisualizationStoreData {
+        const defaultTests: TestsEnabledState = {
+            adhoc: {},
+            assessments: {},
+            mediumPass: {},
+        };
+
+        if (this.visualizationConfigurationFactory != null) {
+            EnumHelper.getNumericValues(VisualizationType).forEach((test: VisualizationType) => {
+                const config = this.visualizationConfigurationFactory.getConfiguration(test);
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/background/initial-visualization-store-data-generator.ts [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+                defaultTests.mediumPass[key].stepStatus = {};
+            });
+        }
+        const tests: TestsEnabledState = persistedData?.tests
+            ? { ...defaultTests, ...persistedData.tests }
+            : defaultTests;
+
+        const defaultValues: VisualizationStoreData = {
+            tests,
+            scanning: null,
```
dbjorge, line 49: This will throw away any persisted values for all these non-test properties, no?

src/background/initial-visualization-store-data-generator.ts
```diff
@@ -0,0 +1,60 @@
...
+            Object.keys(defaultTests.assessments).forEach(key => {
+                defaultTests.assessments[key].stepStatus = {};
+            });
+
+            Object.keys(defaultTests.mediumPass).forEach(key => {
+                defaultTests.mediumPass[key].stepStatus = {};
+            });
+        }
+        const tests: TestsEnabledState = persistedData?.tests
+            ? { ...defaultTests, ...persistedData.tests }
```
dbjorge, line 44: This solves the specific issue with the `mediumPass` key being missing, but I worry that it would lead to a similar issue in the future if we ever added a new ndividual test. Perhaps a lodash `merge()` would be a more robust way to solve both this and the issue below with the non-`tests` properties?

dbjorge: I think it might be even better to more generally update `persistent-store` to default to merging persisted state into default state instead of keeping persisted state as-is, but that's a scary change to do broadly - maybe best to hold off on that until after the initial MV3 release

dbjorge (changes requested):
The restructuring into the separate initializer class LGTM, left a few comments on the initializer merging algorithm

src/background/initial-visualization-store-data-generator.ts
dbjorge: Discussed offline with @madalynrose and @waabid , decided to use a merge strategy for this store only for now and track doing it more generally after the MV3 release in #6233

dbjorge (approved):

PR #6235, opened 2022-12-05
refactor(medium-pass-1): remove usages of assessmentsProvider through props in details-view-content

waabid:
#### Details

Just a minor clean up to not use assessments provider through props but deps, which is how we usually do it.

##### Motivation

Feature work.

##### Context

This should make the next couple of PRs a bit more simple as we look to normalize how we get assessments provider through a get method.

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/details-view-command-bar.tsx
+++ b/src/DetailsView/components/details-view-command-bar.tsx
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { IButton } from '@fluentui/react';
-import { AssessmentsProvider } from 'assessments/types/assessments-provider';
 import { NewTabLinkWithTooltip } from 'common/components/new-tab-link-with-tooltip';
 import { VisualizationConfigurationFactory } from 'common/configs/visualization-configuration-factory';
 import { CardsViewModel } from 'common/types/store-data/card-view-model';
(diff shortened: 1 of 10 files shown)


madalynrose (approved):

madalynrose (changes requested):
After a second look, I think there are few more places that still pass the provider from `deps` into `props`:
* `GetLeftNavSelectedKey`
* `StartOverDialog`
* `RequirementTableSection`, `ManualTestView`, and `FailureInstancePanelControl`
  * none of these three have a `props.deps` but all three are using `deps.assessmentsProvider` in their `props` so I think it makes sense to move them to using deps.

waabid:
> After a second look, I think there are few more places that still pass the provider from `deps` into `props`:
> 
> * `GetLeftNavSelectedKey`
> * `StartOverDialog`
> * `RequirementTableSection`, `ManualTestView`, and `FailureInstancePanelControl`
>   
>   * none of these three have a `props.deps` but all three are using `deps.assessmentsProvider` in their `props` so I think it makes sense to move them to using deps.

Discussed this offline: won't be making these changes right now as we have lots of places where we really should be using the provider through deps instead of props but it's not absolutely necessary that we get this now.

Will create an issue tracking this.

madalynrose (approved):

PR #6278, opened 2022-12-19
chore: Disable `{property|value}-no-vendor-prefix` stylelint rules, clean up prefixes

katydecorah:
#### Details

This pull request will officially disable the stylelint rules `property-no-vendor-prefix` and `value-no-vendor-prefix`. It also audits current prefix usage and removed vendor prefixes that are no longer needed and makes adjustments to existing vendor prefixes as needed.

##### Motivation

##### Alternatives to autoprefixer

- [stylelint-no-unsupported-browser-features](https://github.com/ismay/stylelint-no-unsupported-browser-features) - We would install and enable this stylelint rule to help us learn about unsupported features. I don't like this option as the [recommendation is to set it to "warn"](https://github.com/ismay/stylelint-no-unsupported-browser-features#recommendations) and warnings are easy to miss.
- These rules are meant to work with a secondary build tool, like [autoprefixer](https://github.com/postcss/autoprefixer), however the cost of adding autoprefixer (added dependencies, increase build times) outweighs the benefits.

#### Pull request checklist

- [x] Addresses an existing issue: https://github.com/microsoft/accessibility-insights-web/issues/5187
- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`


src/reports/components/instance-details.scss
```diff
@@ -27,7 +17,6 @@
     outline-style: 'border-style';
     box-shadow: 0 0.6px 1.8px $box-shadow-108, 0 3.2px 7.2px $box-shadow-132;
     margin-bottom: 16px;
-    @include fill-available-width;
```
dbjorge, line 30: Was it intentional to remove this entirely rather than replacing it with `width: fill-available;`?

src/reports/components/instance-details.scss
```diff
@@ -61,8 +50,6 @@
     background: $neutral-0;
     display: table;
     table-layout: fixed;
-    @include fill-available-width;
```
dbjorge, line 64: Was it intentional to remove this entirely rather than replacing it with `width: fill-available;`?

src/reports/components/instance-details.scss
katydecorah: When I tested initially, I found no impact from this declaration. I'm not quite finished validating these changes but will share my findings when I'm ready for review.

dbjorge:
It's a good point that our browserlist entry is outdated, I would support switching our browserslist config to a dynamic option. I'd propose something like `>= 0.2%, last 2 versions, not dead, last 4 chrome major versions, last 4 edge major versions`, but I'd be open to other suggestions.

It might be good to pair this with some sort of cron action to run `npx update-browserslist-db@latest` periodically - that's probably simple enough that it'd be better to just write our own action to run the command, but if I'm underestimating the complexity of having an action run `gh create pr`, there do exist purpose-built actions for that (eg https://github.com/c2corg/browserslist-update-action)

src/reports/components/instance-details.scss
katydecorah: From testing the extension and generated report, I don't believe setting `width` has an impact on this element and appears safe to remove. Please let me know if there's a scenario I'm missing.

katydecorah:
> It might be good to pair this with some sort of cron action to run npx update-browserslist-db@latest periodically 

@dbjorge I love this idea! I started playing around with this idea and will create a separate pull request to propose the new workflow.

katydecorah:
> It's a good point that our browserlist entry is outdated, I would support switching our browserslist config to a dynamic option. I'd propose something like >= 0.2%, last 2 versions, not dead, last 4 chrome major versions, last 4 edge major versions, but I'd be open to other suggestions. [https://browsersl.ist](https://browsersl.ist/#q=%3E%3D+0.2%25%2C+last+2+versions%2C+not+dead%2C+last+4+chrome+major+versions%2C+last+4+edge+major+versions) is a site where you can test different queries and read about best practices.

I also like using a dynamic configuration here to avoid maintenance. With the dynamic configuration, we add significantly more browser and audience coverage, which resulted in some added prefixes to the report snapshot tests (as demonstrated in commit https://github.com/microsoft/accessibility-insights-web/pull/6278/commits/80cb3a0b66939e5b97280c1581cecdc4a3e7bd04).

The next section compares the current browser support and the proposed browser support.


### Current browser support using static configuration

Audience coverage: 0.04 %

- chrome 88
- edge 88
- firefox 87


### Proposed browser support using dynamic configuration

Audience coverage: 93.3 %

- and_chr 108
- and_ff 107
- and_qq 13.1
- and_uc 13.4
- android 108
- android 4.4.3-4.4.4
- chrome 108
- chrome 107
- chrome 106
- chrome 105
- chrome 104
- chrome 103
- edge 108
- edge 107
- edge 106
- edge 105
- firefox 107
- firefox 106
- ios_saf 16.1
- ios_saf 16.0
- ios_saf 15.6
- ios_saf 15.5
- ios_saf 15.4
- ios_saf 15.2-15.3
- ios_saf 14.5-14.8
- ios_saf 14.0-14.4
- ios_saf 12.2-12.5
- kaios 2.5
- op_mini all
- op_mob 72
- opera 92
- opera 91
- safari 16.1
- safari 16.0
- safari 15.6
- safari 15.5
- safari 14.1
- safari 13.1
- samsung 19.0
- samsung 18.0

src/electron/views/common/window-title/window-title.scss
```diff
@@ -14,7 +14,7 @@
     display: flex;
     align-items: center;
     -webkit-app-region: drag;
-    -webkit-user-select: none;
```
katydecorah, line 17: This needs investigation.

katydecorah: I think we need to add autoprefixer to webpack.config.js.

katydecorah:
I'm learning a lot! We build CSS a few different ways:

- [x] webpack
- [x] esbuild
- [ ] grunt

If we decide to move forward, we'll want to setup autoprefixer with each build system. I'm feeling a little unsure about grunt as many packages have been deprecated in this space. I'll keep investigating, but I'm watching how the dependency list is growing here.

src/electron/views/common/window-title/window-title.scss
katydecorah: Added in https://github.com/microsoft/accessibility-insights-web/pull/6278/commits/ccdbddf1d82061ec6ca24706c2cc9d9336f35dae, I confirmed by temporarily commit the drop files to see the diff.

src/reports/components/report-sections/header-section.scss
```diff
@@ -8,7 +8,7 @@
     height: 40px;
     width: 100%;
     display: flex;
-    justify-content: end;
+    justify-content: flex-end;
```
katydecorah, line 11: autoprefixer provided the following warning: "end value has mixed support, consider using flex-end instead"

katydecorah:
It would also be good to compare build times before and after adding autoprefixer.

katydecorah:
After engineering discussion yesterday, I went on a quest to see if we can improve the build time performance of autoprefixer. Seeing no clear opportunities, I looked more carefully at the value autoprefixer provided by comparing compiled CSS files from production against what this pull request would propose. In the end, I didn't see a strong value add with any of the prefixes that we didn't already have.

While autoprefixer would help automate and catch prefixes for us, it doesn't seem to outweigh the costs:

- **Adding 3 new dependencies (autoprefixer, postcss-cli, and postcss-loader).** These new dependencies allow us to enable autoprefixer wherever we build CSS: webpack, grunt, and esbuild. 
- **Increasing build times.** The addition of autoprefixer will add approximately ~7s to build times (16.8% increase for build, 7.6% increase for tbuild).
- **Needing to keep browserslist up-to-date.** Since autoprefixer requires latest updates to the Can I Use database, we need to keep this updated. This happens by [running a command](https://github.com/browserslist/browserslist#update-caniuse-lite) which will update `yarn.lock`. We can automate this process with a GitHub action.

Instead, this PR officially disables the prefix-related stylelint rules and does some prefix clean up.

dbjorge (approved):

src/DetailsView/components/static-content-common.scss
```diff
@@ -10,7 +10,8 @@
     margin-bottom: auto;
 
     ol {
-        -webkit-padding-start: 16px;
```
peterdur, line 13: Is the `static-content-in-details-view` style actually used? I couldn't find it in a search of the repo.

peterdur (approved):
LGTM. One style may be obsolete.
Built locally, generated a report, and looked for visual changes.

src/DetailsView/components/static-content-common.scss
katydecorah: I do see the class used in `src\DetailsView\components\static-content-details-view.tsx` and `src\DetailsView\components\target-page-changed-view.tsx`, however it's not clear to me if the nested `ol` is in use. 

I do see some interesting font-size and style inconsistencies when the target page changes. I'll plan to look into this closer in another pull request. I'm imagining there's an opportunity to consolidate styles here to prevent the style shift when the target page changes.

The screenshot below shows the "Needs review" panel:



The next screenshot shows the same "Needs review" panel, when the target page changes. There is a change to the font-size and style, especially on the italicized "possible":

PR #6279, opened 2022-12-19
chore(null): null checks for web-visualization-config-factory and deps

dbjorge:
#### Details

This PR adds just a single file to the strict null check list (`src/common/configs/web-visualization-configuration-factory.ts`), but I've given it its own PR since one of the changes touched a lot of files. Specifically, the factory involved initializing as null several visualization fields that were present in the `VisualizationConfiguration` type but never used except by ad-hoc checks.

In one case, there was a pattern that I thought was a bit misleading where a bag of UI strings called `displayableData` contained a mix of properties which all visualizations were expected to provide (`title`) vs some which only ad-hoc visualizations were expected to provide (several properties used to store the labels/etc on the ad hoc popup panel). I separated these so the ad-hoc specific ones were under a new `adHoc` grouping and included a comment on the grouping that clarified when this could be expected to be null or not. This involved touching all the files which use these properties.

This isn't quite as good as using separate typings for the ad-hoc vs non-ad-hoc options to make the type system enforce that comment, but I felt it was a reasonable middle ground for how much existing code to re-write in order to make this change.

##### Motivation

#2869

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn null:autoadd`
- [x] Ran `yarn fastpass`


sfoslund (approved):
LGTM!

PR #6284, opened 2022-12-20
chore(null): switch strictNullChecks config from opt-in to opt-out for new files

dbjorge:
#### Details

This PR updates our strictNullChecks config and tooling to include all non-test `.ts` and `.tsx` files in the strict null checks set by default, only excluding those files explicitly marked as having known issues in `tsconfig.strictNullChecks.json`. This:

* Simplifies the `strictNullChecks` config file a lot (it reduces the size by half and is now a flat list of files instead of a mix of files and glob patterns)
* Saves a step of having to manually run `yarn null:autoadd` for most changes, since new files are now implicitly added
* Removes the corresponding item from our PR checklist (I left the automated PR check in place since it doesn't hurt anything and has a self-descriptive error message if it does fail)

To implement this, I wrote a script to perform the conversion from an include list to an exclude list. I included it for review in this PR, but I don't expect it to need to be run again in the future. I verified that `yarn null:progress` reports finding the same counts of checked and eligible files before and after the change:

> ## Web strict-null progress
> **81%** complete (**1273**/1562 non-test files)
> *Contribute at [#2869](https://github.com/microsoft/accessibility-insights-web/issues/2869). Last update: Tue Dec 20 2022*

I ensured that the assorted helper scripts (`yarn null:find`, `yarn null:autoadd`, `yarn null:check`) all still work as expected.

##### Motivation

#2869 and see above

##### Context

I considered renaming `autoadd` to something else now that it's arguably "removing an exclusion" rather than "adding an inclusion", but I decided to leave it as it; it's still "adding new files to the strict null checked set" and it felt like more work/confusion than it was worth to try to update all the issues/team working knowledge/etc about it.

#### Pull request checklist

- ~~[x] Ran `yarn null:autoadd`~~
- [x] Ran `yarn fastpass`


tools/strict-null-checks/write-tsconfig.js
```diff
@@ -1,12 +1,20 @@
...
+const prettier = require('prettier');
 
 module.exports = {
-    writeTsconfigSync: (tsconfigPath, content) => {
+    writeTsConfig: async (tsconfigPath, content) => {
         let serializedContent = JSON.stringify(content, null, '    ');
-        serializedContent += '\n';
 
-        fs.writeFileSync(tsconfigPath, serializedContent);
+        let prettierConfigPath = path.join(__dirname, '..', '..', 'prettier.config.js');
```
dbjorge, line 11: Without this chance, `yarn null:autoadd` would try to write out the `include:` property on multiple lines and conflict with prettier wanting it on one line

madalynrose (approved):
This is awesome!!

Issue #6212, opened 2022-11-28
[Bug] Test page contents are not rendering properly after running automated checks


DaveTryon:
## Describe the bug

We have a web site where running AIWeb causes the content of the nav pane to disappear. It exists in HTML, but does not render as expected.

**To Reproduce**

1. Go to https://tdc.azure.net/welcome (requires MSFT login). Notice that the Home button is displayed
<img width="375" alt="image" src="https://user-images.githubusercontent.com/45672944/204363622-d20ff1dc-e608-454a-9920-691590a14ce7.png">

2. Run FastPass. Notice that the Home button is _not_ displayed
<img width="374" alt="image" src="https://user-images.githubusercontent.com/45672944/204363792-32338b6c-6362-47f4-8a3a-105557b8cfce.png">

3. Inspecting the page when the button is not displaying confirms that it's still there in the HTML, but something is interfering with the rendering.

Additional observation: I positioned the FastPass results and the original page so they were both visible, then refreshed the original page multiple times. 2 consistent patterns emerged:
1. On the first refresh, the Home button flashed momentarily then disappeared. It consistently appeared on the second refresh.
2. Pressing "Start Over" from the automated checks page caused the home button to disappear again.

## Context (please complete the following information)

- OS Name & Version: Win11 22H2 (also happens on MacOS Ventura 13.0.1)
- AI-Web Version & Environment: 2.35.0 Production
- Browser Version: Google Chrome 107.0.5304 (Official Build)
- Target Page: https://tdc.azure.net/welcome

ferBonnin:
triaged with @peterdur we should investigate. Ideally let's create a minimal repro that doesn't depend on this page. We should check if it has anything to do with both sites using Fluent UI

dbjorge:
Resolved by #6294

PR #6294, opened 2022-12-22
fix: prevent styling conflicts with certain fluentui-based target pages

dbjorge:
#### Details

`client-init` contains the following bit of code very early during injected page initialization:

```ts
    const { Stylesheet } = await import('@fluentui/merge-styles');
    const stylesheet = Stylesheet.getInstance();
    stylesheet.setConfig({
        defaultPrefix: 'insights',
    });
```

This code exists because many `@fluentui/react` components' styling works internally by dynamically generating some CSS and injecting it into a document-level fluentui-managed `<style>` element using [`@fluentui/merge-styles`](https://github.com/microsoft/fluentui/blob/master/packages/merge-styles/README.md). Because this is injecting into a document-level style element within the DOM, it is "visible" to the target page itself, not just our injected content, and that means that if the target page already uses fluentui itself, the two fluentui instances can stomp on each other and override each others' styles. Changing the `defaultPrefix` is intended to avoid this by forcing our instance of `fluentui` to use a special, non-default prefix for all the CSS styles it injects (`insights-whatever` instead of its default `css-whatever`).

However, it turns out that this configuration was not quite happening early enough. In particular, before this PR, we were performing this configuration *after* an `await import('../common/fabric-icons')` statement (but before invoking any of the imported code). This turns out to be too late; there are certain `@fluentui` components that start setting up styles and mutating the global state of the page just by being imported. In this specific case, the following import path:

```
src/common/fabric-icons.ts
node_modules/@fluentui/font-icons-mdl2/lib/index.js
node_modules/@fluentui/font-icons-mdl2/lib/fabric-icons.js
node_modules/@fluentui/style-utilities/lib/index.js
node_modules/@fluentui/style-utilities/lib/classNames/index.js
node_modules/@fluentui/style-utilities/lib/classNames/AnimationClassNames.js
node_modules/@fluentui/style-utilities/lib/styles/index.js
node_modules/@fluentui/style-utilities/lib/styles/AnimationStyles.js
node_modules/@fluentui/theme/lib/index.js
node_modules/@fluentui/theme/lib/motion/index.js
node_modules/@fluentui/theme/lib/motion/AnimationStyles.js
```

...runs the following code on import:

```ts
const FADE_OUT: string = keyframes({
  from: { opacity: 1 },
  to: { opacity: 0, visibility: 'hidden' },
});
```

...which ends up conflicting with a class that the target page happened to define for a portion of the target page's nav menu, resulting in the nav menu being given an opacity 0/hidden end state for its flyin transition animation.

This PR fixes the issue by moving our `defaultPrefix` config to happen *before* we transitively import any fluentui bits (besides the `setConfig` functionality itself). This is very non-obvious so I also left a big comment banner explaining the timing requirements for the `setConfig` call.

##### Motivation

Addresses #6212

##### Context

n/a

#### Pull request checklist

- [x] Addresses an existing issue: #6212
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/injected/client-init.ts
+++ b/src/injected/client-init.ts
@@ -16,16 +16,23 @@ async function initClient() {
     }
 
     const { Stylesheet } = await import('@fluentui/merge-styles');
+    const stylesheet = Stylesheet.getInstance();
+
+    // This *must* be set *before* any transitive import of any other fluentui component!
+    //
+    // This configuration modifies the prefix that fluentui will use for any styles that it
+    // injects dynamically into the target page's global styles. It's important that we use a
+    // unique prefix to avoid conflicts with target pages that use fluentui themselves. In some
+    // cases, just *importing* fluentui will transitively create conflicting styles, so this needs
+    // to be configured before that import happens (see #6212).
+    stylesheet.setConfig({
+        defaultPrefix: 'insights',
+    });
+
     const { createDefaultLogger } = await import('common/logging/default-logger');
     const { initializeFabricIcons } = await import('../common/fabric-icons');
     const { MainWindowInitializer } = await import('./main-window-initializer');
     const { WindowInitializer } = await import('./window-initializer');
-
-    const stylesheet = Stylesheet.getInstance();
-
-    stylesheet.setConfig({
-        defaultPrefix: 'insights',
-    });
 
     const logger = createDefaultLogger();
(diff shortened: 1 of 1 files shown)


waabid (approved):
Great find! LGTM.

PR #6302, opened 2022-12-29
feat(medium-pass-2): add store/wiring for transferring data from quick assess to assessment

waabid:
#### Details

Adds a store to contain state that will be used for modules transferring data from quick assess to assessment.

##### Motivation

Feature work.

##### Context

Currently, this shouldn't do anything but exist. In future PRs, we will add functionality wiring this to UI and the module that actually transfers the data.

#### Pull request checklist

- [ ] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [ ] (UI changes only) Added screenshots/GIFs to description above
- [ ] (UI changes only) Verified usability with NVDA/JAWS

PR #6305, opened 2022-12-31
feat(medium-pass-2): add and wire up data going from quick assess to assessment

waabid:
#### Details

Creates a module responsible for sending the quick assess data to the assessment store.

##### Motivation

Feature work.

##### Context

The flow of this is slightly complicated so here is the gist:

1. User action (activates button to transfer data from quick assess to assessment).
2. Action(initializeTransfer) is invoked.
3. Data transfer store is updated with state indicating that transfer has begun.
4. The updated store state triggers actions (transfer data **with** quick assess data as payload) and (finalizeTransfer) in `QuickAssessToAssessmentConverter` (created in this PR).
5. Assessment store is updated with aforementioned quick assess data.

This flow is this way to ensure that we adhere to our flux architecture (and is similar to the flow used for injecting content-scripts into the target page).

Note: transitioning the view to "Assessment" in the switcher is not in the context of this PR (I think it makes sense to make the UI responsible for that).

Below is a gif on my local machine where I use the "Start Over" button to trigger the messaging I've set up in this PR and the previous PR #6302 (and these changes were only for demo purposes and are not in the PR). Additionally, the dialog which prevents overwriting an in-flight Assessment is also not in this PR.

quickAssessToAssessment

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/background/actions/assessment-actions.ts
```diff
@@ -51,4 +52,5 @@ export class AssessmentActions {
     public readonly continuePreviousAssessment = new AsyncAction<number>();
     public readonly loadAssessment = new AsyncAction<LoadAssessmentPayload>();
     public readonly updateDetailsViewId = new AsyncAction<OnDetailsViewInitializedPayload>();
+    public readonly transferAssessment = new AsyncAction<TransferAssessmentPayload>();
```
madalynrose, line 55: It feels a little odd to put this in `assessmentActions` but that might just be a naming thing. Because other actions are operations on the current assessment, `transferAssessment` sounds like we're sending our current assessment's data somewhere else. Something like `importAssessmentData` might more readily describe the action.

madalynrose (changes requested):
Looks good overall! One question and one stray console.log.

src/background/actions/assessment-actions.ts
waabid: Hm, I was thinking about this as well. I was going with transfer because it relates specifically to the `DataTransfer` prefix (as far as store/actions/etc.) because otherwise it feels like it's a synonym for loadAssessment.

I'm going to change it to `loadAssessmentFromTransfer', let me know what you think.

madalynrose (approved):

PR #6324, opened 2023-01-09
chore: Add message to stylelint `selector-class-pattern`, improve pattern, remove deprecated `ms-` classes

katydecorah:
#### Details

This pull request will improve the `selector-class-pattern` regex which missed a few selectors, it also makes the following changes:

- Add custom error message to `selector-class-pattern`.
- Fix a few selectors that did not match the patterns and add `is-lowerCamelCase` FluentUI pattern. One example of this is `is-focusVisible`.
- Remove deprecated classes `ms-fontColor-neutralTertiaryAlt`, `ms-fontWeight-semibold`, `ms-fontColor-neutralPrimary`, `ms-nav-linkButton`.
- Improve markup for `FileUrlUnsupportedMessagePanel` to use an ordered list. While I was updating the component to replace the deprecated `ms-fontWeight-semibold` class, I noticed an improvement to the markup could be made here.

The table below contains two screenshots that describe the changes to `FileUrlUnsupportedMessagePanel`. The first screenshot is the current UI for the `FileUrlUnsupportedMessagePanel` and the second screenshot is for the proposed UI for the same component. In the proposed UI, the steps use an ordered list which changes the spacing of the steps slightly. Also, the text "Allow access to file URLs" is now bolded, which was the intent of the current UI but the class `ms-fontWeight-semibold` no longer pulls in styles from FluentUI.

Current | Proposed
---|---
 | image

##### Motivation

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/styles/root-level-only/global-styles.scss
```diff
@@ -8,20 +8,11 @@
 
 @import '../common.scss';
 
-.ms-Fabric.is-focusVisible .ms-Nav-group .ms-nav-linkButton:focus::after {
-    @include set-focus-ring-styles('border');
-}
```
katydecorah, line 13: To do: double check this. I was unable to find a trace of `ms-nav-linkButton`.

katydecorah:
I need to spend more time with this, but will close for now until I have some slack time to think about this more.

PR #6318, opened 2023-01-06
feat(medium-pass-2): Update next requirement button to navigate properly for Quick Assess

madalynrose:
#### Details

This PR updates the next requirement button in `RequirementView` to navigate properly for Quick Assess.
This work includes:
* adding an object to `DetailsViewSwitcherNav` that contains requirement view components and configuration for Assessment and Quick Assess
* calculating next requirement for quick assess based on `mediumPassRequirementKeys` instead of `assessment.requirements`

Quick Assess:
screenshot of side nav showing "Bypass blocks" and next requirement button showing "Reflow" for Quick Assess
screenshot of side nav showing Automated Checks "aria-allowed-attr" and next requirement button showing "aria-allowed-role" for Quick Assess

Assessment:
screenshot of side nav showing "Bypass blocks" and next requirement button showing "Consistent navigation" for Assessment
screenshot of side nav showing Automated Checks "aria-allowed-attr" and next requirement button showing "aria-allowed-role" for Assessment

##### Motivation

feature work 🚀 

##### Context

A later PR will use the `shouldShowQuickAssessRequirementView` flag to display the "why it matters" box and hide the info icon for `RequirementView`.

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [x] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #6317, opened 2023-01-05
feat(medium-pass-2): Add "why it matters" content for quick assess requirements

madalynrose:
#### Details

This PR adds content that will appear in the "Why it matters" gray box for Quick Assess requirements.

##### Motivation

Feature work 🚀 

##### Context

This PR just adds the new content. A later PR will display this content for quick assess requirements.

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #6328, opened 2023-01-09
feat(medium-pass-2): Quick Assess requirement context section

madalynrose:
#### Details

As part of Quick Assess, we no longer display the "Getting Started" view for Assessments and surface that information along with the info and examples content in a new UI component. This PR:
* creates the `RequirementContextSection` component
  * uses `whyItMatters` text from #6317 
  * uses `helpfulResourceLinks` from #6317 if available
  * follows reflow option 1, where small viewports render this section below other requirement content
* adds functions for `shouldShowInfoButton` and `shouldShowRequirementContextSection` to the `requirementViewComponentConfiguration` object
* updates `RequirementViewTitle`, `ContentPanelButton`, and `ContentLink` to take optional parameters to facilitate hiding and showing appropriate content

##### Examples

* Without resource links:
  * why it matters section for link purpose showing only info and examples link and no additional resource links
  * why it matters section for link purpose reflow state
* With resource links:
  *  why it matters section for instructions showing resource links in addition to info and examples link
  * why it matters section for instructions reflow state

##### Motivation

feature work 🚀 

##### Context

This builds on work from #6317 and #6318

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/requirement-view-component-configuration.tsx
```diff
@@ -7,9 +7,21 @@ import {
...
+} from 'DetailsView/components/requirement-view-should-show-context-box';
+import {
+    ShouldShowInfoButton,
+    shouldShowInfoButtonForAssessment,
+    shouldShowInfoButtonForQuickAssess,
+} from 'DetailsView/components/requirement-view-should-show-info-button';
 
 export type RequirementViewComponentConfiguration = {
     getNextRequirementButtonConfiguration: GetNextRequirementButtonConfiguration;
+    shouldShowInfoButton: ShouldShowInfoButton;
```
waabid, line 23: I think we could simplify the components that are using these flags by instead returning the components that we want instead (as the switcher configuration is already operating as a logical switch anyways). As in to say this becomes something like:

```infoButton: ReactFCWithDisplayName<some-shared-props-type>```
or alternatively
```getShowInfoButton: (props: some-shared-props-type) => JSX.Element```

would then mean components go from:

``` 
{props.shouldRenderInfoButton && <info button component>
```
to
```
const InfoButton = config.infoButton;*
<InfoButton {...props}>
``` 
reducing the need for a logical check.

\* (variable assignment is only necessary because I believe TS has an issue with using the object inline for component declaration)

waabid (approved):
LGTM!

waabid:
I'm running into a couple issues I didn't notice before:

1. Errors when browsing automated checks. It seems to be because the `props.requirementKey === AutomatedChecks.key` check is not doing what I think it is supposed to: the requirement key will be different than the automated checks key and so we should check the assessment key instead.
2. The styling of the instances table seems incorrect; the width should extend further to match the width of the how to test section (based on the mocks). With that said, I wonder if it makes sense to extend them even further ...

image
image

waabid (changes requested):
Largely LGTM, but a couple comments.

madalynrose:
@waabid good catches! These bugs have been addressed.

waabid (approved):
LGTM!

Issue #6128, opened 2022-10-12
User is not told if AI-Web skips unresponsive iframe


karanbirsingh:
## Describe the bug

Our extension makes a best-effort attempt to scan child iframes - we send messages to the child frame and wait for a small timeout. If the child frame does not respond by our timeout, we skip the child iframe.

The user should be notified when we skip an iframe. `axe-core` exposes a rule `frame-tested` that we can use to detect when an iframe was skipped.

**To Reproduce**

1. Go to [this demo page](https://karanbirsingh.github.io/temp-snippets/microsoft/accessibility-insights-web/message-capturing-iframes/same-origin.html)
2. Select either checkbox so that child window message events are disrupted
3. Try scanning the page and granting iframe permissions if needed
4. Notice that no issues are found in the child iframe (even though the empty input is unnamed)
5. Notice there is no indication to the user that the frame was skipped

## Expected behavior

`axe-core` has a rule called [`frame-tested`](https://github.com/dequelabs/axe-core/blob/master/doc/frame-messenger.md#error-handling--timeouts) that we can enable in our scans. By default, results will show up in "needs review" and we can add content to flesh it out.

If we prefer to surface the results elsewhere (a dialog box, a notice banner, etc)  that works too.

## Context (please complete the following information)

- AI-Web Version & Environment: Version 2.35.0 
- Target Page: https://karanbirsingh.github.io/temp-snippets/microsoft/accessibility-insights-web/message-capturing-iframes/same-origin.html

This is a demo target page. We have seen websites in our support channel exhibiting this behavior.

karanbirsingh:
@ferBonnin By default, this rule would have shown up in needs review. With the banner, would we like to omit `frame-tested` from needs-review, or is the banner in addition?

Should the banner follow similar design & behavior as the existing banners (screenshot below)?

![banners in ai-web](https://user-images.githubusercontent.com/7775527/198089696-5bdc408b-f459-40d7-84ab-6b36c065fb67.png)

The existing banners have a call to action for the user. In this case, the user probably can't do much because we aren't giving them any configuration options. They should probably file an issue on our GitHub. Maybe there's a default issue template link we can populate for them and suggest in the banner link?

ferBonnin:
@karanbirsingh since we cannot assume the user will step into needs review, we would like to omit the rule and instead add the banner. Yes, the banner should follow similar design & behaviour as the existing banners.

sfoslund:
This was addressed in #6330

PR #6330, opened 2023-01-10
fix: add banner for unresponsive iframes

sfoslund:
#### Details

This PR adds logic to detect when iframes have been skipped in an axe scan via that frames-tested rule. When iframes have been skipped, it shows a yellow banner to notify the user. 

##### Motivation

Addresses issue [#6128](https://github.com/microsoft/accessibility-insights-web/issues/6128)

##### Context

Screenshot of banner with fastpass (shows up similarly in assessment UI):

[Screenshot description: Yellow banner across top of details view with text "there are iframes in the target page that are non-responsive. These frames have been skipped and are not included in the result.]
iframes-skipped-screenshot

#### Pull request checklist

- [x] Addresses an existing issue: [#6128](https://github.com/microsoft/accessibility-insights-web/issues/6128)
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/iframe-skipped-warning.tsx
+++ b/src/DetailsView/components/iframe-skipped-warning.tsx
@@ -1,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { NamedFC } from 'common/react/named-fc';
+import * as React from 'react';
+
+export const IframeSkippedWarningContainerAutomationId = 'iframe-skipped-warning-container';
+
+export const IframeSkippedWarning = NamedFC('IframeSkippedWarning', () => (
+    <div data-automation-id={IframeSkippedWarningContainerAutomationId}>
+        There are iframes in the target page that are non-responsive. These frames have been skipped
+        and are not included in results.
+    </div>
+));
--- a/src/DetailsView/components/iframe-skipped-warning.tsx
+++ b/src/DetailsView/components/iframe-skipped-warning.tsx
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { NewTabLink } from 'common/components/new-tab-link';
 import { NamedFC } from 'common/react/named-fc';
 import * as React from 'react';
 
@@ -8,6 +9,10 @@ export const IframeSkippedWarningContainerAutomationId = 'iframe-skipped-warning
 export const IframeSkippedWarning = NamedFC('IframeSkippedWarning', () => (
     <div data-automation-id={IframeSkippedWarningContainerAutomationId}>
         There are iframes in the target page that are non-responsive. These frames have been skipped
-        and are not included in results.
+        and are not included in results. If you get this warning unexpectedly in a page where you
+        expect the iframes to be responsive, please
+        <NewTabLink href={'https://github.com/microsoft/accessibility-insights-web/issues/6347'}>
+            let us know.
+        </NewTabLink>
     </div>
 ));
--- a/src/injected/scan-incomplete-warning-detector.ts
+++ b/src/injected/scan-incomplete-warning-detector.ts
@@ -3,6 +3,7 @@
 import { BaseStore } from 'common/base-store';
 import { PermissionsStateStoreData } from 'common/types/store-data/permissions-state-store-data';
 import { ScanIncompleteWarningId } from 'common/types/store-data/scan-incomplete-warnings';
+import { ScanResults } from 'scanner/iruleresults';
 import { IframeDetector } from './iframe-detector';
 
 export class ScanIncompleteWarningDetector {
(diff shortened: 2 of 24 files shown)


src/scanner/scan-parameter-generator.ts
```diff
@@ -20,11 +20,11 @@ export class ScanParameterGenerator {
         };
 
         if (options == null || options.testsToRun == null) {
-            result.runOnly!.values = Object.keys(this.rulesIncluded).filter(
-                id => this.rulesIncluded[id].status === 'included',
-            );
+            result.runOnly!.values = Object.keys(this.rulesIncluded)
+                .filter(id => this.rulesIncluded[id].status === 'included')
+                .concat('frame-tested');
```
dbjorge, line 25: It feels off to split the logic of "which rules get included" between the existing place (`get-rule-inclusions`) and this hard-coded magic rule. I think it would be preferable to keep the knowledge of which rules are included in one place.

I'd recommend adding a new `status` value (maybe `included-always` or `force-included` or something) along with a comment indicating the difference between "included" (in automated checks by default) vs "included-always" (even when running scans of an otherwise limited rule set). Then the reason for always-including `frame-tested` can be documented in `get-rule-inclusions` in its `reason` field, like we already do for all the other cases where we override axe rule enablement.

src/DetailsView/components/iframe-skipped-warning.tsx
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { NamedFC } from 'common/react/named-fc';
+import * as React from 'react';
+
+export const IframeSkippedWarningContainerAutomationId = 'iframe-skipped-warning-container';
+
+export const IframeSkippedWarning = NamedFC('IframeSkippedWarning', () => (
+    <div data-automation-id={IframeSkippedWarningContainerAutomationId}>
+        There are iframes in the target page that are non-responsive. These frames have been skipped
```
dbjorge, line 10: This is a good starting point, but I'm not sure this is actionable enough. What do we want a user who sees this warning to do about it, exactly? Perhaps we should include something like "If you see this warning consistently for a page where it isn't expected, please [file an issue](https://github.com/microsoft/accessibility-insights-web/issues/new)."? @ferBonnin , what are your thoughts?

src/injected/scan-incomplete-warning-detector.ts
```diff
@@ -11,14 +12,20 @@ export class ScanIncompleteWarningDetector {
...
+
         if (
             this.iframeDetector.hasIframes() &&
             !this.permissionsStateStore.getState().hasAllUrlAndFilePermissions
         ) {
             warnings.push('missing-required-cross-origin-permissions');
         }
+
+        if (results && results.framesSkipped) {
+            warnings.push('frame-skipped');
```
dbjorge, line 26: Does it actually make sense for us to show this and the missing-permissions warning at the same time? I think it might make more sense to suppress this message if the missing-permissions case would also be triggered, since that's going to be the root cause of the issue. I think this is doubly true if we do end up adding a "file an issue" action item to the warning - we don't really want to have users report issues about the frame-tested violations if they haven't given the extension the necessary permissions.

dbjorge (changes requested):
Thanks @sfoslund ! Some suggestions inline

src/DetailsView/components/iframe-skipped-warning.tsx
dbjorge: Or perhaps it would be more manageable to create a single known tracking issue and link to it, asking folks to comment on it with the site in question.

sfoslund: Either of those options sound reasonable to me! I'll defer to @ferBonnin's judgement for specific wording/ how to manage issues.

ferBonnin: let's create a tracking issue and let users comment on it, the barrier is lower for the user.
What are the cases where the user would get this message unexpectedly? if there aren't iframes?

for the message, how about:
_If you get this warning in a page <reason to get it unexpectedly>, please [let us know](link to the issue)._

sfoslund: > What are the cases where the user would get this message unexpectedly? if there aren't iframes?

Users shouldn't get this message when there aren't iframes on the page, but they might get it on pages where they expect the iframes to be responsive but they're not, for example, if the iframes are responding but very slowly. 

> for the message, how about: `If you get this warning in a page , please [let us know](link to the issue).`

Sounds good to me! I assume that would be in addition to the existing text?

ferBonnin: not sure I follow completely: in which cases would we want the user to let us know / file a bug?

> Sounds good to me! I assume that would be in addition to the existing text?

edited my message because I was using `<>` and the text between them didn't show 🙂 
Yes, that would be in addition to the text " There are iframes in the target page ..."

sfoslund: Not sure if Dan has other thoughts, but IMO I think it would be interesting to get some example target pages where a user thinks that the iframes should be responsive but for some reason we aren't able to communicate with them (and as a result show this error). So it would be valuable for a user to file an issue if they are seeing this warning unexpectedly and think that the iframes should be/ are responsive. 

In this case, should we be a little more specific in the wording and include something like `If you get this warning unexpectedly in a page where you expect the iframes to be responsive, please [let us know](link to the issue).`

sfoslund: I created a tracking issue to link to here: https://github.com/microsoft/accessibility-insights-web/issues/6347, please feel free to edit it directly if I missed anything!

ferBonnin: That message looks great to me!

dbjorge: LGTM! I think the most common case will actually be that the person running the tool won't really know a lot of details about the iframes causing problems in the page, but I think that message should cover that case okay. Thanks for the update!

dbjorge (approved):

PR #6342, opened 2023-01-11
feat(medium-pass-2): add transfer to assessment button

waabid:
#### Details

Adds the transfer to assessment button for the command bar, in quick assess.

##### Motivation

Feature work.

##### Context

Screenshot:
image
High-contrast screenshot:
image

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
+++ b/src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
@@ -127,6 +127,34 @@ describe('Details View ->', () => {
             });
         });
 
+        describe.each([true, false])('With high contrast mode=%s', highContrastMode => {
+            const { commandBarMenuButtonSelectors } = navMenuSelectors;
+            const commandBarWindowWidth = narrowModeThresholds.collapseCommandBarThreshold - 1;
+
+            describe.each`
+                componentName           | componentSelectors               | width
+                ${'command bar button'} | ${commandBarMenuButtonSelectors} | ${commandBarWindowWidth}
+            `('With $componentName visible', ({ componentName, componentSelectors, width }) => {
+                beforeAll(async () => {
+                    await resizeDetailsView(width, componentSelectors.collapsed);
+                });
+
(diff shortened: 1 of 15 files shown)


madalynrose (changes requested):
This looks awesome! Pulled down locally and checked, we need to make sure this button is also available in narrow mode.

narrow mode command bar showing only "start over" for action buttons

waabid:
> This looks awesome! Pulled down locally and checked, we need to make sure this button is also available in narrow mode.
> 
> narrow mode command bar showing only "start over" for action buttons

Good catch! Updated.

src/tests/end-to-end/tests/details-view/reflow-ui.test.ts
```diff
@@ -127,6 +127,34 @@ describe('Details View ->', () => {
             });
         });
 
+        describe.each([true, false])('With high contrast mode=%s', highContrastMode => {
```
waabid, line 130: There was an automated check issue that was being missed in fastpass without these tests.

Effectively, we were loading empty divs with role set to menuitem in FastPass for the load and save assessment buttons. This is because the actual methods would return null BUT we were only checking whether the function was null (which happens in Electron but not in FastPass) and not the return value of the function.

The command-bar-buttons-menu has been modified to take in the JSX.Elements instead and since the details-view-command-bar was already rendering/generating the JSX.Elements for the non-narrow case, it made sense to also do so for the narrow case.

src/tests/unit/tests/DetailsView/components/details-view-command-bar.test.tsx
```diff
@@ -224,16 +194,6 @@ describe('DetailsViewCommandBar', () => {
         expect(rendered.getElement()).toMatchSnapshot();
     });
 
-    test('renders with save assessment button', () => {
```
waabid, line 227: These tests were removed because they are not testing the DetailsViewCommandBar and were not moved to their respective test files because they do not cover anything novel.

madalynrose (approved):

PR #6364, opened 2023-01-19
fix: save assessment button high contrast styling

madalynrose:
#### Details

The "Save Assessment" button is a different color than the rest of the command bar buttons when using Windows high contrast themes. This PR fixes the styling to be consistent.

 State | Before | After|
-------|--------|-------------|
Default | Screenshot showing command bar buttons in white and save assessment in light blue | screenshot showing all command bar buttons in white
Hover | Screenshot showing command bar buttons in white and save assessment in light blue | screenshot showing all command bar buttons in white and the save assessment text in light blue

##### Motivation

quick fix

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/components/controls/insights-command-button.scss
+++ b/src/common/components/controls/insights-command-button.scss
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 @import '../../../common/styles/colors.scss';
-@import '../../../common/styles/common.scss';
 
 .insights-command-button {
     .command-bar-button-icon {
@@ -21,11 +20,9 @@
             color: $primary-text;
         }
     }
-}
 
-// high contrast overrides to make a.insights-command-button
-// match button.insights-command-button
-a.insights-command-button {
+    // high contrast overrides to make a.insights-command-button
+    // match button.insights-command-button
     @media screen and (forced-colors: active) {
         color: buttontext;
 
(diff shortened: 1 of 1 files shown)


src/common/components/controls/insights-command-button.scss
```diff
@@ -1,8 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 @import '../../../common/styles/colors.scss';
+@import '../../../common/styles/common.scss';
```
katydecorah, line 4: Double checking: does this file pull in anything from `common.scss` or ok to remove?

src/common/components/controls/insights-command-button.scss
```diff
@@ -21,3 +22,21 @@ button.insights-command-button {
         }
     }
 }
+
+// high contrast overrides to make a.insights-command-button
+// match button.insights-command-button
+a.insights-command-button {
```
katydecorah, line 28: From what I can tell the selector at line 6 and this selector at line 28 are one-in-the-same. What do you think about combining them? (For example, moving the media query and its contents before the closing bracket at line 24.)

katydecorah (approved):
I tested in all Windows 11 high contrast themes and the button state performs as expected! I left two non-blocking questions/suggestions.

src/common/components/controls/insights-command-button.scss
madalynrose: I had it that way to begin with and separated them for clarity but am happy to move it back!

src/common/components/controls/insights-command-button.scss
madalynrose: oh! this was originally used for the fluent ui high contrast override but I ended up not needing it. good catch!

PR #6334, opened 2023-01-10
feat(axe-core-4.6): update to axe-core 4.6.3, promote link-in-text-block and meta-viewport

dbjorge:
#### Details

This PR updates `axe-core` to its latest version, 4.6.2, from 4.4.1.

Notable code changes beyond the simple version bump include:

* `link-in-text-block` failures have been promoted from "needs review" to "automated checks" (removed from needs review with explicit code change, added to automated checks implicitly due to new axe-core tagging)
* `meta-viewport` failures have been promoted from "unreported" to "automated checks" (implicitly due to new axe-core tagging)
* WCAG mappings for a few rules have changed (we pull these automatically from axe-core's tags, so there is no corresponding code change)
* Per discussion with PM, `link-in-text-block` incomplete results are staying unreported (not being promoted to "needs review")
* `frame-title-unique` and `no-autoplay-audio` are explicitly disabled. They were promoted in axe-core from experimental/best-practice to "wcag-correspondant, but needs-review-only", but we want to omit them from our needs review checks pending further investigation

##### Motivation

See 1960423

##### Context

* We are considering a separate PR to pull in a temporary version of the `aria-required-children` message improvements we've suggested upstream as [axe-core#3842](https://github.com/dequelabs/axe-core/issues/3842) (which would hopefully be removed later once we update to axe-core 4.7). Leaving that for a separate PR.
* While testing this locally, I noticed an issue where the target page console would show an error during scans suggesting axe-core scans were being run multiple times concurrently. I also observed this is `main` without this PR - I don't think it's a regression and will be following up separately.

#### Pull request checklist

- [x] Addresses an existing issue: 1960423
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #6388, opened 2023-01-31
feat(axe-core-4.6): Expose axe-core relatedNodes as "Related paths" card row

dbjorge:
#### Details

This PR implements a feature to add a new "Related paths" row to result cards that represent axe-core results which included any check results with `relatedNodes` properties.

It adds this information to cards in the fast pass automated checks and needs review results, as well as exported reports.

It also adds a (backwards-compatible) property that the service may include as part of combined report reports - if specified, it would add the same field to combined report cards. The decision of whether it is safe to include that data in practice is intentionally out of scope for this PR and delegated to the service - we might decide to always include it if available, or to only include it if it's both available and consistent between all grouped results. The service will have the responsibility of deciding that, like it already has responsibility for the rest of result-grouping logic.

Screenshot of FastPass Automated Checks showing an `aria-required-children` violation with Related paths:
screenshot of FastPass card with "Related paths" card row added in between "Snippet" and "How to fix" rows, showing a list of targets formatted similarly to the old "Path" row

Screenshot of FastPass Report Export showing the same info:
screenshot of FastPass report export with "Related paths" card row added in between "Snippet" and "How to fix" rows, showing a list of targets formatted similarly to the old "Path" row

Screenshot of FastPass Automated Checks' `color-contrast` results on a page where `axe-core` only reports related paths for some results - note that cards for results where the data is present show it, and results where axe-core didn't report that data omit it.
screenshot of two FastPass cards, one with Related paths and one without

##### Motivation

This is primarily to improve the actionability of the `aria-required-children` "unknown" violation path which we added support for during our 4.6.3 update in #6334. That path uses a how to fix message of the form `"Element has children which are not allowed (see related nodes)"`, which isn't very actionable without this PR since we don't expose "related nodes" anywhere. This specific message might be updated in a future version of axe-core (see https://github.com/dequelabs/axe-core/issues/3842)

This secondarily improves the actionability of a few other rules that axe-core exposes related nodes for - particularly, `color-contrast` violations where axe knows the node that is responsible for the background color of a violation will now present that background node as a "related path".

##### Context

In addition to this change, we'll want to make a docs change to https://accessibilityinsights.io/info-examples/web/aria-required-children/ to make this specific form of `aria-required-children` violation easier for a user to understand and take action on.

#### Pull request checklist

- [x] Addresses an existing issue: 2020748
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/common/components/cards/related-paths-card-row.scss
+++ b/src/common/components/cards/related-paths-card-row.scss
@@ -1,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+.path-list {
+    padding-left: 16px;
+    list-style-type: disc;
+}
--- a/src/common/components/cards/related-paths-card-row.tsx
+++ b/src/common/components/cards/related-paths-card-row.tsx
@@ -1,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { SimpleCardRow } from 'common/components/cards/simple-card-row';
+import { CardRowProps } from 'common/configs/unified-result-property-configurations';
+import { NamedFC } from 'common/react/named-fc';
+import { isEmpty } from 'lodash';
+import * as React from 'react';
+import styles from './related-paths-card-row.scss';
+
+export interface RelatedPathsCardRowProps extends CardRowProps {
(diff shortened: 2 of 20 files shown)


JGibson2019 (approved):
LGTM

JGibson2019 (approved):

PR #6395, opened 2023-02-01
chore: add axe-config-generator package

sfoslund:
#### Details

Add a `axe-config-generator` package, which generates an `axe-config.json` file containing the current [ScanOptions](https://www.deque.com/axe/core-documentation/api-documentation/#options-parameter) we provide to axe.run for fastpass scans. 

##### Motivation

Keep tool in centralized location

##### Context

We have been noticing confusion from users about how/ why the set of rules that the web extension uses differs from the set of rules that axe-core provides. This tool is the first step in increasing clarity about which rules the extension is running. 

#### Pull request checklist

- [x] Ran `yarn fastpass`


dbjorge (changes requested):
I think it would make more sense to set up the name/contents of the package according to a user's perspective. From a user's perspective, this package is just a wrapper around the config file, and the `index.js` bundle that builds the config file is an implementation detail of how this package is built, not something they would ever want to run themselves. I'd recommend:

* Remove "generator" from the package name
  * We should consider using namespacing rather than a hyphenated prefix, it's more consistent with how most monorepos are laid out; eg, `@accessibility-insights/axe-config`
* Make a new `index` that just exists to export the contents of the JSON file as a const variable, since that's how most users will want to use it in practice
* Ensure that only that thin index file and the JSON file end up in `drop/`
* Treat the current `index` file as a build step, rename it accordingly, ensure it doesn't end up in `drop/`

src/packages/axe-config-generator/index.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dbjorge, line 1: Did you look into whether it would be possible for this to just live in `/packages/axe-config-generator` directly, rather than making the separate `/src/packages/axe-config-generator` directory? The UI and report packages use the `/src/` directories for historical reasons - if it's technically feasible, it would be better to just keep the config generation in one place, I think

dbjorge: I don't think it's worth spending a ton of time on it, though; if you tried it and it creates difficulties for build/bundler/etc configuration, it's fine to leave it here

sfoslund: I know when I created the assessment-validator package recently someone had a similar suggestion and I explored it and found it to be technically problematic... I don't remember all of the details offhand but I believe the basic issue that importing src code into the packages dir causes a lot of complexity with dependencies/building. That's why the package code that stands alone and does not require src imports (like the new index.js added in this PR) lives in top level packages dir whereas the rest of the code lives in src/packages. 

Let me know if you're interested in more details and I can dig a little further to refresh my memory :)

package.json
```diff
@@ -33,6 +33,7 @@
         "build:package:report": "lerna --scope accessibility-insights-report run build",
         "build:package:ui": "lerna --scope accessibility-insights-ui run build",
         "build:prod": "grunt build-prod",
+        "generate-axe-config": "grunt generate-axe-config",
```
dbjorge, line 36: Maybe `build:package:axe-config` for more consistency with other packages?

dbjorge (approved):
Thanks, this looks great! Approved with 1 nit

PR #6399, opened 2023-02-02
feat(quick-assess): start over should be specific to quick assess

waabid:
#### Details

The start over dropdown is specific to assessment; this changes that to be specific to be Quick Assess and generalizes the start over text.

##### Motivation

Feature work.

##### Context

What it looks like now:
startOverQuickAssess

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/components/start-over-dropdown.tsx
+++ b/src/DetailsView/components/start-over-dropdown.tsx
@@ -12,7 +12,7 @@ import { InsightsCommandButton } from 'common/components/controls/insights-comma
 import { StartOverDialogType } from 'DetailsView/components/start-over-dialog';
 import * as React from 'react';
 
-import { DetailsRightPanelConfiguration } from './details-view-right-panel';
+import { StartOverContextMenuKeyOptions } from './details-view-right-panel';
 
 export interface StartOverState {
     isContextMenuVisible: boolean;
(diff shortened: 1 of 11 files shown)


src/DetailsView/components/start-over-dropdown.tsx
```diff
@@ -88,23 +90,33 @@ export class StartOverDropdown extends React.Component<StartOverProps, StartOver
...
-                key: 'test',
-                name: `Start over ${testName}`,
-                onClick: this.onStartOverTestMenu,
-            },
-        ];
-
-        return rightPanelConfiguration
-            .GetStartOverContextualMenuItemKeys()
-            .map(key => items.find(item => item.key === key));
+        const {
```
waabid, line 93: This is slightly complicated behavior: basically, we don't show the "test" item in the dropdown depending on which rightPanel we're showing (Overview vs. Testview). However, with the advent of Quick Assess, that differential behavior now also applies to the selection of the switcher (QuickAssess vs Assessment).

As such, both config options/preferences must agree for us to show them. If there's an easier/simpler way to do this without removing the right panel config (which I'd be open to, just not here), I'm open to changing this.

sfoslund: Agreed, I don't immedietly see a simpler alterative without some more serious refactoring, so this approach seems fine to me.

sfoslund (approved):

PR #6447, opened 2023-02-21
feat(automated-checks): refactor test view container into config

sfoslund:
#### Details

Replace switch statement in test-view-container with a config function for getting the appropriate test view container. This will be useful in future feature work as we will be making changes to the test view container used for assessment/ quick assess automated checks.

##### Motivation

Feature work

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/itest-view-container-provider.tsx
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export interface ITestViewContainerProvider {
+    createStaticTestViewContainer(props: any): JSX.Element;
```
waabid, line 5: I think we should look to create a props type that captures all the props necessary for each component.

i.e. something like `const TestViewContainerProviderProps = AdhocIssuesTestViewProps & AdhocTabStopsTestViewProps & [...]`

src/DetailsView/components/itest-view-container-provider.tsx
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export interface ITestViewContainerProvider {
```
waabid, line 4: I think we have moved away from not using the `I` prefix for interfaces.

src/DetailsView/components/itest-view-container-provider.tsx
sfoslund: This is where I was running into the strict null check errors. All of these props have imports which have not been fixed up for strict null checks so when we import them into a file that is included we get ~200 errors. Obviously I'd prefer not to have to fix all of those in this PR, so if we want to import those types here the alternative is to exclude all of the files in this import hierarchy from strict null checks, which I took a stab at but seemed like a bit of a mess to figure out.

lisli1 (approved):

PR #6449, opened 2023-02-22
chore: remove unnecessary axe-config rules

dbjorge:
#### Details

This PR:

* Reverted this change, see comment below: ~~Adds an exclusion for the `bypass` rule, which axe-core changed in 4.2.0 to be a "reviewOnFail" rule (ie, it only ever returns "pass" or "needs review"), but which we were still running anyway and just never reporting any results for (since we don't have needs review content for bypass). Service already excluded this rule explicitly.~~
* Updates the new `@accessibility-insights/axe-config` package to not include an entry for `frame-tested`. We include this rule in AI4Web's scans, but then post-process the results such that we don't present it to users as an automated checks failure (we instead present it as a special warning bar). This change means that a user relying on axe-config won't include `frame-tested`, which will avoid the appearance of issues reported by `axe-config` but not by automated checks.

##### Motivation

* Preparation for [using the axe-config output in service](https://github.com/microsoft/accessibility-insights-service/pull/2402)
* ~~marginal perf improvement by not running unnecessary rule~~

##### Context

n/a

#### Pull request checklist

- [x] Ran `yarn fastpass`


katydecorah (approved):

dbjorge:
The failures here are legitimate; it turns out that removing an automated checks rule from assessment breaks backwards compatibility with old `.a11yassessment` files containing results for that check. The e2es are failing in flows that involve loading assessments because the extension enters an error dialog state rather than the states the e2es are expecting.

PR #6453, opened 2023-02-22
chore: track how long grunt tasks take

karanbirsingh:
#### Details

I was exploring a build update and wanted to know how it affected execution time. GH actions and ADO tasks have time for each task, but a `yarn build` can contain many grunt tasks. This PR introduces a basic library to track how long each `Grunt` task takes. We write the output to disk in case we want to upload/analyze this later.

##### Motivation

Begin collecting some data we can use to inform future changes to the build process

##### Context

Here's some sample output: https://github.com/microsoft/accessibility-insights-web/actions/runs/4247080990/jobs/7384694962#step:8:576

Considerations:
- One drawback of this library is that a multitask like 'drop' doesn't include the full execution time of contained tasks. `grunt.task.run` changes to a new task and the counter resets. I still think the tracking is helpful for granular tasks, but analysis of this data should keep this in mind.
- Another drawback that I didn't realize until the PR check is that the individual tasks inside `concurrent:compile-all` report their time individually as they execute, so their JSON files get overwritten later. The final JSON does include a full time for `concurrent:compile-all` but it doesn't help us track the individual task times. Maybe one way to mitigate this is to create a special GH action or ADO task like `build-reliability` that invokes particular tasks for the sole purpose of measuring performance over time. We could do that without this PR anyways. The PR was already complete so I've left it for review, but if these two drawbacks feel overwhelming I am OK with closing it and looking for a different strategy.
- I don't want the JSON file to inadvertently get pulled into any artifacts. There is a separate Gruntfile for deploy and I have intentionally left this out.

#### Pull request checklist

- [ ] Ran `yarn fastpass`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ drop/
 obj/
 test-results/
 analysis/
+grunt-task-time.json
 *.scss.d.ts
 original-manifest-copy.json
 generated-validate-assessment-json.js
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -8,6 +8,15 @@ const sass = require('sass');
 const targets = require('./targets.config');
 
 module.exports = function (grunt) {
+    require('time-grunt')(grunt, (stats, done) => {
+        // This library tracks execution time of grunt tasks. Note that
+        // grunt.task.run() resets the counter. This means the execution
+        // time for multi-tasks like 'drop' is quite short because it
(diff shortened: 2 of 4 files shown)


karanbirsingh:
After thinking more about this I'm closing the PR primarily because of the behavior around `concurrent` tasks. Before introducing something like this, I will explore the native task run times in ADO / GH.

PR #6457, opened 2023-02-24
feat(automated-checks): AssessmentCardSelection infrastructure

lisli1:
#### Details

This PR adds the infrastructure for AssessmentCardSelection: actions, messages, action creator, message creator, and store.

##### Motivation

Part of feature work

##### Context

The AssessmentCardSelection classes and data types were based off the existing CardSelection and NeedsReviewCardSelection classes and data types.

Will add listeners for AssessmentActions in the AssessmentCardSelectionStore in a future PR.

#### Pull request checklist

- [x] Addresses an existing WI: 2032785
- [x] Ran `yarn fastpass`


src/common/types/store-data/assessment-card-selection-store-data.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { CardSelectionStoreData } from 'common/types/store-data/card-selection-store-data';
+
+export type AssessmentCardSelectionStoreData = CardSelectionStoreData;
```
waabid, line 6: This should instead be an object with keys being assessment configuration keys and the properties being CardSelectionStoreData.

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -0,0 +1,219 @@
...
+        this.assessmentCardSelectionActions.toggleVisualHelper.addListener(this.toggleVisualHelper);
+        this.assessmentCardSelectionActions.getCurrentState.addListener(this.onGetCurrentState);
+        this.unifiedScanResultActions.scanCompleted.addListener(this.onScanCompleted);
+        this.assessmentCardSelectionActions.resetFocusedIdentifier.addListener(
+            this.onResetFocusedIdentifier,
+        );
+        this.assessmentCardSelectionActions.navigateToNewCardsView.addListener(
+            this.onNavigateToNewCardsView,
+        );
+        this.tabActions.existingTabUpdated.addListener(this.onResetStoreData);
```
waabid, line 59: We should not reset data on tab being updated as this is for assessment.

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -0,0 +1,219 @@
...
+            this.toggleRuleExpandCollapse,
+        );
+        this.assessmentCardSelectionActions.toggleCardSelection.addListener(
+            this.toggleCardSelection,
+        );
+        this.assessmentCardSelectionActions.collapseAllRules.addListener(this.collapseAllRules);
+        this.assessmentCardSelectionActions.expandAllRules.addListener(this.expandAllRules);
+        this.assessmentCardSelectionActions.toggleVisualHelper.addListener(this.toggleVisualHelper);
+        this.assessmentCardSelectionActions.getCurrentState.addListener(this.onGetCurrentState);
+        this.unifiedScanResultActions.scanCompleted.addListener(this.onScanCompleted);
```
waabid, line 52: We should instead be using assessment actions here.

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -0,0 +1,219 @@
...
+import { StoreNames } from '../../common/stores/store-names';
+import { RuleExpandCollapseData } from '../../common/types/store-data/card-selection-store-data';
+import {
+    CardSelectionPayload,
+    RuleExpandCollapsePayload,
+    UnifiedScanCompletedPayload,
+} from '../actions/action-payloads';
+import { UnifiedScanResultActions } from '../actions/unified-scan-result-actions';
+
+export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCardSelectionStoreData> {
```
waabid, line 20: We cannot use the exact same types as we do with CardSelectionStore. Specifcally we should create new payloads which should specify which assessment needs to be updated, using assessment keys (i.e. "automated-checks"). This is in relation to the comment specifying that this store contains card selection store data for all tests in an assessment.

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -0,0 +1,183 @@
...
+            this.onResetFocusedIdentifier,
+        );
+        this.assessmentCardSelectionActions.navigateToNewCardsView.addListener(
+            this.onNavigateToNewCardsView,
+        );
+    }
+
+    public getDefaultState(): AssessmentCardSelectionStoreData {
+        const defaultValue: AssessmentCardSelectionStoreData = {
+            testKey: {
```
sfoslund, line 57: I think the testKey state value is being misused throughout the store. With this current implementation there isn't really any point to having a testKey property at all, because we only ever reference `this.state.testKey....`. I think the ideal usage is that the testKey value corresponds to different assessment requirements. So right now, the only testKey in use will be automated checks, but in the future we might have AssessmentCardSelectionStoreData state that looks more like this:

```
{
   'automated-checks': { rules: {....}, visualHelperEnabled: false, focusedResultUid: ''},
   'landmarks': { rules: {....}, visualHelperEnabled: false, focusedResultUid: ''},
   'focus': { rules: {....}, visualHelperEnabled: false, focusedResultUid: ''},
....
}
```

So throughout this store instead of always referencing `this.state.testKey....` it should be more like `this.state['test-key-value-we-get-from-payload']....`. This change will also require us to add the test key to many of the new action payloads that are missing this property right now.

src/tests/unit/tests/background/stores/assessment-card-selection-store.test.ts
```diff
@@ -0,0 +1,392 @@
...
+    }
+});
+
+describe('AssessmentCardSelectionStore Test', () => {
+    let initialState: AssessmentCardSelectionStoreData = null;
+    let expectedState: AssessmentCardSelectionStoreData = null;
+
+    beforeEach(() => {
+        const defaultState: AssessmentCardSelectionStoreData = {
+            testKey: {
```
sfoslund, line 46: Might make sense to use a test default state with multiple test keys to ensure that the logic will work in the future. So instead of 
```
testKey: {....testData}
```
this might be
```
'test-key-1': {....testData1},
'test-key-2': {....testData2}
```

sfoslund (changes requested):
See comments about test key usage. Otherwise, the logic looks good :)

src/background/actions/action-payloads.ts
```diff
@@ -242,13 +242,22 @@ export interface UnifiedScanCompletedPayload extends BaseActionPayload {
 
 export interface RuleExpandCollapsePayload extends BaseActionPayload {
     ruleId: string;
+    testKey?: string;
```
sfoslund, line 245: nit: I think it would be better to have 2 separate payloads here: a `RuleExpandCollapsePayload` that only has the ruleId that is used by the needs review and automated checks card selection store, and a separate `AssessmentSingleRuleExpandCollapsePayload` which **requires** both the ruleId and testKey. I think it's worthwhile to separate them out since I'm concerned that we would be confused about when the testKey is actually necessary if we have it as an optional parameter.

src/background/actions/assessment-card-selection-action-creator.ts
```diff
@@ -0,0 +1,86 @@
...
+import * as TelemetryEvents from '../../common/extension-telemetry-events';
+import { getStoreStateMessage, Messages } from '../../common/messages';
+import { Interpreter } from '../interpreter';
+import { TelemetryEventHandler } from '../telemetry/telemetry-event-handler';
+import {
+    AssessmentCardSelectionPayload,
+    BaseActionPayload,
+    RuleExpandCollapsePayload,
+} from './action-payloads';
+import { CardSelectionActions as AssessmentCardSelectionActions } from './card-selection-actions';
```
sfoslund, line 14: With this current import we're importing the existing card selection actions but just renaming them to match the new assessment card selection action names, instead of actually importing the new assessment card selection actions. This should be `
import { AssessmentCardSelectionActions } from 'background/actions/assessment-card-selection-actions';`. I think this is where a lot of the payload mixups are coming from- fixing this import should cause fastpass to start reporting the places where the payloads are wrong.

src/background/actions/assessment-card-selection-action-creator.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    private onRuleExpansionToggle = async (payload: RuleExpandCollapsePayload): Promise<void> => {
+        await this.assessmentCardSelectionActions.toggleRuleExpandCollapse.invoke(payload);
+        this.telemetryEventHandler.publishTelemetry(
+            TelemetryEvents.RULE_EXPANSION_TOGGLED,
+            payload,
+        );
+    };
+
+    private onToggleVisualHelper = async (payload: BaseActionPayload): Promise<void> => {
```
sfoslund, line 72: Wrong payload here (also on lines 79 and 84). Also, if we refactor the RuleExpandCollapsePayload, then we will need to update line 64 as well.

src/common/message-creators/assessment-card-selection-message-creator.ts
```diff
@@ -0,0 +1,84 @@
...
+        };
+
+        this.dispatcher.dispatchMessage({
+            messageType: Messages.AssessmentCardSelection.RuleExpansionToggled,
+            payload,
+        });
+    };
+
+    public collapseAllRules = (event: SupportedMouseEvent) => {
+        const payload: BaseActionPayload = {
```
sfoslund, line 53: These are still all the old payloads, we want to make sure that these payload types match the type associated with the message we're sending and include the `TestKey`. For example, this `collapseAllRules` message eventually triggers the `collapseAllRules` callback in the `AssessmentCardSelectionStore`, which accepts a `AssessmentExpandCollapsePayload`. With this current implementation we're sending it a `BaseActionPayload` instead, so we want to change line 53 to be `const payload: AssessmentExpandCollapsePayload = {`.

This will also result in a type checking error that we need to include the testKey, which we can add as a parameter to this method similarly to how you did in the `toggleCardSelection` method on line 21. Also note that this change will need to be reflected in the `CardSelectionMessageCreator` interface- so all of those interface methods will have an additional optional test key parameter after this change is finished. 

The same goes for the other payloads in this file (lines 41, 64, and 75), we need to make sure their types match what the store callbacks expect.

src/common/message-creators/card-selection-message-creator.ts
```diff
@@ -8,6 +8,7 @@ export interface CardSelectionMessageCreator {
         ruleId: string,
         resultInstanceUid: string,
         event: React.SyntheticEvent,
+        testKey?: string,
     ) => void;
     toggleRuleExpandCollapse: (ruleId: string, event: React.SyntheticEvent) => void;
```
sfoslund, line 13: These other interface methods will have to be updated with an optional testKey parameter too, see my comment in the assessment card selection message creator for context.

src/tests/unit/tests/common/message-creators/assessment-card-selection-message-creator.test.ts
```diff
@@ -0,0 +1,139 @@
...
+            .returns(() => telemetryStub);
+
+        testSubject.toggleCardSelection(ruleId, resultInstanceUid, eventStub, testKey);
+
+        dispatcherMock.verify(handler => handler.dispatchMessage(expectedMessage), Times.once());
+    });
+
+    test('toggleRuleExpandCollapse', () => {
+        const ruleId = 'test-rule-id';
+        const payload: RuleExpandCollapsePayload = {
```
sfoslund, line 64: Let's update these payloads to match what we actually expect to get in the wild, so add a testKey to all of them and update the types.

src/tests/unit/tests/background/stores/assessment-card-selection-store.test.ts
```diff
@@ -0,0 +1,564 @@
...
+                        cards: {
+                            sampleUid1: false,
+                            sampleUid2: false,
+                        },
+                    },
+                },
+                visualHelperEnabled: false,
+                focusedResultUid: null,
+            },
+            testKey2: {
```
sfoslund, line 69: Thanks for updating this test data, looks good!

src/tests/unit/tests/background/actions/assessment-card-selection-action-creator.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+        ruleExpansionToggleMock.verifyAll();
+        telemetryEventHandlerMock.verify(
+            handler => handler.publishTelemetry(TelemetryEvents.RULE_EXPANSION_TOGGLED, payload),
+            Times.once(),
+        );
+    });
+
+    test('onToggleVisualHelper', async () => {
+        const payloadStub: BaseActionPayload = {};
```
sfoslund, line 93: These payload types need to be updated throughout this file

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -0,0 +1,237 @@
...
+            this.onResetFocusedIdentifier,
+        );
+        this.assessmentCardSelectionActions.navigateToNewCardsView.addListener(
+            this.onNavigateToNewCardsView,
+        );
+    }
+
+    public getDefaultState(): AssessmentCardSelectionStoreData {
+        const defaultValue: AssessmentCardSelectionStoreData = {
+            'automated-checks': {
```
sfoslund, line 64: I haven't given much thought to what the default state should be, but I had assumed that it would be `{}` and we would add testKeys along the way. What is the rationale behind including the `automated-checks` value from the beginning?

sfoslund (changes requested):
The changes you made in the last round look great, but there are quite a few more places where we need to be sure to update the payloads to include the testKey. Thanks for your patience!

src/background/actions/assessment-card-selection-action-creator.ts
lisli1: Good catch!

src/background/stores/assessment-card-selection-store.ts
lisli1: I just wanted to have something there for now but I can take it out. I will be adding the store initialization logic in another PR.

src/background/actions/assessment-card-selection-action-creator.ts
```diff
@@ -0,0 +1,93 @@
...
+        this.telemetryEventHandler.publishTelemetry(
+            TelemetryEvents.RULE_EXPANSION_TOGGLED,
+            payload,
+        );
+    };
+
+    private onToggleVisualHelper = async (
+        payload: AssessmentCardToggleVisualHelperPayload,
+    ): Promise<void> => {
+        await this.assessmentCardSelectionActions.toggleVisualHelper.invoke(null);
```
sfoslund, line 78: Need to pipe payload through (also on line 85 and 90). Seems like tests should have caught this- might mean we have to update the tests as well?

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -0,0 +1,231 @@
...
+    private onNavigateToNewCardsView = async (
+        payload: AssessmentNavigateToNewCardsViewPayload,
+    ): Promise<void> => {
+        if (!payload || !payload.testKey || !this.state[payload.testKey]) {
+            return;
+        }
+
+        this.state[payload.testKey].focusedResultUid = null;
+
+        if (this.state[payload.testKey].rules) {
```
sfoslund, line 217: Random musing: since a lot of this logic seems to be duplicated across all the card selection stores, it might make sense to refactor to consolidate to one place, maybe make the card selection store an abstract class with this logic instead of an interface. 
For the purposes of this PR/ feature, I don't want to block on it, but might be something worthwhile to file an engineering issue over.

src/tests/unit/tests/background/stores/assessment-card-selection-store.test.ts
```diff
@@ -0,0 +1,564 @@
...
+            ['invalid testKey', 'invalid-test', 'sampleRuleId1'],
+            ['invalid ruleId', 'testKey1', 'invalid-rule-id'],
+            ['invalid testKey and ruleId', 'invalid-test', 'invalid-rule-id'],
+            ['null testKey', null, 'sampleRuleId1'],
+            ['undefined testKey', undefined, 'sampleRuleId1'],
+            ['null ruleId', 'testKey1', null],
+            ['null testKey and ruleId', null, null],
+        ];
+
+        it.each(testCases)('does nothing with payload: %s', async (testName, testKey, ruleId) => {
```
sfoslund, line 128: Love these invalid values test cases, very thorough, great work 🎉

PR #6482, opened 2023-03-08
feat(automated-checks): add test view container for assessment automated checks

sfoslund:
#### Details

Add single page view for assessment automated checks. This implementation adds `createAssessmentAutomatedChecksTestViewContainer` to the test view container provider and reuses existing `AdhocIssuesTestView` and `FailedInstancesSection`. Note that these changes are currently behind the automatedChecks feature flag. 

##### Motivation

Feature work

##### Context

This UI component will eventually need to be hooked up to the AssessmentCardSelection infrastructure being added in https://github.com/microsoft/accessibility-insights-web/pull/6457. For now, it is populated with fastpass automated checks data. 

UI after these changes, with feature flag enabled:
[Screenshot description: Assessment details view with automated checks selected. Automated checks is not collapsible in the left nav bar and the main content UI is consistent with fastpass automated checks UI. ]
AssessmentAutoChecksContainer

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [x] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #6480, opened 2023-03-08
feat(automated-checks): add assessment data to cards view model converter

sfoslund:
#### Details

Add logic to convert assessment store data to cards view model. This logic will be used for visualizing automated checks failures with the new assessment automated checks page/ infrastructure.

##### Motivation

Feature work.

##### Context

Implementation notes: 

- This implementation built off of @waabid's branch: https://github.com/microsoft/accessibility-insights-web/compare/main...waabid:accessibility-insights-web:assessmentStoreDataTransformer2
- Right now assessment store data is never passed into getCardViewData, so the new logic is not being exercised in practice. Eventually we will hook it up similarly to the [issues/ needs review logic in details-view-content](https://github.com/microsoft/accessibility-insights-web/blob/main/src/DetailsView/components/details-view-content.tsx#L114-L137), but this is blocked until https://github.com/microsoft/accessibility-insights-web/pull/6457 is merged as we will need to pass in AssessmentStoreData and the new AssessmentCardSelectionStoreData that is being added in that PR.
- This implementation reuses the `convertStoreDataForScanNodeResults` logic added in https://github.com/microsoft/accessibility-insights-web/pull/6476.

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #6476, opened 2023-03-06
feat(automated-checks): add assessment data to element based view model converter

sfoslund:
#### Details

Add logic to convert assessment store data and card selection store data to element based view model. This logic will be used for visualizing automated checks failures with the new assessment automated checks page/ infrastructure. 

##### Motivation

Feature work

##### Context

Implementation notes:

- Right now assessment store data is never passed into `getElementBasedViewModel`, so the new logic is not being exercised in practice. Eventually we will hook it up similarly to the [issues/ needs review logic in SelectorMapHelper](https://github.com/microsoft/accessibility-insights-web/blob/main/src/injected/selector-map-helper.ts#L96-L107), but this is blocked until https://github.com/microsoft/accessibility-insights-web/pull/6457 is merged as we will need to pass in `AssessmentStoreData` and the new `AssessmentCardSelectionStoreData` that is being added in that PR. As a result, I expect there may still be some bugs in this logic that will have to be investigated once we can hook things up end to end. 
- Much of the logic for `getElementBasedViewModel` is the same regardless of the type of data that is passed in, the main difference is just how that data is structured. As a result, to reduce duplication, I'm using an adapted template pattern here such that we normalize both types of data into one universal format that can them be used to generate the element based view model. 

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

PR #6501, opened 2023-03-14
feat(automated-checks): wire up assessment card selection store

sfoslund:
#### Details

This PR integrates the AssessmentCardSelectionStore added in https://github.com/microsoft/accessibility-insights-web/pull/6457 into the data converters and new test view container added in https://github.com/microsoft/accessibility-insights-web/pull/6482, https://github.com/microsoft/accessibility-insights-web/pull/6480, and https://github.com/microsoft/accessibility-insights-web/pull/6476. 

Wiring up this infrastructure also exposed a bug in which scanning would be triggered over and over in the new assessment automated checks view because the `rescanVisualizationWithoutTelemetry` message creator was being triggered over and over in the `IssuesTable`. This PR also fixes that bug by refactoring so that we send different messages depending on the type of data provided (that is, when we're using assessment data, we trigger an assessment rescan instead of a unified rescan). 

##### Motivation

Feature work

##### Context

After this PR there are never any results displayed in the assessment automated checks test view. This is because we still need to hook up more action listeners for the `AssessmentCardSelectionStore`, so right now the store is never populated with any data. 

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/DetailsView/components/details-view-content.tsx
```diff
@@ -134,6 +137,18 @@ export const DetailsViewContent = NamedFC<DetailsViewContentProps>('DetailsViewC
             ),
         );
 
+        const assessmentCardsViewData = props.deps.getCardViewData(
+            props.storeState.assessmentStoreData,
```
waabid, line 141: We should use the `assessmentStoreData` we get on line 170 (since that will appropriately switch for Quick Assess and Assessment). We should also look to mimic that pattern for the card selection data.

src/background/stores/assessment-store.ts
```diff
@@ -442,7 +442,10 @@ export class AssessmentStore extends PersistentStore<AssessmentStoreData> {
             this.generateDefaultState(null),
         );
         this.state.assessments[test.key] = defaultTestStatus;
-        this.state.assessmentNavState.selectedTestSubview = test.requirements[0].key;
+
+        if (this.state.assessmentNavState.selectedTestType === payload.test) {
```
waabid, line 446: Hm, reading through the description, I'm not sure I follow this. When would this not be true?

src/injected/element-based-view-model-creator.ts
```diff
@@ -48,7 +48,9 @@ export class ElementBasedViewModelCreator {
         return this.getSelectorToVisualizationMap(
             cardSelectionData,
             results,
-            'platformInfo' in storeData && storeData.platformInfo ? storeData.platformInfo : null,
+            storeData && 'platformInfo' in storeData && storeData.platformInfo
```
waabid, line 51: nit: could this be replaced with `storeData?.platformInfo || null`

src/injected/selector-map-helper.ts
```diff
@@ -105,6 +116,14 @@ export class SelectorMapHelper {
                     cardSelectionStoreData,
                 );
                 break;
+            case VisualizationType.AutomatedChecks:
```
waabid, line 119: I'm not sure if this is the right place for this. The function name refers to Adhoc visualizations (and so does the `isAdHocVisualization` method) above. Either we should rename these methods appropriately or we should move this to the assessment method below (`getFilteredSelectorMap`)

Alternatively, we could also move these to the configurations as well if we so choose, like the `getTestViewContainer` but that might be a future refactor.

src/background/stores/tab-context-store-hub.ts
```diff
@@ -177,6 +179,16 @@ export class TabContextStoreHub implements StoreHub {
             persistStoreData,
         );
         this.needsReviewCardSelectionStore.initialize();
+
+        this.assessmentCardSelectionStore = new AssessmentCardSelectionStore(
```
waabid, line 183: I'm not opposed to the idea but are we sure the assessment card selection store operating at the tab context would work? The assessment stores operate within the global context is why I ask.

waabid (commented):
Largely looks good; left some comments around trying to make sure that quick assess is also included and some other considerations.

src/background/stores/assessment-store.ts
sfoslund: I was running into a bug where the startOver message would be triggered from the issues table right before the details view switched to a different assessment requirement. So it would cause us to get into a state where the selected test type was something like `landmarks` or some other assessment requirement and the selected test subview was a rule not associated with that requirement. This is a fix for that bug.

src/injected/element-based-view-model-creator.ts
sfoslund: Since this could be assessment store data the typechecker requires us to do the in check, so I think it has to stay this way unfortunately :(

src/background/stores/tab-context-store-hub.ts
sfoslund: It looks like @lisli1 included the tabID in the assessment card selection store constructor, but I would be fine with removing that and making this global. @lisli1 do you have any thoughts?

lisli1: We can remove it since it isn't used

src/DetailsView/components/test-view-container.tsx
```diff
@@ -72,6 +82,15 @@ export const TestViewContainer = NamedFC<TestViewContainerProps>('TestViewContai
         automatedChecksCardSelectionMessageCreator:
             props.deps.automatedChecksCardSelectionMessageCreator,
         needsReviewCardSelectionMessageCreator: props.deps.needsReviewCardSelectionMessageCreator,
+        assessmentCardSelectionMessageCreator:
```
waabid, line 85: We can use the `assessment-functionality-switcher` instead. An example of what that would look like can be seen in `iframe-warning`.

src/injected/selector-map-helper.ts
```diff
@@ -58,6 +60,15 @@ export class SelectorMapHelper {
             );
         }
 
+        if (visualizationType === VisualizationType.AutomatedChecks) {
```
waabid, line 63: I think we could use `AutomatedChecks.key` (from the `automated-checks/assessment.tsx`) instead because that would work for both QuickAssess.

Additionally, we'd have to modify `visualization-state-change-handler.ts` to pass the appropriate store's card selection data.

waabid:
> @waabid per our discussion I'm going to remove the work around for the analyzer issue for now, so that we can make a separate (smaller) PR for that change. In the meantime everything is working normally when the feature flag is disabled but when the feature flag is enabled there is no data in the assessment/ quick assess automated checks details view we get the following error in the console:
> 
> ```
> Error while processing browser RuntimeOnMessage event:  TypeError: Cannot read properties of undefined (reading 'key')
>     at Object.getIdentifier (web-visualization-configuration-factory.ts:117:48)
>     at VisualizationStore.enableTest (visualization-store.ts:183:49)
>     at visualization-store.ts:163:20
>     at async-action.ts:24:48
>     at Array.map (<anonymous>)
>     at AsyncAction.invoke (async-action.ts:24:32)
>     at Object.insights/visualization/rescanVisualization (action-creator.ts:489:61)
>     at async EventResponseFactory.applyEventTimeout (event-response-factory.ts:34:16)
>     at async BackgroundBrowserEventManager.forwardEventToApplicationListener (background-browser-event-manager.ts:153:20)
> ```
> 
> I think this is probably fine for now and should not block this PR since we will plan on fixing it in a subsequent PR, but let me know if you disagree.

PR LGTM. As long as it doesn't completely break the extension, I think that's fine.

waabid (approved):

PR #6510, opened 2023-03-16
feat(automated-checks): add listeners for assessment actions in AssessmentCardSelectionStore

lisli1:
#### Details

This PR adds listeners for various assessment actions (scanCompleted, resetData, resetAllAssessmentsData, loadAssessment, and loadAssessmentFromTransfer) and updates the AssessmentCardSelectionStore state based on the payload. 

The `store-data-to-scan-node-result-converter` was updated to take in the `testKey` as a param, since we would want to populate the AssessmentCardSelectionStore with data from all the assessment tests, not just the selected test.

##### Motivation

Part of feature work

##### Context

It uses `store-data-to-scan-node-result-converter` to convert the assessment store data to card selection store data for the load actions.

Added `assessment-card-selection-store.ts` to the exclude list in `tsconfig.strictNullChecks.json` which is not ideal, but there were failing null checks from `initial-assessment-store-data-generator.ts` (already in the exclude list) that would require changing a lot of files that could potentially impact existing functionality.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/background/assessment-card-controller.ts
```diff
@@ -0,0 +1,23 @@
...
+        private readonly assessmentCardSelectionMessageCreator: AssessmentCardSelectionMessageCreator,
+    ) {}
+
+    public initialize(): void {
+        this.assessmentStore.addChangedListener(this.onAssessmentStoreChanged);
+    }
+
+    private onAssessmentStoreChanged = async (): Promise<void> => {
+        const assessmentStoreState = this.assessmentStore.getState();
+        console.log(assessmentStoreState);
```
sfoslund, line 19: Left behind from debugging?

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -56,6 +59,15 @@ export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCard
         this.assessmentCardSelectionActions.navigateToNewCardsView.addListener(
             this.onNavigateToNewCardsView,
         );
+        this.assessmentCardSelectionActions.assessmentStoreChanged.addListener(
+            this.onAssessmentStoreChanged,
+        );
+    }
+
+    public override initialize(initialState?: AssessmentCardSelectionStoreData): void {
+        this.state = initialState || (this.persistedState ?? this.getDefaultState());
```
sfoslund, line 68: 2 things here: 
- The normal pattern would be to override the `generateDefaultState` method instead of initialize. What is the reason for using initialize here? It looks like we're pretty much just duplicating the persistent store initialize logic. 
- This doesn't include the logic to get the state based off of the assessment store state- are you going to add that in a separate PR?

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -64,6 +76,28 @@ export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCard
         return defaultValue;
     }
 
+    private onAssessmentStoreChanged = async (
+        payload: AssessmentStoreChangedPayload,
+    ): Promise<void> => {
+        if (!payload || !payload.assessmentStoreData) {
```
sfoslund, line 82: nit: check that payload.assessmentStoreData.assessment is non-null and has more than 0 entries here too

src/background/assessment-card-controller.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { BaseStore } from 'common/base-store';
+import { AssessmentCardSelectionMessageCreator } from 'common/message-creators/assessment-card-selection-message-creator';
+import { AssessmentStoreData } from 'common/types/store-data/assessment-result-data';
+
+export class AssessmentCardController {
+    constructor(
```
sfoslund, line 8: It doesn't look like this is ever getting constructed/ initialized yet. Is there any reason to put this off to another PR? Have you tested end to end with this being initialized?

src/common/messages.ts
```diff
@@ -238,6 +238,7 @@ export const Messages = {
         ExpandAllRules: `${messagePrefix}/assessmentCardSelection/expandAllRules`,
         ToggleVisualHelper: `${messagePrefix}/assessmentCardSelection/toggleVisualHelper`,
         NavigateToNewCardsView: `${messagePrefix}/assessmentCardSelection/navigateToNewCardsView`,
+        AssessmentStoreChanged: `${messagePrefix}/assessmentCardSelection/assessmentStoreChanged`,
```
sfoslund, line 241: I believe we will want to have 2 assessment card controllers eventually- one for assessment data and one for quick assess data. As such, we will have to duplicate a lot of this code including the following:

- Add an additional AssessmentStoreChanged message to the QuickAssesCardSelection section
- Mirror the logic that this PR is adding in the assessment-card-selection-action-creator in the quick-assess-card-selection-action-creator
- Initialize the controller twice- once with the assessmentCardSelectionMessageCreator and once with the quickAssessCardSelectionMessageCreator

If you prefer, I'm okay with putting this off until another PR though. I forgot to make the message creator able to handle both assessment and quick assess messages in the last PR I made, so those changes are not yet in main for you to integrate with, which would be a blocker. Up to you if you prefer to wait until my next PR is merged or if you want to just port that change to this PR. Here's the necessary change: https://github.com/microsoft/accessibility-insights-web/pull/6501/files#diff-63c0ad049698d3b0edac710cab5bfa621ff664673990ab907f5280ca8d0bb5ceL11-L90

src/tests/unit/tests/background/stores/assessment-card-selection-store.test.ts
```diff
@@ -85,6 +87,181 @@ describe('AssessmentCardSelectionStore Test', () => {
...
+        it('sets the store state with assessment data', async () => {
+            const payload: AssessmentStoreChangedPayload = {
+                assessmentStoreData: createAssessmentStoreDataWithStatus(ManualTestStatus.FAIL),
+            };
+
+            const storeTester =
+                createStoreForAssessmentCardSelectionActions(
+                    'assessmentStoreChanged',
+                ).withActionParam(payload);
+            await storeTester.testListenerToBeCalledOnce(null, expectedState);
```
sfoslund, line 144: This test/ test name seems misleading- the title implies that we're setting the state based on the assessment data but we end up verifying that the final state is the expectedState, which is just the default state. Am I misunderstanding?

src/tests/unit/tests/background/stores/assessment-card-selection-store.test.ts
```diff
@@ -85,6 +87,181 @@ describe('AssessmentCardSelectionStore Test', () => {
...
+
+        it('does nothing when there is no payload', async () => {
+            const storeTester =
+                createStoreForAssessmentCardSelectionActions(
+                    'assessmentStoreChanged',
+                ).withActionParam(null);
+            await storeTester.testListenerToNeverBeCalled(initialState, expectedState);
+        });
+
+        it('does nothing when there is no assessment store data in payload', async () => {
```
sfoslund, line 184: nit: Could probably do a it.each here with all of the cases where we expect the initial and expected state to end up the same- including when the assessmentStoreData.assessments are null and when they are empty.

sfoslund (commented):
Logic looks good over all, a few small comments

src/background/stores/assessment-card-selection-store.ts
lisli1: I was thinking about this too but wasn't completely sure if we wanted to clear the card selection store state if the assessment store state was cleared (on starting over assessment), but I'll double check how the assessment store state looks after such an event.

src/background/assessment-card-controller.ts
lisli1: I'll add this

src/tests/unit/tests/background/stores/assessment-card-selection-store.test.ts
sfoslund: Ah I see, thanks for the clarification!

src/background/stores/assessment-card-selection-store.ts
lisli1: It appears that the assessment store state should never be empty or null, so I'll add the check for payload.assessmentStoreData.assessments as well. Thanks!

src/background/stores/assessment-card-selection-store.ts
lisli1: Discussed offline; we decided that we can pass the persisted assessment store state in the constructor. Will update to make these changes

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -58,6 +80,53 @@ export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCard
         return defaultValue;
     }
 
+    private onAssessmentStoreChanged = async (
```
waabid, line 83: I am not sure if this is the right approach here due to, what I imagine, will be the perf implications; converting the assessment store completely after EACH change seems quite cost-intensive.

Was the approach of adding listeners to assessment actions (and updating accordingly) not feasible?

I suspect we'll have to do a conversion like this every so often (after say load assessment) but not necessarily every change.

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -18,19 +29,29 @@ import {
...
         persistedState: AssessmentCardSelectionStoreData,
+        assessmentStoreData: AssessmentStoreData,
         idbInstance: IndexedDBAPI,
         logger: Logger,
         persistStoreData: boolean,
         indexDBKey: string,
         storeName: StoreNames,
     ) {
         super(storeName, persistedState, idbInstance, indexDBKey, logger, persistStoreData);
+        this.persistedStateFromAssessmentStore =
```
waabid, line 53: nit: I think you could do this in `initialize` since that is usually where we load the persisted data.

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -221,4 +282,134 @@ export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCard
 
         await this.emitChanged();
     };
+
+    private onScanCompleted = async (payload: ScanCompletedPayload<any>) => {
+        if (
+            !payload ||
```
waabid, line 288: Is there a reason we are validating so much of the payload? We don't usually do that ...

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -221,4 +253,106 @@ export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCard
...
+            return;
+        }
+
+        const testKey = assessment.key;
+        const ruleId = payload.key;
+
+        if (!this.state[testKey]) {
+            this.state[testKey] = {
+                rules: null,
+                visualHelperEnabled: false,
```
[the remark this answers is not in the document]
waabid: I think false is best for now; in Assessment we don't usually enable the visual helper by default (but do run a scan). I could see us changing that though, if PMs want that change.

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -50,12 +69,55 @@ export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCard
...
+            return null;
+        }
 
-        return defaultValue;
+        const assessmentCardSelectionStoreData: AssessmentCardSelectionStoreData = {};
+
+        forOwn(assessmentStoreData.assessments, (assessment, key) => {
+            const cardSelectionStoreData: CardSelectionStoreData = assessmentCardSelectionStoreData[
+                key
+            ] ?? {
```
waabid, line 104: nit: it probably makes sense to extract this to a private `getDefaultTestSelectionData` method.

src/common/store-data-to-scan-node-result-converter.ts
```diff
@@ -25,17 +25,21 @@ export type ScanNodeResult = UnifiedResult & {
...
 export const convertResultsToCardSelectionStoreData: ConvertResultsToCardSelectionStoreDataCallback =
-    (state: CardSelectionStoreData, results: ScanNodeResult[]): CardSelectionStoreData => {
+    (state: CardSelectionStoreData, results: ScanNodeResult[] | null): CardSelectionStoreData => {
         if (results) {
             results.forEach(result => {
                 if (result.status !== 'fail' && result.status !== 'unknown') {
                     return;
                 }
 
+                if (state.rules === null) {
```
waabid, line 39: I'm not sure I follow why this is necessary? This code is invoked in the fast pass (needs review/issues) card selection stores as well and would be hesitant to make any behavioral changes that might impact them.

src/tests/unit/tests/background/stores/assessment-card-selection-store.test.ts
```diff
@@ -553,6 +541,147 @@ describe('AssessmentCardSelectionStore Test', () => {
         });
     });
 
+    describe('onScanCompleted', () => {
+        const stubRuleId = 'sampleRuleId1';
+        const stubTestKey = 'testKey1';
```
waabid, line 546: nit: you should look to move these stub keys/ids out if you use them in multiple tests to reduce duplication and also makes it easier for future modifications.

src/common/store-data-to-scan-node-result-converter.ts
lisli1: Removed

src/background/stores/assessment-card-selection-store.ts
```diff
@@ -221,4 +252,90 @@ export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCard
...
+        forOwn(selectorMap, result => {
+            if (result.ruleResults && result.ruleResults[key] && result.ruleResults[key].id) {
+                cards[result.ruleResults[key].id!] = false;
+            }
+        });
+        return cards;
+    }
+
+    private onResetAllAssessmentsData = async (): Promise<void> => {
+        this.state = {};
```
waabid, line 302: I'm okay with leaving this as is but I wonder if in the future we'll need to change this to 'go through each test key and just set it to default'. Regardless, I think `onScanCompleted` would be able to handle a completely empty object as well.

waabid (approved):

PR #6525, opened 2023-03-21
feat(automated-checks): remove storeDataKey and make keys consistent across assessment and visualization configuration objects

waabid:
#### Details

This removes the storeDataKey properties from the assessment objects. This is because we want the visualization configuration objects to be consistent with the assessment objects.

##### Motivation

Feature work.

##### Context

There's a lot of context to this, so here's a summary:

First off, reasoning.

The assessment object has two keys before this PR: the `storeDataKey` and the `key`. When an assessment provider looks up a key, it looks for the `key` property. This is used to store data in the `AssessmentStoreData`, which is persisted and utilized for backwards compatibility. The `storeDataKey` is not directly used from the assessment object but instead is transferred to the visualization configuration object as the `key` property and later utilized to store data within the `VisualizationStore` which is only persisted for MV3 reasons.

There is no tangible benefit that can be seen that was gained from utilizing two different keys here. However, we do run into issues with this current way of operation. The visualization configuration factory will return an object that's `key` will align with the `assessmentObject.storeDataKey` while the assessment providers will instead return an object which will return an object with `key` aligning to `assessmentObject.key`. This can cause confusion as we treat both visualization configuration objects and assessment providers as the source of truth/a way to identify tests. While we can continue to use two different keys (by adding a new `someOtherKey` property to VisualizationConfiguration objects), in this PR, I'm removing them because there doesn't seem to be a clear benefit to doing so while it increases confusion within the codebase.

So, we are now using the assessment object `key` property as the **true** key. This makes sense because that is the key utilized in assessment store data and is persisted/backwards compatible.

The visualization store uses the `storeDataKey` but is only persisted per tab. I did make sure to test the scenario where I started an assessment without these changes and then updated the extension with these changes: the extension worked as expected and did not cause an error (because the store state is permissive: i.e. the new state for that tab had both types of keys and used the new key for changes/state).

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/assessments/adaptable-content/assessment.tsx
+++ b/src/assessments/adaptable-content/assessment.tsx
@@ -56,5 +56,4 @@ export const AdaptableContentAssessment: Assessment = AssessmentBuilder.Assisted
         TextSpacing,
         HoverFocusContent,
     ],
-    storeDataKey: 'adaptableContentAssessment',
 });
--- a/src/assessments/assessment-builder.tsx
+++ b/src/assessments/assessment-builder.tsx
@@ -147,12 +147,12 @@ export class AssessmentBuilder {
         };
 
         const visualizationConfiguration: AssessmentVisualizationConfiguration = {
+            key,
             testViewType: 'Assessment',
             enableTest: AssessmentBuilder.enableTest,
             disableTest: AssessmentBuilder.disableTest,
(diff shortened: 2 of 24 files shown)


sfoslund (approved):
LGTM! 🥳

PR #6570, opened 2023-04-06
fix: prevent spurious "no tab with id errors" in console

dbjorge:
#### Details

This PR removes some spurious error spew from the service worker console along the lines of:

```
Unchecked runtime.lastError: no tab with id: 12345678
sendTabUrlUpdatedAction: tab with ID 12345678 not found, skipping action message
```

The `sendTabUrlUpdatedAction` part happens because `TargetPageController` is printing a message when it sees a browser update/navigate event for a tab that it can't successfully query the browser for information about. However, that case happens commonly in normal operation - for example, the browser sends such a navigation event every time you open a new tab because it sends us a navigation event as it preloads the new tab content in a tab that isn't visible to our extension. The message looks like an error but it isn't; less confusing to omit it.

The `Unchecked runtime.lastError` part happens because `WebextensionBrowserAdapter.getTabAsync` was still using the non-async `chrome.tabs.getTab` API and manually converting it to a Promise-based API, but was doing so incorrectly in that it wasn't checking `chrome.runtime.getLastError` on error like it's supposed to. This PR fixes that by just converting it to use the async version of the API directly.

While I was there, I also normalized all the rest of the async `WebextensionBrowserAdapter` implementations to consistently use `await` instead of directly returning browser promises. This has no functional impact but makes it easier to debug these methods - it enables better stack traces when an error occurs there and also improves the behavior of breakpoints there.

##### Motivation

Avoid misleading looking errors in console

##### Context

The changes that are particularly interesting to review are in `target-page-controller.ts` - most of the individual lines changes are just:
* typings changes from `chrome.tabs.*` types to equivalent `Tab.*` types to accomadate the use of the async API.
* updating the name of the API from `getTabAsync` to just `getTab` (for consistency; none of the other async browser adapter methods use the `*Async` naming convention)

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/DetailsView/details-view-initializer.ts
+++ b/src/DetailsView/details-view-initializer.ts
@@ -193,7 +193,7 @@ initializeFabricIcons();
 
 if (tabId != null) {
     void browserAdapter
-        .getTabAsync(tabId)
+        .getTab(tabId)
         .then((tab: Tab): void => {
             const telemetryFactory = new TelemetryDataFactory();
 
--- a/src/background/extension-details-view-controller.ts
+++ b/src/background/extension-details-view-controller.ts
@@ -2,11 +2,12 @@
 // Licensed under the MIT License.
 import { DetailsViewController } from 'background/details-view-controller';
 import { IndexedDBDataKeys } from 'background/IndexedDBDataKeys';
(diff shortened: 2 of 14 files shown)


peterdur (approved):
Looks good!
Built and smoke tested.

Issue #6524, opened 2023-03-21
Launching web extension sometimes gets stuck in loading spinner


dbjorge:
## Describe the bug

A few users have reported scenarios where they aren't able to successfully launch the extension (ie, click the browser action icon to load the extension popup). We aren't sure exactly what conditions lead to this case yet. In most cases, the issue appears to get resolved by reloading the extension; however, we have had at least [one report from a user that sees the issue reliably even after uninstalling and reinstalling the extension](https://stackoverflow.com/q/75751002/2825668).

We are suspicious that one possible explanation could be users with invalid stored application state which causes the extension to error while loading. However, this wouldn't explain cases that persist across uninstall/reinstall. It's possible that there may be multiple issues causing similar symptoms.

@nang4ally, @lisli1, and myself have all seen this symptom at least once before, but don't have a current repro or working repro steps. Our next steps would be some combination of:

* Reach out to user(s) with consistent error state and request console error logs + a dump of application state.
* Spend some time attempting to reproduce the issue locally (maybe with some attempts at uninstalling/reinstalling)
* Add some sort of self-healing mechanism to detect this class of issue and offer to clear some/all application state to try to work around it

**To Reproduce**

Not yet known

## CodePen repro example

n/a

## Expected behavior

Application should be able to load successfully, even if stored state is somehow invalid

## Screenshots

n/a

## Context (please complete the following information)

- OS Name & Version: Dan has reproed on Windows 11, unsure if other platforms have reproed
- AI-Web Version & Environment: 2.37.3 Production (but also sometimes on dev builds)
- Browser Version: Unknown
- Target Page: Unknown if this is a factor

## Are you willing to submit a PR?

Yes

## Did you search for similar existing issues?

Yes

## Additional context

* [Related public StackOverflow question](https://stackoverflow.com/q/75751002/2825668)

dbjorge:
If anyone does encounter this issue, it would be very helpful if you could provide a dump of the application state of Accessibility Insights. To gather this data:

1. **WARNING**: This debugging data may contain information identifying the pages open in your browser as well as any in-progress work in Accessibility Insights, which may include HTML snippets from page(s) being tested. Do not share this debugging data unless you are comfortable with this! You will have the opportunity to review the data (in step 5) before sharing it (in step 6).
2. Activate Accessibility Insights by clicking its icon in the browser toolbar (or equivalent keyboard shortcut)
3. Right click on the popup with the loading spinner and select "Inspect"
[...]

PR #6571, opened 2023-04-07
fix: prevent UI getting stuck loading after service worker upgrades

dbjorge:
#### Details

This PR addresses an issue that could occur when upgrading the extension (including when rebuilding dev locally without reloading) where you could end up in a state with multiple service worker instances running and store initialization messages getting lost due to not targeting the authoritative one. All the app's UIs would then get stuck in loading spinners waiting for initial store state messages to get sent in reply to messages that are never received.

To achieve this, we add a new component near the beginning of the `service-worker-init` path which detects the case where the initializing service worker is not yet considered "activated" by the browser (because an old worker still exists), and forces the browser to treat the new worker as the "active" one and "claim" the communications of any pre-existing clients (details views, content scripts, etc) from the old worker to the new one.

I expect this to resolve the "stuck on loading spinner" issues that we see during regular development in dev builds, but I do *not* necessarily expect this to fix the version of #6524 that persists across uninstalling + reinstalling the extension like was reported in [this SO post](https://stackoverflow.com/questions/75751002/insights-wont-launch-just-buffers). I suspect what's probably happening there is some error getting thrown during service worker initialization, which would cause the same symptom.

I also included a minor fix to the `TelemetryEventHandler` I noticed while working on this where it accidentally masks the true stacks of exceptions that occur early in initialization before the telemetry client is initialized.

##### Motivation

Partially addresses #6524

##### Context

Reference material on service worker lifetimes I used while writing this:

* [MDN docs on updating service workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#updating_your_service_worker)
* [Helpful blog post tying together several different aspects of managing worker lifetime](https://web.dev/service-worker-lifecycle/)

#### Pull request checklist

- [x] Addresses an existing issue: #6524 (partially)
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/service-worker-init.ts
+++ b/src/background/service-worker-init.ts
@@ -48,6 +48,7 @@ import { title } from 'content/strings/application';
 import { IssueFilingServiceProviderImpl } from 'issue-filing/issue-filing-service-provider-impl';
 import UAParser from 'ua-parser-js';
 import { deprecatedStorageDataKeys, storageDataKeys } from './local-storage-data-keys';
+import { ServiceWorkerActivator } from './service-worker-activator';
 import { cleanKeysFromStorage } from './user-stored-data-cleaner';
 
 let logger: Logger;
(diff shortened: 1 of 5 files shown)


src/background/service-worker-init.ts
```diff
@@ -89,6 +95,9 @@ function initializeSync(): void {
 }
 
 async function initializeAsync(): Promise<void> {
+    // This should happen *before* other initialization steps that involves persisted data
```
DaveTryon, line 98: minor typo
```suggestion
    // This should happen *before* other initialization steps that involve persisted data
```

DaveTryon (approved):

PR #6610, opened 2023-05-03
chore: Refactor to remove transitive dependency in report package

DaveTryon:
#### Details

https://github.com/microsoft/accessibility-insights-action/pull/1624 is failing with the following error:

```
Cannot find module '@microsoft/applicationinsights-core-js' from '../../../node_modules/accessibility-insights-report/drop/index.js'
```

This dependency was added in #6476, which introduced `isNullOrUndefined` from the telemetry module. While using other's code is a great idea, it adds a strange dependency (and one that isn't called out) to the report package. I've added our own version of `isNullOrUndefined` instead of using the one from applicationinsights.

##### Motivation

We should be able to update the report package at any point in time

##### Context

As mentioned above, we could just add the transitive dependency but it seems like the wrong option here. Other places in the code explicitly check for `foo === null || foo === undefined`, which is just nasty here.

#### Pull request checklist

- [x] Ran `yarn fastpass`


src/common/is-null-or-undefined.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export function isNullOrUndefined(value: any): boolean {
```
dbjorge, line 4: If you make the return type a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) instead of just `boolean`, you can avoid needing to use `results!` on L66 of `store-data-to-scan-node-result-converter.ts`.

```suggestion
export function isNullOrUndefined(value: any): value is null | undefined {
```

src/common/store-data-to-scan-node-result-converter.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { isNullOrUndefined } from '@microsoft/applicationinsights-core-js';
+import { isNullOrUndefined } from 'common/is-null-or-undefined';
```
dbjorge, line 4: I still think it's more consistent with the rest of the codebase to just use `foo == null` instead of `isNullOrUndefined(foo)`. I would rather omit the helper and replace L97-104 with something like:

```ts
    if (
        (selectedTest == null) ||
        (assessmentStoreData?.assessments?.[selectedTest] == null)
    ) {​​
        return null;
    }​​
```

...and L62 with:

```ts
if (unifiedScanResultStoreData == null || results == null) {
```

If we want to discuss replacing `== null` with something that doesn't rely on the `== null` behavior matching `undefined`, I'd be okay with that as a more general change done consistently throughout the codebase (add a helper, use a codemod to update existing instances throughout, remove `{ null: 'ignore' }` from our `eqeqeq` eslint rule config), but I'd rather not create two conventions where there is currently one.

src/common/is-null-or-undefined.ts
DaveTryon: As discussed, this method has gone away

src/common/store-data-to-scan-node-result-converter.ts
```diff
@@ -95,10 +94,9 @@ export function convertAssessmentStoreDataToScanNodeResults(
     ruleInfoMap?: ScannerRuleInfoMap,
 ): ScanNodeResult[] | null {
     if (
-        isNullOrUndefined(assessmentStoreData) ||
-        isNullOrUndefined(assessmentStoreData.assessments) ||
-        isNullOrUndefined(selectedTest) ||
-        isNullOrUndefined(assessmentStoreData.assessments[selectedTest])
+        selectedTest == null ||
+        assessmentStoreData?.assessments == null ||
+        assessmentStoreData?.assessments[selectedTest] == null
```
DaveTryon, line 99: I tried `assessmentStoreData?.assessments?[selectedTest] = null`, but it produces the following message: "This condition will always return false since JavaScript compares objects to reference, not by value. ts(2839)"

src/common/store-data-to-scan-node-result-converter.ts
DaveTryon: Agreed that `== null` is the better approach at this time

src/common/store-data-to-scan-node-result-converter.ts
dbjorge: That error is happening because it's intepreting `assessmentStoreData?.assessments` and `[selectedTest] == null` with a `?` in between as the first two clauses of a ternary expression; what you'd instead want here is the optional chaining operator (`?.` with a period after) in between those clauses, like this:

`assessmentStoreData?.assessments?.[selectedTest] == null`

That said, I think the version as written here is at least as clear to read, I'm fine with leaving it as-is if you'd prefer.

dbjorge (approved):

PR #6717, opened 2023-06-16
chore: have dependabot ignore updates from @fluentui/react until we find a solution for enzyme

madalynrose:
#### Details

Update `dependabot.yml` to ignore all updates from `@fluentui/react`

##### Motivation

Updates to `@fluentui/react` are blocked until we replace `enzyme`.

#### Pull request checklist

- [n/a] Ran `yarn fastpass`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -56,10 +56,11 @@ updates:
   - dependency-name: "@types/node"
     versions:
     - ">=15.0.0"
-    # Major version upgrades of fluentui will require extra validation
+    # Ignore fluentui/react updates until we come up with a solution for either
+    # working with enzyme or replacing it
   - dependency-name: "@fluentui/react"
     versions:
-    - ">= 9.0.0"
+    - ">= 0"
     # Keeping react at 16 due to compatibility issues with enzyme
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -56,10 +56,11 @@ updates:
   - dependency-name: "@types/node"
     versions:
     - ">=15.0.0"
-    # Major version upgrades of fluentui will require extra validation
+    # Ignore fluentui/react updates until we come up with a solution for either
+    # working with enzyme or replacing it
   - dependency-name: "@fluentui/react"
     versions:
-    - ">= 9.0.0"
+    - ">= 0"
```
DaveTryon, line 63: Just a thought, but would keeping the current version convey more specific information?
```suggestion
    - ">= 8.96.1"
```

DaveTryon (approved):
Made 1 suggestion, but your call as to whether or not to implement it

.github/dependabot.yml
madalynrose: Oh good call! I followed the pattern of using "0" that we're using for a couple other libraries that require manual upgrades but I don't think there's harm in using the current version number.

PR #6743, opened 2023-06-21
DRAFT: Android Removal burndown

DaveTryon:
This PR serves as a reference for the changes needed to fully remove Android support, and as a playground to understand how the pipelines respond to the changes. We will _not_ merge this PR, but will split it into more manageable pieces until this PR has nothing left in it. In that sense, it will serve as a sort of burndown to indicate how much work still remains.


DaveTryon:
Migration is complete, so this is no longer needed

PR #6747, opened 2023-06-22
feat(remove-android): Reduce android entanglement

DaveTryon:
#### Details

Deprecating Android support will touch hundreds of files, so we're breaking it up into more manageable units. This PR removes some more files from `src\electron`, focusing on things that are clearly unique to the Electron client. Specifically, this removes the persistence of window state, window bounds, and ADB location. These all flow through the same set of message-passing components.

##### Motivation

Android deprecation feature

#### Pull request checklist

- [n/a] Addresses an existing issue (part of Android deprecation feature)
- [x] Ran `yarn fastpass`


src/injected/visualization/focus-indicator-creator.ts
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { Point } from 'electron';
```
DaveTryon, line 4: There's zero reason for a connection to Electron here. The [equivalent unit test definition of Point](https://github.com/microsoft/accessibility-insights-web/blob/main/src/tests/unit/tests/injected/visualization/focus-indicator-creator.test.ts#L15) uses the definition in `injected/visualization/point`, so that's what I used here.

As an aside, this class is not used anywhere in the product, but since it's not related to Android, I'm making no attempt to remove it here.

src/common/stores/store-names.ts
```diff
@@ -16,16 +16,8 @@ export enum StoreNames {
     PathSnippetStore,
     UnifiedScanResultStore,
     CardSelectionStore,
-    DeviceStore,
```
DaveTryon, line 19: All of the removed stores are Android-specific. The build will break if any of these were referenced in the existing code.

src/common/stores/store-names.ts
```diff
@@ -16,16 +16,8 @@ export enum StoreNames {
     PathSnippetStore,
     UnifiedScanResultStore,
```
codeofdusk, line 17: Does "unified" here refer to the "unified" (i.e. Android) client? Should it therefore be removed?

codeofdusk (approved):

Issue #5251, opened 2022-03-09
Assessment instructions need to be updated for windows 11/current browser settings flows


sfoslund:
## Describe the bug

Some assessment `How to test` instructions need to be updated for windows 11 and current browser settings. 

**To Reproduce**

1. Begin an assessment
2. Go to `Adaptable Content, High Contrast Mode` (can also be seen on `Adaptable Content, Resize Text` and possibly others)
3. See step 2 of `How to test` instructions, which says `Use Windows Settings > Ease of Access > High contrast to apply a high contrast theme.`. Settings differ with windows 11 and these instructions are no longer applicable with that version. 

## Expected behavior

Instructions should be updated to reflect the correct settings in:

* The latest version of Windows
* The latest version of Chrome and Microsoft Edge

This should include updating references to "Windows 10", "the new Microsoft Edge", and "Microsoft Edge Insider" to the current values

DaveTryon:
Nandita to drive this and get updated text

DaveTryon:
@nang4ally, tagging you after adding the "needs PM input" tag

DaveTryon:
This has been released to prod

PR #6840, opened 2023-07-26
fix: Update os-specific instructions

DaveTryon:
#### Details

As called out in #5251, we were providing instructions that were targeting only Windows 10, not Windows 11 or macOS. We were also referring to "Microsoft Edge" as either "new Microsoft Edge" or "Microsoft Edge Insider", since that was the appropriate name when the code was written. Times have changed, so we've updated the following:
- In Assessment 17.1 (title: "High contrast mode"), we have replaced the Windows 10 instructions with a link that opens a new page containing instructions for selecting a high contrast theme, with instructions for Windows 10 or Windows 11. We have also added a note that macOS high contrast settings are not picked up in the browser (at least it didn't when I tested it).
- In Assessment 17.2 (title: Resize text), we have replaced the Windows 10 instructions with a link that opens a new page instructions for setting the logical screen size on both Windows or macOS. The macOS instructions are less precise because of how macOS settings work, especially if using an external monitor. For example, on my external 4K display, the smallest available width is 1920.
- Consistently use "Microsoft Edge" as the displayed browser name

##### Screenshots

High contrast:
Description | Screenshot
--- | ---
Old | image
New | image
Instructions | image

Text Resize:
Description | Screenshot
--- | ---
Old | image
New | image
Instructions | image

##### Motivation

Address #5251

##### Context

Something to note: The Windows-recommended way to increase text size (in both Win 10 and Win11) is via the Resize Text option in accessibility. However, that option does not constrain the viewport, so running the test that way on a full desktop display doesn't really cover the scenario that is supposed to be tested here. For that reason, we persist with the way you'd do this on Win7 and earlier.

#### Pull request checklist

- [x] Addresses an existing issue: #5651
- [x] Ran `yarn fastpass`


madalynrose (changes requested):
Looks like some e2e snapshots need to be updated but otherwise this looks great!

DaveTryon:
@madalynrose e2e snapshots have been updated :)

madalynrose (approved):
Awesome!

PR #6300, opened 2022-12-28
refactor: remove obsolete persistStoreData flag from PersistentStore

dbjorge:
#### Details

When we originally introduced `PersistentStore` during the MV2->MV3 transition, it included a `persistStoreData` constructor flag which allowed us to gradually transition different stores to using persistence depending on whether we were working with an MV2 or MV3 build. This functionality is now obsolete - all `PersistentStore`-based stores always use this with a constant `true` input. This PR removes the obsolete flag.

This refactor is fairly simple but touches a lot of files, so I wanted to give it its own PR for ease of review.

##### Motivation

Clean up obsolete code

##### Context

Noticed while working on #6233

#### Pull request checklist

- [n/a] Addresses an existing issue: #0000
- [x] Ran `yarn fastpass`
- [x] Added/updated relevant unit test(s) (and ran `yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] PR title *AND* final merge commit title both start with a semantic tag (`fix:`, `chore:`, `feat(feature-name):`, `refactor:`). See `CONTRIBUTING.md`.
- [n/a] (UI changes only) Added screenshots/GIFs to description above
- [n/a] (UI changes only) Verified usability with NVDA/JAWS

Issue #6233, opened 2022-12-02
Persistent stores should merge default states into persisted states by default


dbjorge:
## Describe the bug

This is a tracking issue for a bit of engineering followup to #6231 that @madalynrose, @waabid, and I agree that we want to defer til after our initial MV3 release (due to being somewhat risky) but implement soon afterwards (for correctness in future upgrade scenarios).

#6231 updates the `VisualizationStore` initialization path to produce initial store state by *merging* any persisted state on top of the store's default state, rather than using the `persistent-store`'s current default behavior of keeping persisted state and ignoring default state when persisted state is available. This is necessary to handle an upgrade scenario where a tab might have persisted store state that doesn't include the new `tests.mediumPass` property.

This issue happened to hit `VisualizationStore` today because it's the store that we happened to update recently, but the underlying issue isn't specific to any one store; this could occur for any store that receives new properties that are expected to be always-initialized in a default state.

We think the correct fix is to update `PersistentStore`'s default initial state calculation behavior to handle "there is persisted state available" by merging persisted state into default state, instead of taking persisted state as-is like it does today. This should make it feasible to simply `VisualizationStore` and revert most of the new behavior added in #6231. It might also enable a similar simplification for other cases that use custom `getDefaultState` behavior (I think only `AssessmentStore` currently).

dbjorge:
Marking as blocked until the initial MV3 release has happened and had a little bake time.

PR #6864, opened 2023-08-04
refactor: remove obsolete persistStoreData flag from PersistentStore, take 2

codeofdusk:
#### Details
When we originally introduced `PersistentStore` during the MV2->MV3 transition, it included a `persistStoreData` constructor flag which allowed us to gradually transition different stores to using persistence depending on whether we were working with an MV2 or MV3 build. This functionality is now obsolete - all `PersistentStore`-based stores always use this with a constant `true` input. This PR removes the obsolete flag.

##### Motivation
Clean up obsolete code

##### Context
This is a takeover of abandoned #6300, which was closed due to the flag's usage in Unified. Now that Unified is gone, this change is unblocked.

This diff is quite large – many files are modified, but each modification is relatively small. Happy to split in any way desired to ease review.

Noticed while working on #6233

#### Pull request checklist
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/background/stores/assessment-card-selection-store.ts
+++ b/src/background/stores/assessment-card-selection-store.ts
@@ -45,11 +45,10 @@ export class AssessmentCardSelectionStore extends PersistentStore<AssessmentCard
         private readonly initialAssessmentStoreDataGenerator: InitialAssessmentStoreDataGenerator,
         idbInstance: IndexedDBAPI,
         logger: Logger,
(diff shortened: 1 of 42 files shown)


DaveTryon (approved):
Seems reasonable. There are a few places where we pass in null for `persistStoreData`, which would have the same result as passing in `false`, but these are all in test code and won't impact the actual product

Issue #6636, opened 2023-05-23
Improve web release steps' NOTICE generation flow


dbjorge:
## Describe the bug

Web embeds a `/src/NOTICE.html` file which is (mostly) generated automatically in `/build.yaml`'s `generate NOTICE.html file` step using an OSPO build task based on data from Component Governance and ClearlyDefined. ClearlyDefined a the project that regularly "harvests" new project releases (eg, npm package releases", identifies their license information, and stores that information in a publically accessible database/API, which the OSPO notice generator uses as part of its NOTICE generation flow.

We used to be able to just accept the generated NOTICE.html and directly overwrite a checked in template `NOTICE.html` file during CI builds before releasing. However, a few months ago, we started noticing that the OSPO build task was emitting warnings along the lines of:

```
##[warning]Components missing from notice, add manually (4):
See: https://docs.opensource.microsoft.com/content/using/required-notice-template.html
Not harvested in ClearlyDefined (4):
npm/npmjs/-/react-router-dom/6.11.0
npm/npmjs/-/react-router/6.11.0
npm/npmjs/-/tabbable/6.1.2
npm/npmjs/@remix-run/router/1.6.0
```

This happens when we accept a dependabot update to a new version of a component prior to ClearlyDefined's automated "harvester" harvesting data about that component.

In practice, ClearlyDefined's harvesting is often very far behind NPM, and sometimes never catches an update of a component - they have a long backlog queue of updates to process and if the queue is full, a given update might simply never be processed.

This means that our current release process involves looking up whether one of these warnings occurred, and if so, manually fixing up the generated NOTICE file, checking it into /src/NOTICE.html, and then spinning up the release based on that NOTICE update merge commit.

This isn't ideal; it requires manual work and it would be easy for a release driver to accidentally skip that work and do a release without an up to date NOTICE file. It's also not great that our canary builds are regularly publishing with inaccurate NOTICE files.

ClearlyDefined has a UI at https://clearlydefined.io/harvest where you can manually request the harvesting of a specific component, and if a human does so, it appears to prioritize the request in the queue; I've usually seen data appear only an hour or two after an explicit request.

I think the ideal update here would be:

* Change our dependabot acceptance flow to require requesting a harvest of new versions explicitly at update time
* Add a step to PR builds for dependabot PRs that will fail until clearlydefined has data for the new dependency version
[...]

PR #6866, opened 2023-08-04
fix: Return to auto-generated NOTICE.html file

DaveTryon:
#### Details

Now that we check package versions in dependabot PR's (see #6858), we want to go back to using auto-generated NOTICE.html files. This PR basically reverts #6410, but it was done in 3 separate commits to enable testing the CI build after each step. I've confirmed that the generated `NOTICE.html` file is properly reflecting our dependencies. ClearlyDefined still has some gaps caused by several of the Microsoft-owned packages associated with ApplicationInsights version 2.8.15, but I've used the ClearlyDefined UI request these (they'd be non-blocking anyway since we own them).

##### Motivation

Progress on #6636 

##### Context

Completing #6636 completely will require a contribution to the notice generator, so that the CI build will fail if ClearlyDefined hasn't yet harvested the versions of the packages that we need.

#### Pull request checklist

- [x] Addresses an existing issue: #6636
- [x] Ran `yarn fastpass`


codeofdusk (approved):
Seems fine to me!

PR #6905, opened 2023-08-19
fix: Fix typo in tab stop instructions

codeofdusk:
#### Details
This PR fixes a small typo in the tab stop instructions.

##### Context
Discovered during 2.39.0 release validation.

#### Pull request checklist
- [x] Ran `yarn fastpass`
- [x] (UI changes only) Verified usability with NVDA


JGibson2019 (dismissed):
It seems the guidance-content.test snapshots haven't been updated and that's causing a failure; we'd need to update that in the e2e folder and then we should be good to go! Other than that LGTM

madalynrose:
On second thought, I actually think this is as designed. The ad-hoc test does not involve `DetailsView` and therefore there is no window that the user needs to return to after testing the page. They just need to traverse the page and notice if the visualization marks any unexpected behavior.

madalynrose (approved):

PR #6976, opened 2023-09-22
chore: Remove inline snapshots

DaveTryon:
#### Details

Due to https://github.com/jestjs/jest/issues/14305, our inline snapshots are causing problems when using `yarn test -u`. We discussed this internally and decided that the simplest approach was to convert the inline snapshots to traditional snapshots. For most cases, this was just removing "Inline" from the method name and removing the string that provided the snapshot. A small number of test cases were checking multiple inline snapshots in the same test, so the approach there was to build an array of the things that were previously checked individually, then having a single snapshot for the test case.

I ran `yarn test -u` and `yarn format:fix` after making all of the changes, just to make sure that we're ready for future changes.

It will probably be easiest to review this one commit at a time--each of the 32 commits represents the conversion of a single test case.

#### Pull request checklist

- [x] Ran `yarn fastpass`

### DIFF
--- a/src/tests/unit/common/__snapshots__/application-properties-provider.test.ts.snap
+++ b/src/tests/unit/common/__snapshots__/application-properties-provider.test.ts.snap
@@ -1,0 +1,16 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`createToolData returns proper tool data 1`] = `
+{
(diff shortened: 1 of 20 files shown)


madalynrose (approved):
Thank you for splitting this into easily digestible commits!

PR #7090, opened 2023-11-09
chore: Initial work for migrating tests from Enzyme to React Testing Library

madalynrose:
#### Details

This PR adds a [jscodeshift](https://github.com/facebook/jscodeshift) [codemod](https://github.com/facebookarchive/codemod) to the repo that can be run using `yarn codemod` to migrate test files (set in the paths variable in tools/codeshift.js) that use [Enzyme](https://enzymejs.github.io/enzyme/) to use [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) instead. 

While this codemod will migrate the low hanging fruit in the tests, it can't cover every scenario perfectly. Tests will need to be incrementally migrated to cut down on PR size. This PR migrates tests in the `src/tests/unit/tests/assessments/` directory to show an example of what that incremental migration will look like.

Because our tests make heavy use of Enzyme's `shallow` rendering and React Testing Library does a full mounted render into raw HTML, we needed a solution to keep our snapshots scoped to the actual component we are testing. To address this, this PR includes new mock helper utilities:
* `mockReactComponents`: adds a `mockImplementation` for each React Component in the array passed in that renders a React Element instead of the full HTML of that component.
* `getMockComponentCall`: retrieves `SomeComponent.mock.calls[callNum]` for SomeComponent and callNum passed in for a mocked React Component
* `getMockComponentClassPropsForCall`: replaces `someReactComponent.props()` functionality for a mocked React Component
* `expectMockedComponentPropsToMatchSnapshots`: a convenience method to assert that a mocked React Component's props match a snapshot

Note: I recommend reviewing this commit by commit in order to understand the flow and changes best.

##### Motivation

migrating Enzyme to eventually upgrade React

##### Context

After all tests have been migrated and all mentions of Enzyme are gone, the migration can continue:
* remove enzyme dependency
* upgrade React version
* upgrade React Testing Library version, which requires a newer version of React

#### Pull request checklist

- [x] Ran `yarn fastpass`


tools/jscodeshift/convert-enzyme-to-rtl.js
```diff
@@ -0,0 +1,311 @@
...
+    const root = j(fileInfo.source);
+    // FIND ENZYME IMPORT
+    const enzymeImportDeclaration = root.find(j.ImportDeclaration, { source: { value: 'enzyme' } });
+    if (enzymeImportDeclaration.length === 0) {
+        return;
+    }
+
+    const getFirstNode = () => root.find(j.Program).get('body', 0).node;
+
+    // Save the comments attached to the first node then remove them
```
DaveTryon, line 118: Comments in this file are sort of inconsistent--is it worth having a consistent style? (definitely not a blocker)

Line 110: `// FIND...`
Line 123: `// add...`
Line 126: `//remove...`
Line 133: `///remove...`

DaveTryon (approved):
Made a minor observation about comments, but looks great overall. Thanks for tackling this!

Issue #7007, opened 2023-10-06
No links to needs-review doc pages?


DaveTryon:
## Describe the bug

None of the needs-review document pages are referenced anywhere inside the web extension. Some of the rules that show up in the __Needs review__ tab link to their counterparts of  __Automated checks__ document pages.

Specifically:
- https://accessibilityinsights.io/info-examples/web/needs-review/aria-input-field-name/ is never referenced
- https://accessibilityinsights.io/info-examples/web/needs-review/color-contrast/ is never referenced. If a `color-contrast` is detected, the "More information about color contrast" opens https://accessibilityinsights.io/info-examples/web/color-contrast/, which is a different page
- https://accessibilityinsights.io/info-examples/web/needs-review/link-in-text-block/ is never referenced. If you create the scenario described on that page, it appears in Automated Checks, not Needs Review, and the "More information about link-in-text-block" link goes to a page owned by deque
- https://accessibilityinsights.io/info-examples/web/needs-review/p-as-heading/ is never referenced
- https://accessibilityinsights.io/info-examples/web/needs-review/th-has-data-cells/ is never referenced

## Expected behavior

Since these pages exist to add more detail on fixing these issues, I'd expect the pages to be used as reference material by the web extension

v-viyada:
This is released in version 2.42.0

PR #7133, opened 2023-12-05
fix: No links to needs-review doc pages? #7007

v-sharmachir:
#### Details
Bug fix to use needs-review document pages for the Needs review tab link. None of the needs-review document pages were referenced anywhere inside the web extension. Some of the rules that show up in the Needs review tab link have document page available but the url under needs review were of either automated check rule url.

The fix used is, In the rule-resources.tsx file we are checking if the tab selected is needs-review on the basis of the outcome type and added a new file named needs-review-rule-resources.tsx is used to create the needs review url on the basis of rules present in the document page.

Screenshot 2023-12-07 191019

##### Motivation

fix: No links to needs-review doc pages? #7007

#### Pull request checklist
- [x] Addresses an existing issue: #7007
- [x] Ran `yarn fastpass`
- [NA] (UI changes only) Added screenshots/GIFs to description above
- [NA] (UI changes only) Verified usability with NVDA/JAWS


src/tests/unit/tests/common/configs/needs-review-rule-resources.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        expect(resourceUrl).toBe(expectedUrl);
+    });
+});
+
+describe(isOutcomeNeedsReview, () => {
+    const ruleId = 'color-contrast';
+    const outcomeType = 'review';
+
+    it('for outcome that is review and ruleId is eligble for needsReview', () => {
+        const expectedBoolean = true;
```
DaveTryon, line 29: 2 thoughts:
1. `expectedResult` would be a far better name than `expectedBoolean`, since it captures the intent, not just the type.
2. In a simple test like this, having everything parameterized just makes it harder to read, especially with a screen reader. There are at least 2 ways to improve this. Either (1) simply the test to something like 
```
        expect(isOutcomeNeedsReview('color-contrast', 'review')).toBe(true);
```
or (2) keep the parameters but use the it.each pattern to specify the parameters as part of an input array, similar to what is done in [this test](https://github.com/microsoft/accessibility-insights-web/blob/main/src/tests/unit/common/element-snapshot-formatter.test.ts#L14-L28)

src/tests/unit/tests/common/components/cards/rule-resources.test.tsx
```diff
@@ -54,6 +54,7 @@ describe('RuleResources', () => {
                 deps: {
                     LinkComponent: linkComponents[testCase.linkComponent],
                 } as RuleResourcesDeps,
+                outcomeType: 'pass',
```
DaveTryon, line 57: Do we need test cases where `outcomeType` is not `pass` ?

src/common/components/cards/rule-resources.tsx
```diff
@@ -8,6 +8,11 @@ import { UnifiedRule } from 'common/types/store-data/unified-data-interface';
 import { isEmpty } from 'lodash';
 import * as React from 'react';
 
+import { InstanceOutcomeType } from '../../../reports/components/instance-outcome-type';
+import {
+    getNeedsReviewRuleResourcesUrl,
+    isOutcomeNeedsReview,
```
DaveTryon, line 14: For testing purposes, should `isOutcomeNeedsReview` be passed in as part of `RuleResourcesDeps`? That would decouple the consumption of `isOutcomeNeedsReview` from the implementation and allow the 2 ends to be tested independently.

src/common/configs/needs-review-rule-resources.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+const needsReviewRuleResourcesPath =
+    'https://accessibilityinsights.io/info-examples/web/needs-review';
+
+export const isOutcomeNeedsReview = (ruleId: string, outcomeType: string) => {
+    return needsReviewRules.includes(ruleId) && outcomeType === 'review' ? true : false;
+};
```
DaveTryon, line 9: Does `needsReviewRules` even need to exist? It seems like this could potentially be simplified to only consider the outcomeType:

```suggestion
export const isOutcomeNeedsReview = (outcomeType: string): boolean => {
    return outcomeType === 'review';
};
```

v-sharmachir: Some of the rules which sometime come to needs review tab don't have their respective documentation URL in the need review URL so `needsReviewRules` is required to check if page exist or not otherwise it will give 404 error. 
image
image

src/tests/unit/tests/common/configs/needs-review-rule-resources.test.ts
v-sharmachir: Refactored the test file like
        expect(isOutcomeNeedsReview('color-contrast', 'review')).toBe(true);

src/common/components/cards/rule-resources.tsx
v-sharmachir: resolved the comment and passed  isOutcomeNeedsReview and getNeedsReviewRuleResourcesUrl as parts of RuleResourcesDeps

src/common/configs/needs-review-rule-resources.ts
DaveTryon: @v-sharmachir, let me dig into the label-content-name-mismatch issue. If it appears only in Needs Review, then I'd argue that its documentation should also appear there. Ultimately, I'm hoping to avoid having 2 separate places that need updating if a rule moves between automated checks and needs review.

v-sharmachir: removed the `needsReviewRules` and updated the test files as well

src/common/configs/needs-review-rule-resources.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+const needsReviewRuleResourcesPath =
+    'https://accessibilityinsights.io/info-examples/web/needs-review';
+
+export const isOutcomeNeedsReview = (outcomeType: string) => {
```
DaveTryon, line 7: Can we add the output type here?
```suggestion
export const isOutcomeNeedsReview = (outcomeType: string) : boolean => {
```

src/tests/unit/tests/common/components/cards/__snapshots__/rule-resources.test.tsx.snap
```diff
@@ -190,10 +197,13 @@ exports[`RuleResources renders with {
 exports[`RuleResources renders with {
   url: 'test-url',
   guidanceLinks: [ { href: 'test-href' }, [length]: 1 ],
-  linkComponent: 'ExternalLink'
+  linkComponent: 'ExternalLink',
+  outcomeType: 'pass'
 }: GuidanceTags props 1`] = `
 {
   "deps": {
+    "GetNeedsReviewRuleResourcesUrl": [Function],
```
DaveTryon, line 205: Hmm....

We don't have direct unit tests for `needs-review-rule-resources`, and are instead testing that code via `rule-resources-test.tsx`. The preferred pattern is for 1 test class to exercise 1 production class, and to mock out dependencies. If we were following our usual pattern, both `GetNeedsReviewRuleResourcesUrl` and `IsOutcomeNeedsReview` would be `MockFunction` instead of `Function`, and there would e a `needs-review-rule-resources.test.tsx`

@madalynrose, I'd love to hear your thoughts on this...

src/common/configs/needs-review-rule-resources.ts
v-sharmachir: resloved

src/common/configs/needs-review-rule-resources.ts
```diff
@@ -0,0 +1,13 @@
...
+    return outcomeType === 'review';
+};
+
+export const getNeedsReviewRuleResourcesUrl = (ruleId: string) => {
```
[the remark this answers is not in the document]
v-sharmachir: resolved

src/tests/unit/tests/common/components/cards/__snapshots__/rule-resources.test.tsx.snap
v-sharmachir: changed the `GetNeedsReviewRuleResourcesUrl` and `IsOutcomeNeedsReview` to be `MockFunction`. we have a seperate test file `needs-review-rule-resources.test.tsx` for the above functions.

DaveTryon (approved):
Thanks, @v-sharmachir! LGTM

PR #7250, opened 2024-02-26
chore: Typescript Cleanup Part 2 of 3 for TabId error

v-prachin:
#### Details

Typescript Cleanup Part 2 of 3 
We have total 140 errors in 104 files. Thus, targeting each error and fixing it. 

##### Motivation
Existing User Story: [User Story: 2132090](https://dev.azure.com/mseng/1ES/_workitems/edit/2150742)

Tab Id error in below files observed after removing deprecated properties from tsconfig file:

1.src/background/extension-details-view-controller.ts
2.src/tests/unit/tests/background/extension-details-view-controller.test.ts
3.src/background/feature-flags-controller.ts
4.src/background/user-configuration-controller.ts
5.src/tests/unit/tests/background/feature-flags-controller.test.ts
6.src/tests/unit/tests/background/user-configuration-controller.test.ts
7.src/background/keyboard-shortcut-handler.ts
8.src/background/target-page-controller.ts
9.src/tests/unit/tests/background/background-message-distributor.test.ts
10.src/background/keyboard-shortcut-handler.test,ts
11.src/background/target-page-controller.test.ts

We already have a Message interface which has two components messageType & payload. InterpreterMessage is a type which has Message & TabId. As TabId was not used so often in Message thus this change must have be done by someone (I assume). So as part of TabId error, replacing Message with InterpreterMessage.

#### Pull request checklist

- [x] Addresses an existing issue: [User Story: 2132090](https://dev.azure.com/mseng/1ES/_workitems/edit/2150742)
- [x] Ran `yarn fastpass`


src/common/message.ts
```diff
@@ -6,7 +6,7 @@ export interface Message {
     payload?: any;
 }
 
-export type InterpreterMessage = Message & { tabId?: number };
+export type InterpreterMessage = Message & { tabId?: any };
```
v-viyada, line 9: Keep the type to number and made related changes in other files

madalynrose (changes requested):
I'm fairly certain we want to find a solution that preserves `tabId` in as many places as possible. I would recommend *not* removing `tabId` from any objects and instead turning any `Message` objects that are causing errors into `InterpreterMessage`s.

src/background/user-configuration-controller.ts
```diff
@@ -25,10 +24,9 @@ export class UserConfigurationController {
         const payload: SetTelemetryStatePayload = {
             enableTelemetry,
         };
-        const message: Message = {
+        const message: InterpreterMessage = {
             messageType: Messages.UserConfig.SetTelemetryConfig,
             payload: payload,
-            tabId: null,
```
v-rakeshsh, line 31: guess we need tabId here?

v-prachin: We removed tabId from here as was getting yarn null check error
NullCheck

src/common/message.ts
v-prachin: Made the suggested change

PR #6611, opened 2023-05-03
chore: Remove deprecated TypeScript options

DaveTryon:
#### Details

Determine the expected impact of removing deprecated TypeScript options. This will not be merged

PR #7256, opened 2024-03-01
chore: Typescript Cleanup- Part 2 for isSelected,Requirement,fixInstructionProcessor

JeevaniChinthala:
#### Details

As part of Typescript migration added/removed the error causing fields.

##### Motivation

TypeScript 5.0 has marked some options as deprecated. We can temporarily override these deprecations, but the documented plan is for the ability to override the flag to go away in TypeScript 5.5

##### Context

1. **Requirement**

-   **Error:** Type '{ test: VisualizationType; requirement: string; telemetry: AssessmentTelemetryData; }' is not assignable to type 
'ToggleActionPayload'.
- Object literal may only specify known properties, and 'requirement' does not exist in type 'ToggleActionPayload'.  
- In AssessmentToggleActionPayload interface the methods enableFastPassVisualHelperWithoutScan and rescanVisualizationWithoutTelemetry uses the requirement property which was an optional property. To avoid null check error a new interface named RequirementToggleActionPayload was introduced, in which the requirement property is declared as either string or undefined.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              

2. **isSelected**

-  **Error:** Type '{ isSelected: boolean; status: InstanceResultStatus; ruleId: string; uid: string; identifiers: { conciseName: string; identifier: string; 'css-selector': string; target: Target; }; descriptors: { ...; }; resolution: { ...; }; rule: { ...; }; }' is not assignable to type 'ScanNodeResult'.
- Object literal may only specify known properties, and 'isSelected' does not exist in type 'ScanNodeResult'.      
- Added isSelected in ScanNodeResult in order to get required data.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            

3.   **fixInstructionProcessor**

- **Error:**  Type'{ deps: InstanceDetailsGroupDeps; fixInstructionProcessor: FixInstructionProcessor; recommendColor: RecommendColor; rule: CardRuleResult; userConfigurationStoreData: null; targetAppInfo: { ...; }; cardSelectionMessageCreator: CardSelectionMessageCreator; }' is not assignable to type 'InstanceDetailsGroupProps'.
-  Object literal may only specify known properties, and 'fixInstructionProcessor' does not exist in type 'InstanceDetailsGroupProps'.
- Upon verifying in UI, fixInstructionProcessor is defined in 'deps' property. So, removed from code.

                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
#### Pull request checklist
<!-- If a checklist item is not applicable to this change, write "n/a" in the checkbox -->
- [x] Addresses an existing issue: #(https://github.com/microsoft/accessibility-insights-web/pull/6611)
- [ ] Ran `yarn fastpass`


src/common/store-data-to-scan-node-result-converter.ts
```diff
@@ -21,6 +21,7 @@ import { ScannerRuleInfo, ScannerRuleInfoMap } from 'scanner/scanner-rule-info';
 
 export type ScanNodeResult = UnifiedResult & {
     rule: UnifiedRule;
+    isSelected?: boolean;
```
madalynrose, line 24: Correct me if I'm wrong, but I looked through the code and I don't think `isSelected` is referenced anywhere. It is needed for `CardResult` but not for `ScanNodeResult` and is set in `rule-based-view-model-provider.ts` `createCardResult`, which accepts a `ScanNodeResult`. I think we can remove it from the object returned in `convertAssessmentResultToScanNodeResult` below.

src/DetailsView/actions/details-view-action-message-creator.ts
```diff
@@ -203,7 +204,7 @@ export class DetailsViewActionMessageCreator extends DevToolActionMessageCreator
         test: VisualizationType,
         requirement?: string,
     ): void {
-        const payload: ToggleActionPayload = {
+        const payload: RequirementToggleActionPayload = {
```
madalynrose, line 207: Why create a new type instead of using `AssessmentToggleActionPayload`?

madalynrose (changes requested):
I have a couple of questions/suggestions before I approve.

src/common/store-data-to-scan-node-result-converter.ts
JeevaniChinthala: Agreed and removed isSelected property from function convertAssessmentResultToScanNodeResult

src/DetailsView/actions/details-view-action-message-creator.ts
JeevaniChinthala: Updated

PR #7274, opened 2024-03-14
feat(axe-core 4.8): Axe core version upgrade for Web

SaanicaG:
#### Details
This PR updates axe-core to its latest version, [4.8.4](https://github.com/dequelabs/axe-core/releases/v4.8.4), from [4.7.2](https://github.com/dequelabs/axe-core/releases/v4.7.2). It also bumps the accessibility-insights-report package version. As part of the axe-core update:

- The deprecate & disable duplicate-id / duplicate-id-active rule has been deprecated (https://github.com/dequelabs/axe-core/pull/4071).
- moved duplicate-id-aria to needs review. (https://github.com/dequelabs/axe-core/pull/4075)
image

- checks: enable help-same-as-label, but remove from rules (https://github.com/dequelabs/axe-core/pull/4096)

- Various improvements were made to the types. Potentially the most impactful of which is that the target and ancestry property now return as UnlabelledFrameSelector instead of as string[],

- Major version upgrade 4.8.0 details https://github.com/dequelabs/axe-core/releases/tag/v4.8.0

##### Motivation

This change is part of https://dev.azure.com/mseng/1ES/_workitems/edit/2157659/

#### Pull request checklist

- [ ] Ran `yarn fastpass`


src/scanner/axe-extension.d.ts
```diff
@@ -31,6 +30,23 @@ declare module 'axe-core/axe' {
...
+        getRolesByType: Function;
+        lookupTable: any;
+    }
+
+    interface Text {
+        accessibleText: Function;
+        isHumanInterpretable: Function;
+        sanitize: Function;
+        subtreeText: Function;
+    }
```
madalynrose, line 49: What impact do these changes have? Was there some kind of error without the interfaces? It seems odd to me that the calls that use these functions didn't need to be updated but I'm not that familiar with setting up a .d.ts file for a library like we've done here.

src/tests/unit/tests/scanner/__snapshots__/get-rule-inclusions.test.ts.snap
```diff
@@ -16,9 +16,18 @@ exports[`getRuleInclusions getRuleInclusions matches snapshotted list of product
     "reason": "best practice rule that was investigated with no known false positives, implemented as an automated check.",
     "status": "included",
   },
+  "aria-braille-equivalent": {
```
madalynrose, line 19: Enabling new rules usually includes adding documentation to info-examples. Have we done that for these?

madalynrose (commented):
The upgrade itself looks solid but I have a couple questions about implementation.

src/tests/unit/tests/scanner/__snapshots__/get-rule-inclusions.test.ts.snap
v-viyada: PM team is working on the documentation for these. We will release both Web extension and docs in sync. @nang4ally to add more details if required.

src/scanner/axe-extension.d.ts
v-rakeshsh: Hi @madalynrose 
Yes correct, we will get errors.
Because for 4.7.2 the above properties were part of axe-extension.d.ts, because in 4.7.2 package these properties were not there. So its added in the code repo.
But in 4.8.4, these properties are added in package, but not fully like few properties are not added fully which is used by our code. So those missing properties are added above.
Please let me know if need further information.

Thanks

madalynrose (approved):

Issue #7598, opened 2025-05-12
[False Positive] aria-hidden-focus on elements with data-tabster-dummy


ultrababs:
## Describe the bug

There is a [known issue](https://dev.azure.com/microsoftdesign/fluent-ui/_wiki/wikis/fluent-ui.wiki/334/Fluent-v9-known-issues?anchor=%60aria-hidden-focus%60-on-elements-with-%60data-tabster-dummy%60) with Fluent where they use Tabster which uses hidden focusable elements. These elements cause a numerous false positives for `aria-hidden-focus` on WCP accessibility scans. There have been quite a few open tickets about this over the years. Is there any possibility of hiding this false positive from WCP scans to prevent failures going forward?

## Screenshots

![Image](https://github.com/user-attachments/assets/ac7074d1-7700-4463-a238-fad25cf4039f)

![Image](https://github.com/user-attachments/assets/4b12d836-d141-4017-b6e9-140bf54ae3f2)

## Did you search for similar existing issues?

Yes, related tickets:
* https://github.com/microsoft/fluentui/issues/25133
* https://github.com/microsoft/fluentui/issues/30670
* https://github.com/microsoft/fluentui/issues/27056
* https://github.com/microsoft/fluentui/issues/25133
* https://github.com/microsoft/fluentui/issues/25133

Related Fluent teams channel threads:
* https://teams.microsoft.com/l/message/19:0201c9e5433b400b8eb01674ebd508b9@thread.skype/1746653580404?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=ffe264f2-14d0-48b5-9384-64f808b81294&parentMessageId=1746653580404&teamName=Fluent%20Community&channelName=Accessibility&createdTime=1746653580404
* https://teams.microsoft.com/l/message/19:0201c9e5433b400b8eb01674ebd508b9@thread.skype/1745431404068?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=ffe264f2-14d0-48b5-9384-64f808b81294&parentMessageId=1745431404068&teamName=Fluent%20Community&channelName=Accessibility&createdTime=1745431404068
[...]

PR #7605, opened 2025-05-20
fix: suppress fluent ui tabster false positive

v-viyada:
#### Details

This PR adds the function to identity fluent ui false positive for aria-hidden-focus. It exclude the violation result for html containing `data-tabster-dummy` for aria-hidden-focus rule.

##### Motivation

addresses issue #7598 

#### Pull request checklist

- [x] Addresses an existing issue: #7598 
- [x] Ran `yarn fastpass`

### DIFF
--- a/src/scanner/rule-processor.ts
+++ b/src/scanner/rule-processor.ts
@@ -29,7 +29,7 @@ export class RuleProcessor {
 
     public suppressFluentUITabsterResult(result: AxeRule): AxeRule {
         /**
-         * [False Positive] aria-hidden-focus on elements with data-tabster-dummy #2769
+         * [False Positive] aria-hidden-focus on elements with data-tabster-dummy #7598
          * Resolves a known issue with Fluent UI, which uses Tabster to manage focus.
          * Tabster inserts hidden but focusable elements into the DOM, which can trigger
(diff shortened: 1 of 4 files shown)


src/scanner/rule-processor.ts
```diff
@@ -27,6 +27,22 @@ export class RuleProcessor {
         return rule;
     }
 
+    public suppressFluentUITabsterResult(result: AxeRule): AxeRule {
+        /**
+         * [False Positive] aria-hidden-focus on elements with data-tabster-dummy #2769
```
v-viyada, line 32: please correct the issue#.

scottaohara:
there is at least one more version of this tabster bumper that uses the attribute `data-is-focus-trap-zone-bumper="true"`

it's used throughout fluent v8's components that can constrain focus.  for instance, the [v8 dialog component](https://developer.microsoft.com/en-us/fluentui#/controls/web/dialog)

```html
<div aria-hidden="true" tabindex="0" data-is-visible="true" data-is-focus-trap-zone-bumper="true" style="pointer-events: none; position: fixed;"></div>
```

checking for both of these attributes would help a lot, especially since there are still so many products using v8

homitc (approved):

PR #7704, opened 2025-11-28
Fix text spacing visual helper not applying to input/button elements

### LINKED ISSUE #4800: Visual helper for text spacing fails to change the spacing for text in input fields and buttons
## Visual helper for text spacing fails to change the spacing for text in input fields and buttons

**To Reproduce**

1. Open a page with at least a native input field and a button, enter some text in the input field.
2. Start an Accessibility Insights for Web assessment.
3. Navigate to step 17.6, Adaptable content > Text spacing.
4. Toggle the Visual helper toggle to the "On"state.
5. The spacing of the text in the input field and on the button does not change.

## Expected behavior

The spacing of the text in the input field and on the button does change.

## Screenshots
Before:
image

After:
image

- Browser Version: Google Chrome Version 94.0.4606.81 (Official Build) (32-bit)
- Target Page: https://www.wikipedia.com will suffice as target for the assessment

Yes

## Additional context

This bug is caused because the tool attempts to use inheritance to change text spacing. This however is very weak and easily overridden. A colleague of mine created an html code snippet which pins down the underlying problem:
```
<style>
    *:not(.not-universal, input) {
        color: blue;
    }
    .container {
        color: fuchsia !important;
    }
    .test {
        color: green;
    }
</style>
<div class="container">
    This is the exact element, i.e. no inheritance, so fuchsia
    <p>
        This is blue because the universal selector is stronger than inheritance
    </p>
    <p class="test">
        This is green because the class selector is stronger than inheritance.
        <span class="not-universal">
            Still green, because the "test" class is closer. (And the universal selector is explicitly not applied here)
        </span>
    </p>
    <input value="black due to user agent">
</div>
```

v-sohmondal:
## Summary

Fixes #4800

Applies the WCAG 2.1 SC 1.4.12 text spacing metrics (`line-height`, `letter-spacing`, `word-spacing`) via a dedicated `TextSpacingDrawer` instead of relying on CSS inheritance from a container class. The drawer walks the target subtree and writes the properties onto each element's own inline style with `important` priority, so they apply to form controls and to elements that carry their own inline / CSS-in-JS declarations — including a pre-existing inline `!important`, which a stylesheet rule can't override.

A `MutationObserver` re-applies the spacing to nodes added or restyled after the helper is toggled on (paused around the drawer's own writes to avoid reacting to them). On toggle-off, each element's original inline value and priority are restored exactly, and the tool's own injected UI is skipped.

## Testing

**Reproduced the bug on main branch**
Screenshot 2025-11-28 112505

### Fix Branch (After Fix)
- Input fields now properly get text spacing
- Buttons now properly get text spacing
Screenshot 2025-11-28 113515

## Details

Added `TextSpacingDrawer` (`src/injected/visualization/text-spacing-drawer.ts`) and wired the text spacing step to it via `DrawerProvider.createTextSpacingDrawer`. The base container rule in `injected.scss` is retained for `p { margin-bottom }` (not one of the inherited spacing properties) and as a lightweight base; the earlier `input, button, select, textarea` selector block is removed since the drawer now covers those controls and more. Unit tests cover applying the properties, overriding an inline `!important`, exact restore on toggle-off, dynamic re-apply, and skipping the tool's own UI.

### Motivation

Addresses issue #4800 - users reported that the text spacing visual helper (used in WCAG 2.1 SC 1.4.12 assessments) was not working on input fields, buttons, and elements with CSS-in-JS styling.

### Context

Text spacing properties are inherited, and an inherited value is the weakest in the cascade — any element with its own declaration (user-agent styles on form controls, or inline / CSS-in-JS styles) keeps its own value and doesn't inherit the container's. Writing the properties directly onto each element's inline style with `important` priority is the only declaration that overrides those cases, including an inline `!important`.
## Pull request checklist

- [x] Addresses an existing issue: #4800
- [x] Ran `yarn fastpass`


src/injected/styles/injected.scss
```diff
@@ -20,6 +20,15 @@
     p {
         margin-bottom: 2em !important;
     }
+
+    input,
+    button,
+    select,
+    textarea {
```
katharinakoal: The PR description says this fixes inputs, buttons, *and* elements with CSS-in-JS styling, but this rule only targets `input`, `button`, `select`, and `textarea`. Anything styled with an inline `style="..."` (CSS-in-JS) will still override these inherited values and won't get re-spaced.

For an a11y tool that is a bit risky, since a tester could think all text responded when some inline-styled text quietly didn't, and pass a page for SC 1.4.12 that shouldn't be.

Could we either:
- scope the description down to "form controls only" and note CSS-in-JS as a known limitation / follow-up, or
- if we really want the inline-style case covered, revisit it (note: even `!important` in the stylesheet won't beat an inline `style` with `!important`, so that likely needs a JS pass, not CSS alone)?

src/injected/styles/injected.scss
```diff
@@ -20,6 +20,15 @@
     p {
         margin-bottom: 2em !important;
     }
+
+    input,
+    button,
+    select,
```
katharinakoal: One thing worth double-checking before this merges: not all of these properties reliably affect native form controls across browsers.

- `select`: `letter-spacing` / `word-spacing` on the collapsed control is inconsistent, and on a lot of platforms the OS draws the text and ignores these entirely, so this line may just do nothing.
- `input`: single-line inputs have historically ignored `line-height` for layout.

That is fine if it is a conscious "best effort" call, but right now a tester might see some controls change and others not, with no signal about which is a tool limitation vs. a real page problem.

Could we verify the actual rendered behavior per property on Chromium, Firefox and Safari, then either drop whatever probably does nothing (`select` is my main suspect) or add a short comment noting these are best-effort on native controls? Keeps the rule honest about what it actually does.

src/injected/styles/injected.scss
v-sohmondal: Replaced the CSS-inheritance approach with a `TextSpacingDrawer` that walks the target subtree and writes `line-height`/`letter-spacing`/`word-spacing` onto each element's own inline style with `important` priority, so it overrides even a pre-existing inline `!important`. A `MutationObserver` re-applies to nodes added or restyled after toggle-on (paused around our own writes); on toggle-off each element's original inline value and priority are restored exactly; the tool's own UI is skipped.

Verified in the loaded extension (Chrome/Edge): `<input style="letter-spacing:0.02em !important">` becomes `0.12em !important` on, and restores to `0.02em !important` off **(cases D & E in the attached repro)**.
Please check the attached repro html page for the cases tested on.
[text-spacing-repro.html](https://github.com/user-attachments/files/29742970/text-spacing-repro.html)

This is an attempt at the JS-pass approach as suggested - 9d8082bdb45d7a32de88f3527a55334ed0304204

src/injected/styles/injected.scss
v-sohmondal: Tested the new JS-based approach in Chromium-based browsers (Chrome & Edge); couldn't install the extension in Firefox.

`<select>` — **toggled off**
select off

`<select>` — **toggled on** (spacing renders and the control widens)
select on

The only part that doesn't reflect it is the OS-drawn dropdown popup (the open option list), which is a platform limitation. The drawer applies spacing to all elements, so these controls stay in scope.
image


`<input>` — **toggled off**
input off

`<input>` — **toggled on**
input on

For Firefox, since the extension wouldn't install, I applied the properties directly to the controls to check the engine renders them — it does:
firefox direct

katharinakoal (approved):

src/injected/styles/injected.scss
katharinakoal: Thanks for checking and providing screenshots. That's super helpful!

